PR #27, opened 2016-12-09
Added Nubian Keyman keyboard

LornaSIL:
This keyboard was updated to include an onscreen keyboard and touch. It
required changes to 3 keystrokes to work with touch better.


mcdurdin:
Thank you for this PR and apologies for the delay in responding. This is a great starting point for a keyboard layout but there are a number of changes I'd like you to make so we can merge the PR. 

Renaming and removing files (rationale: http://help.keyman.com/developer/keyboards/#toc-use-ascii-characters-for-keyboard-base-names):

1. The folder name needs to be the same as the project name, all lower case, and not using punctuation apart from underscore (this is so the name can be a valid identifier in Javascript and we don't run into casing issues across operating systems). I suggest renaming the .kmn, .kpj, .kps and folder to **sil_nubian**) You'll need to update the file references in the .kpj and .kps files. It's a good idea to rename the other related files as well for consistency -- sil_nubian.bmp; sil_nubian-layout.js, sil_nubian-1.0.json.

2. Remove Nubian.kmx, Nubian-1.2.js, Nubian-1.2_load.js, Nubian-.json from the source folder (these will be built into the build folder)

The final folder tree should look something like:

```
sil_nubian\build.cmd
sil_nubian\HISTORY.md
sil_nubian\LICENSE.md
sil_nubian\README.md
sil_nubian\sil_nubian.kpj
[sil_nubian\sil_nubian.kpj.user] -- excluded from git by .gitignore
sil_nubian\source
sil_nubian\source\readme.htm
sil_nubian\source\sil_nubian-layout.js
sil_nubian\source\sil_nubian-1.2.json
sil_nubian\source\sil_nubian.bmp
sil_nubian\source\sil_nubian.kmn
sil_nubian\source\sil_nubian.kps
sil_nubian\source\sil_nubian.kvk
sil_nubian\source\welcome
sil_nubian\source\welcome\welcome.htm
```

sil_nubian.kpj (rationale: http://help.keyman.com/developer/keyboards):

1. Add sil_nubian.kps to the project.
2. Remove Nubian.kmn and re-add sil_nubian.kmn

sil_nubian.kmn:

1. Add reference to bitmap `store(&BITMAP) 'sil_nubian.bmp'`
2. Update visual keyboard reference `store(&VISUALKEYBOARD) 'sil_nubian.kvk'`
3. Update layout file reference `store(&LAYOUTFILE) 'sil_nubian-layout.js'`
4. Targets only needs to have 'any' `store(&TARGETS) 'any'`
5. The license does not match LICENSE.md
6. There are two target lines?
7. Should name be "SIL Nubian" or "Nubian"? I'm not sure on the right answer here.
8. The OSK/touch layout font should be the same on all platforms: use Keyboard/Fonts dialog to correct this.

sil_nubian.kps (rationale: some points in http://www.tavultesoft.com/keymandev/quality/ -- this document needs updating but is still relevant):

1. Remove the .kmn from the sil_nubian.kps (source files shouldn't be included in the package) as they will be online in the git repository.
2. Remove the .kmx and re-add it from the build\ folder (compile the .kmn first to make sure it is there)
3. Remove SN.bmp (not needed for distribution)
4. Add welcome\welcome.htm
5. Don't include uninstall shortcut in Start Menu (this is deprecated and may be removed in KMDev 10)

welcome.htm:

1. Remove `..\` from the reference to readme.htm (after installation they will be in the same folder)
2. License details in the file don't match the LICENSE.md
3. The keyboard layout should really be included in the welcome.htm and not in the readme -- the readme is shown at install time and in a smaller window which is not really helpful for viewing the layout.

Fonts (rationale: http://www.tavultesoft.com/keymandev/quality/)

As far as I can tell, neither Windows nor Mac include an Old Nubian font. Is that right? If so, this means that this keyboard will not be useful without a font, so a default font should be included (I know this is may require further discussion).

I know there is a fair bit to process here; hope it is all clear and makes sense. Where possible I've included documentation on rationale behind the changes, but not all of this is well documented yet. Once you've made the changes above, we will need to go through one more PR review which I will prioritise.

LornaSIL:
Since the folder name itself is changing and so many other changes, we decided it would be best to close the PR and submit a new one.

PR #29, opened 2017-01-18
sil-power-g-1-1

rik-shaw:
* removed 'RALT' trigger for less-used forms.  Added 2 alternate methods:
	* use '[' after letter, such as s[ to produce variant s form
	* use unused capital letter, such as 'J' to produce variant s form (matches SIL Ethiopic)
* modified 'special diphthongs' to use 'WI' trigger
* added 'single Amharic quotes'
* cleaned up welcome.htm
* cleaned up bmp image
* renamed display name to 'SIL Power-G' ('SIL Ethiopic Power G' got confusing for users that want to try both systems).  I am petitioning to include the "-" between Power-G... if not acceptable can be renamed to "Power G".


rik-shaw:
@mcdurdin Not certain why my 'welcome.htm' appears to be a 'binary file'.  In copy and paste did I mess up the encoding on the file?

Also, I dropped the .json file for now as I wasn't keeping it updated correctly and the list of supported languages isn't quite right anyway.

rik-shaw:
@mcdurdin in a bit of research it may be that the .htm is thought to be a "binary file" because there are some really long lines in it (I was just taking the SIL Ethiopic .htm file and manually tweaking from there but probably should clean it up more).  I COULD use a .gitattributes file in order to force it to be recognized as .html, but I don't think that is necessary at this point.  BTW I made a minor commit to change the encoding from utf16 to utf8.

sil_ethiopic_power_g/source/sil_ethiopic_power_g.kmn
```diff
@@ -1,30 +1,23 @@
...
 store(&TARGETS) 'windows macosx linux'
-store(&KEYBOARDVERSION) '1.0'
-store(&ETHNOLOGUECODE) 'amh tir bst'
+store(&KEYBOARDVERSION) '1.1'
+store(&ETHNOLOGUECODE) 'amh bcq bst dwr gdl gez gmv gof guk kxc mdx suq tig zay'
 store(&BITMAP) 'sil_ethiopic_power_g.bmp'
-store(&COPYRIGHT) '© 2016 SIL Ethiopia'
-store(&MESSAGE) 'The SIL Ethiopic Power G Unicode keyboard enables typing in all Ethiopic Fidel based languages.'
-
+store(&COPYRIGHT) '� 2017 SIL Ethiopia'
```
mcdurdin: Encoding appears to be incorrect on the (C) symbol?

rik-shaw:
@mcdurdin thanks for finding that.... I have *no idea* how the encoding got so mixed up.... somehow bouncing back and forth between Windows and Linux and it got mixed up somehow :-(  Anyway, please see the newest commit.

rik-shaw:
OK I don't know much about BOM it seems :-)  My reading says that for UTF-8 BOM isn't recommended: http://stackoverflow.com/questions/2223882/whats-different-between-utf-8-and-utf-8-without-bom

So I am happy to defer to you and can add it :-)

rik-shaw:
@mcdurdin have now added the UTF-8 BOM using vim.... yikes I have no idea about any other text file in Linux but think the BOM isn't there.... again happy to defer to you on this for Keyman.  I have a few more keyboards coming in the near future so will try to remember to add the BOM to them as well.

mcdurdin:
Yeah, UTF-8 BOM is pretty much a weird Windows thing. Keyman Developer relies on it because there are too many .kmn files out there in legacy encodings and I don't really feel safe trying to detect it heuristically for a keyboard source file! So we're stuck with it for .kmn files.

mcdurdin:
Now published at http://keymankeyboards.com/?id=706

Signed installer will be available soon

PR #30, opened 2017-01-20
Added Torwali keyboard

LornaSIL:
This is a new keyboard. I'd like the Scheherazade font installed for the
Web keyboard and the Android keyboard since this keyboard has a few
special characters that are not in most fonts.

PR #31, opened 2017-01-24
Changed sil_nubian.kpj back to original status

LornaSIL:


mcdurdin:
See #30 -- not required

PR #37, opened 2017-02-26
updating to 1.2 and renaming to sil_power_g_ethiopic

rik-shaw:
Sorry to rename once again but sil_ethiopic_power_g was getting confused with sil_ethiopic.  It is my hope that it will be more clear now with the "power_g" first, but still indicating "ethiopic" at the end.

Various fixes included and welcome.htm re-worked to be more maintainable.


mcdurdin:
Would it be possible to leave the package/keyboard filename as sil_ethiopic_power_g and just change the name inside the keyboard and the package? Because:

* Changing the filename is a breaking change that will be complicated within the deployment repositories; existing users will not get the update and we will have two maintain two names for a period of time for sync purposes
* The name sil_ethiopic_power_g groups the keyboard with other sil_ethiopic keyboards, whereas otherwise it tends to get lost in the other sil_* keyboards which are starting to appear.

PR #38, opened 2017-02-26
sil_ethiopic initial commit

rik-shaw:
This is an initial commit for SIL Ethiopic.  Previously SIL NRSI had maintained this keyboard but I have taken it over with their blessing.  After this is finalized I'll need to communicate with them to update various SIL sites that have this available for download, and also get it to the Linux packages.sil.org repository for kmfl.


sil_ethiopic/README.md
```diff
@@ -0,0 +1,23 @@
+SIL Ethiopic Keyboard
+=============================
+
+Copyright (C) SIL Ethiopia
+
+Version 2.0
+
+The “SIL Ethipoic” keyboard was developed to input Ethiopic Fidel scripts.  This keyboard is designed to work with the Abyssinica SIL font (version 1.500). However, unless you need to use the newest Unicode 6.0 Ethiopic codepoints (for Basketo, Gumuz and Gamo-Gofa-Dawro) or the Private Use Area (PUA) characters in this font, you can use this keyboard with any Ethiopic Unicode font.
```
mcdurdin: typo - "SIL Ethipoic"

mcdurdin: Also; there may well be other Ethiopic Unicode fonts that cover more of the Ethiopic range and we'd encourage font developers to support those additional codepoints in the future.

It would be better to restructure the sentence to document that Abyssinica SIL covers all the codepoints that the keyboard supports, in particular for languages X, Y and Z and that some other Unicode fonts may miss some of the Unicode 6.0 codepoints.

sil_ethiopic/HISTORY.md
```diff
@@ -0,0 +1,9 @@
+SIL Power-G Ethiopic Change History
+===================================
+
+2.0 (25 Feb 2017)
```
mcdurdin: Is there any earlier change history? Perhaps also note the change in dev ownership?

sil_ethiopic/source/sil_ethiopic.kps
```diff
@@ -0,0 +1,2 @@
...
...SIFileName><MSIOptions></MSIOptions></Options><Buttons><Button><ButtonType>install</ButtonType><Caption>&amp;Install</Caption><URL></URL><Left>0</Left><Top>0</Top></Button><Button><ButtonType>cancel</ButtonType><Caption>Cancel</Caption><URL></URL><Left>0</Left><Top>0</Top></Button><Button><ButtonType>about</ButtonType><Caption>&amp;About...</Caption><URL></URL><Left>0</Left><Top>0</Top></Button></Buttons><Registry/><StartMenu><Folder></Folder><Items/></StartMenu><Info><Name URL="">SIL Ethiopic</Name><Version URL="">2.0</Version><Copyright URL="">© SIL Ethiopia</Copyright><Author URL="mailto:computer_ethiopia@sil.org">SIL Ethiopia</Author><WebSite URL="http://www.silethiopia.org">http://www.silethiopia.org</WebSite></Info><Files><File><Name>welcome.htm</Name><Description>File welcome.htm</Description><CopyLocation>0</CopyLocation><FileType>.htm</FileType></File><File><Name>sil_ethiopic.bmp</Name><Description>File sil_ethiopic.bmp</Description><CopyLocation>0</CopyLocation><FileType>.bmp</FileType></File><File><Name>..\build\sil_ethiopic.kmx</Name><Description>Keyboard SIL Ethiopic</Description><CopyLocation>0</CopyLocation><FileType>.kmx</FileType></File></Files><Strings/></Package>
```
mcdurdin: The ReadmeFile should be a short summary of what the keyboard contains; perhaps just the initial paragraph or two from the welcome.htm, in a separate file. As it stands you can see the readme is hard to view because it has too much content.

image

sil_ethiopic/source/welcome.htm
```diff
@@ -0,0 +1,1654 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
mcdurdin: The welcome.htm is comprehensive. This is good. However, is it possible to save it in a more repo-friendly structure?Can LibreOffice generate cleaner html or at least add linebreaks into its generated html?

rik-shaw:
@mcdurdin I have made the changes as requested.

welcome.htm: I used "tidy" on the generated .html file from LibreOffice (the result is still really complicated with all the defined styles, but at least wrapping is in place and it should be relatively readable).  I added the source .odt file to the "source" directory and give instructions in the bottom of it (and thus the welcome.htm as well) on how to run the tidy process.  This way others could contribute to the file.  I still find it valuable to keep the source as .odt so that it can do things like when printing have the header row at the top of each page, etc.  In short it makes a much better .pdf (which I will be distributing on the Linux side).

I haven't found any change history for SIL Internationals prior development of the keyboard, but when tracked down I will add it.

Also I have just copied the README.md into source in order to provide as a "Read Me" for the package.

mcdurdin:
Close :) I think you left out the source/README.md and source/SIL-EthiopicKeyboardMap.odt files. Also, don't include the .odt file in the .kps, just in the repository.

However ... README.md is not a workable format for the readme, as it is treated as a downloadable file. So either README.txt or README.html ... preferably .html though it's a tiny bit more work.

(I would also normally say: remember the different target audiences: README.md in the repo is targeted at keyboard developers; the readme.html in the package is targeted at an end user installing the keyboard package. But in this case the content in README.md is fine for both audiences.)

rik-shaw:
@mcdurdin I have made the requested changes.  I have also included a .pdf of the .odt just to make sure that any users without the correct fonts will still be able to reference the document.  This may not be needed but seemed "nice to have?"

mcdurdin:
http://keymankeyboards.com/?id=711

Issue #41, opened 2017-03-22
Mismatches in documentation for sil_ethiopic_power_g keyboard


mcdurdin:
There are a few documentation discrepancies for this keyboard.

1. In the instructions ‘We’ is printed instead of ‘Wy’:

![image](https://cloud.githubusercontent.com/assets/4498365/24177660/96c25e34-0ed6-11e7-93b6-01a12bc84ea9.png)

The 'K#' variations all work as expected. 

ኸ 
ኹ 
ኺ 
ኻ 
ኼ 
ኽ 
ኾ 

ዀ 
ዂ 
ዅ 
ዃ 
ዄ 

ⷐ 
ⷑ 
ⷒ 
ⷓ 
ⷕ 
ⷖ 

All the above work correctly. 

2. However, in the instructions ‘H[‘ OR ‘K’ should produce the following – 

![image](https://cloud.githubusercontent.com/assets/4498365/24177669/a833b668-0ed6-11e7-9fba-c8f4af582350.png)

K works as above but ‘H[‘ does not work and produces the following: 

ሐ[ H[ 
ሐ[ኡ H[u 
ሐ[ኢ H[i 
ሐ[ኣ H[a 
ሐ[ H[y 
ሐ[እ H[e 
ሐ[ኦ H[o 

ሐ[ኧኦ H[Wo 
ሐ[ኧኢ H[Wi 
ሐ[ኧኡ H[Wu 
ሐ[ኧ H[W 
ሐ[ኧ H[Wy 

3. 'H#' should produce the following, but does not. 

![image](https://cloud.githubusercontent.com/assets/4498365/24177672/b031b1c6-0ed6-11e7-9329-4f0abb0f9e6b.png)

ሐ# H# 
ሐ#ኡ H#u 
ሐ#ኢ H#i 
ሐ#ኣ H#a 
ሐ# H#y 
ሐ#እ H#e 
ሐ#ኦ H#o 

So basically the 'H[' and 'H#' keys do not work.

rik-shaw:
@mcdurdin I have corrected these items in the sil-power-g-1-2 branch (it seemed easier to just create a new clean branch rather than re-name everything back to sil_ethiopic_power_g in this branch).  So this branch can be deleted?

mcdurdin:
Per #42:
* Documentation has been updated to show `Wy`
* `H[` now works as well as `K`
* `H#` has been corrected to `H[#`

PR #42, opened 2017-03-27
Sil Power-G 1.2

rik-shaw:
* Corrections made for "H[" and "H[# / K#" sequences.
* Keyboard display renamed to "SIL Power-G Ethiopic" so users not confused between SIL Ethiopic and SIL Power-G Ethiopic if they have both installed.
* readme.htm added
* .odt / .pdf of keyboard chart added


mcdurdin:
Fixes relate to #41

PR #43, opened 2017-04-01
initial version

rmlockwood:
This is a keyboard with some special characters for languages in Ghana. More characters will likely be added in the near future, but this initial version will suffice for now.


mcdurdin:
Thank you for your PR! I am currently at a conference and will review as soon as I get a chance.

mcdurdin:
This looks good. I've accepted this PR with only one important change:

* Removed source/keymandesktop90.msi, source/ghana-.json, source/ghana-1.0.json

The two .json files are not a big deal but it's important not to include the .msi.

mcdurdin:
https://secure.tavultesoft.com/keyman/downloads/keyboards/details.php?KeyboardID=712 - desktop download
http://keymanweb.com/#hag,Keyboard_ghana - web keyboard; also available now in Android and iOS apps under "Hanga" language.

rmlockwood:
Thanks Marc, it’s nice to have this keyboard globally available.

 

I’ve got a couple questions. First of all, why is it that on an Android keyboard the touch+hold popup key is always in uppercase? I would like it to be in lowercase if shift is not on and uppercase if the shift is on.

 

For this Ghana keyboard, they want the capital eng (014A)  to be displayed in this form not in this form when it pops up as a person holds the n key. The “correct” behavior happens in the Euro Latin keyboard. How does one do that?

 

Lastly, I’m wondering about use of deadkeys with tablet/phone keyboards. When someone uses a device, especially a tablet, they may have an external keyboard. How can I have the deadkey system work for these devices?

 

Thanks,

Ron

mcdurdin:
1. The uppercase issue is a bug in Keyman for Android; I have reported it and it will be addressed in a future update.

2. I'm not sure what you mean by "they want the capital eng (014A)  to be displayed in this form not in this form"?

3. Keyman for Android automatically switches to 'desktop' style rules when an external keyboard is connected to the device and it receives input with that keyboard. I have just tested this and it worked without problems.

rmlockwood:
Hi Marc,

 

Thanks for the info. Regarding #2 did the images come through on my last email? On my Android tablet using the EuroLatin 2 Keyman keyboard the Eng capital form looks like a big version of the small Eng which is what my colleague wants. On the new Ghana keyboard the Eng looks like a capital N with a tail. There must be something different in the keyboard definition to produce this. I’m wondering what that is.

 

Thanks,

Ron

LornaSIL:
The glyph for the uppercase Eng that is used is completely dependent on the font, not the keyboard. Some fonts use one style and other fonts another style. The keyboard needs to output U+014A in either case. You could consider packaging a font with the desktop application, but I don't think that will help you with an Android font.

rmlockwood:
But what I don’t understand is why it looks “correct” in the EuroLatin Keyman keyboard. There’s no font installed for that. On your Android hold down the n key on the EuroLatin keyboard and see if the Eng looks like a large version of the small one.

 

Ron

LornaSIL:
You can actually distribute a font for use with the Keyman app on the Android, but it only works when you use the Keyman app. Maybe that's the case here? When I type N using that keyboard I do see the uppercase n-style of eng and I get that style of glyph in the Keyman app. When I switch to using the keyboard in IMing, I still see the uppercase ENG style, but when I actually type it, it displays as the N-style eng, that's because it's not using the font. That's what I think is going on, but I guess only Marc really knows.

rmlockwood:
Can you try the Ghana (Hanga) keyboard in the Keyman app and IMing? If it’s like mine you don’t see the uppercase n-style of Eng, rather the N-style with a tail displayed as the pop-up letter when you press and hold the n key. I realize what actually gets typed is font dependent. I would at least like to have the n-style as the letter that pops-up. Ron

LornaSIL:
The Hanga keyboard works, but I see the uppercase style of ENG, not the style you want. What I did with the sil_nubian keyboard was to put the fonts I wanted distributed in the "source" folder (https://github.com/keymanapp/keyboards/tree/master/sil_nubian/source). Then I asked Marc to distribute them with the Android keyboard. I probably did that in the Pull Request. In the Keyman app I now see the Coptic letters. In other apps on Android I see square boxes because there is no font that supports the script. The Hanga case would be similar. If you put "Charis SIL Afr" in the "source" folder you would at least have a font that displays properly in the Keyman app even though it won't display anywhere else.

andjc:
@rmlockwood @LornaSIL Fonts are problematic for Android devices. By default on Android 4.1 on wards, the device will be using the Roboto font. On older versions it would be using Droid Sans. From memory neither support n-form of Capital Eng.

To add more complications device manufacturers may change the system fonts themselves.

TO change the system font, you need to root the phone, a complicated process for the average users.

Some devices allow you to select a different UI font form fonts pre-installed.  For instance Samsung licensed Monotype's FlipFont technology to allow them to add fonts and allow users to change system fonts.

Some app developers have leveraged off this and have created apps that allow you to add fonts to a device.

All told it can be messy and there is limited information available.

mcdurdin:
In terms of the European Latin keyboard, this is using the DejaVuSans.ttf font in the On Screen Keyboard and within the Keyman app, whereas the Ghana keyboard is currently using the system default font (this is controlled on the Keyman Cloud Server). We can update the Ghana keyboard registration to use DejaVuSans.ttf but then the Capital Eng will differ on the keyboard and the display e.g. in IM apps. Which way do you think you'd prefer to go?

rmlockwood:
I’d like to go with the DejaVuSans font for the Ghana keyboard. This way at least users will see the capital Eng the way we want on the keyboard.

 

Thanks,

Ron

andjc:
The DejaVu Sans LCG version would be more suitable

mcdurdin:
I've added the font registration to the Ghana keyboard; you'll need uninstall the keyboard, flush the Keyman cache and reinstall the keyboard to get the font immediately. You can flush the cache in the Android app settings -- https://www.cnet.com/how-to/how-to-clear-app-cache-and-app-data-in-android-6-0-marshmallow/

mcdurdin:
@andjc - can you elaborate?

andjc:
The DejaVu fonts come in two flavours: the full font, and a subset font restricted to the Latin, Greek and Cyrillic repertoires supported by the font. Both sets are available from the DejaVu download pages

PR #56, opened 2017-06-23
submission of Tem keyboard tem_kdh

DavidLRowe:


DavidLRowe:
I worked with Andy to resolve the issues you raised. I think it's ready.

mcdurdin:
I made one small change: I removed leading zeros from the version numbers, which matches the pattern for other keyboards in the repository. Now processing the keyboard for publication.

PR #65, opened 2017-07-31
Dari Keyboard

MakaraSok:
TODO in each already-migrated keyboard:

- [ ]  update HISTORY.md
- [ ]  update README.md
- [ ]  update welcome.htm
- [ ]  add MIT copyright to the repo
- [ ]  update keyboard layout where applicable


LornaSIL:
Would it be best to split this into several pull requests? It would be easier to process since there are different issues for each keyboard. I see several misspelled welcome.htm files (wecome.htm)
Also, I thought the following files did not belong in the repo? Perhaps I'm remembering wrong.
gff_el_naija_7/extra/1.0/naijanfd10.js
gff_el_naija_7/extra/1.0/naijanfd10_load.js
khmer_angkor/source/khmer_angkor-1.0.js
khmer_angkor/source/khmer_angkor-1.0_load.js

Should the README.md file be updated with actual information on that project?
Should the LICENSE.md files be updated with copyright date and copyright holder?
Do the HISTORY.md files have anything real in them?

I'm concerned that if we just pull in the template, rather than adding useful information, that we'll not have solid information on history, copyright holders, links to web pages, BCP-47 codes, etc.

If we can split the PR into the separate keyboards it would be easier to do them one by one.

andjc:
The naming convention is wrong.  The projects are using the prefix ```gff_el_``` it should be just ```el_```. All the existing layouts have the digit seven as a suffix, ie ```gff_el_naija_7/```. I assume this refers to the version of Keyman is was originally developed for. When the keyboards are upgraded, will they new projects/layouts have to be started?  Or would it be better to just name them as ```el_naija/``` etc?

The Burmese layout was not indented to be migrated across.

MakaraSok:
@LornaSIL @andjc Thanks a for the comments/suggestions/corrections. 
1. The file/folder name will be changed to `el_<language-name>`
2. @andjc You may know the detail for README and HISTORY of each keyboard, could you please help updating those once the PR for each keyboard is up?
3. One pull request will be created for each keyboard. I don't know how to split the PR, but I'll find a way to make sure we have one PR for one keyboard.

Is there anything else missing here?

MakaraSok:
@andjc the content of this PR is now solely for 'Dari Keyboard'. The rest will come later in separate PRs.

MakaraSok:
There is something wrong with this PR, as @LornaSIL pointed out, there are irrelevant commits and files, thus this PR is closed. For the Dari Keyboard, please go to PR #77.

PR #71, opened 2017-08-01
[el_osmanya] Osmanya Keyboard

MakaraSok:


darcywong00:
I've created the kps file for this keyboard

MakaraSok:
Osmanya numeral doesn't show up correctly. It needs to be double-checked.

LornaSIL:
The keyboard seems to work. 
* One issue is that keyboard_layout.png must be included in the .kps file so that welcome.htm displays properly.
* The other issue is one you mentioned above. The digits are not displaying in the OSK, nor in the keyboard_layout.png. Keyman itself seems to have a problem with the digits. It's displaying U+104A0 as U+D801 U+DCA0 and because the font doesn't have those codepoints they become square boxes. The character viewer also displays the digits the same way, but the other Osmanya characters are displayed properly. So, something to fix in Keyman? Every time I clicked on the digit in the OSK Keyman would crash. It did this in both Keyman 9 and 10. I'm using Noto Sans Osmanya which has the digits in the font.
image

LornaSIL:
The keyboard itself works, I just can't edit the OSK without a crash. So, I've modified the graphics so that the digits are showing on the keyboard layout, the welcome.htm and the .kps and I'll attach those here. They should be replacements for what is in Makara's PR.
* welcome.htm
* keyboard_layout.png
* el_osmanya.kps

The OSK still shows boxes for the digits, but it does output the correct characters.

[el_osmanya.zip](https://github.com/keymanapp/keyboards/files/1318608/el_osmanya.zip)

LornaSIL:
@MakaraSok if you could make the changes requested above (Sep 20) I think this one would be ready to accept.

MakaraSok:
Done, thou the issue of not having numerals shown up correctly is still here.

@DavidLRowe or @LornaSIL please approve this PR and then label is as 'ready to merge' :)

mcdurdin:
Versioning questions need to be resolved before we merge this keyboard. There may be an existing newer binary version.

release/el/el_osmanya/el_osmanya.keyboard_info
```diff
@@ -0,0 +1,10 @@
+{
+    "license": "mit",
+    "languages": {
+        "som": {}
```
darcywong00: Going from ISO 639-3 to BCP 47 tags, the language tag will need to change per the CI server
```
'som' is a valid tag but is not canonical: 'som' should be 'so'
```

release/el/el_osmanya/el_osmanya.keyboard_info
```diff
@@ -0,0 +1,10 @@
+{
+    "license": "mit",
+    "languages": {
+        "so": {}
+    },
+	"description": "<p>A unicode based Osmanya keyboard layout using the Somali Latin orthography as the input mechanism.</p>",
+    "related": {
+      "osmanya10": { "deprecates": true }
```
darcywong00: Should also make a file `legacy/o/osmanya10/DEPRECATED.md` that has the line
```
This keyboard has been deprecated and replaced by release/el/el_osmanya.
```

PR #72, opened 2017-08-01
[el_pan_sahelian] Pan-Sahelian Keyboard

MakaraSok:


darcywong00:
I've added the kps file for this keyboard

MakaraSok:
OK, but note that the ISO Code and Language Name  fields are empty.

mcdurdin:
Versioning questions need to be resolved before we merge this keyboard. There may be an existing newer binary version.

PR #76, opened 2017-08-01
Yolngu Keyboard

MakaraSok:


LornaSIL:
The graphic needs to be included in the .kps file (in addition to being called from the welcome.htm file). As it is now, when I install the package the welcome.htm just shows a blank box where the image should be.

I would like to see the keyboard name just be "Yolngu". That would mean removing "Unicode" and the version number from the keyboard name in the .kmn, .kpj and .kps files. That is consistent with the existing online keyboard (http://www.tavultesoft.com/keyman/downloads/keyboards/details.php?KeyboardID=372). The README.md and HISTORY.md files call the keyboard "Yolngu Unicode" (without the version number) and those two files need the word "Unicode" to be removed.

mcdurdin:
Available for download from http://keymankeyboards.com/372

PR #77, opened 2017-08-02
[el_dari] Dari Keyboard

MakaraSok:


LornaSIL:
@andjc we are missing some significant information on this keyboard.

* Can you tell us the copyright years? That is, what year was this keyboard created?
* Can you tell us what version the keyboard should be?

As far as I can tell it was a webbased keyboard and there was no .kmp package so I can't find the information that I might have found in the .kmp.

LornaSIL:
It would definitely be worth our associating this Arabic script keyboard with an Arabic script Windows language so that it inherits RTL behavior.

MakaraSok:
@LornaSIL , I was not able to find metadata for this keyboard. It doesn't exist in the legacy collection.

PR #87, opened 2017-08-08
[gff_sebatbeit] Sebatbeit Keyboard

MakaraSok:
(update branch name)


LornaSIL:
Rename sebatbeit.bmp to gff_sebatbeit.bmp
Change gff_sebatbeit.kmn to use gff_sebatbeit.bmp
Change gff_sebatbeit.kpj to use gff_sebatbeit.bmp

In the Sebatbeit Typing Table the following things need changing:
* qye, qye, qye, qye > qyi, qya, qyie, qy (the glyphs/fidels are correct, the keystrokes are wrong)
* in the "be" row, change "mue" to "bui" (the glyph is correct)
* in the "ae" row the glyph for the "u" keystroke is wrong.
* change "fuu" to "fui" and the second "fue" to "fuu"
* change "puu" to "pui" and the second "pue" to "puu"

I would still like the SebatbeitTyping-English.pdf document to be updated to reflect single keyboard distribution.

I would like the welcome.htm to be updated with the correct Abyssinica SIL url: https://software.sil.org/abyssinica/. The font filename has changed to "AbyssinicaSIL-R.ttf" but it will likely change again to "AbyssinicaSIL-Regular.ttf so I would recommend against including the filename in the documentation. The fontname will remain the same.

MakaraSok:
@LornaSIL Do we have the office version of the PDF file? I don't see an editable file of the document (SebatbeitTyping-English.pdf).

LornaSIL:
I think @dyacob is working on getting all the gff keyboards into better shape with source files. I suspect it's taking some time to split the packages into individual keyboards.

dyacob:
@LornaSIL and @MakaraSok , I've put updating gff keyboards on hold until https://github.com/MakaraSok/keyboards/pull/10 (for gff_geez) makes it through the review process and into the keyman/keyboards main repository.  This will serve as a baseline where I can then apply updates to the remainder of the group as needed and also minimize the review burden.

DavidLRowe:
@dyacob will submit a new PR for this keyboard

PR #89, opened 2017-08-08
Blin Keyboard

MakaraSok:
(update branch name)


dyacob:
This PR would be ready following the processing of the dependent https://github.com/MakaraSok/keyboards/pull/3

MakaraSok:
@dyacob I have played with this keyboards a bit and it looks like the keystrokes for punctuation and numbers do not behave as expected. For example, double semi colon does not give ፤, but semi colon itself. 
image

Could you please double check this? I am happy to help with updating the BlinTyping-English.pdf if you confirm that the instruction is outdated.

MakaraSok:
@dyacob Thanks a lot for your work on updating the documentation. The PR has been merged, but I have just double checked the keystrokes, and I found some keystrokes have the actual outputs different from what is mentioned in the documentation. See the table below:

keystrokes | output in the doc. | actual output
----------- | ------------------ | --------------
u | ኣ | ኡ
ea | ኣ | እኣ
;; | ፤ | ;
,,, | ፥ | ፡
:+ |  ፠ | ፡+
:# | ፨ | ፡#

Everything else is OK.

dyacob:
Hi @MakaraSok , thanks for the merge!  I think something is amiss, two things possibly.  Are you referring to the welcome.htm document or BlinTyping-English.pdf, or both when you mention the documentation? In both of these I don't see a mapping mentioned for "ea" and ",,,".

I am using Keyman Developer and Desktop 10.0.933.0, and with it the keystrokes are correct, I do not get the error column in your table.  I assume you rebuilt the .kmn file?

I'll update my local Keyman and check that the output is as expected.

MakaraSok:
@dyacob The Keyman Developer I'm using is V.10.0.961.0 (the latest one). The third and the last two keystrokes outputs as the outputs stated in the BlinTyping-English.pdf
> keystrokes | output in the doc. | actual output
----------- | ------------------ | --------------
u | ኣ | ኡ
ea | ኣ | እኣ
;; | ፤ | ;
,,, | ፥ | ፡
:+ |  ፠ | ፡+
:# | ፨ | ፡#

After checking out the branch and re-compile this keyboard, I still encounter three issues. Please see them below.

keystrokes | output in the doc. | actual output | description
----------- | ------------------ | -------------- | -----------
u | ኣ | ኡ | the output is different from what is stated in the PDF. Which one should be the correct output of "u"? If it's ኡ, let's update the PDF.
ea | ኣ | እኣ | This እኣ appears when typing "ea".
,,, | ፥ | , | typing this "," three times and the output is a comma, not the ፥.

dyacob:
@MakaraSok thanks for all the checking, I have V.10.0.961.0 installed now and can reproduce the issues, and see the errors in the doc.  I should have updates in this evening for Blin and Ge'ez.

dyacob:
Corrections to the documentation have been pushed in.

MakaraSok:
@DavidLRowe @LornaSIL Please take a look at the update and if nothing is a concern to you, let's label this as 'ready-to-merge'.

LornaSIL:
@dyacob 

The BlinTyping-English.pdf seems correct.

The welcome.htm file has some issues:
* The "Special Note" talks about the apostrophe, but the example doesn't actually work. It turns into the 1st form, not the 6th form of the glottal series.
* The "Q" keystroke (left column) has an upside down "R" instead of "Q"

@MakaraSok 
I don't know how the keyboard_info file was generated, but it is calling for two font files that are not part of the package. 
* wookianos.ttf
* ETHIOPI0.eot

I assume these should either be deleted from the keyboard_info file or added to the repo.

I would like the newer version of Abyssinica SIL to be distributed rather than this version. The font can be downloaded here: https://software.sil.org/downloads/r/abyssinica/AbyssinicaSIL-1.500.zip

dyacob:
@LornaSIL Thanks for the review, I had caught the upside down "R" over the weekend also.  I've updated the welcome file for a few minor changes.

On the "Special Note" issue, when you say "**_it_** turns into the 1st form..." what is the "it" ?  I think you mean l' is becoming ለ instead of ል, but I'm not able to reproduce this so seeking clarification.

I'm not familiar with the keyboard_info file or how it gets generated.  I wouldn't think were be in references to fonts in this package.

MakaraSok:
@LornaSIL 

> I don't know how the keyboard_info file was generated, but it is calling for two font files that are not part of the package.
wookianos.ttf
ETHIOPI0.eot

The metadata included in the .keyboard_info was drawn from the existing gff-byn-7 keyboard in the 'legacy' folder. @dyacob are these fonts needed at all? If not, we may want to remove them from the .keyboard_info.

LornaSIL:
@dyacob Typing ```mel'ak``` produces መልኣክ instead of መልእክ

dyacob:
@LornaSIL I see, I think is both an issue with a copy and paste from the Ge'ez welcome.htm into Blin that I forgot to update (መልአክ => መልኣክ), and an issue with letter clarity at small sizes (አ vs እ).

A merge from this morning fixes the word form in the welcom.htm.

dyacob:
With Keyman Developer 10 I removed included fonts from the project file (.kps I believe) and have submitted the change in https://github.com/MakaraSok/keyboards/pull/23 .  My understanding is that the package should be keyboard only and fonts would be included in a "Power Pack" distribution.

I couldn't see how to safely edit the gff_blin.keyboard_info file with Keyman so I've left it untouched.  It appears to have data that keyman web would use.  I believe it would safe to remove the file for now -I'll leave this to @LornaSIL or @MakaraSok .

As a side note, I saw that the keyboard_info file has references to the wookianos font which in turn has some Unicode encoding issues, limited letter coverage for Blin (and related), and undesirable glyph shapes for Blin users.  All the more reason to remove it.  In parallel I'm working on updates to wookianos and the font collection that it is part of.  The updates would be ready when we turn to updating the Power Packs.

MakaraSok:
> With Keyman Developer 10 I removed included fonts from the project file (.kps I believe) and have submitted the change in MakaraSok#23 . My understanding is that the package should be keyboard only and fonts would be included in a "Power Pack" distribution.

I've talked to @mcdurdin about this and it turns out we have to have at least one font in the .kps so that people can use the keyboard right away, otherwise the keyboard would produce square boxes. With that said, let's bring `AbyssinicaSIL-R.ttf` and `FreeSerif.otf` back in. 

Regarding the "Power Pack", you may include as many fonts as you'd like.

> I couldn't see how to safely edit the gff_blin.keyboard_info file with Keyman so I've left it untouched. It appears to have data that keyman web would use. I believe it would safe to remove the file for now -I'll leave this to @LornaSIL or @MakaraSok . When this is sorted out I can make the corresponding updates needed for the Ge'ez and Tigrinya-ER keyboards.

I'll remove the two fonts (i.e. `wookianos.ttf` and `ETHIOPI0.eot`) from the .keyboard_info since, as you said, they seem outdated.

dyacob:
@MakaraSok thanks for the explanations, I'll apply keyboard_info changes to the other gff keyboards accordingly.

MakaraSok:
Sorry @LornaSIL but some changes have been made and that cuased your review to be dismissed. Could you please review this PR again. Thanks!

MakaraSok:
@LornaSIL Please re-review this too.

mcdurdin:
Published keyboard:
 - Home: https://keyman.com/keyboard/gff_blin
 - Legacy: http://keymankeyboards.com/?id=728
 - Documentation: https://help.keyman.com/keyboard/gff_blin/

PR #90, opened 2017-08-08
Ge'ez Keyboard

MakaraSok:
(update branch name)


dyacob:
This PR would be ready following the processing of the dependent PR: https://github.com/MakaraSok/keyboards/pull/1

LornaSIL:
This keyboard seems about ready to publish. I have a few small quibbles.

In the doc/pdf documentation 
* it took me 5 x ":" to get the regular colon (:::::) - this also needs fixing in welcome.htm (or fixing in the keyboard)
* for the comma and semicolon, the keystroke and result are turned around, the keystrokes should be on the bottom row.

That's all I see that needs fixing.

dyacob:
@MakaraSok I may not be able to resume work for another week still as I'm chasing another project deadline; that and my Windows box can no longer connect to the internet after an update, so that slows me down.

I looked at the manual and will want to regenerate it, the typeface is not very nice.   SIL Abyssinica font is the target font. The Latin input for # and ' I believe depends on input context and will not always work with a single keystroke.  I'll review this also.   When the usual single keystroke can be used for a keyboard symbol (like @ % / [ ] + - = etc), I keep it out of the manual.

Foreign punctuation in a Ge'ez language document is a little contrived anyway, I'm considering removing the tabular part and just have a general note to use repeated key strikes to arrive at the keyboard native punctuation. I believe this simpler and cleaner.

MakaraSok:
@dyacob There are two things here to check again. The first issue is similar to the first one in #89, but the second is a brand new issue. It may have happened because of the modification of the keyboard.

keystrokes | output in the doc. | actual output
----------- | ------------------ | --------------
u | ኣ | ኡ
_ | N/A | [1᎐2᎓3᎒4᎙5᎑6᎔7᎖8᎕9᎗0᎘_፟]

Please double check this. Thank you, Daniel!

dyacob:
Documentation updates have been made for the Ge'ez keyboard.  The menu output with _ is explained so it should not alarm the user.  The Word file has been updated and saved with Word 2016 for the first time where 2010 had been used previously.  The change to 2016 broke layout in a number of places.  I believe that I have fixed all the layout issues but some may still linger.

MakaraSok:
@DavidLRowe @LornaSIL Similar to the gff_blin, please take a look at the updates and if nothing is a concern to you, let's label this as 'ready-to-merge'.

LornaSIL:
@MakaraSok 

The keyboard_info file is calling for these fonts. Please remove them:
* wookianos.ttf
* ETHIOPI0.eot

The README.md keyboard name should be changed to: ግዕዝ (Ge'ez)
The HISTORY.md keyboard name should be changed to: ግዕዝ (Ge'ez)

Otherwise the keyboard will be ready to merge. One interesting thing I learned in checking this keyboard is that Notepad on Win10 does not like the ```SS``` keystroke. I get a popup that says "Do you want to open Game bar? Yes, this is a game." I can't use the ```SS``` keystroke in Notepad, but it works elsewhere (like in Word). And the author has given us another way to access those keys using ```D``` so that is acceptable.

dyacob:
I've made the changes in https://github.com/MakaraSok/keyboards/pull/24

Strange issue with notepad, I just tried it with Win10 and Keyman 10.0.961.0 and no issue with SS, it works as expected.  I'm not familiar with the Game bar, I uninstall all the games that come with Windows, so maybe that helped me avoid it.

MakaraSok:
@dyacob thanks a lot for the changes. 👍 

@LornaSIL Like Daniel, I didn't experience the problem you mentioned in Notepad on Win10.

MakaraSok:
@LornaSIL We've made a little change to the keyboard. Abyssinica SIL and FreeSerif fonts were removed, and now they are restored to where they've been. Sorry for the ad-hoc.

MakaraSok:
@LornaSIL Please re-review this PR as well.

mcdurdin:
Published keyboard:
 - Home: https://keyman.com/keyboard/gff_geez
 - Legacy: http://keymankeyboards.com/?id=729
 - Documentation: https://help.keyman.com/keyboard/gff_geez/

PR #94, opened 2017-08-08
Khmer Angkor Keyboard

MakaraSok:


MakaraSok:
RALT key combos to output rarely used characters haven't yet been in place. When the RALT is pressed, the keyboard may trigger an unintended behavior.

For the time being, Ctrl+ALT key combo is a substitute of RALT.

MakaraSok:
@LornaSIL it is working OK for me. Could you please try again using Chrome browser? After compiling the keyboard and click "Open debugger in local browser", it will take you to your default browser (in this case, Chrome) and then please press F12 and then F5.

MakaraSok:
@LornaSIL When it's changed to '10.0', the keyboard fails to compile. I think we may not want to change it just yet. @mcdurdin Should we worry about this?

mcdurdin:
The `'9.0'` value refers to the minimum version of Keyman Desktop that can be used with the keyboard, not the version of Keyman Developer to use. At this point, all of us should be trying to use Keyman Developer 10, because it has many bug fixes and improvements over version 9.0, even though it is still in alpha.

All keyboards created with Keyman Developer 10 are still backwardly compatible with Keyman Desktop 9, so the `&VERSION` value should still be `'9.0'`.

MakaraSok:
nts
- [ ] The keycaps of '១២៣' 'កខគ' and '᧠᧡᧢' need to be updated to look like the keycaps of other function keys.

- [ ]  The backspace on the touch layout doesn't delete the subscript. For instance, when pressing backspace at the end of this character sequence ឆ្ក, it only deletes ក leaving ឆ្; while it should leave only ឆ.

- [ ]  Finalize documentation in both EN and KH (i.e. get clear screenshot of the keyboard on mobile) and include the latest update of them in the package.

LornaSIL:
I think the only change that is needed now is that utf-8 isn't specified in welcome.htm and the Khmer isn't displaying properly. If you add a line like this to the htm file it displays properly"
```
  <meta http-equiv="Content-Type" content="text/html; charset=utf-8">

```

MakaraSok:
@LornaSIL The welcome.htp was updated as suggested.

mcdurdin:
The package fails to install ... with a very curious error:

image

The source files all look fine with a visual inspection. However, when I opened kmp.json (generated with latest alpha of Keyman Developer, as a member file of the kmp file), I saw the following:
````js
    {
      "name" : "KAK\u200B Documentation EN.pdf",
      "description" : "File KAK\u200B Documentation EN.pdf"
    },
````

`\u200B` is [Zero Width Space](http://www.fileformat.info/info/unicode/char/200B/index.htm).

The kmp.inf file, which is the metadata file inside the .kmp, for legacy reasons, does not support Unicode characters in the filenames it references. So the ZWSP gets converted into a `?`. Oops.

We need to rename both of the pdf files to remove the ZWSP and fixup the references in the .kps file.

I'm not sure how this was missed in testing. @LornaSIL did you experience any issues installing the .kmp file?

LornaSIL:
It looks like I may have just installed the .kmx file without testing the kmp. I know I separately tested the web keyboard, Android keyboard, read all the text files and .htm files, etc. I'm getting a similar message with the .kmp.

MakaraSok:
Just updated the filenames and the screenshots of the mobile layout in the documentation. @LornaSIL could you please try the .kpm again and then re-review this.

MakaraSok:
@LornaSIL Could you please re-review this because the last revision of the keyboard has dismissed your review?

mcdurdin:
NOTE! This keyboard currently requires Keyman 10.

Published keyboard:
 - Web: https://keymanweb.com/?tier=alpha#khm,Keyboard_khmer_angkor
 - Home: https://keyman.com/keyboard/khmer_angkor
 - Legacy: http://keymankeyboards.com/?id=724
 - Documentation: https://help.keyman.com/keyboard/khmer_angkor/

PR #96, opened 2017-08-15
Easy_chakma - replace RibengUni font with updated version

mcdurdin:


DavidLRowe:
- [ ] Still have some license issues:
* .kmn has GPL
* user manual includes restrictions on copying; is this compatible with MIT license?
* LICENSE.md has GPL

- [ ] Copyright date in .kmn and README.md is 2015, but HISTORY.md has dates in 2016 and 2017. 

- [ ] Targets: README.md lists windows, web, mobile; .kmn has entire list (including 'any'), but there is no `-layout.js` file. Suggest changing to windows only.

- [ ] @mcdurdin How critical are English typographical errors in .kmn comments and user manual?

DavidLRowe:
Also user manual lists version 1.15, and includes a lot of information about Keyman itself (some of it out-of-date).

MakaraSok:
Please kindly review this PR again, since it has just been rebased to the master.

MakaraSok:
@DavidLRowe I have updated the keyboard based on your comments, please check it again. A couple of notes though:
- Easy Chakma OSK shows boxes on each key, is it because I don't have the font installed?
- I am not sure about the inclusion of the restriction on copying in the User Manual.
- The manual is more about how to set up Keyman Desktop 9. Should we have it with our latest version?

DavidLRowe:
- .kmn includes GPL license (rather than MIT) in opening comments. Do we have permission to release this keyboard and supporting files under MIT license?
 - .kmn has "any" as well as all target options, but there's no -layout.js file in project. Is that OK?
 - User manual (.pdf file) has copying restrictions -- is that compatible with MIT license?
 - User manual has references to Keyman 9 installation that will soon be out-of-date. Instructions for actual keyboard use suggest invoking the on-screen keyboard.

MakaraSok:
@mcdurdin Could you please take a quick look at @DavidLRowe's comments?

mcdurdin:
> @mcdurdin How critical are English typographical errors in .kmn comments and user manual?

Not super important. I don't think we need to block acceptance on them unless they are seriously problematic.

> .kmn includes GPL license (rather than MIT) in opening comments. Do we have permission to release this keyboard and supporting files under MIT license?

Yes, we have written permission so we should update the .kmn file accordingly.

> .kmn has "any" as well as all target options, but there's no -layout.js file in project. Is that OK?

It's not great but the keyboard will be usable on touch devices, just not optimised.

> User manual (.pdf file) has copying restrictions -- is that compatible with MIT license?

No, that's not compatible with MIT. We'll need to get that updated.

> User manual has references to Keyman 9 installation that will soon be out-of-date. Instructions for actual keyboard use suggest invoking the on-screen keyboard.

Yes, should consider removing version-specific details for manual.

DavidLRowe:
@mcdurdin Who will change the .kmn license info and update the user manual?

MakaraSok:
@mcdurdin @DavidLRowe The user manual and the .kmn have been updated. ptal

DavidLRowe:
@mcdurdin The welcome.htm file has `<meta http-equiv=Content-Type content="text/html; charset=windows-1252">` but contains characters that aren't in Windows 1252. Is this a problem? It seems to display fine (though I wouldn't know if the Chakma characters are rendering correctly!), but gives a warning if I try to open it with a text editor.

When I tried the mobile layout (as hosted by Keyman Developer 10.0.968.0) on an iphone 4s, the keycaps displayed with Chakma characters, but all I got was the QWERTY keyboard output. Same for the Open debugger in local browser. When I installed the compiled .kmp with Keyman Configuration, I can select the Chakma keyboard (and I get the green icon), but the keyboard only produces QWERTY keyboard characters. I'm not sure if the problem is with the keyboard files or with my test setup.

mcdurdin:
> @mcdurdin The welcome.htm file has <meta http-equiv=Content-Type content="text/html; charset=windows-1252"> but contains characters that aren't in Windows 1252. Is this a problem?

@DavidLRowe It is likely that the file format is being sniffed by the browser and the meta tag is being ignored. We do need to fix this; the file should be in UTF-8 for consistent cross-platform results.

@MakaraSok if you can repro these issues on your computer, we can take a look together at root causes.

MakaraSok:
@mcdurdin @DavidLRowe 

'utf-8' has been added to the charset.

The problem of having the touch keyboard output QWERT characters (i.e. English chars) may be because we haven't set up touch layouts yet.

DavidLRowe:
Not sure what is in the welcome.htm file, but just changing the charset value wasn't the solution! I'm now seeing lots of U+FFFD displayed, including for the copyright symbol in the first paragraph. Perhaps we should revert the last change and leave the erroneous 1252 ref? If not the whole file needs to be reviewed and offending characters converted to UTF-8.

And that copyright statement still has: "No part of this book may be reproduced by photocopying, typing, printing or using any other device to serve any commercial purpose without the prior written permission..." which I don't think is compatible with MIT license.

What would you think of changing the &TARGETS to just Windows (that is, drop support for web and touch, which don't seem to be working)?

mcdurdin:
Having looked more closely at welcome.htm, it's a Word-generated document which is hugely bloated. I suggest we run it through something to clean Word's mess up and hopefully sort out the character encoding at the same time. There are a number of tools available to cleanup Word HTML but I am not sure which is most useful (even Word has a "Save as HTML (Clean)" option).

DavidLRowe:
Tidy won't work on welcome.htm because the encoding is messed up (mixture of CP1252 with other stuff). I can't open it in gedit for the same reason. Not sure where to go. Maybe retry with the version before the charset=utf-8 change?

DavidLRowe:
@mcdurdin @MakaraSok What's the status of this keyboard?

- welcome.htm 
  - wrong encoding
  - bloated
  - license issues (still has restrictions on copying that are not compatible with MIT license)
- mobile
  - no specific mobile layout
  - test default (maybe I got no results because I didn't have a suitable font??)

MakaraSok:
According the title, it seems this PR only addresses the font issue. If there are these many issue, I am not quite sure about it.

Probably we can close this PR, because there is a new PR on this keyboard as well.

@bivuti Could you please confirm if this PR #96 is outdated and we should work on PR #167 instead?

Bivuti:
@MakaraSok you are right! Yes, The PR #96  is outdated. We should work on PR #167 for the latest updated the Chakma Keyboard and font.

MakaraSok:
OK. Let's close this PR then.

PR #63, opened 2017-07-20
Kayan keyboard

purple-monkey:
This is a guess at what I'm supposed to upload to get a keyboard online.

PR #97, opened 2017-08-16
Kayan Keyboard - Fixed touch keyboard

purple-monkey:
A number of keys on the touch keyboard triggered rules that were made with the desktop keyboard in mind. I decided to make new names for the touch keyboard keys when they differed from the desktop keyboard keys to avoid triggering the desktop keyboard rules.


purple-monkey:
Sorry. I meant to update the Kayan Keyboard #63 pull request.

PR #105, opened 2017-09-15
sil_el_1-2-2

rik-shaw:
kmn:
- added U+0294 ʔ
- added store(&LAYOUT) 'default'

.odt:
- corrected styles

.htm:
- re-generated from .odt and tidy'd


rik-shaw:
@mcdurdin @DavidLRowe I had done a "git pull upstream master" before creating my new branch, so not sure how I got the residual files in there.  Maybe I did something wrong somehow....

Regarding the `store(&LAYOUT) 'default'` this is absolutely required for ibus-kmfl (Linux), which is my biggest target.  We spent a long time debugging why a US underlying keyboard was assumed before we added that line, which will correctly set the underlying keyboard to match the system keyboard (e.g. when a German keyboard is used, without that addition the y and z will be "US style" on the keyman keyboards otherwise).  Again this may be "Linux only" but I would like the keyboard to not have to be different between the various platforms....

Regarding the version number, as I only added one point I didn't really feel a big version bump was needed, but I could move it to 1.3 instead if necessary.

Regarding the copyright years, I could add that in: I was avoiding it so that I didn't have to change it all the time, but this would be fine.  So I'll add it in after I get a bit of clarification on the "default layout", version number, and residual file cleanup.

mcdurdin:
In Keyman, we deal with the base layout changing with the [`store(&mnemoniclayout)` header store](https://help.keyman.com/developer/language/reference/mnemoniclayout). So that clearly is somewhere where KMFL has diverged which I was not aware of. 

Keyman does not support `store(&layout) 'default'` so you could use [compile targets](https://help.keyman.com/developer/language/guide/compile-targets) to avoid the problem, although I don't know if KMFL supports them at present. If KMFL doesn't support compile targets then the only real option is 2 source files.

We're not quite ready to start tackling the question of Keyman and KMFL compatibility yet. It's a bit of a shame when platform divergence happens but we'll work on straightening those kinks out next year I think. For now, we'll just have to work around the challenges in keyboard source files.

rik-shaw:
@mcdurdin I already have a single difference between the kmfl and the keyman .kmn files: namely that kmfl hasn't been updated to support version 9 (so I have to specify a version of 7 for kmfl).  So I will just leave out this line for the master repo .kmn.  Our location specific wasta-custom-eth packages distributes the .kmn files so I have full control to add it there.

I'll wait for your clarification on the other issues.  Thanks again for everything!  Hopefully my "problems" help others in the future :-)

mcdurdin:
@rik-shaw In terms of the version number, so long as it fits the dotted number format (a.b.c.d) with up to 4 components, that'll work well for us. A third component is entirely appropriate for a minor change.

In terms of the copyright, the SIL experts in the area have recommended always including a date range - i.e. initial version through to last modified. That can be ambiguous at times but we just do the best we can.

When it comes to time to look at KMFL, I'll be keen to pick your brain on your experiences with keyboard distribution with KMFL and how we can best work the deployment of keyboards into the model.

LornaSIL:
@rik-shaw There are a few rules that didn‘t work for me. 
* The ```;H``` rule produces ```Ĩ``` and not ```Ħ```.
* The 3rd rota for ```"``` and ```'``` just toggles back to the first result, it doesn't give the 3rd result.
Otherwise the keyboard works and the files seem correct to me.

MakaraSok:
@LornaSIL I fixed the two issues.
@rik-shaw Please double check and see if the fixes are what you expect the keyboard to behave.

Besides the two issues, all keystrokes work as expected.

MakaraSok:
One thing is that the date should be included in the copyright lines.

LornaSIL:
@MakaraSok Please add a copyright date in the .kmn, .kps, welcome.htm, readme.htm.

The .odt, .pdf and welcome.htm do not have a copyright year after the (c) symbol, but they do have a date at the end of the line. It might be nice to fix those up too if possible.

The keyboard changes I requested are working well now.

MakaraSok:
@LornaSIL Done. PTAL.

rik-shaw:
@mcdurdin @LornaSIL @MakaraSok THANKS for your following up on this and getting it merged.  In maintaining the "parallel" version for Linux I think I unintentionally re-introduced the "bug" that Lorna found: "The ;H rule produces Ĩ and not Ħ"  But I didn't have time to come back here and make a comment or fix it up.  So thank you again for your extra help.

I have a request from a colleague here to add diacritics for the barred I and barred U characters, so will work on that on the Linux side and then *hopefully* submit a much cleaner pull request the next time around :-)

I am still not sure why many extra files were pulled in on my original pull request.  My process has been / is:
<working on my own private fork>
git fetch upstream (to sync with the Keyman master here before starting any changes)
git checkout -b <name> to get a new branch going.
....
git push origin <branch> to upload my branch to my fork.
Use github web interface to create the pull request to Keyman.

Anything amiss with that process?

rik-shaw:
@mcdurdin @LornaSIL @MakaraSok THANKS for your following up on this and getting it merged.  In maintaining the "parallel" version for Linux I think I unintentionally re-introduced the "bug" that Lorna found: "The ;H rule produces Ĩ and not Ħ"  But I didn't have time to come back here and make a comment or fix it up.  So thank you again for your extra help.

I have a request from a colleague here to add diacritics for the barred I and barred U characters, so will work on that on the Linux side and then *hopefully* submit a much cleaner pull request the next time around :-)

I am still not sure why many extra files were pulled in on my original pull request.  My process has been / is:
<working on my own private fork>
git fetch upstream (to sync with the Keyman master here before starting any changes)
git checkout -b <name> to get a new branch going.
....
git push origin <branch> to upload my branch to my fork.
Use github web interface to create the pull request to Keyman.

Anything amiss with that process?

LornaSIL:
I have found that when I do a lot of work and some time passes before I commit, I need to fetch again so that I keep my system in sync with upstream. Otherwise I end up doing a commit with old master files. It's also always useful to do a git status to see what I'm actually going to commit to make sure it's not committing files unintentionally. I think we've all done that at times.

mcdurdin:
Published keyboard:
 - Home: https://keyman.com/keyboards/sil_el_ethiopian_latin
 - Legacy: http://keymankeyboards.com/?id=713
 - Documentation: https://help.keyman.com/keyboard/sil_el_ethiopian_latin/

mcdurdin:
Regarding sync, you need to do the following steps to keep both your own repos (local and github) in sync:
```
git checkout master
git fetch upstream
git merge upstream/master
git push origin master
```

PR #106, opened 2017-09-15
Balochi Scientific Initial version

rmlockwood:


LornaSIL:
This PR included kayan and balochi_phonetic folders. Those will need removing from the PR.

The .kpj file has this record which should be removed 

```
    <File>
      <ID>id_bb799b27b00f70be9fe2c4578b57a866</ID>
      <Filename>Untitled.kmn</Filename>
      <Filepath>Untitled.kmn</Filepath>
      <FileVersion>1.0</FileVersion>
      <FileType>.kmn</FileType>
      <Details/>
    </File>
```

The .kvk file doesn't have any characters in it.
The .kmn file doesn't doesn't include the .kvk file yet.

The characters for all of the quote (') key in the keyboard don't match the documentation. For instance 'a is supposed to produce ā according to the documentation, but the keyboard itself gives the output as æ. I haven't yet tried the desktop keyboard out, but the web keyboard did indeed produce æ. All the non-quote keystrokes were correct on the web keyboard.

I'll do more testing Monday.

MakaraSok:
@rmlockwood @LornaSIL I  have updated the following in this keyboard:

- added .kvks
- added touch layouts
- updated welcome.htm and README.md
- updated balochi_scientific.pdf to match the specification in the .kmn
- updated .kps

But one problem is that I do not have permission to push to this **master** branch on rmlockwood/keyboards. Thus, I created a PR for it. Please take a look at it here #125, if everything looks good to you there, we may want to close this PR #106.

PR #108, opened 2017-09-28
Areare initial version

LornaSIL:
New keyboard.


MakaraSok:
We do not have to two separate layouts for phone and tablet. Since the touch layout for phone has been neatly done, we will keep it; but the one for tablet is a bit off, thus it has been removed.

Metadata has also been updated.

Please take a look and see if everything lgty, it should be labeled as 'ready to merge'.

release/sil/sil_areare/sil_areare.kpj
```diff
@@ -7,15 +7,15 @@
...
     <File>
-      <ID>id_5a3a81088f3ad10f84e3ef28e3a9dfea</ID>
+      <ID>id_5824f6df49f46571f19d55e2c53f7e65</ID>
       <Filename>sil_areare.kmn</Filename>
       <Filepath>source\sil_areare.kmn</Filepath>
       <FileVersion>1.0</FileVersion>
       <FileType>.kmn</FileType>
       <Details>
-        <Name>ɁAreɂare</Name>
+        <Name>?Are?are</Name>
```
LornaSIL: I just have one change in the sil_areare.kpj file the name has somehow changed to using question marks. It should be: 
```
<Name>ɁAreɂare</Name>
```

MakaraSok: It looks like you have changed from what is correct (**ɁAreɂare**, with glottal stops) to what is incorrect (**?Are?are**, with question marks) here.

LornaSIL:
@MakaraSok I changed the keyboard name to use the glottals. I looked at the phone touch layout and that is in good shape so I think it's fine.

mcdurdin:
Published keyboard:
 - Web: https://keymanweb.com/#alu,Keyboard_sil_areare
 - Home: https://keyman.com/keyboard/sil_areare
 - Legacy: http://keymankeyboards.com/?id=722
 - Documentation: https://help.keyman.com/keyboard/sil_areare/

PR #120, opened 2017-10-16
Kayan Keyboard - added semicolon back

purple-monkey:
We had changed the semicolon key to have the value of U+02CA to fix a diacritic issue. Then we were told to avoid that value by the NRSI team. Then we learned that the semicolon is used in the Kayan orthography. So we added the semicolon back.


DavidLRowe:
The changes of this PR replaced by #178

PR #121, opened 2017-10-18
Keyboards CI - use 7z instead of 7za (already on build agents)

mcdurdin:


ci.sh
```diff
@@ -12,6 +12,7 @@ KEYBOARDROOT="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 JQ="$KEYBOARDROOT/tools/jq-win64.exe"
 RSYNC="$KEYBOARDROOT/tools/rsync.exe"
 CI_CACHE="$KEYBOARDROOT/.cache"
+APP7Z="/c/Program Files/7-Zip/7z.exe"
```
darcywong00: What about using the environment variable `7Z_HOME` as a build agent requirement? That's what the other build configs do.

mcdurdin: Good idea. One issue, we can't really use an environment variable starting with a digit...

I'll compromise for now and use `7Z_HOME` as an agent requirement but not update the script just yet...

PR #122, opened 2017-10-18
Make code sign timestamping more robust with randomization and more server endpoints and fix bugs in algorithm

mcdurdin:


ci.sh
```diff
@@ -12,7 +12,12 @@ KEYBOARDROOT="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 JQ="$KEYBOARDROOT/tools/jq-win64.exe"
 RSYNC="$KEYBOARDROOT/tools/rsync.exe"
 CI_CACHE="$KEYBOARDROOT/.cache"
-APP7Z="/c/Program Files/7-Zip/7z.exe"
+
+if [ ! -z "$SEVENZ_HOME" ]; then
+  APP7Z="$SEVENZ_HOME/7z"
```
darcywong00: Does this need to be `7z.exe`?

mcdurdin: No, .exe will be implied.

PR #67, opened 2017-08-01
Harari-Ethiopic Keyboard

MakaraSok:

PR #129, opened 2017-11-29
Updated Harari (Ethiopic) keyboard to match documentation

LornaSIL:
I think this can replace PR67


andjc:
@LornaSIL, looking at the kmn file, this looks like a very old version of the keyboard layout pre-version 2.0

The kmn file is not version 2.1

Changing the version to 2.2 means that it could bump version 2.1 replacing a valid 2.1 layout with an older layout that does not reflect the current orhtography of the language.

@mcdurdin, this is related to the discrepancy in keyboard layouts with some of the layouts being added reflecting older versions of the layouts than exist on keyman.com

LornaSIL:
@andjc I updated the .kmn file to produce what the documentation says it should produce. 

Thus, I changed the results for ```h]``` and ```h```, added rules for ```t]``` ```xh```, ```d]```, ```zl```, fixed some rule regarding ```dh```, added rules for the acute accent, and fixed the rules for producing the numbers. 

So, if that isn't sufficient then we need to abandon all the PRs until you have time to find the most recent source files. This was our attempt to try to move this forward.

andjc:
Having a quick look at it ...we have a situation where a 1.x layout is being partially adapted to match 2.x documentation.

Not sure how the 2.x documentation ever got matched up with 1.x source  files since the two were written a few years apart.And would have never coexisted.

It also looks like the problem has moved upstream wrt the Harari Latin layout. 

Best solution is to wind back kmn to original 1.1 version bumbling version back to 1.1 or 1.2. Dumping the 2.x documentation, and doing a PR for 1.x

I will then fork it axe the kvk file, update documentation and source files for 2.x

andjc:
@LornaSIL not sure what do do with Latin version though, since things are now out of complete whack there. May have to do something similar with some of the other PRs

mcdurdin:
Why don't we just start again with the 2.x version? I don't see a lot of value pushing forward with an old version of the source. Andrew, do you want to submit a PR to obsolete this one?

DavidLRowe:
Shall we just close PR #67 and this one (PR #129)? Then start fresh with the right version?

LornaSIL:
That seems reasonable.

PR #131, opened 2017-12-08
Deprecate legacy yolngu20 and yolngu21, replaced by el_yolngu

mcdurdin:


mcdurdin:
Please wait for discussion on what we do with old versions of keyboards and whether we keep them available somewhere and indexed, before merge.

PR #132, opened 2017-12-08
Deprecate legacy dinkaweb11 and dlia25bas, and add dlia25bas as a related deprecated keyboard to el_dinka

mcdurdin:


mcdurdin:
Please wait for discussion on what we do with old versions of keyboards and whether we keep them available somewhere and indexed, before merge.

PR #135, opened 2017-12-08
[libtralo] LIBTRALO keyboard

LornaSIL:
New keyboard.


DavidLRowe:
A few minor items:

* The `extras/libtralo.doc` file looks as though it's really RTF format.

* In the `README.md` file, under Links, the self-referential `keyman.com` link could be removed (along with the whole Links section) or, if some other address exists, replaced with a non `keyman.com` address.

* Compiling .kmn gives warning:
```
libtralo.kmn: Warning: line 0  warning 2092: Key "T_new_425" on layer "numeric", platform "phone", is a custom key but has no corresponding rule in the source.
```
Does this need an entry in the .kmn file?

* On touch layout, shift layer, the "shift" key is blank. Usually it uses the same uparrow to give the indication that pressing it will take one out of shift mode.

[Also curious why "dia" key is on symbol layer, rather than on default and shift layers (to the right of "L", for example). It seems that you'd want the means to type diacritics close to the letters over which they appear.]

LornaSIL:
I've made the changes you suggested including the "dia" key on both default and shift layer (and removed it from the 3rd layer.

DavidLRowe:
I'm not able to get the right curly bracket on the "]" longpress key (on the numeric layer of the touch layout) to work. The left curly bracket is defined:
```
"text": "{",
"id": "U_007B"
```
while the right curly bracket is defined:
```
"text": "}",
"id": "K_RBRKT",
"layer": "shift"
```
I'm not sure if that's the problem (or why it would make a difference).

Other than that minor item, everything looks okay.

LornaSIL:
@DavidLRowe can you review this change?

PR #139, opened 2017-12-11
Torwali Keyboard - minor revisions

LornaSIL:
Updated two keystrokes
* . > U+06D4
* E > U+0654
Per request from rac. The requests seemed reasonable.


rachitralli:
@LornaSIL Thanks

DavidLRowe:
For the desktop layout, the Shift+A and the Shift+\ produce the same character (U+0622). I wonder if the Shift+\ should be producing U+007C (vertical line).
```
+ [SHIFT K_A] > 'آ'
+ [SHIFT K_BKSLASH] > 'آ'
```
Similarly, both the shifted and unshifted = key produce +. I wonder if the unshifted = key should be producing = U+003D (EQUALS SIGN).
```
+ [K_EQUAL] > '+'
+ [SHIFT K_EQUAL] > '+'
```
On the "phone" touch layout (as tested on an iphone4s, served by KD 10.0.977.0), I'm seeing some incorrect output on the symbol layer. I think this is linked to the desktop issues above.

In the second row of keys, the one labeled "=" is producing "+". 
```
"id": "K_EQUAL",
"text": "=",
"layer": "default"
```
I think this is because it is triggering this rule in the .kmn file:
```
+ [K_EQUAL] > '+'
```
which probably ought to be producing '='.

Also in the second row of keys, the one labeled "|" is producing U+0622 instead of U+007C.
```
"id": "K_BKSLASH",
"text": "|",
"layer": "shift"
```
which seems to be because of this rule in the .kmn file is producing the wrong character:
```
+ [SHIFT K_BKSLASH] > 'آ'
```

LornaSIL:
I accidentally created a new PR when I made changes. #145 should take precedence.

PR #141, opened 2017-12-12
[sil_ipa] SIL IPA keyboard

LornaSIL:
This keyboard should take the place of older SIL IPA keyboards.


DavidLRowe:
@mcdurdin Will having `"languages": []` in the .keyboard_info file cause a problem? Not sure what else to put, though!

Minor: In README.md, under Links for Support, drop the self-referencing link to keyman.com, or change to `fonts@sil.org` as in the .kps file.

With that minor change (and assuming that having an empty "languages" field is OK), this is ready to merge.

mcdurdin:
Yes, for a non-language-specific keyboard such as IPA, it's hard to know what to put int the `languages` section. In the past, we have used a custom language code such as `qaa`. We could consider doing something similar again. We do have the `-fonipa` BCP 47 variant subtag available, but I don't know how useful that is. In the absence of a better suggestion, we could go with `qaa-fonipa` (and we should note `qaa` as our private use reservation for pan-language?

LornaSIL:
Marc suggested we ```Use und-fonipa in the .keyboard_info file and und-Latn in the .kps file``` so I have added those there. Could you review the keyboard @DavidLRowe ?

darcywong00:
Does this deprecate the "SIL IPA" keyboard in `legacy/i/ipauni111/`?

darcywong00:
It's probably better if you add them since this PR is in your repo.

See [gff-amh-powerpack-7](https://github.com/keymanapp/keyboards/blob/master/legacy/gff/gff-amh-powerpack-7/DEPRECATED.md) for an existing DEPRECATED.md

mcdurdin:
Note, please make sure you add the deprecated keyboards to the related section of the .keyboard_info file, e.g. (from gff_amh_7):
```
  "related": {
    "gff-amh-7": {
      "deprecates": true
    }
  }
```

DavidLRowe:
@LornaSIL Darcy pointed out that some additional information about deprecated keyboards needs to be added. I'm not sure I fully understand all the changes needed.

PR #147, opened 2018-01-14
[hausa_kano] Hausa Kano keyboard (My first keyboard submission)

shahzahm07:
My first keyboard submission, kindly check and let me know if there is anything to fix. 
Thanks


mcdurdin:
Congratulations on your first pull request and keyboard submission! Our team look forward to reviewing your keyboard and sending it out to the world.

MakaraSok:
I've moved the keyboard to the `release` directory and also added `.keyboard_info` to it. The keyboard is now complete, but one thing needs to be fixed. As you can see, it doesn't get built successfully. It may be not only because of the language tag issue, but also a missing `.kvk` in the `source` directory caused by the new implemented functionality in Keyman Developer. Let's wait for now!

mcdurdin:
I'm investigating the issue with `.kvks`

There is also an file `release\h\hausa_kano\hausa_kano.kps` that should be removed; the file release\h\hausa_kano\source\hausa_kano.kps` is the proper file.

mcdurdin:
Additional comments:
* welcome.htm and readme.htm need to be filtered and paths removed from the included files (right now, the included resources are in welcome_files and readme_files respectively)
* The project needs to be updated to reference the correct .kps file 
* Once https://github.com/keymanapp/keyman/pull/536 is merged and the compiler updated in the keyboards repo, the .kvks issue will be resolved.

DavidLRowe:
As I understand, @mcdurdin noted that in the `hausa_kano.keyboard_info` file, the language code needs to be `ha` (rather than `hau`).

LornaSIL:
@shahzahm07 the keyboard seems to work well. It is not yet compiling correctly though so I'll go through some changes that you should make.

In the .kmn file:
* please change the "Name" from "hausa_kano" to "Hausa Kano" or just "Hausa". This will be the name the users see when they install the keyboard. I like having "Kano" in the name so that people who use a different orthography (like in Niger) won't get confused.
* For Targets, please do not check "Any". Check "Windows", "macos" and Web. If you wish the keyboard to work for "any" you will also need to add a "Touch optimized keyboard" and do the work on that.

In the .kps file:
* Remove the Readme.md file
* Add the readme.htm file
* Remove the .kvk file
* Add the .kvks file
* If you wish "hausa_kano.pdf" to be included in the package, add it.
* Under the "details" tab change the package name to "Hausa Kano"
* Under "Readme file" choose readme.htm

In the .kpj file:
* remove the .kmp lines
```
    <File>
      <ID>id_2388242b7e75754bbc711af13d8b0fd6</ID>
      <Filename>HausaKeyboard.kmp</Filename>
      <Filepath>..\..\Users\Hamza\Documents\HausaKeyboard.kmp</Filepath>
      <FileVersion></FileVersion>
      <FileType>.kmp</FileType>
    </File>
```
* Change the "Name" from "<Name>hausa_kano</Name>" to "<Name>Hausa Kano</Name>" (or whatever name you chose to use in the .kmn file.
* Change the path for the .kps file to "<Filepath>source\hausa_kano.kps</Filepath>"
* Change the path for the .kvk file to "<Filepath>source\..\build\hausa_kano.kvk</Filepath>"
* Change the path for the .kmx file to "<Filepath>source\..\build\hausa_kano.kmx</Filepath>"

*In your "source" folder you should delete the two folders (and contents) called "readme_files" and "welcome_files". I think you should delete each individual file in the folders using ```git rm <filename>``` to do this. 

I think once you do all these changes the keyboard will about ready.

mcdurdin:
@LornaSIL Two small notes:

1. The .kps file should contain references to the built keyboard files, i.e. `..\build\hausa_kano.kvk`, `..\build\hausa_kano.kmx`, not the source .kvks file.
2. There is no need to manually edit the .kpj file. It will be updated by Keyman Developer automatically -- in fact, I would advise against manual editing, because it's easy to make a typo.

DavidLRowe:
@shahzahm07 In order to get your keyboard into the automatic build system, you'll need to make some changes to the project. The .kpj (Keyman Project) file needs to be named `hausa_kano` -- the same as the project folder. In Keyman Developer, from the Project menu, select Project Setttings. In the dialogue box that comes up, change the Output path to `$PROJECTPATH\build` and tick the checkbox labelled "Treat compiler hints and warnings as errors". 

Other changes to clean up or rearrange files:
- delete the extra `Hausa Kano.kpj` from the `source` folder. (There should only be one .kpj file and it should be in the project folder, not the source subfolder.)
- delete `hausa_kano.kmx`, `hausa_kano.kmp`, `hausa_kano.kvk` from the `source` folder. (When you next compile the keyboard, these will be created but now will be placed in the `build` folder.)

With the change listed above to use a separate `build` folder, you will now see all the compiled items end up in that folder and you will need to change the Keyman package to reference them from the `build` folder. (This is the information that ends up in the .kps (Keyman package source) file, which controls what files end up in the .kmp (Keyman package) file.)
- The path to the `hausa_kano.kmx` needs to be `hausa_kano\build\hausa_kano.kmx` (that is, referencing the file in the `build` folder, not the `source` folder
- Same for `hausa_kano.kvk`

@mcdurdin Can you verify whether the names `welcome.html` and `readme.html` are permitted, or whether they need to be changed to `welcome.htm` and `readme.htm`? Thanks.
@shahzahm07 Depending on Marc's reply, you may need to rename the files from `.html` to `.htm`.

Other files:
- `hausa_kano.kvks` is not needed in the package and should be removed. (The compiled `hausa_kano.kvk` file will be included from the `build` folder.)
- `welcome.htm` should be included in the package 
- `hausa_kano.pdf` doesn't seem to be referenced by any other file, so should probably be removed from the package (unless you end up referencing it from `welcome.htm`). This file could be placed in a folder called `docs` if you want to include it with the keyboard. Or if you do reference this file from `welcome.htm`, then you would need to leave it in the `source` folder and leave it in the Keyman package source.

So, most of these changes are not about the keyboard itself, but about fitting it into the automatic build and distribution system. Let me know if you have any questions.

mcdurdin:
`welcome.html` must be named `welcome.htm`, however the readme filename is determined from the `ReadmeFile` reference in the package source file, so does not need to be changed.

shahzahm07:
Hi, I tried adding a Touch optimized keyboard and i came across these errors

hausa_kano.kmn: Warning: line 0  warning 2092: Key "T_new_411" on layer "default", platform "phone", is a custom key but has no corresponding rule in the source.

DavidLRowe:
For touch layouts:
 - A K_xxx entry will use the corresponding rule in the .kmn file (if one exists), with modifications (shift/default, etc).
 - If there is no correspoinding rule, then a K_xxx entry will generate a character based on the base keyboard (e.g. K_A -> 'a').
 - A U_ABCD entry will generate the corresponding Unicode character, U+ABCD.
 - A T_xxx entry will have no default output, and needs a rule in the .kmn file to produce output.

You may find it easier to use a U_ABCD entry (rather than T_new_411) to generate a Unicode character from a touch key.

DavidLRowe:
Here are a couple of suggestions and one error to correct.

(1) You currently have three pretty much identical touch layouts (keyboard, tablet and phone). I'd suggest that you eliminate two of them and that would leave you with just one layout to maintain. So, for example, keep the phone layout and delete the other two, then the phone layout will be used on phones, but also on tablets, etc. You only need to have multiple layouts if you're making them different (for example, putting more keys per row on a tablet layout), but that doesn't seem to be the case here.

(2) The second suggestion is to change the K_PERIOD key on the shift layer to make it identical to the K_PERIOD key on the default layer. So on either layer the user would have access to the period (full stop) and other punctuation (rather than replacing the period on the shift layer with ">"). To do this you can change:
```
"id": "K_PERIOD",
"text": ">",
```
to:
```
"id": "K_PERIOD",
"text": ".",
"layer": "default",
```
on the shift layer.

(3) As for the error, the "$" on the longpress for "₦" is producing "₦" rather than "$". The code for that key is:
```
"text": "$",
"id": "K_4",
"layer": "shift"
```
but because the `id` is `K_4 and the `layer` is `shift`, this uses the logic from the .kmn file for SHIFT K_4 (which is what "$" works out to):
```
+ "$" > U+20A6
```
and so produces "₦" rather than "$".

One solution would be to change the entry to specify:
```
"text": "$",
"id": "U_0024"
```
which will output U+0024, which is "$".

LornaSIL:
https://keyman.com/keyboards/hausa_kano

PR #154, opened 2018-01-23
template: update two files per David's proposal

MakaraSok:
The README.md and template.keyboard_info have been updated per David's proposal.


DavidLRowe:
@MakaraSok Can you make these changes?

DavidLRowe:
@MakaraSok Were you able to make the two minor changes Lorna requested ("in" to "on", and "this" to "the")? I'm not seeing them, but maybe I'm not looking at the right thing.

PR #166, opened 2018-01-31
[ntl_onekey] initial release of NTL OneKey (smartphone + desktop)

ericjding:
This is a first attempt at submitting the NTL OneKey keyboard, developed by Adrian Cheuk (desktop) and Eric Ding (smartphone), to the Keyman keyboard repository.  The NTL OneKey desktop keyboard is based on a legacy keyboard layout used for the New Tai Lue script (Ethnologue code: khb).  Please contact me at eric_ding@sil.org for further discussion or questions.


darcywong00:
The continuous-integration build is failing with this error
>  Keyboard info file ntl_onekey.keyboard_info is missing

Keyboard ID's use `_` characters and not `-`, which means your keyboard ID should be `ntl_onekey`. I think if you rename your `ntl-onekey...` files to `ntl_onekey...` and their references, that should satisfy the build server.

ericjding:
ok, I've changed ntl-onekey to ntl_onekey in all filenames and references as you've suggested. Hope this works better now! :-)  Thanks!

darcywong00:
@DavidLRowe or @MakaraSok , when you get a chance, can you provide feedback on this experimental keyboard PR?

LornaSIL:
Looks like a useful keyboard. I have a few comments:
* .kpj and .kmn file
```<Message>This NTL OneKey (新傣文一键) keyboard is distributed under the X11 free software license.</Message>```
All Keyman keyboards must be distributed under the MIT license and other places indicate you are willing to distribute under MIT so that should be changed.
* Copyright dates in multiple files: some have 2017 and some 2018. They should be made consistent, probably 2017-2018.

* In the .kmn file delete the second one of these lines ```group(main) using keys```. Once I did that I could start typing the script.

There is no documentation so I didn't attempt to test the keyboard further.

ericjding:
Thanks for your detailed feedback! Regarding documentation: I hadn't thought of it when submitting, but I do have a visual keyboard map that I made way back when Adrian first made the desktop keyboard for our project. I made it in LibreOffice, so it's available in either ODT or PDF format; which version should I actually check into the keyboard's source directory and include in the package file?

ericjding:
I have checked the appropriate licensing and copyright date changes you've requested, as well as fixing up the kmn file (which somehow got broken since Adrian first passed it on to me, even though I've made no substantive changes to it).  The X11 license is actually essentially the same as the MIT license, so changing the licensing wording was a trivial issue.

I've also added a visual keyboard map for the desktop NTL OneKey keyboard to the package, and checked in the ODT source for that file as well.

LornaSIL:
That's a beautiful keyboard layout. And it works. I was a bit confused about how to get the things after the "/" (for the "w" key and the double and single quotes) until I realized you just type the key twice. I would recommend adding some text to explain that in English, Chinese and/or New Tai Lue. I guess adding an explanation is up to you since I eventually figured it out.

We are encouraging people to include a readme.htm file: This file can be displayed in the Install dialog and should be included with in the package (you'll need to add it to the .kps and .kpj file too). The readme.htm file should contain some basic information about the keyboard, for example, the keyboarding style, copyright information, acknowledgements, perhaps links to fonts, etc. It should not include instructions on how to use the keyboard – this information belongs in the Welcome documentation. The readme may be the same as the Keyboard Description on the keyman.com website. You can also link to the pdf of the layout, so people will be able to find it easily.

In the .kps, under "Package details" you can add the readme.htm under Optional Information and that readme will then show up in the package installer.

I sort of think you should rename the .odt and .pdf to either ```ntl_onekey.odt``` or to ```welcome.odt```. That would follow the patterns set for other keyboards.

You are allowed to include fonts if you wish. You might want to consider it if (!) there are not to many NTL fonts. They would go in the source folder and you would need to also include files (OFL.txt, fontlog.txt, etc) that came with the font. Of course, that means you must have permission to redistribute the font (and you do have that permission if it's an OFL font). At a minimum it would be good to include links to download appropriate fonts in the readme.htm. If you choose to add the fonts you need to remember to add all those files in the .kps file.

ericjding:
Thanks for your feedback, Lorna. Full disclosure: my main motivation to submit this keyboard is not for the desktop keyboard (as beautiful a job as Adrian did on it), which pretty much only our project uses for various reasons, but rather for the smartphone keyboard to be more more readily available. But when I have time over the next week or two, I'll work on improving the documentation to make the overall package (desktop + smartphone layouts) more ready for distribution.

darcywong00:
I note @ericjding is also submitting this keyboard as `experimental`, which should be a "lower" bar to approval than `released`, right?

DavidLRowe:
@darcywong00 In principle, yes. @mcdurdin Is the experimental section of the keyboards repository working now?

LornaSIL:
@darcywong00 that's right. I just didn't know how to test the keyboard since I'm not familiar with how the script works.

@ericjding I'm happy to approve the keyboard, but I will give you a bit more feedback and you can choose whether you want to fix it now or later.

As far as I can tell the default and shift layers of the mobile layout are working. The numeric layout is not working right. The following keys need the "Code" changed:
* longpress on the digits aren't working (1234567890)
* AT sign
* pound sign
* hyphen
* plus
* equal 
* left angle
* right angle
* $
* ^
* ( ) (and the longpress keys)
* / (and the longpress keys)
* % 
* asterisk
* : and ; are switched

IMO, the easiest way to fix these is to use the Unicode value (U_xxxx) rather than try to figure out what key and shift state they were on in the desktop version.

Let me know if you want the keyboard approved as is or if you want to fix the numeric layer first.

FYI: testing on Android with KitKat, none of the script shows up, no font is installed. Testing on an old IPad (I don't know about versions on IPad) there is a font so I could see the script.

ericjding:
@LornaSIL No rush, let me fix up the numeric layer first.  I am not seeing all the issues you're seeing, actually, but I did encounter a few others. So I'll do some more testing and then get back to you.

As for as testing: yes, we are aware that this keyboard requires Android 6.0 and above or iOS 10.0 and above (in order to get proper font support for New Tai Lue). Even if we included a font, or if you test on an old iPad, you may encounter other issues because of some important changes in Unicode 8.0. So we're simply going to inform our planned audience that these are the minimum system requirements for using the mobile keyboard.

mcdurdin:
@DavidLRowe yes `experimental` is working now. The projects are treated the same as `release` folder in that they need to compile, but the bar for the metadata should be lower.

ericjding:
@LornaSIL I'm finally getting around to writing up the documentation for the NTL OneKey keyboard. I'm creating a readme.htm file as well as additional files in the welcome subdirectory (inc. both English and, once I get the translation finished, Chinese versions of the welcome.htm file). I'll also delete the ntl_onekey.odt and ntl_onekey.pdf files, since those are replaced by the welcome documentation. Can you explain in more detail what steps I need to follow to make sure these files are visible and accessible to users? Thanks!

DavidLRowe:
@ericjding
- The readme.htm file is displayed when the user is on the point of installing the keyboard. It can be a couple of sentences.
- The welcome.htm is displayed after the user has installed the keyboard or asked for help. It should have enough information to enable the user to discover all the key combinations (although those combinations do not need to be listed exhaustively). The information can be in charts, prose or keyboard graphics. If the welcome.htm is the only file, you can put it directly in the source folder. If the welcome.htm has a number of other files it references (keyboard images, for example), it is preferable to make a `welcome` folder (under the `source` folder) and place welcome.htm and all its files there.
- Under the `source` folder, create a `help` folder, and in that folder place ntl_onekey.php. Often the text for this file is extracted from welcome.htm and given a few lines of header info. This file is used for the online keyboard help. It has the added feature of producing keyboard images directly (using `<div id='osk' data-states='default shift'/>`).
- The ntl_onekey.keyboard_info file also needs a description entry.

I'd suggest looking at `release/sil/sil_arabic_phonetic for an example of these files, but please feel free to pose additional questions here!

PR #168, opened 2018-01-31
Allow bypass of .exe build in CI, add support for injecting sourcePath to .keyboard_info, push to s.keyman.com

mcdurdin:
This PR covers some CI requirements:
1. Support injection of sourcePath into .keyboard_info
2. Allow bypass of .exe build in CI for test purposes
3. Push new .js keyboards to s.keyman.com


mcdurdin:
kmcomp.exe is now updated and as you can see, build passed.

resources/s-keyman-com.sh
```diff
@@ -0,0 +1,132 @@
...
+fi
+
+#
+# Check command line
+#
+
+for key in "$@"; do
+  case "$key" in
+    -f)
+      FORCE=true
```
darcywong00: The `-f` option sounds pretty powerful. Since it's rare to ever need it, what are your thoughts on making the flag `-force` as a precaution on accidentally forcing.

mcdurdin: It's unlikely to be an issue I think. This script can only be run if you have write access to s.keyman.com anyway, and needs environment setup to make it happen. Given it's a git transaction, we can always roll back if the worst was to happen.

PR #169, opened 2018-02-01
Check minimum Keyman Desktop version before bundling

mcdurdin:


resources/util.sh
```diff
@@ -110,4 +110,23 @@ function parse_args {
...
+
+function verlt {
+  [ "$1" = "$2" ] && return 1 || verlte $1 $2
+}
+
+#verlte 2.5.7 2.5.6 && echo "yes" || echo "no" # no
+#verlt 2.4.10 2.4.9 && echo "yes" || echo "no" # no
+#verlt 2.4.8 2.4.10 && echo "yes" || echo "no" # yes
+#verlte 2.5.6 2.5.6 && echo "yes" || echo "no" # yes
+#verlt 2.5.6 2.5.6 && echo "yes" || echo "no" # no
```
darcywong00: Is `verlte()` comparing that *a* is "less than, equal to" *b*? 

From these test cases, it seems the function is more like `vergte`

Did you also check this?
```
#verlte 2.5.7 2.5.7 && echo "yes" || echo "no"
```

mcdurdin: Ugh, not sure how I managed to break that before. Try again now.

resources/util.sh
```diff
@@ -110,4 +110,29 @@ function parse_args {
...
+function verlt {
+  [ "$1" = "$2" ] && return 1 || verlte $1 $2
+}
+
+# verlte 2.5.7 2.5.6 && echo "yes" || echo "no" # no
+# verlt 2.4.10 2.4.9 && echo "yes" || echo "no" # no
+# verlt 2.4.8 2.4.10 && echo "yes" || echo "no" # yes
+# verlte 2.5.6 2.5.6 && echo "yes" || echo "no" # yes
+# verlt 2.5.6 2.5.6 && echo "yes" || echo "no" # no 
+
```
darcywong00: looks like you can remove lines 128-132 now.
And include one for
```
# if $(verlte 2.5.7 2.5.7) ; then echo "yes" ; else echo "no" ; fi # ...
```

ci.sh
```diff
@@ -222,8 +231,12 @@ function upload_keyboard {
   
   if [[ ${package_filename##*.} == kmp ]]; then
     # We only upload a combined installer for .kmp files
-    create_package_installer "$buildpath" "$buildpath/$installer_filename" "$buildpath/$package_filename" "$package_name" "$package_version"
-    prepare_for_upload "$buildpath/$installer_filename" "$installer_upload_path"
+    if $(verlte "$KEYMANDESKTOP_VERSION" "$min_required_desktop_version"); then
```
darcywong00: Now that `verlte` compares a <= b, doesn't this become
```
if $(verlte "min_required_desktop_version" "$KEYMANDESKTOP_VERSION"); then
```

resources/util.sh
mcdurdin: `verlte 2.5.7 2.5.7` is the same as `verlte 2.5.6 2.5.6` is it not? The test cases are commented out, so they are just for reference now; wanted to leave both patterns.

ci.sh
mcdurdin: Argh. Not my day for comparisons.

resources/util.sh
darcywong00: ic. I just wanted to make sure there were 3 test cases each for `verlt` and `verlte`

PR #173, opened 2018-02-02
Enable push to s.keyman.com by removing early return from commit_and_push

mcdurdin:


resources/s-keyman-com.sh
```diff
@@ -117,6 +116,9 @@ function commit_and_push {
   git commit -m "Keyboard deployment (automatic)" || return 1
   git push origin master
```
darcywong00: Do you want to return/print error if `git push` fails?

Issue #156, opened 2018-01-24
Lazuri keyboard should be tagged with Latn script


mcdurdin:
Because this keyboard is designed for Latin transcription of the Lazuri language, the language tag should probably include `-Latn`

mcdurdin:
Adding a rider. In the documentation, the links to PDFs should not be prefixed with `file:`

darcywong00:
@mcdurdin can this be closed?

DavidLRowe:
@darcywong00 Yes.

Issue #160, opened 2018-01-24
Kayan keyboard has incorrect language tags in .keyboard_info


mcdurdin:

darcywong00:
@mcdurdin can this be closed?

DavidLRowe:
@darcywong00 Yes

PR #178, opened 2018-02-08
[kayan] correct .keyboard_info and add version 1.4 files

DavidLRowe:
Correction to BCP 47 codes in .keyboard_info (issue #160), plus changes for version 1.4 originally submitted on #120


DavidLRowe:
closes #156

PR #179, opened 2018-02-09
[lazuri] Add -Latn to BCP 47 code in README.md and .keyboard_info files

DavidLRowe:
Addresses issue #156


DavidLRowe:
fixes #156

PR #181, opened 2018-02-12
[boonkit] Initial upload of SIL Boonkit keyboard

sdysart:
Initial upload of Boonkit keyboard.  Desktop keyboard was created by @mhosken some years ago.  I've added the visual keyboard, fonts and package stuff.  I had asked Martin H. to review the Readme.md file but I do not think that is necessary as I'm following the pattern of a readme.md from another sil project.


DavidLRowe:
I haven't got as far as reviewing the actual keyboard, but here are some things that need fixing on a higher level:

- There needs to be a sil_boonkit.keyboard_info file in the sil_boonkit folder. (See `release/template` for an example, although the changes in #154 are trying give more explanation about what that file should contain.)

- The .kmx file should not be included. This file will be built from your source. 

- In the README.md file, you can remove the Links section. Current policy is to include links only if they don't reference `keyman.com`. Removing those leaves no links left, so you can remove the whole section.

- In the README.md file, the Keyman version line is not needed

- The copyright statement needs to include the year (or year range) and should be the same everywhere it occurs. README.md and LICENSE.md are missing the year. The welcome.htm file has 2008-2017. The .kmn file has 2018. I'd suggest that all four places should have 2008-2018.

- The version number needs to be the same everywhere it occurs. HISTORY.md has 0.6.1, README.md has 0.6 and .kmn file has 0.5.4.

- The .kpj file should have CompilerWarningsAsErrors set True. This can be done from inside the Keyman Developer application using the Project | Project Settings menu item. This is the setting that will be used when your keyboard built by the automated system.

- One warning that will then become an error is line 15 of the .kmn file: `HOTKEY "^+B"`. This line can probably be removed.

- The welcome.htm file refers to two .png files. The files themselves have .PNG extensions. This works on Windows but fails on Linux (or any operating system where file names are case sensitive). You should change either the extensions on the files to be `.png`, or else change the welcome.htm file to reference files with extension `.PNG`.

DavidLRowe:
All the changes look good. For the .keyboard_info file, you'll need to fill in values for "languages" and "description". This format might be easier to work with:
```
{
    "license": "mit",
    "languages": ["nod-Lana"],
    "description": "This keyboard layout is designed for Northern Thai with Lana Script."
}
```
(I took those values from the README.md file. You should verify them.)

DavidLRowe:
The changes look good. The .keyboard_info file needs values for "languages" and "description". For example, using values from README.md, you could have:
```
{
    "license": "mit",
    "languages": ["nod-Lana"],
    "description": "This keyboard layout is designed for Northern Thai with Lana Script."
}
```

There's still a mismatch between the .PNG extension on the images in the source/welcome folder and the .png extension in the references to these files found in the welcome.htm file.

DavidLRowe:
The .kps file should not include: HISTORY.md, README.md, LICENSE.md

Package version needs to be set to keyboard version (either manually, or using the `Package version follows keyboard version` tick box in Keyman Developer (Package Details).

Package name includes an old version number. You'll probably be better off removing the version number from the name in both .kmn and .kps files.

I tested the keyboard as much as my limited knowledge of the script allows. The U+1A58 character on the 7 key on the Boonkitshifted.PNG image seems to have been clipped. In Keyman Developer you can add a space after the character to bring it fully onto the key cap. @mcdurdin Is this an issue that can be corrected in KD? I've noticed it with some (but not all) Latin combining diacritics.

(Sorry for the double posting above. I made the first comment and it wasn't showing on the page, so I thought it was lost and wrote it again.)

mcdurdin:
The position of combining diacritics on key caps can be tricky. Because the glyph typically falls outside its own character bounding box, it's hard to reliably determine its position. The positioning differs across platform and font and renderer versions as well! So the most appropriate fix in this situation is probably as you say to add spaces after the character on the OSK to shift it over to the left. Experiment 😉

DavidLRowe:
@sdysart Just a point of clarification: "All checks have failed" = the automated build is failing because the boonkit.keyboard_info file is incorrect AND (as I just noticed) is not named correctly. Message is:
```
Failed to validate sil_boonkit.keyboard_info in /c/BuildAgent/work/612377aa60a1616b/keyboards/release/sil/sil_boonkit/
```
(though I'm not sure whether that is visible outside the team).

In fact, in addition to needing to rename keyboard_info to sil_boonkit.keyboard_info, the following files:
- boonkit.kpj
- boonkit.kmn
- boonkit.kps
- boonkit.kvk 
should all be renamed to sil_boonkit.* (to match the folder name). This is a requirement of the automated build system. Sorry I didn't notice that earlier.

DavidLRowe:
There's a trailing comma on the "description" line in the sil_boonkit.keyboard_info. This makes the file invalid, so the build fails.

I'm not positive it's required, but it might be helpful to rename the .ico and .kvk files as sil_boonkit.* as well. You'll then need to change the .kmn file to use these new file names.

The sil_boonkit.kpj still references some files as boonkit.* (rather than the renamed sil_boonkit.*) -- including boonkit.kmn and boonkit.kps. I think you'll want to open Keyman Developer, select the renamed .kmn and .kps files, then make sure all the files that those reference have the proper name. Then save the project, which should put the proper info in the sil_boonkit.kpj which is what is needed for the automated build.

Sorry the process of publishing the keyboard is so complicated!

DavidLRowe:
(1) The trailing comma after the description string makes the sil_boonkit.keyboard_info file invalid.
```
{
    "license": "mit",
    "languages": ["nod-Lana"],
    "description": "This keyboard layout is designed for Northern Thai with Lana Script.",
}
```
needs to change to:
```
{
    "license": "mit",
    "languages": ["nod-Lana"],
    "description": "This keyboard layout is designed for Northern Thai with Lana Script."
}
```

(2) There's an unwanted change to `release/template/template.keyboard_info`.  This file should be removed from the the PR.

(3) The project file (sil_boonkit.kpj) references boonkit.kmn (and .ico) rather than the renamed sil_boonkit.kmn (and .ico)

(4) Verify that the Keyman package (which gets written to sil_boonkit.kps file) references the sil_boonkit.kvk file in the build folder (and not in the source folder).

PR #182, opened 2018-02-12
Update .keyboard_info to rename minKeymanDesktopVersion to minKeymanVersion and update associated tooling

mcdurdin:
A lot of files changed, but the changes are very straightforward. In fact, the `minKeymanVersion` field is optional but it is still helpful for looking at the source, so in most cases I have opted to leave the field there. A few legacy keyboards had mismatched version information (5.1 vs 5.0 minimum), and I opted to delete their `minKeymanVersion` field rather than updating it; the data will be present in the build/ version of the .keyboard_info files.


darcywong00:
Should all of these `"minKeymanVersion": "5.0"` be removed? There's 10 remaining in the `legacy` folder on this PR

mcdurdin:
@darcywong00 There's probably no reason to remove them for now. The ones removed are ones that were incorrect -- they should have been 5.1, not 5.0.

PR #189, opened 2018-02-24
New compiler version supports .js within .kmp for metadata, orders chiral layers correctly, improves error handling for warnings

mcdurdin:
This minor update to the compiler fixes a number of issues, including:

* support for .js within .kmp when building .keyboard_info files (keymanapp/keyman#591)
* fixes the order of chiral layers (keymanapp/keyman#602)
* improves error handling when warnings are produced (keymanapp/keyman#620)

The Khmer Angkor keyboard needed an update to 1.0.1 and a rebuild for two reasons:

1. to fix the chiral layer ordering, with a recompile
2. in order to remove legacy language metadata that was warned about, but didn't fail the build, but now (correctly) fails the build


release/k/khmer_angkor/source/khmer_angkor.kps
```diff
@@ -68,8 +68,9 @@
       <Name>Khmer Angkor</Name>
       <ID>khmer_angkor</ID>
       <Version>1.0</Version>
+      <RTL>True</RTL>
```
darcywong00: I wouldn't think Khmer was RTL

mcdurdin:
Appears that latest TIKE may set `RTL=1` on package metadata. Don't merge yet.

mcdurdin:
See https://github.com/keymanapp/keyman/pull/622 for fix to RTL bug. Need to get rebuilt kmcomp.exe to resolve this safely.

mcdurdin:
Okay, build 10.0.1026.0 corrects the `RTL` bug.

PR #191, opened 2018-02-26
Add php help for each keyboard from help.keyman.com to this repo and implement CI to send it back again

mcdurdin:
This tiny little PR seeds the repo with the documentation files associated with each keyboard on help.keyman.com and provides the CI script resources/help-keyman-com.sh to do the copying.

The corresponding TeamCity CI update will need to be made to call help-keyman-com.sh along with configuring a new version control root for the help.keyman.com git repo.


darcywong00:
Should we strip out the Thumbs.db files?

darcywong00:
Are all these files eventually going back to help.keyman.com? I note some links to
`<a href="/go?language=...>`

legacy/e/ekwbamuni/source/help/srii/index.php
```diff
@@ -0,0 +1,3 @@
+<?php
+  header('location: ../');
```
darcywong00: Is this redirect needed?

mcdurdin:
1. Yes, we should strip out Thumbs.db. Working on it now.

2. These files do get copied back to help.keyman.com. This will be the master version. This allows us to capture all changes for a keyboard in a single PR. So links such as the example you give are fine. In the future, I anticipate being able to improve the format and processing of these files, but right now the aim is just to get the source files in the same place, and synchronise the versions and keyboards available between the keyboards repo and the help.keyman.com site.

legacy/e/ekwbamuni/source/help/srii/index.php
mcdurdin: Possibly not, but I am not trying to fixup individual files; this is a reasonably pure copy of the existing content. This would have been there because there was a link to the page from some other site in the past.

darcywong00:
`legacy/e/eurpoean*` contain some `uppercase.PNG` files. Do they need to be renamed to `uppercase.png`?

mcdurdin:
We don't really need to rename or change filenames, etc, at this time.

PR #194, opened 2018-03-05
[jorai] Jorai keyboard

LornaSIL:
New keyboard for Jorai.


DavidLRowe:
- I think README.md need to change to exclude web
- Date in LICENSE.md needs correction (2108 to 2018) :smile: 
- welcome.htm displays headings for `CTRL+ALT` and `CTRL+ALT+Shift`, but keyboard works with Right Alt and Right Alt + Shift
- Touch layout (only tested phone, but tablet seems identical), numeric layer:
  - longpress on `[` for `{`gives U+01AF
  - longpress on `]` for `}` gives U+01A0

Also, on the touch layout, shift layer, what about having a `.` longpress (identical to default layer) instead of the existing `>` longpress? (Or add `<` and `.` to the existing `>` longpress?) Just a thought!

Since the phone and tablet layouts are identical, I think you can omit one and the other remaining one will be used for both purposes. (@mcdurdin Can you verify this?) The advantage is not having to make changes in two places, if indeed the layouts are intended to be identical.

mcdurdin:
Correct, if the phone and tablet layouts are identical, you should remove one (it doesn't matter which one, really). The other one will be sourced by default.

PR #124, opened 2017-10-25
[rac_kalasha] Moves files to rac folder and made changes to version 1.1. 

LornaSIL:
Moves files to rac folder and made changes to version 1.1. This replaces PR 102 and takes into account that changes that Makara made in the previous PR.

PR #202, opened 2018-03-15
[rac_kalasha] Lorna rac kalash version 1.1

mcdurdin:
Replacing #124 which had too much history in it and mixed up the folder move and the 1.1 version change.

@LornaSIL may want to validate that the changes here are the correct ones for the 1.1 version change.


mcdurdin:
And ... replaced with one I can write to: #203

PR #204, opened 2018-03-16
[sil_el_ethiopian_latin] sil-el-1-3-1 Added ǵ and Ǵ

rik-shaw:
Added ǵ and Ǵ.


LornaSIL:
One line in the .kps file still refers to version 1.3.0. Please change that to 1.3.1. Then the keyboard will be ready. I‘ve tested the additional rules.

rik-shaw:
@LornaSIL I corrected the file version of the .kps file documented in the .kpj file.  I think this was what you were seeing.  It does raise a question as to if the other files in the .kpj should have values for the <FileVersion> field?  Right now most of those are blank.  If the version numbers are not needed there, then maybe I should remove that value from the .kps file in the .kpj (so that the version number isn't referenced in the .kpj)?

LornaSIL:
@rik-shaw I think the 3 versions in the .kpj file are needed. At least, the other keyboards I just looked at had version numbers for .kmn, kps and in the "details". I think it helps build the package properly.

rik-shaw:
@LornaSIL thanks for the feedback and thanks for the merge

Issue #198, opened 2018-03-15
[osage_nation] missing php help


LornaSIL:
Need to create help file

PR #209, opened 2018-03-22
[osage_nation] added php help file

LornaSIL:
Added php help file


LornaSIL:
This closes #198

release/o/osage_nation/source/help/osage_nation.php
```diff
@@ -0,0 +1,64 @@
+<META http-eq<?php
```
darcywong00: The first line can be 
```
<?php
```

Issue #200, opened 2018-03-15
[libtralo] missing php help


LornaSIL:
Need to create help file

PR #211, opened 2018-03-22
[libtralo] added php help file

LornaSIL:


LornaSIL:
This closes #200

release/l/libtralo/source/help/libtralo.php
```diff
@@ -0,0 +1,1767 @@
+<META http-eq<?php
```
darcywong00: I wonder if you can just make the first line
```
<?php
```

LornaSIL:
I removed ```<META http-eq``` from the first line.

PR #222, opened 2018-04-06
Add keyboard name for NTL OneKey in keyboard_info file

ericjding:
I've added a name (in Chinese) for the NTL OneKey keyboard to the keyboard_info file.


LornaSIL:
I don't see any change in the keyboard on my phone. I wonder if it's because we didn't bump the version number.

ericjding:
I see the new keyboard name now in the list of available keyboards in Keyman on my Android.

PR #223, opened 2018-04-06
[masaram_gondi] Early test version of Masaram Gondi Keyman keyboard (currently incomp…

markpenny:
…lete)


LornaSIL:
@markpenny Thanks for your submission. Are you ready for this keyboard to be reviewed? Since I don't have a font for testing with it, I wonder if you could include a font (if it has an open license) in the PR, or you could send it to me via email for testing if you don't want the font distributed. Adding a font will allow people to use the keyboard and if you add the font to the .kps and .kpj I think people will see the font on Keyman for Android.

I did notice the README.md has an older version number in it.

If you add this store to your keyboard you won't get the warning messages about minimum engine:
```store(&VERSION) '9.0'```

If you want to add a "name" field to the keyboard_info file you'll see the keyboard name rather than "masaram_gondi" 
```"name": "Masaram Gondi",```
Add that right before the description.

Basically, it seems to work, so if you don't want further testing I think I could approve it since it is considered experimental.

keyman-server:

markpenny:
I'm closing this pull request because I have realized that it would be better to re-name this keyboard from marasam_gondi to gondi_masaram (in order that all the Gondi related keyboards are easy to find under 'G'). I will re-submit this soon with a new name.

PR #228, opened 2018-04-16
[sil_cameroun_azerty] Initial source files

andjc:
Currently, PHP help files are missing.


LornaSIL:
@andjc 
* the README.md needs a description and probably remove the "todo" list if you did it?
* the readme.htm is almost empty.
* I think the .kps file belongs in "source".

andjc:
@LornaSIL 
* Latest commit fixes the errors, I pushed wrong version of branch. Includes the help files.
* Re F208 and F218, I was unable to locate them in the kmn file. Any further info on those characters?
* Yes underlying French keyboard is necessary, it is a mnemonic layout rather than positional layout, so will not work unless underlying layout is a standard fr-FR keyboard layout. In your keyman, you will need to set the base keyboard layout to test.

```
Keyman Destop Configuration > Options > Base Keyboard > French
```

Currently testing on macOS

LornaSIL:
U+F218 > U+0244 (line 26 after the "Ƴ") - it's a square box
U+F208 > U+2C6D (line 27 after the "g") - it's a square box
U+F196 > U+02C8 (line 48)

Everything else looks like it should work, but I really don't know how to test it. I associated it with a French-France keyboard and the base characters work, but not the combining marks. I plugged in an old French keyboard I had, but it is qwerty and didn't work. Can you give me more hints on how to test it?

andjc:
@LornaSIL 
Pushed the fixes for PUA characters. And fixed a bug in my PUA regex in the process. Regarding testing. I can arrange to be online one of your mornings, and guide you through the setup I used.

andjc:
@LornaSIL

 Looking at the kmn file and the comments in it, it would appear that there was a change in the keys used, but the documentation wasn't updated to match. Will test to make sure, if problem is with documentation will fix that.

Andree

LornaSIL:
https://keyman.com/keyboards/sil_cameroon_azerty

PR #234, opened 2018-04-20
[sil_cameroon_qwerty] Added help files

andjc:
Added help files.


LornaSIL:
It works fine, but I think we are supposed to remove ``</body>`` and ``</html>`` tags at the bottom of the file since we removed them from the top.

andjc:
@LornaSIL , fixed.

LornaSIL:
https://help.keyman.com/keyboard/sil_cameroon_qwerty

PR #235, opened 2018-04-20
Make build script more robust for nearly-empty folders

darcywong00:
* Add build parameter `-w` to treat warnings as errors
* When building keyboards, skip if no .keyboard_info file exists or directory is empty

### DIFF
--- a/resources/compile.sh
+++ b/resources/compile.sh
@@ -18,6 +18,14 @@ function build_keyboards {
 
   local shortname
   for shortname in "$KEYBOARDROOT/$group/"*/ ; do
+    if [ ! "$(ls -A $shortname)" ]; then
+      if [ "$WARNINGS_AS_ERRORS" = true ]; then
+        die "$shortname is empty."
+      fi  
+      echo "$shortname is empty. skipping..."
+      continue
+    fi
+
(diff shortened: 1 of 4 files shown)


resources/compile.sh
```diff
@@ -68,11 +76,24 @@ function build_keyboard {
 
   pushd "$keyboard"
   
+  #
+  # Check if .keyboard_info doesn't exist
+  #
+  keyboard_infoFilename="$base_keyboard.keyboard_info"
+  if [ ! -f "$keyboard_infoFilename" ]; then
+    echo "- No $keyboard_infoFilename file. Skipping..."
```
mcdurdin: Can you move this line below the WARNINGS_AS_ERRORS test and either remove the hyphen here or add it to the shortname report above?

resources/util.sh
```diff
@@ -92,14 +93,17 @@ function parse_args {
...
+          ;;          
         -h|-?)
           display_usage
           ;;
         -t)
           lastkey=$key
           ;;
         *)
-          TARGET="$1"
+          TARGET="$1" # should this be "$key"?
```
mcdurdin: Good question. Can you test and determine the answer?

Issue #246, opened 2018-04-28
[infrastructure] Update packages to use the follow-keyboard-version flag


darcywong00:
* ~Update compiler to 10.0.981.0 or later for fkv support~
* Update keyboards to use FKV flag (or do we use this on an as-needed basis?)

(edit) Current kmcomp is already later than 10.0.981.0

darcywong00:
@DavidLRowe and @LornaSIL can you clarify which keyboard packages in the repo should **not** get this flag?
`<FollowKeyboardVersion/>`

LornaSIL:
I can't think of any reason they shouldn't all get the flag. That should be
the default in my opinion. At least for ones in release. We may need to
review legacy ones as sometimes those have more than one keyboard per
package.

DavidLRowe:
Most of the keyboards in the release folder are updated by PR #407 . See that PR for several odd cases.

LornaSIL:
Still need to fix:
* gff_ethiopic_7 (should let @dyacob to fix)
     * gff_ethiopic_7/source/gff_ethiopic_7.kps has version "1.0" and no FKVflag, but gff_ethiopic_7/source/gff_ethiopic_7.kmn contains no &keyboardversion statement -- seems as though the .kmn needs to be updated to have a version number before the FKVflag can be set (or is the default version number 1.0?).
* sil_tchad (but Jeff H wants to update it, so should we wait?)
     * sil_tchad/source/sil_tchad.kps has version "3.3.1" and FKVflag set, but sil_tchad/source/sil_tchad.kmn has "store(&KEYBOARDVERSION) '3.3' " -- seems as though the version in the .kmn file needs updating.

The powerpacks will be deprecated so it's probably okay not to update them:
* release/packages/gff_amh_powerpack_7/source/gff_amh_powerpack_7.kps -- has multiple keyboards

PR #247, opened 2018-04-28
Update kmcomp to 10.0.1061.0 beta

darcywong00:
Update compiler to [10.0.1061.0 beta](https://downloads.keyman.com/developer/beta/10.0.1061.0/) for #246


darcywong00:
kmckmp update not needed

PR #257, opened 2018-05-05
[sil_hebrew] Inital migration of sil_hebrew keyboard

andjc:


andjc:
Re version number: there wasn't any real versioning but there were comments about when changes were made ...  started at 1.0 increment by a minor version number for each version and adding one for the migration. Unless i missed something that took me to version 1.6

LornaSIL:
Version 1.6 is fine.

LornaSIL:
Several places in the .kps and .kpj file it says version 1.0. Other places 1.6 and that's what is ending up in the package. So, I don't know how important it is, but I think it would be good to fix that from 1.0 > 1.6.

The README.md and HISTORY.md should have the keyboard name changed from "Hebrew" to "SIL Hebrew".

I think with those changes the keyboard will be good to go.

sewhite:
I just discovered an issue with the SIL Hebrew  keyboard installing into WIndows 10. The keyboard installs but Windows does not assign it to a language so you do not see it on the list in the main Keyman window. Is "hbo" still a valid language tag? Sometimes also WIndows requires a script tag as part of the language tag, maybe this is one of those times.

andjc:
@sewhite  I will look at it. 

hbo is a valid BCP-47 language tag.

In theory, it could be assigned to hbo-Hebr or maybe hbo-Hebr-001 may be more correct. It could also be assigned to modern Hebrew.

PR #266, opened 2018-05-10
[balochi_urdu] New Balochi keyboard based on the Urdu layout.

rmlockwood:


LornaSIL:
* On the "tablet" layout
     * You have many different layers. All but "default" and "shift" could be deleted since those are the only two layers you need. You could consider removing the CTRL and ALT keys also. Just make the K_SPACE wider.
     * On the tablet layout, default layer, the K_QUOTE key is outputting U+0652. If that is the correct output, could you change the keycap it be sukun rather than straight quote?
     * On the table layout, shift layer, the K_6 key is outputting U+200C. Is that correct? The keycap is ```^``` but I don't have any good ideas of what it should be!
     * On the tablet layout, shift layer, the K_BKQUOTE is outputting U+0654. If that is correct can you change the keycap to display hamza instead of tilde?
     * On the table layout, shift layer, the K_PERIOD is outputting U+2039. Can you change the keycap to match? (similar to how you have K_COMMA)
* On the phone layout:
     * The shift "shift K_6" is not outputting anything. Do you want to add U+200C? Also, add the same thing on the keycap as you used for the tablet one.
* So Balochi doesn't use the U+06F0..U+06F9 digits? They use the "English" digits?
* You should add a ```balochi_urdu.keyboard_info``` file similar to how it was done in ```balochi_phonetic.keyboard_info```
* Consider adding screenshots of the default and shift Desktop layers to the welcome.htm and balochi_urdu.php files (you can export the graphics from the On-Screen tab and then click Export)
* In the .kps file:
     * If you added screenshots, add them into the files
     * Add readme.htm to files
     * Click on Keyboards, add the BCP-47 codes there, along with the language names. That will help people when they look for the mobile layouts.
     * Click on "Details" and add the readme.htm file under "Optional Information"
* Reopen the .kpj file to make sure it inherits everything it needs to from the changes you've made.

I realize this is a lot of feedback, but it seems like it's in great shape. Thanks for your submission.

rmlockwood:
Hi Lorna,

 

Thanks for the feedback which makes my keyboard much better! I had forgotten to look over the tablet keyboard. According to what I saw on the Urdu Windows keyboard, there are only the “English” digits. A decision was reached to only use English digits.

 

Let me know how the keyboard looks now with the changes I’ve committed and synced.

 

Thanks,

Ron

LornaSIL:
Thanks!

I'm wondering about the shift \ key that outputs U+06C1. The On screen keyboard has U+FBA7 on the keycap. Could that be changed? Then the OSK graphic needs regenerating for the welcome file.

Also, on SHIFT+CTRL ` you are outputting U+FEF2. I'm just wondering why. You're already outputting U+064A on the "L" key.

* It looks like none of the characters from the desktop CTRL+SHIFT layer are on the touch layouts. I don't know how important that is, because the following characters are on both the default and the CTRL+SHIFT layer
     * damma U+064F
     * fatha U+064E
     * kasra U+0650
     * sukun U+0652
     * alef hamza U+0623
* So the characters on the desktop CTRL+SHIFT layer that aren't on the tablet/phone layouts are (consider adding them to a longpress):
     * U+FEF2 (which I asked above about)
     * U+06D9
     * U+064B
     * U+064C
     * U+064D
     * U+06E2
     * U+0625
* On the phone layout I noticed that the default / top row / far right key outputs U+06C6 and U+0648 which are also output somewhere else (default / middle row / 2nd from left key). You could make the top far right key a diacritic longpress instead of having the waws there (and then include the above diacritics). (Maybe use the fatha U+064E on the keycap and as the default output.)
     * on that same default / top row /far right key, there is a thing in the longpress that has the glyph of U+0654, but it is outputting U+0658. I think you want U+0654 instead of U+0658.

I'm sorry I didn't actually test the touch layouts yesterday.

It would also be good to open the .kps file and on the Keyboards tab add the BCP-47 tags for the 4 languages. I think that will make it so people can find the mobile keyboards by language name and not just tag.

rmlockwood:
The FBA7 is the image I want. In its isolated form it looks exactly like the other heh so I wanted to distinguish it.

 

The SHIFT+CTRL ` is just what was already on the Urdu keyboard and I was trying to keep things the same. I think I’ll delete it though. I don’t know why they would want a presentation form outputted.

 

I do have the main diacritics on the touch layouts. For phone they are long-press on the K_P key. For tablet, damma, kasra, fatha and sukun are on the right side on the default layer. There’s a couple more on the shift layer.

 

For phone:

*	U+06D9 – now on long press of K_P
*	U+064B – on long press of K_P
*	U+064C – now on long press of K_P
*	U+064D – now on long press of K_P
*	U+06E2 – now on long press of K_P
*	U+0625 –  on long press of K_J

For tablet:

*	U+06D9 – now on long press of K_LBRKT
*	U+064B – now on long press of K_ LBRKT
*	U+064C – now on long press of K_ LBRKT
*	U+064D – now on long press of K_ LBRKT
*	U+06E2 – now on long press of K_ LBRKT
*	U+0625 –  on long press of K_J

The vav behavior is what I want. We want the composed character outputted. I changed 0658 to 0654. Thanks for catching that.

 

I’ve committed the changes.

 

Thanks,

Ron

LornaSIL:
Ok, thanks! LGTM.

release/b/balochi_urdu/source/balochi_urdu.kps
```diff
@@ -0,0 +1,85 @@
...
+      <Name>welcome\welcome.htm</Name>
+      <Description>File welcome.htm</Description>
+      <CopyLocation>0</CopyLocation>
+      <FileType>.htm</FileType>
+    </File>
+    <File>
+      <Name>..\build\balochi_urdu.kmx</Name>
+      <Description>Keyboard Balochi Urdu</Description>
+      <CopyLocation>0</CopyLocation>
+      <FileType>.kmx</FileType>
```
darcywong00: With Keyman Developer 10, you can also include the compiled touch keyboard in the keyboard package
`..\build\balochi_urdu.js`.

This way, users can pass around the same .kmp file and it will install on Desktop and mobile apps.

release/b/balochi_urdu/README.md
```diff
@@ -0,0 +1,36 @@
...
+ * Windows
+ * Macosx
+ * Web
+ * iOS
+ * Android
+
+Todo
+----
+
+* Change "home" url when there is a release on keyman site
```
darcywong00: You can use `https:/help.keyman.com/keyboard/balochi_urdu`

The URL will be valid when this PR gets merged.

release/b/balochi_urdu/source/balochi_urdu.kps
LornaSIL: @rmlockwood do you want to go ahead and do this? If not, I can go ahead and merge. If you want to do that, I'll wait. You might need to update your version of Keyman Developer to the latest.

rmlockwood:
Hi Lorna,

 

I made the changes Darcy suggested.

 

Thanks,

Ron

LornaSIL:
The "-Arab" was removed from the .kps BCP47 tags. Was that on purpose? I think they are needed to show up properly in the Keyboard taskbar on Windows. 

I think everything else is okay.

rmlockwood:
fixed

PR #267, opened 2018-05-10
[rac_kalasha] Added example text and id changed to include language name

LornaSIL:


rachitralli:
@LornaSIL thanks

release/rac/rac_kalasha/source/rac_kalasha.kps
```diff
@@ -106,7 +106,7 @@
       <ID>rac_kalasha</ID>
       <Version>1.0</Version>
```
darcywong00: Do you know if the keyboard/package version needs to be bumped. To force a rebuild of the package.

LornaSIL: You know better than I do. I was thinking meta data changes didn't need a version bump, but if it affects the build, then I guess it does.

LornaSIL:
@darcywong00 Unfortunately, with a bump of the version I also felt that we should fix the language name, BCP codes, using the shared fonts, etc, etc. (so it's no longer just adding the "example" to the keyboard_info file).

rachitralli:
@LornaSIL thanks for adding the example text.

PR #272, opened 2018-05-11
[rac_palula] Initial Palula keyboard

rachitralli:
Please add palula keyboard, I have uploaded all the files here https://github.com/rachitralli/Rachitrali-Palula-Keyboard


rachitralli:
@MakaraSok sir all checks have passed successfully, please check and guide me, thanks

rachitralli:
@DavidLRowe @Makarasok All checks have passed, pls check when you have spare time please

MakaraSok:
@rachitralli As an author of the keyboard, you have to ensure that the performance of the keyboard is up to what you have created it to work. The checks here is only the systematic issues such as file structures and  organization.

I would suggest that you test the keyboard on various devices (i.e. Windows, macOS, iOS, Android and web) to ensure that everything is working as expected. 

Our team will double check on all that as well. Please do not expect us to catch all error as we do not speak/know the language of the keyboard. :)

DavidLRowe:
- The .kps file needs to include FONTLOG.txt, OFL-FAQ.txt, OFL.txt, Scheherazade_README.txt
 from `release/shared/fonts/sil/scheherazade`
- source/help/rac_palula.php is not in the correct format (needs PHP header in place of `<head>` element). 
- source/help/RAC.jpg is not referenced; it should either be referenced by a file or else removed from the PR.
- The Thumbs.db file should be removed from the source/ folder.
- touch layout, shift layer, rightmost key on third layer has text "\" but produces "|".
- touch layout, shift layer, shift key doesn't take you out of shift layer, so there's no way to get back to default layer!

rachitralli:
@MakaraSok please include fonts as suggested by @DavidLRowe and check  other above minor issues when you are free

MakaraSok:
Here is the link to test things out: https://drive.google.com/file/d/1UfUMILqcSxBpCL2v3-7UDtBm7Wvd7Pl2/view?usp=sharing

By the way, would you like to learn how to modify/create touch layout? I think it may be better for your to know how to customize the key positions and maneuver around the touch development process. It may give you more control over what/how you like the touch layout to be like.

rachitralli:
@MakaraSok in the touch layout shift layer the key \ producing | please delete the entire key as the \ and | are already there, after this amendment this is ready to be approved. thanks

MakaraSok:
@rachitralli Your layout has two |. That's why the touch has two of it too.

I have re-arranged the touch key layout. Here is how each layer looks:

Default layer:
image

Shift layer:
image

Symbol layer:
image

What do you think about them?

rachitralli:
@MakaraSok is now ok, please change the font for touch layout

DavidLRowe:
It seems that the help/rac_palula.php uses different graphics than welcome.htm. Is that intentional? That's not a problem, though it does mean more work to update both if the keyboard is ever revised.

I'll approve this as is, and if you can update the help/rac_palula.php at a later date if desired.

PR #278, opened 2018-05-14
[osage_nation] Changed touch key and .kvk to .kvks

LornaSIL:
Changed touch key laout
Changed .kvk to .kvks
Refreshed .kps
Refreshed .kpj
Removed history from welcome.htm and help
Updated version to 1.2


DavidLRowe:
Touch layout, shift layer, "." produces ">"

Minor: Touch layout, numeric layer: keys in rows 2 and 3 slightly offset from the other two layers

DavidLRowe:
Should soon be available at https://keyman.com/keyboards/osage_nation

PR #284, opened 2018-05-15
[sil_bari] Initial Bari keyboard

LornaSIL:


DavidLRowe:
seems to be a problem with .kmn entry for " that includes an extra "Z":
```
+ [SHIFT K_QUOTE] > '"Z'
```
This shows up on touch layout on . longpress, but fixing it in .kmn should fix touch layout as well.

On tablet layout, shift layer, the "." key produces ">". Changing the layer to "default" should correct that.

The different layers on the touch layout don't have the same configuration, so when changing between layers, the keys slide around. The shift and numeric are the same, but on the default layer, the second row is all the way to the right and some keys changes size. It would be better if all the layers had the same configuration.

LornaSIL:
https://keyman.com/keyboards/sil_bari

PR #290, opened 2018-05-18
[el_pasifika] porting to crossplatform version to GitHub

andjc:


LornaSIL:
* ```README.md``` the link has a typo: https://enabling-langauges.github.io (change "languages")
* readme.htm is not html, but a plain text file.
* the link at the bottom of the welcome.htm file for the "State Library Victoria" is not working. It has a spurious "w" in it: httpw://www.slv.vic.gov.au/

I didn't get a chance to test the keyboard. I'll do that next week.

LornaSIL:
The requested fixes were done properly, so no need to check again.

* I am not getting any output for the Right-Alt nor Shift Right-Alt desktop or OSK
* For the phone and tablet layouts:
     * On the default and shift layers / fullstop longpress / the ```\``` is outputting U+02BB
     * On the numeric layer / ```$``` longpress / the Euro and pound sign are not being output
     * On the numeric layer / ```&``` Key the output is U+005E
     * On the numeric layer / ```|``` Key there is no output
     * On the numeric layer / ```\``` Key there is no output
     * On the numeric layer / ```(c)``` Key there is no output
     * On the numeric layer / ```[``` Key there is no output
     * On the numeric layer / ```[``` longpress key there is no output for "{"
     * On the numeric layer / ```]``` Key there is no output
     * On the numeric layer / ```]``` longpress key there is no output for "}"

If the phone and table layouts are the same, there is no reason to include both. I'd just use the phone layout and delete the table layout. That way you don't have to fix it in both places. The only reason to use both layouts is if you want them different.

I think I've checked everything, so once these are fixed we can approve the keyboard.

andjc:
Hi Lorna,

fixed the key name problems relating to:
* On the default and shift layers / fullstop longpress / the \ is outputting U+02BB
* On the numeric layer / & Key the output is U+005E

Other mobile problems would be related to the AltGr and Shift+AltGr problems, keys not being caught by nomatch rule in constraints group. Added a rule to explicity pass them to the group. This seems to work in desktp. Haven't had a chance to check on mobile yet, but should fix the problem. My mobile is glitzing at the moment. Time for a factory reset.

LornaSIL:
There is one thing left that needs changing and then it will be ready.
* On the mobile layout, shift layer, the "." longpress key, the fullstop is outputting ```>``` and the ```\``` is not outputting anything. Rather than try to figure out all the layers, when I have something like this, I like to just change the code to using "U_xxxx".

LornaSIL:
https://keyman.com/keyboards/el_pasifika

PR #292, opened 2018-05-19
[rac_kashmiri-shina] Initial Kashmiri-Shina Keyboard

rachitralli:
please add Kashmiri-Shina Keyboard


rachitralli:
All the relevant files have been upload here as suggested https://github.com/rachitralli/keyboards/tree/rac_kashmiri-shina/release/rac/rac_kashmiri-shina

rachitralli:
All the relevant files have been upload here as suggested https://github.com/rachitralli/keyboards/tree/rac_kashmiri-shina/release/rac/rac_kashmiri-shina

LornaSIL:
@rachitralli 
* How is this keyboard different from the rac_shina keyboard? 
     * Both keyboards use the scl language code. 
     * Does this one also support the Kashmiri language? 
* the ```keyboard_info``` file says that this keyboard deprecates the "Shina" keyboard. But no such keyboard is on the Keyman site. Is this keyboard supposed to deprecate the rac_shina keyboard?
* Should this be version 0.6 or 1.0?

rachitralli:
@LornaSIL 
this keyboard different from the rac_shina keyboard, this is Kashmir Shina, spoken in Kashmir and the rac_shina is spoken in Pakistan, the letters and diactrics and vowels are also different.
Yes both keyboards use the scl language code.
No this is not supporting the Kashmiri language (kas)
Yes there is a Kohistani Shina language (plk) but this keyboard is different and developed the Kashmiri-shina spoken in Kashmir.
You are right that there is not a Kashmiri Shina language although it appears the Shina language is in the Kashmiri province.
This is  Shina language spoken in Kashmir and I named this language as Kashmir-Shina to differentiate it for Pakistani shina language.
Yes, you are right that the keyboard_info file says that this keyboard deprecates the "Shina" keyboard. But no such keyboard is on the Keyman site. this keyboard supposed to deprecate the rac_shina keyboard and this be version 1.1
Note: the special characters of shina language spoken in Pakistan are different from Kashmiri-Shina spoken in Indian Occupied Kashmir State.

rachitralli:
@LornaSIL both the keyboard are needed, I agree with your suggestions, please proceed. thanks

LornaSIL:
@rachitralli Thanks for the response. I'm now looking at the keyboard layout which I will attach here.

image

* I am surprised at the keys that output a base character plus a sukun (noon, jeem, etc). There are 4 of these. Can those be removed and we just add a keystroke for sukun? 
* I am concerned about the two shift keys (% and &) that output U+FEF1. We shouldn't be using characters in the presentations form block. Scheherazade does not support presentation forms either. FEF1 is "ARABIC LETTER YEH ISOLATED FORM" and I believe we should be using the U+064A character. Is it okay to change that?
* There is a key for yeh barree + ^ (I) and reh + ^ (l). You already have yeh barree on the "y". And you have U+065B on the "F" key.
* Then there is a key with U+0622 + maddah (Q). I've never seen two maddahs on an alef. Is that really needed or can we remove that? Is there a need for us to add maddah to be on a keystroke by itself?
* On the "}" key you have waw + U+0657 + U+0657. I don't think you would want two of them (0657) do you? Should I remove one? There is already waw + U+0657 on the "$" key, so maybe we could remove what is on the "}" key?
* In place of some of the above keys would you like me to add the dagger alef (above and below) on two of those keys?

Do you see any other changes you would like me to make?

rachitralli:
@LornaSIL The characters plus sukun are the special characters of Shina language spoken in Indian occupied Kashmir, the Unicode consortium has not approved these characters, there is no keyboard available in this language, I have added these special character to facilitate the community to write in their language. If you want to remove these special characters then the user fist type the Noon then sukun, this is time consuming. Therefore I have added letters plus sukun to type the special characters of shina spoken in Kashmir. This is up to you, if you want to remove, I have no objection but keystroke for sukun may kindly be added.

👍 With regard to two shift keys (% and &) that output U+FEF1. please change that as desired.

: You are right that there is a key for yeh barree + ^ (I) and reh + ^ (l). these are also the special characters and the Unicode consortium has not approved these special characters, please change as desired.

: double maddah is also the special character of Kashmir Shina, you are right that yoy have never seen two maddahs on an alef. This is really needed but if you want to remove this please check the Alif maddah after inserting another additional maddah.

:On the "}" key you have waw + U+0657 + U+0657. this is Kashmir-Shina waw, this is Ok, there is no Unicode available for this waw, I have created this waw for shina.


image

rachitralli:
image

image

LornaSIL:
@rachitralli I've left most of the keys as you requested. I changed the reh with caret to U+06EF.

You can test the keyboard here: https://drive.google.com/open?id=190TDSjjpBm2w0GqASLv049yw2vydEnsh

Let me know if you need changes.

I'll submit a new PR when you let me know it's okay.

LornaSIL:
PR #602 replaces this keyboard

PR #303, opened 2018-05-24
[kmhmu_2008] migration to github

andjc:


LornaSIL:
* README.md
     * I think it should be version 1.1 (not 1.2)
     * Add a description (Lao-script Kmhmu language keyboard)
* Is the usage.htm file necessary? If not, please delete.
* kmhmu_2008.xml file necessary? If not, please delete.
* In readme.htm the link for the welcome.htm file should be changed from ```<a href="welcome/welcome.htm">``` to ```<a href="welcome.htm">```. This is because when it's in the package it is no longer in the welcome folder.
* In welcome.htm in the first two rows (for Y and U) I do not see the same results as what is in the results column.

Everything else seems fine.

andjc:
@LornaSIL 
You wrote: "In welcome.htm in the first two rows (for Y and U) I do not see the same results as what is in the results column."

What are you seeing?

Looking at the relevant rules in he kmn I see:
```
+ "Y" > U+0E81 U+0ECC
+ "U" > U+0E8D U+0ECC
```

And looking at the text in welcome.html using Ishada's Uniview tools, I see:
```
ກ  ‎0E81  LAO LETTER KO
໌  ‎0ECC  LAO CANCELLATION MARK
```
and
```
ຍ  ‎0E8D  LAO LETTER NYO
໌  ‎0ECC  LAO CANCELLATION MARK
```

Which matches. Will do some font testing to see if the issue is font related.

PR #304, opened 2018-05-24
[lao_2008_basic] migration to github

andjc:


LornaSIL:
We have a bit of a problem with the folder naming. There is a folder in ```legacy\l\lao_2008_basic```. This folder has the same name, but it is under ```release```. I think that might be a problem? @mcdurdin ?

You named the PR lao_basic_2008 which might be an alternate option. If so, all the filenames need to change.

Looking in the ```lao_2008_basic.keyboard_info``` file, I see
```    "related": {
      "lao 2008": { "deprecates": true }
     }
```
I wonder if it should also deprecate ```lao_2008_basic``` (the one in legacy). Both folders need a ```DEPRECATED.md``` file.

LornaSIL:
* Even if the folder name/filenames have to change, the keyboard name doesn't have to change.
* I don't know how much it matters, but the readme.html should be renamed to readme.htm (and reloaded in .kps). Actually, when I try to edit it, it seems to be in some binary format.
    * In the link for "keyboard layout instructions" remove the ```welcome\" since it won't be in the welcome folder when it's distributed in the package.
* In the .kps file add the readme.htm in the details tab where it asks for a "Readme file:"
Otherwise, keyboard seems to work as advertised.

mcdurdin:
The PR name and branch name is fairly irrelevant in the scheme of things. I'd prefer to see the filename not simply changing component order (lao, 2008, basic) without a good reason.

* The folder name MUST match the filename of the keyboard, package and project.
* If the keyboard has the same filename as one from legacy, then delete the keyboard from legacy in the same PR. (Obviously, this is assuming the keyboards are related. Otherwise, the new keyboard needs a new name).
* If the keyboard has a different filename, but is intended to replace a keyboard from legacy, then use the deprecated model.

Note: I need to check this a little further; if the keyboard actually does have an identical filename to one in legacy, then the build should not have passed.

Re readme.html, the only files that have the .htm requirement are welcome.htm and usage.htm.

mcdurdin:
Indeed, I see this in the build log:
```
[23:26:01][Step 1/1] Keyboard lao_2008_basic exists in more than one location:
[23:26:01][Step 1/1] /c/BuildAgent/work/612377aa60a1616b/keyboards/legacy/l/lao_2008_basic/lao_2008_basic.keyboard_info
[23:26:01][Step 1/1] /c/BuildAgent/work/612377aa60a1616b/keyboards/release/l/lao_2008_basic/lao_2008_basic.keyboard_info
[23:26:01][Step 1/1] Aborting with error 0
```

See #308; we should re-run the status check when that issue is corrected and fix is merged.

mcdurdin:
@andjc can you rebase on current master for this branch so we get the latest build script updates from #309?

mcdurdin:
I did the rebase and push to validate the build was working; have now removed the legacy keyboard as well. So you'll just need to pull the changes in to your local copy ,aDnrew.

LornaSIL:
@andjc 
It all looks good. There's a DEPRECATED.md file in ```legacy\l\lao_2008_basic```. Since we deleted the legacy keyboard I guess we don't need that. Can that be deleted? (And presumably the folder?)

andjc:
Deeleted. Oddly it didn't exist in my local repo any more, and was unable to push or pull changes. Had to deelete directly form github. An odd side effect of the rebasing perhaps.

LornaSIL:
URL will be: https://keyman.com/keyboards/lao_2008_basic once the merge is finished.

PR #310, opened 2018-05-28
Khmer angkor updates

MakaraSok:
Updates:

- [x] update the EN/KH manual
- [x] add pressing the spacebar twice for a white space
- [x] add script and region tags to the .kps
- [x] add Mondulkiri font to the .kps

Known issues:
- the RALT bring the cursor out of the text edit preventing one to type, unless ESC.


LornaSIL:
* ```.kvk``` file should be deleted from the repo since now you are using ```kvks``` source
* Typo in welcome.htm: "These make Khmer Angkore uniquely..." change "Angkore" to "Angkor" (do the same fix in the help file)
* It might be useful to include all the Khmer Mondulkiri faces and not just regular. (http://scripts.sil.org/cms/scripts/render_download.php?format=file&media_id=Mondulkiri-7.100&filename=Mondulkiri-7.100.zip)

LornaSIL:
@MakaraSok I made some suggestions which need to be addressed before I can approve the keyboard.

DavidLRowe:
Copyright dates in README.md, LICENSE.md, don't seem to agree with 2015 date in HISTORY.md. Shouldn't the date range be 2015-2018?

See @LornaSIL 's previous comments about typo in welcome.htm and need to delete .kvk file from repo.

MakaraSok:
@LornaSIL @DavidLRowe Updates have been made per your suggestions.

DavidLRowe:
In the path to the font files: `release/shared/fonts/khmer/mondulkiri-7.100/` please remove the version number from the folder name: `release/shared/fonts/khmer/mondulkiri`. Otherwise a new release of the font would mean making a new folder and changing all the keyboards that use it.

DavidLRowe:
The four .ttf files plus FONTLOG.txt, OFL.txt and OFL-FAQ.txt appear twice: once in `khmer_angkor/source/` and again in `release/shared/fonts/khmer/mondulkiri`. The first set should be removed from `khmer_angkor/source/` and all references in the package (.kps) should be to the shared folder.

DavidLRowe:
The folder structure is a bit different from what is recommended:
- welcome.htm and the files it references (currently in source/) could be moved to a new source/welcome/ folder
- the documentation/ folder and its contents could be moved up a level (so that you have khmer_angkor/documentation/ instead of khmer_angkor/source/documentation/ )

MakaraSok:
> the documentation/ folder and its contents could be moved up a level (so that you have khmer_angkor/documentation/ instead of khmer_angkor/source/documentation/ )

The documentation here doesn't belong to the keyboard itself, but the font family: Mondulkiri.

DavidLRowe:
Almost there!
- khmer_angkor.php should not be in the package (.kps)
- welcome.htm has references to .pdf documentation files that are located in the help/ folder (rather than the welcome/ folder). These links don't work when the package is compiled. Instead the .pdf files should be in the welcome/ folder.

PR #311, opened 2018-05-28
[mozhi malayalam] migration to release

andjc:


LornaSIL:
* Added DEPRECATED.md file to all 4 legacy mozhi keyboards folders
* You could add this example text (taken from mozhi_1_1_0.keyboard_info) to the new ```keyboard_info``` file
```      "example": {
        "keys": "malayaaLam",
        "text": "\u0D2E\u0D32\u0D2F\u0D3E\u0D33\u0D02",
        "note": "Name of language"
      }
```
* Readme.htm 
     * Change title to Mozhi Malayalam
     * remove the link to the Keyman font helper and change text to say ```Keyman Desktop Font Helper``` 
     * I think the keyboard tables could be removed from the readme.htm since the welcome.htm describes how to type (remove everything after fonts). But do we maybe need to move the atomic chillu information to welcome.htm?
* OSK
     * Add `, digits, -, = [, ], \ to default layer (keyboard supports them)
     * Add ^ to shift layer (on 6) and + (above equal)
     * Add {, }, |, :, ", <, >, ? to relevant keys on OSK as the keyboard supports them.
* In the welcome.htm there is a link: "Click here to see other keyboard layouts for Malayalam. " That link doesn't go anywhere. I think there's a way for us to search by language.
* It seems like they would want to type digits. If the keyboard doesn't support digits we could add them to the unshifted layer and then regenerate the screenshot used in welcome.htm. I noticed I can't see the whole character in the OSK or the exported graphic. One way around this is to add a space after the text in the OSK and it makes the character show up smaller so it fits on the keycaps. That's a bit of a hack, but currently the best solution.
* You can delete this file ```old.mozhi_malayalam.kmx``` from the PR, unless there's a reason you included it?

LornaSIL:
I think the reason the build is failing is in the keyboard_info file the languages section (lines 3-9) should be 
```
    "languages": {
         "ml-IN": {
            "example": {
               "keys": "malayaaLam",
               "text": "\u0D2E\u0D32\u0D2F\u0D3E\u0D33\u0D02",
               "note": "Name of language"
            }
        }
    },
```

I'll check other things separately.

LornaSIL:
You could add chillu chart to help file too if you want.
Otherwise keyboard seems to be working ok. readme.htm and welcome.htm changes are fine.

PR #314, opened 2018-05-31
[sil_lepcha] port source to github

andjc:


LornaSIL:
* HISTORY.md needs keyboard name
* LICENSE.md need copyright date and copyright holder
* Since you removed the font (at my suggestion), the readme.htm needs modifying. Maybe link to the Mingzat font (http://scripts.sil.org/Mingzat) and link to https://smartfonts.com/noto-sans-lepcha.font
* There's a problem with 4 cells in the "Consonant Conjuncts" table of the welcome.htm. It's displaying some of the html code starting at line 315.
* Change title of welcome.htm to "Lepcha (SIL) Keyboard"
* In the Consonant Conjuncts table in both welcome.htm and help file the "AY" should be changed to "aY"
* Unless the font is installed the help file displays squares. Thus, I think it is best to do screenshots for the help file since it's on the web. 
* The documentation is a bit hard to follow if one doesn't know the script. Whoever wrote it should consider updating it to be more readable (but we'll leave it for now).

Otherwise it looks fine.

PR #316, opened 2018-05-31
Update Keyman Developer compiler to 10.0.1099.0

mcdurdin:
This PR goes along with keymanapp/keyboards-starter#5


darcywong00:
The CI server is failing with
```
lazuri.kps: Warning: The keyboard lazuri has a non-canonical language ID "lzz" (lzz), should be "lzz-Latn".
```

mcdurdin:
Note: now fixing the related compile errors that arise from more stringent checks of language tags.

darcywong00:
Adding script tags in the .kps files raises 2 more questions:
1. Won't we need to update the package versions?
2. Does this affect the language tags in the .keyboard_info files too?

mcdurdin:
1. Yes. Good catch.
2. No, at this stage I don't mind if they diverge just on script metadata... We could discuss this further but it shouldn't block this PR.

PR #317, opened 2018-05-31
[sil_vai] Source migration to github

andjc:


LornaSIL:
* In the welcome.htm file and in the help file
     * We could probably remove the sentence "The keyboard was developed for Keyman 6.0 (or greater) on any keyboard (US, UK, French, etc). "
     * please remove the sentence "For complete instructions on keyboard installation and usage, go to http://scripts.sil.org/KeyboardInstallationAndUse#KM_install."
     * In table "Other" the Keystroke should be ```X``` (uppercase) instead of ```x```
* For the help file, it would be too difficult to do screenshots so we will just leave it using the chart form and hope people have fonts that support Vai.
Everything else seems okay.

LornaSIL:
https://keyman.com/keyboards/sil_vai

Issue #225, opened 2018-04-12
[hieroglyphic] Hieroglyphics keyboard has outdated instructions on Addins


mcdurdin:
Per https://community.software.sil.org/t/display-egyptian-hieroglyphs/766/3, the hieroglyphics keyboard instructions are outdated as they refer to UI elements only present in Keyman 8 and earlier.

LornaSIL:
@andjc has this keyboard on his list of keyboards to move to "release".

andjc:
Do we know whether the Egyptian Hieroglyph Format Controls block will be in the next version of Unicode or a later one?

LornaSIL:
There are not listed for version 11.0.

LornaSIL:
PR #322 fixes this.

PR #322, opened 2018-06-04
[hieroglyphic] migrated to github

andjc:


LornaSIL:
* In welcome.htm 
     * Change ```use the Keyman Font Helper.``` to ```use the Keyman Desktop Font Helper.``` and remove the link. (also change in help file)
     * the link at the bottom should be changed to ```http://jsesh.qenherkhopeshef.org/``` (also needs changing the help file)
* help file
     * bullets at bottom don't seem to be correct.

I tested many of the keystrokes and they seem to work. Metadata seems fine. BCP47 is fine. So, with updating the above, I think it's ready.

LornaSIL:
Fixes #225

andjc:
@LornaSIL, you wrote that in the help file teh bullets at bottom don't seem to be correct. Could you clarify? The markup appears to be identical in welcome and help files. There is an ordered list, no unordered list, so am uncertain what you mean by the bullets appear incorrect.

LornaSIL:
The bullets look like this in welcome.htm:
image

and like this in the help file:
image

I would suggest removing all the bulleted list info in the help file:
image

But, that's the way the "legacy" keyboard help file was so it's not a big deal. Let me know your preference.

andjc:
@Lorna, will remove, that will be easiest. But it's odd, definitely an ordered list in the markup, not an unordered list. Even if there was some problem in the code at some point that. But would like to understand where the bullets are coming from, and why. Its odd.

LornaSIL:
Should be available here soon: https://keyman.com/keyboards/hieroglyphic

PR #326, opened 2018-06-05
[ipauni11-dep] Fixing links in help files and deprecating ipauni11

### LINKED ISSUE #254: [ipauni11] Link to Charis font is broken
@tombogle writes
> The link that is supposed to open a PDF about the Charis font is broken (and no such PDF seems to be in the repo). I think it would be better to just have a link to http://software.sil.org/charis/. The corresponding link for Doulos works, but assuming that PDF is readily available via the Doulos home page, it might b better to just link to that site instead. That gives a better chance of sending the user to the most up-to-date info. There is no link for Gentium. One should be added to take the user to http://software.sil.org/gentium/.

> While we're on the topic of broken links, Marc has suggested:
"It would be helpful to do a consolidated effort on fixing broken links in the help documentation using an automated tool such as Xenu Link Sleuth"

LornaSIL:
* Fixes #254 
* Removed unnecessary files from help folder
* Added DEPRECATED.md file to ipauni11
* Added ```"ipauni11": { "deprecates": true },``` to ```sil_ipa.keyboard_info``` file
     * No changes to sil_ipa package so I didn't bump the sil_ipa keyboard version.


legacy/i/ipauni11/DEPRECATED.md
```diff
@@ -0,0 +1 @@
+This keyboard has been deprecated and replaced by release/packages/sil_ipa (note underscores).
```
darcywong00: The path should be `release/sil/sil_ipa`

PR #331, opened 2018-06-12
[sil_bwe_karen] initial creation of Bwe Karen Keyboard 

sdysart:


DavidLRowe:
This pull request seems to reference `release/sil/sil_ipa/source/welcome.htm` Is that intentional?! It seems as though that file should be deleted from the pull request.

In the .keyboard_info file, we would strongly encourage you to change the name to "Bwe Karen (SIL)" so that the name is sorted by Bwe (rather than SIL) to make it easier for people to find the keyboard when searching the list.

Folder sil_bwe_karen/source/Welcome/ needs to change to lowercase "w": "welcome".

Minor point: Current text in welcome.htm: `...with combining diacritics q is replaced...` might be clearer with the insertion of a colon (or other punctuation): `...with combining diacritics: q is replaced...`

The combining acute accent doesn't appear to show on the screenshot of the unshifted keyboard.

Desktop layout seems to work as documented.

Touch layout, symbol layer
- key with `_` produces `+`
- key with `"` produces `'`

Some additional comments about touch layout:
- It seems that pressing a key on the shift layer takes you back to the default layer, however the longpress `?` and longpress `.` on the `.` key don't (although longpress `,` on that key does).
- The differing number of keys/row between layers means the keys "jump" when switching layers.

sdysart:
Made changes based upon DavidLRowe's suggestions

Regarding the " gives ' issue.  If I understand it right the " was 
assigned the K_Quote designator.  Changed it to be a U_#### with #### 
being unicode for the "

Not sure what to do about the symbol layer having different number of 
keys.  The user I'm working with thinks it's okay, but it might become 
obnoxious for someone using it consistently.

Scott

DavidLRowe:
The different number of keys per row is not a show-stopper. You (or your user) can always make a version 2.0 with changes based on user feedback. Will try to look at this tomorrow.

DavidLRowe:
Should be available here soon: https://keyman.com/keyboards/sil_bwe_karen

PR #341, opened 2018-06-14
[sil_kvl_kayaw]  Initial Creation

sdysart:
Initial creation of the Kayaw Keyboard for Keyman.  This has a Touch Screen keyboard as well as Desktop Keyboard.


DavidLRowe:
Copyright in README.md has 2002-2018, but in .kmn, .kpj, .kps and LICENSE.md (and HISTORY.md), date is just 2018. Date(s) of copyright in all places should agree, and should cover from date of first version to current work.

.kmx, .kvk, .kmp files should not be included in the source folder (or anywhere in the pull request). 

welcome.htm refers to `kayawUnshifted.jpg` but file is really named `kayawunshifted.jpg` so, although it works on Windows, it fails on Linux (and perhaps Mac as well). 

@darcywong00 Are the two .jpg files in welcome/ also needed in help/ (since help.php references them)?

Touch layout, symbol layer, there are two keys that produce : 
- "=" produces - (U+002D)
- "_" produces + (U+002B)

Other notes (not necessarily things to change)
- Keyboard installs as `Brek Karen (kvl-Latn-MM) - Kayaw`
- On shift layer, most keys take you back to default layer immediately after the key is typed. The exception is the top row !@#$%^&*()_  where typing these keys leaves you on the shift layer. 
- There are different numbers of keys/row on the different layers so keys "jump" a bit switching layers.

DavidLRowe:
@LornaSIL CharisSILCompact-R.ttf is included in the .kmp. Should a copy of the OFL be included as well? What else? 

@mcdurdin Is there anywhere we document what needs to accompany a font that's included in a .kmp?

darcywong00:
> Are the two .jpg files in welcome/ also needed in help/ (since help.php references them)?

I think we can replace those references similar to what sil_bwe_karen has:

```
<div id='osk-container'>
  <div id='osk' data-states='default shift'></div>
</div>

  <div id='osk-phone' data-states='default shift Symbol'></div>
</div>

  <div id='osk-tablet' data-states='default shift Symbol'></div>
</div>
```

darcywong00:
`source/help/sil_boonkit.php` needs to be renamed to `source/help/sil_kvl_kayaw.php`. I'll do that when I replace the layout jpgs

mcdurdin:
> @mcdurdin Is there anywhere we document what needs to accompany a font that's included in a .kmp?

@DavidLRowe, no I don't think is documented anywhere. That information really depends on the font, e.g. all SIL OFL fonts need the OFL license text as well and a readme. But that's certainly not consistent. In cases where that is not so obvious, perhaps we include a README.md in the same folder as the font file?

Of course, including the OFL in the KMP hardly makes it accessible, as it is installed into a hidden folder on Windows, and basically completely inaccessible on iOS/Android. But let's not get into that at this time.

DavidLRowe:
welcome.htm still has reference to "kayawUnshifted.jpg" which fails on Linux since the file is named "kayawunshifted.jpg". Reference in the welcome.htm file needs to change.

Other changes look good.

DavidLRowe:
@sdysart When including an SIL font, such as Charis SIL Compact, it is customary to include the OFL.txt file containing the OFL license. 

You could look at https://github.com/keymanapp/keyboards/tree/master/release/sil/sil_ipa/source for an example. There OFL.txt, OFL-FAQ.txt, FONTLOG.txt and the README.txt as well as the Regular, Bold, Italic and Bold-Italic .ttf files were included.

It's a difficult tradeoff between providing complete information and keeping the resulting distribution file small.

sdysart:
David,

I've just added the OFL related .txt files that are in the .zip file 
available on the SIL website for Charis SIL Compact.  They will be 
included in the next update.

Scott

LornaSIL:
Make sure the name is consistent across all the files. I didn't do a review, but just looked at some of the files online. It probably should be "Kayaw (SIL)". I also saw "SIL Kayaw" and also "Kayaw".

DavidLRowe:
The changes to the desktop layout to add more accent keys needs to be reflected in the welcome.htm and help.php. (BTW, from Keyman Developer, On-Screen tab, there's an Export... button at the top that allows you to export .png images of the keyboard layouts.)

These changes also mean that the touch layout is wrong. For example, on the default layer, the "x" key is labeled U+0302, but actually produces U+0302 U+030C. And on the shift and Symbol layers, the key labeled "& actually produces "7".

Now I'm wondering if I was premature in reviewing this keyboard. Perhaps you're still in the middle of making updates. Let me know when you want another review.

LornaSIL:
Yes, updating the desktop layout will affect the mobile layout if the
mobile layout refers to [K_T], etc. So, if you change what is on the "t"
key, then the mobile layout changes too. However, if you use U_0660 (or
whatever Unicode value) on the mobile layout, then the mobile layout will
not change. I've been bitten by this multiple times.

darcywong00:
Any thoughts on moving the Charis SIL Compact font files to **release/shared/fonts/sil/sil_charis_compact**?

DavidLRowe:
@darcywong00 Are there other keyboards that include Charis SIL Compact? If so, then there could be some space savings by referencing a common shared font. There would also then be the advantage of only needing to update the font in one place, rather than multiple places.

Lorna won't be back until later this week. We can raise the issue with her then.

DavidLRowe:
- Please eliminate the README.txt file from this PR and reference `release/shared/fonts/sil/charis_sil_compact/Charis_SIL_Compact_README.txt`. 
- Similarly you can include `release/shared/fonts/sil/charis_sil_compact/FONTLOG.txt`
- Is there any reason not to include the bold, italic and bold-italic .ttf files as well as the regular?

Otherwise the PR is ready to go.

DavidLRowe:
The .kpj file seems to reference only the .kmn and .kps files (in contrast to previous commits where the .kpj file referenced font files, build items, welcome.htm, etc). Everything compiles fine with Developer (10.0.1200.0) but it saves a .kpj file with lots more files specified, in fact all the ones in the previous commit, with the exception of those deliberately changed. 

I'll go ahead and approve this PR because it seems to work correctly, but was curious about the difference.

PR #342, opened 2018-06-14
[arabic_izza] Improved phone layout-final

MaleBou:
- Corrected the five Keys ("، |، {، } ، [) on numeric layer that weren't working because of small letters used instead of capital ones to identify them.
- Swapped some keys on third line of default layer to match those on physical keyboard.
- Put letter ة with ه, and the comma in place of ة, and full stop in place of comma in the fourth line.
- Added a key on the fourth line of the numeric layer, that would enable shifting back to default layer.


DavidLRowe:
(1) Phone layout, default layer:
4th row down, 3rd key from left: text on the key is ".", but it produces U+0648. Current code is:
```
"id": "K_COMMA",
"text": ".",
```
which invokes the .kmn rule:
```
+ "," > U+0648 
```
I think if you add `"layer": "shift"` to make
```
"id": "K_COMMA",
"text": ".",
"layer": "shift",
```
it should invoke the .kmn rule:
```
 + "<" > U+002E
```
which should give the "." that corresponds to the "text".

(In Keyman Developer, for the Touch Layout, there are two ways to do it. On the "Design" tab, you would change the "Modifier" from "(layer default)" to "shift". The equivalent change on the "Code" tab would be adding the `"layer": "shift"` line.)

(2) Phone layout, numeric layer (minor point):
1st row, 1st key on left ("1") has
```
"sp": "1"
```
which causes the "1" key to be a different colour than the normal keys. I think you should remove this. (In Keyman Developer, for the Touch Layout, there are two ways to do it. On the "Design" tab, you would change the "Key Type" from "Special" to "Default". The equivalent change on the "Code" tab would be removing the `"sp": "1"` line.)

Everything else looks good!

MaleBou:
As you said the changes are quite simple:
(1) 
"id": "K_COMMA",
"text": ".",
"layer": "shift",
(2) 
"Key Type" from "Special" to "Default"
Perhaps no need for a Pull request.

DavidLRowe:
The best way to proceed is for you to make these changes and push them to the arabic_izza branch on github.com/MaleBou/keyboards.git Then those changes will be integrated into the existing pull request. Once they are in place, that pull request can be merged and the new arabic_izza keyboard published.

There also needs to be a change of version (from 1.1 to 1.2 for example). That, unfortunately shows up in quite a few places: 
```
arabic_izza.keyboard_info
arabic_izza.kpj
HISTORY.md
source/arabic_izza.kps
source/arabic_izza.kmn
```
The version number change lets devices know that there is a new version so that they can update automatically.

MaleBou:
Are there any easy steps to push the changes?

MaleBou:
I have already changed it in layout:
store(&KEYBOARDVERSION) '1.2'

MaleBou:
I have just updated:
arabic_izza.keyboard_info
and
HISTORY.md
Is that enough before pushing the changes?

DavidLRowe:
@MaleBou You would push the changes to HISTORY.md and arabic_izza.keyboard_info (and other files below where the version needs to change) in the same way that you pushed the changes to arabic_izza-layout.js and arabic_izza.kmn.

The changes you've made to arabic_izza-layout.js and arabic_izza.kmn look good. Now we just need to get the version number updated. (We're sorry that the changes to the version number are needed is so many places. The developers are working on reducing that.)

The version number change (from 1.1 to 1.2) is needed in these files:
- arabic_izza.keyboard_info
- arabic_izza.kpj (The version number appears in two places in arabic_izza.kpj; one has been changed to 1.2, but the other is still at 1.0.)
- HISTORY.md (I think you've already changed this file, but the changes need to be pushed)
- source/arabic_izza.kps (The version number appears in two places in arabic_izza.kps; one is at 1.1, the other at 1.0 --  I guess I missed it last time!)

MaleBou:
I have pushed the last updates for version number 1.2.

PR #343, opened 2018-06-14
[el_nuer] refactored and updated layout, including cross platform support

andjc:


LornaSIL:
The naming of this keyboard is inconsistent.
* ```Nuer``` README.md, HISTORY.md, readme.htm, welcome.htm and help file
*  ```Thok Nath``` .kpj .kmn 
* Both ```Nuer``` and ```Thok Nath``` in different places in .kps
* On the Desktop layout 
     * unshifted: for the ```z``` key, I am getting a ```z``` and not a ```;``` as the welcome file indicates I should.
     * shifted: for the ```Z``` key, I am getting a ```Z``` and not a ```:``` as the welcome file indicates I should.

andjc:
@LornaSIL , no problems i think i know where the problem is. In theory it should be working, but i suspect I know where the problem lies.

LornaSIL:
It is much improved. I'm still having a few issues.

* On phone layout - default and shift layers
     * longpress epsilon, I'm not getting epsilon with diaeresis nor epsilon with macron and diaeresis
     * I don't see the ellipsis on the phone layout
* On tablet
     * default and shift layers: longpress epsilon, I'm not getting epsilon with diaeresis nor epsilon with macron and diaeresis
     * longpress capital open o, I'm getting a **capital O** with macron instead of **capital open O** with macron.
     * shift layer: the combining macron and combining diaeresis aren't working (these aren't really needed are they, since you've added characters to the longpress keys.

andjc:
@LornaSIL re the diacritic keys on tablet, i added them since it is quicker and easier to type than longpress esp. for users familiar with desktop equivalents. Added longpress to keep it compatible with mobile phone layout which i wnated fewer keys on.

LornaSIL:
Will be available here when the build is completed: https://keyman.com/keyboards/el_nuer

PR #344, opened 2018-06-14
[sil_lisu_basic] port to github repo

andjc:


LornaSIL:
Everything looks good except that in the .kmn you have added a "Touch-Optimized keyboard" for mobile use, but none of the keycaps have the Lisu characters. I think you should delete that from the .kmn file and then delete this file: sil_lisu_basic-layout.js from the repo.

LornaSIL:
Does this keyboard support the whole Lisu block? If so, we should consider adding the other languages that use the script (atb, lbc, lpo, nxq) - would need to add to .kmn, kps and keyboard_info.

andjc:
@LornaSIL will check coverage.

andjc:
@LornaSIL  forgot to comment out the Touch-Optimized keyboard"  was playing with options that I haven't pushed. The scaffolding is in place just need to decide on a final design, in terms of how best to handle tones, ie put all tones on default/shift layers or add a single key with longpress for each tone.

PR #345, opened 2018-06-14
[sil_lisu_standard] Port source to github repo

andjc:


LornaSIL:
* The name in .kmn, .kps, kpj should be changed from ```Lisu Standard``` to ```Lisu Standard SIL)```
     * welcome.htm, readme.htm, README.md, HISTORY.md all have ```(SIL)``` in the name.
* No fonts are in this package. Should there be?
* There is no help file

Does this keyboard support the whole Lisu block? If so, we should consider adding the other languages that use the script (atb, lbc, lpo, nxq) - would need to add to .kmn, kps and keyboard_info.

PR #347, opened 2018-06-15
[sil_kayah] Initial porting of source code to git

andjc:


LornaSIL:
I think I would make this version 1.4 based on the weird versioning in the comments in the .kmn file. Somebody got mixed up and I have no idea if the 1.2 and 1.3 versions were ever published.

LornaSIL:
The keyboard works fine. 

I'm rethinking the keyboard naming. It looks like they have a keyboard for Mymr and Latn scripts too.
Should we do:
* Kayah [Kali] (SIL)
* Kayah [Mymr] (SIL)
* Kayah [Latn] (SIL)

Or do you have another suggestion?

andjc:
@LornaSIL my understanding is that the history in the kmn was a mix of the pre Unicode (PUA) version and the Unicode version. I think PUA got to 1.3 then Unicode restarted at 1.0 and 1.1 but I could be very wrong

andjc:
I would use full script name rather than script code.

LornaSIL:
Since the name is changing anyway, I think version is okay. I'm happy to
have full script name If you think that's best.

LornaSIL:
@andjc there is a conflicting file (el_nuer.kpj). Once you remove that the keyboard seems good to go.

Issue #340, opened 2018-06-14
[cheyenne] Deprecate keyboard and replace with chey-unicode


darcywong00:
From https://community.software.sil.org/t/please-delete-old-cheyenne-keyboard/987
 > Please delete the old Keyman Cheyenne keyboard from this webpage:
>
>https://secure.tavultesoft.com/keyman/downloads/keyboards/search.php?Submit=search&Search=cheyenne&x=34&y=16
>
>I am working with the Cheyenne tribe right now and they are confused about which keyboard to download for typing their language. The old keyboard was designed before Unicode was available and required use of three special fonts designed by SIL. Several Cheyennes have downloaded both keyboards but they find this confusing. Since most Cheyennes today do not have the old SIL fonts, the old keyboard produces wrong results.
>
>It would be fine to rename the remaining “Cheyenne - Unicode” keyboard to simply “Cheyenne”.

------------------

Also, from a separate PM.

Currently, their Keyman Android users use the European2 keyboard (listed under the "Aleut" language). We should be able to associate "Cheyenne" (chy?) in the metadata for the keyboard.

LornaSIL:
Wayne sent me the source file for Chey-Unicode at the end of April and gave permission to move it to MIT license. It just hasn't made it to the top of my priorities. I'll work on doing this.

darcywong00:
More input from Wayne if we don't remove `Cheyenne`.

> If you keep the old Cheyenne keyboard, please try to position it on the Cheyenne keyboard page so that it does not show up right away. Please rename it to something like “Cheyenne - old keyboard that requires fonts no longer available”. Then please rename the good keyboard as simply “Cheyenne”. Cheyennes do not understand technical labels so they are confused by words like Unicode. We have been using Keyman for many years. Cheyennes are finding it helpful as they become more computer literate. Thank you to all the Keyman team for maintaining Keyman and its keyboards for many years with different operating systems

mcdurdin:
Copying my comment from community discussion:

Here’s my thoughts on this:

1. I’m not keen on removing keyboards because there will always be someone who needs to edit old data.
2. We can certainly make it much clearer that the Cheyenne (non-Unicode) keyboard is obsolete. We can update the keyboard_info for both keyboards and that should make this a lot clearer.
[...]

PR #348, opened 2018-06-16
[sil_cheyenne] new keyboard to replace old ones

LornaSIL:
- Added OSK
- Added mobile layout
- Fixed code which produced diaeresis instead of macron on a few characters.
- Added documentation
- Deprecated chey-unicode
- Attempted to make changes to "cheyenne" (legacy keyboard) to make it more obvious not to use the keyboard.


LornaSIL:
Closes #340

DavidLRowe:
When the build completes, this keyboard will be available at: https://keyman.com/keyboards/sil_cheyenne

PR #351, opened 2018-06-18
[ansi_obsolete] Ansi obsolete

LornaSIL:
* I updated the keyboard_info file to indicate these ANSI keyboards are obsolete. If a Unicode keyboard is available, I gave a suggested alternate URL.

* I changed the mohawk keyboard from "ansi" to "unicode" in the metadata as it turns out it is a Unicode keyboard.


LornaSIL:
@mcdurdin I'm not sure why the check failed. All I did for ```mohawk_u.keyboard_info``` was to change the encodings from "ansi" to "unicode" which seems to be a valid keyterm.

LornaSIL:
Because I'm not sure how the Kayah files got in this PR I am closing it and this PR is replaced with PR #375

PR #353, opened 2018-06-19
[sil_yupik_cyrillic] Sil yupik cyrillic

JBrase:
For ess-Cyrl (Central Siberian Yupik written with the Cyrillic script).
This keyboard has been in use for about 12 years, though it has not been used heavily. (The SIL language project worked primarily in the Latin script.) The changes made at this time were primarily focussed on reorganizing the source files and updating the documentation.


DavidLRowe:
Consider changing the name to "Yupik Cyrillic (SIL)" (while leaving the project name as "sil_yupik_cyrillic"). This will cause the keyboard to appear in the list under "Y" (as "Yupik..."), rather than under "S". This appears in the .kmn file (&NAME) as well as in the .kpj and .kps files.

Here's a really picky point: The welcome.htm text says "...any ASCII character can be typed by pressing the + key followed by the ASCII key" isn't strictly true, since it only applies to the keys whose normal value on the US English keyboard has been replaced by a Cyrillic character. But I can't think of a better, concise way to state that. Wait! Your sil_yupik_cyrillic.pdf file says: "The original character assigned to a remapped key can be input by typing an “+” in front of it." What would you think of using that in the welcome.htm?

You might also want to use the option to have the package version number follow the keyboard version number. This checkbox option is in the Package Details. It reduces one or two places where the version number needs to be changed when updating the keyboard.

Otherwise, everything looks really good.

DavidLRowe:
- Excellent CapsLock treatment! (Most keyboard authors just ignore it.) Now I wonder if the following line should now be removed from the .kmn file:
```
c I want this to behave like a standard US keyboard, but it insists on generating upper case for the SHIFT + CAPS combination.
```
- The Keyman package source (.kps file, "Files" tab) contains a reference to sil_yupik_cyrillic.pdf which no longer exists in the docs/ folder. This causes the build to fail. This reference should be removed.
- There is currently no help file (help/sil_yupik_cyrillic.php).
  - We could approve this keyboard without the help file, then make a separate pull request to create it from the welcome.htm file (which would mean making an update to the keyboard and its version number).
  - Or you could look at (for example) release/sil/sil_boonkit and compare source/welcome/welcome.htm and source/help/sil_boonkit.php to see how to take the contents of the <body> HTML element from the welcome.htm file and stick a PHP header on it to create the .php file. @LornaSIL is the expert in this area!

DavidLRowe:
More on the .php file: The welcome.htm has a lot of style information in the `<head>` element. Not sure how to get that into the .php file. Sounds like a job for @LornaSIL !

LornaSIL:
Doing a separate PR for the help file thankfully does not require a bump in
version number.

So, we can easily do that for you.

DavidLRowe:
Will be available here when the build is completed: https://keyman.com/keyboards/sil_yupik_cyrillic

PR #359, opened 2018-06-22
[mayan_us] Initial version of mayan_us keyboard

philleckrone:
This is a trial of the keyboard used with Mayan languages of Guatemala. It is based on the US keyboard


release/m/mayan_us/mayan_us.kpj
```diff
@@ -0,0 +1,96 @@
+<?xml version="1.0" encoding="utf-8"?>
+<KeymanDeveloperProject>
+  <Options>
+    <BuildPath>C:\keyboardprojects\keyboards-starter\mayan_us\release\m\mayan_us\build</BuildPath>
```
darcywong00: Can you make the BuildPath relative?
```
<BuildPath>$PROJECTPATH\build</BuildPath>
```

The build servers use a different folder.

DavidLRowe: - As Darcy indicated, the build path needs to be set to `$PROJECTPATH\build`. This can be set in Keyman Developer from the Project menu, Project Settings menu item.
- See the "ToDo" section at the bottom of the README.md file for a list of things to complete in the README.md, HISTORY.md and LICENSE.md files
- The list of languages in the mayan_us.keyboard_info file should match the BCP 47 codes contained in the mayan_us.kps file (Keyman package source).
- In the Keyman package source (.kps)
  - it is now recommended that you tick the box (on the Package Details tab) to have the package version follow the keyboard (.kmn) version. This is helpful for future updates since it means one less place the version number needs to be changed.
  - The .kmn and .ico files should not be included in the .kps file (they have already been compiled into the .kmn file, so these source files are not needed).
- In the mayan_us.kmn file, the `store(&NAME)` doesn't have to be `'mayan_us'` but can be something more readable if you want.
- Touch layout:
  - default layer
    - key labeled '?' produces '/'
  - shift layer
    - key labeled '\', the five underlined capital vowels, dieresis over S, N, Z, and the accented dieresis-U produce no output. The problem with the '\' is because the key id is 'K_b' (a code that doesn't exist, because the 'b' is lowercase) and should be 'K_BKSLASH' along with "layer": "default". The problem with the others is that you reference 'T_xxx' on the shift layer, so the reference is to [SHIFT T_xxx] in the .kmn file. You need to include "layer": "default" to access the .kmn rule you want. (Note that dieresis-X worked because the reference is `"id": "U_1E8C"`.)
  - minor point: the default and shift layer key layouts and those of the numeric and symbol layers are slightly different so the keys appear to 'jump' a bit when changing from default to numeric. (That could be left for a later revision of the keyboard.)
- There is currently no help file (help/mayan_us.php). We could approve this keyboard without the help file, then create it from the welcome.htm file in an update to the keyboard.

PR #363, opened 2018-06-26
[sil_kayah_mymr] migrate source to Github

andjc:


LornaSIL:
* readme.htm is empty
* Add to .kps: ```sil_kayah_mymr.js``` (it's being created, but not included in the package).

Otherwise, keyboard works as advertised.

andjc:
Didn't include it in package because it is only for Web (desktop) not for touch devices, but can add it if kmp is being used as delivery system for KMW now.

LornaSIL:
This is fine now unless we find out we need to add the .js file.

DavidLRowe:
Darcy's comment: "ATM, it's only meaningful for touch layout js files to be in the kmp since kmw doesn't use packages."

LornaSIL:
I didn't realize it before, but there is a file conflict for the readme.htm.

LornaSIL:
"Fixing" the conflict lost all the files except for the readme.htm.

LornaSIL:
When built, it will be available here: https://keyman.com/keyboards/sil_kayah_mymr

PR #364, opened 2018-06-26
[sil_kayah_latn] Migration of source

andjc:


LornaSIL:
* HISTORY.md ```* Source migrated ot GitHub``` - fix "ot" > "to" (not a showstopper if this is the only issue)
* I think the .kps needs to include sil_kayah_latn.js so it goes in the package. It is building it, it's just not including it in the package.

Everything else looks good.

andjc:
If layout does not support touch devices, is js file required in kmp?

LornaSIL:
Changes are good. Will wait to hear if we need to add .js to package.

LornaSIL:
When merged, keyboard will be available here: https://keyman.com/keyboards/sil_kayah_latn

PR #375, opened 2018-07-03
[ansi_obsolete_new] Obsolete ANSI keyboards

LornaSIL:
* I updated the keyboard_info file to indicate these ANSI keyboards are obsolete. If a Unicode keyboard is available, I gave a suggested alternate URL.
* Despite Mohawk having Unicode in the name @mcdurdin assures us it is ANSI.


darcywong00:
Would it be clearer to name all of these `(obsolete non-Unicode)`? atm I'm seeing 3 variations on this PR.

PR #376, opened 2018-07-03
[sil_nigeria_odd_vowels] Initial commit of Nigeria Odd Vowels (SIL) k…

LornaSIL:
…eyboard

Deprecated old keyboard

Please approve all 3 nigeria keyboards together as I've linked them together in the readme.htm file.


LornaSIL:
* From .kmn
     * Added rules for nasal mM
     * Added 4 lines to NFC section to make oOuU with double acute to be composed characters. 
     * Changed rule for three apostrophes.
     * Removed vnNeedTone store as it is not used
* In .kvks removed the one unneded key in the ctrl level.
* In welcome.htm:
     * Added documentation for ```^``` deadkey
     * Modified .css to use standard fonts instead of strange Tibetan fonts.
     * Removed text about "Deadkeys may be combined to get two combining marks..." as that is documented elsewhere and this was incomplete.
* Help folder:
     * Deleted .css as it is unnecessary
     * Removed unnecessary head and body tags.
     * Copied text from welcome.htm into the php to pick up editorial changes.
* Deleted .css in source folder as it is unnecessary
* Removed ```\"``` from README.md and keyboard_info files

PR #380, opened 2018-07-04
[rac_dameli] Shift key is not working

rachitralli:
The touch layout, shift layer, shift key doesn't take you out of shift layer, so there's no way to get back to default layer! Please enable the shift key so that the user can get back to the default layer. Thanks


DavidLRowe:
Shift now works on touch layout to get from shift layer back to default layer.

Please change `dml` to `dml-Arab` in .keyman_info file. (It's already correct in .kps file.)

WIth that change this should be ready.

rachitralli:
@Makarasok pls remove Arab, the dml is ok

MakaraSok:
@rachitralli I've just updated the touch layout of the keyboard. 

Test it here [rac_dameli.zip](https://github.com/keymanapp/keyboards/files/2196698/rac_dameli.zip).

@DavidLRowe I experience an issue with this keyboard that when you are using the keyboard in one application and then change to other application, the keyboard changes to English. Could you test this and see if you see this same issue.

DavidLRowe:
BTW, the change to `dml-Arab` is still needed in the `.keyman_info` file

DavidLRowe:
@MakaraSok If I have Notepad open and select this keyboard I can type and get Arabic characters (the display in the task bar is "EN" plus the icon for the rac_dameli keyboard). If I then switch to another application, type something, and switch back to Notepad, the keyboard has reverted to English (and the display in the task bar is "EN" plus the generic keyboard icon).

This is in contrast to what happens with other keyboards. 

I removed other keyboards using Keyman Configuration. I then used Windows Settings, Region & Language Settings to delete those other keyboards from the Languages list. After rebooting, the behaviour was the same as above.

When I click on EN in the task bar, I see:
- EN English (United States)
- ZZ Unknown Locale (dml-Arab-PK)

So, I agree that -- when switching from an application using this keyboard to another and then back to the first one -- the keyboard is no longer active, which is not what I would expect (and not what happens with other keyboards).

MakaraSok:
@DavidLRowe I don't know how to fix that, but the team may be able to help diagnose this. @mcdurdin @tombogle

rachitralli:
@DavidLRowe @mcdurdin @tombogle @MakaraSOK I am unable to fix this, please fix this issue

rachitralli:
After the above minor changes suggested by @DaviLRowe this PR is ready to be merged

LornaSIL:
Closed by #549

PR #386, opened 2018-07-05
[bj_oji_cree] Update to keyboard to support Desktop

LornaSIL:
* Got rid of warnings by creating new stores. 
* Added windows and macos as targets.
* kvk > kvks
* changed bmp to ico


DavidLRowe:
Just a few suggestions about comments in the .kmn file:

(1) Remove this comment (since you've fixed the problem):
c compiler says that "The rule will never be matched because its key code is never fired."

(2) I think this section:
```
c initial consonant keys (The Î is U+14D1 which is the sh consonant)
c (The ÿ is U+15AE which is the th consonant that looks like y)

c Remove Î and ÿ from store since it's not on a keyboard (5 July 2018)
store(ConsKey)    "ptkcmnsyf"
store(CapConsKey) "PTKCMNSYF"

c Remove  U+15AE and U+14D1 from store since you can't type it. It will be handled in other rules using "CC" store.  (5 July 2018)
C Add this store to replace "CC" store
c                 p      t      k      c      m      n       s      y       f      
store(CCKey)      U+144A U+141F U+1420 U+1428 U+14BC U+1423  U+1422 U+153E  U+155D 
```
might be clearer as something like:
```
c initial consonant keys which can be typed on keyboard
store(ConsKey)    "ptkcmnsyf"
store(CapConsKey) "PTKCMNSYF"
c                 p      t      k      c      m      n       s      y       f      
store(CCKey)      U+144A U+141F U+1420 U+1428 U+14BC U+1423  U+1422 U+153E  U+155D 
```

PR #388, opened 2018-07-06
[el_dinka] Added cross platform support

andjc:
version 3.0 of layout, cross platform.


LornaSIL:
* Consider changing name in README.md (from "Dinka Basic (DLIA)" to "Thuɔŋjäŋ"
* Consider changing name in HISTORY.md (from "Dinka" to "Thuɔŋjäŋ"
* On the mobile layout (phone):
     * Shift "O" - the double OO with diaeresis is not being output
     * Shift "A" - the double AA with diaeresis is not being output
     * Shift "X" (open o) - the 2nd open o is outputting an epsilon with diaeresis (instead of open o) and the double open o is outputting two epsilons with diaeresis (not double open o with epsilon)
     * numeric layer, the ```"``` key is outputing ```|``` (long press keys are outputting correctly)
     * numeric layer, the ```'``` key is outputing ```\``` (long press keys are outputting correctly)

I can't test the tablet tonight, but it's possible the same errors on the phone are happening on the tablet. I'll check tomorrow.

LornaSIL:
The things I reported last night are fixed.

On the tablet layer there is a key for combining underline, but it doesn't produce anything and you don't use that in the Desktop or phone layout. Is that a remnant from another keyboard? I think it needs removing from both default and shift layers.

PR #390, opened 2018-07-06
[rac_brahui-1] broken pages in keyman help file 

rachitralli:
The rac_brahui help file  is not opening properly please see this https://help.keyman.com/keyboard/rac_brahui/1.0/rac_brahui.php

here is the screenshot 
image


LornaSIL:
@mcdurdin This displays fine in Firefox and Chrome. It looks like he's using Opera and it's truncated. I am getting the same results as he is in Edge. I checked several of the "rac_" keyboard help files. Some are working and some are truncated.

The ones that are truncated still have a ```<body>``` tag in them. Is that the problem?

LornaSIL:
Fixed by PR #397

mcdurdin:
@LornaSIL yes that sounds correct. Sorry for the delayed response; I have been travelling and not keeping up with the email backlog.

PR #396, opened 2018-07-06
[sil_myanmar_my3] migration of Myanmar3 source to GitHub

andjc:


LornaSIL:
* HISTORY.md: type: "Sourrce" > "Source"
* readme.htm: change header and title to ```Myanmar3 (SIL)```
* welcome.htm: change header and title to ```Myanmar3 (SIL)``` (also in help file header and title)
     * What does the footer mean? That the documentation is under Creative Commons?

Keyboard works as advertised.

andjc:
@LornaSIL latest commit should fix issues and also made corrections as per email.

Re question, footer is essentially a creative commons blurb. Who worte doc, who modified it, and which license it was released under.

LornaSIL:
Your fixes look good. When I open the welcome.htm the latin seems conflated (on top of each other). See screenshot. When I open the help php file it doesn't have that problem. Is it using another font than Padauk with no Latin or what could cause this issue?
image

When I changed the font in the welcome.html file to Padauk Book it looks fine:
image


Do you have any ideas?

I think the preference is to have welcome.htm and not welcome.html. If you have to change this file, go ahead and change the name and references to it (kps, etc), if you don't feel the file needs changing, don't feel like you have to change the name.

LornaSIL:
Oh, I had old versions of Padauk installed. Installing the newest version removed this problem. So, I can go ahead and approve this keyboard if you want.

PR #397, opened 2018-07-09
[rac-help] Remove body tag to fix cropped images

LornaSIL:
* ```<body>``` tag needed removing. On Opera, Edge and Chrome the page was getting cropped. 
* Khowar keyboard name needed updating.


LornaSIL:
Fixed #390

PR #398, opened 2018-07-10
[rac_sindhi] Sindhi Phonetic Keyboard

rachitralli:
(This PR created using keyboards-starter)


MakaraSok:
@rachitralli 

Let's try to avoid the following issues for the next keyboard submission.

- [ ] It looks like .kpj is missing.
- [ ] the .kps has no .kmx, .kvk and .js referred from the build folder
- [ ] .php and files related to it should not be included in the .kps
- [ ] fonts included should be referred from the existing `shared` folder
- [ ] in the .kps the `package version follows keyboard version` has to be checked.

MakaraSok:
@rachitralli I am looking into the touch keyboard.

- [ ] The key next to the spacebar key in the shift layer is different from the one in the default layer, is that expected. Even so, we may reserve that key for the Arabic Fullstop (U+06D4)
image

- [ ] An Arabic Fullstop to the `ctrl-a` layer, just so it is consistent with the default layer.
image

- [ ] When comparing the touch layout with the desktop layout, I see one character (i.e. `U+0651 ARABIC_SHADDA`) shown up twice in the shift layer, it's on `SHIFT+K_M` and `SHIFT+K_W`. Please let me know which one should be removed and what should be put in place of either of these.
image

MakaraSok:
@rachitralli 

Issues with the OSK:

- [ ] `U+08FF ARABIC_MARK_SIDEWAYS_NOON_GHUNNA` doesn't get produced when `K_SLASH` is pressed. @DavidLRowe I don't see anything wrong with the configuration, but the Keyman Developer doesn't seem to work as expected for this very character.
image

- [ ] `U+002A ASTERISK` exists twice in the SHIFT layer and the CTRL-ALT layer (i.e. `CTRL ALT K_U` and `SHIFT K_8`). One of the two needs to be removed, and find a character to be put in place of either of the two.

- On the SHIFT layer:
image

- On the CTRL ALT layer:
image

rachitralli:
@MakaraSok please change the following

Shift version touch layout, please remove U_0683 and insert arabic full stop
move U_0683 to in place of U-0670  and move U-0670 to ctrl alt version in place of U_06E1 and move U_06E1 to ctrl alt in place of *

MakaraSok:
- [ ] ٰ `U+0670 ARABIC_LETTER_SUPERSCRIPT_ALEF` doesn't exist in the desktop layout, but it exists in the touch layout. Is that expected? Why is that? 
**It is advisable to have the same number of characters in both desktop and touch layout.**

- [ ] When "move U_06E1 to ctrl alt in place of *", asterisk is no longer in your touch layout. Is that what you really want?

Also, what should be done with the OSK? (The characters which occur more than once in the desktop layout.) Please make sure each character should occur once only, unless you have a good reason to do otherwise.

rachitralli:
@MakaraSOK @DavidLRowe please change the font and font size as already done in rac_khowar, thanks

DavidLRowe:
replaced by #587

PR #399, opened 2018-07-10
[libtralo] Updated keyboard

LornaSIL:
* Updated to include Charis SIL Dan fonts
* Removed Charis SIL fonts and changed target to ```shared/sil/charis_sil`` fonts
* Added On screen keyboard (.kvks)
* Changed touch layout file to use new name: libtralo.keyman-touch-layout (no change in layout)
* Deleted .docx and .pdf
* Updated documentation to include a welcome.htm rather than a .pdf


DavidLRowe:
- In readme.htm, it seems that 'but' should be replaced by 'the' in the sentence:
```
In order to display the proper capital hooked B, but <b>Charis SIL Dan</b> font is required.
```
- In the .kps and .keyboard_info file, `men` and `tem` don't have `-Latn`. Is this intentional?
- The .kps file references two copies of OFL.txt. It doesn't seem that both are needed.
  - shared\fonts\sil\charis_sil_dan\CharisSILDan_OFL.txt
  - shared\fonts\sil\charis_sil\OFL.txt
- This PR adds OFL-FAQ.txt to shared\fonts\sil\charis_sil_dan\, but the .kps file references the copy of OFL-FAQ.txt in shared\fonts\sil\charis_sil\

LornaSIL:
* Made changes to readme.htm to clarify when to use the "Dan" font
* *men* and *tem* are both understood by CLDR and BCP47 to be Latin script, so they don't need *-Latn* appended to the tag
* I removed the ``CharisSILDan_OFL.txt``` from the .kps as it is exactly the same as the OFL.txt file in the charis_sil folder.
* I removed the ``CharisSILDan_README.txt``` from the .kps as it is exactly the same as the README.txt file in the charis_sil folder.
* I deleted the ```OFL-FAQ.txt``` file from the ```libtralo\source``` folder as that was a remnant from when the font was in that folder. In this keyboard I am using the ```OFL-FAQ.txt``` from the ```shared\fonts\sil\charis_sil``` folder 
* Even though I don't use ```OFL-FAQ.txt``` in the ```shared\fonts\sil\charis_sil_dan``` folder I include it there in case someone will only include that font in a keyboard package later on.

PR #404, opened 2018-07-20
[easy_chakma] Update Help Page

MakaraSok:
The [Guideline.pdf](https://github.com/keymanapp/keyboards/files/2212158/Guideline.pdf) has been embedded in the easy_chakma.php. The current one is not readable at all.


Bivuti:
@MakaraSok Thanks. Exactly right now. Please update accordingly. This guideline is perfect for help to user.

LornaSIL:
This is one way to do it, but it looks pretty bad. In Firefox I don't see any of the layout:
image

In Chrome and Edge I just see a tiny bit of the pdf:
image

But that's not real useful either.

The issue seems to be that for the existing help file, if you don't have the font installed, you see square boxes:
image

This is not just a problem for the keyboard, I notice we have the same issue for the hieroglyphic keyboard. 

I feel like we need a better solution. @mcdurdin Can we use webfonts? Or do you have any better ideas of how to handle the online help files? Once a person installs the Desktop keyboard they will have the font installed and then they can see the webpage just fine. But if someone is browsing the help files to see which keyboard they want to use, they will see square boxes.

MakaraSok:
image

It looks OK when I view it locally. The set-up is as followed:

```
<div></div><embed src="Guideline.pdf" width="100%" height="95%" type='application/pdf'></div>
```

@LornaSIL The previous version of the help page is outdated. We need to explore a way to better embed the updated version given in PDF format.

MakaraSok:
@LornaSIL Can you please test this page again and see if it looks better? I use:
```
<object data="Guideline.pdf" type="application/pdf" width="100%" height="100%">
    <embed src="Guideline.pdf" type="application/pdf">
    <p>This browser does not support PDFs. Please download the PDF to view it: <a href="Guideline.pdf">Download PDF</a>.</p>
    </embed>
</object>
```

And locally, on my side, it looks like so in 
- FireFox:
image

- Chrome:
image

LornaSIL:
Firefox won't display it at all. It just suggests you download the .pdf.

Chrome does the same thing with this file as the previous file:

image

You *can* scroll down and see the pdf, but you have a 2 inch tall Window to see it in.

MakaraSok:
@LornaSIL Were you able to adjust the height of the PDF? It's currently set to 100%. I have no idea why only 2 inches of it is displayed. I am wondering what would happen if the height is set to 700px.

MakaraSok:
@LornaSIL HTML converted from .docx is now in place of the PDF.

LornaSIL:
@MakaraSok you'll need to commit the "images" folder and the files in there as I'm not seeing the images.

mcdurdin:
I don't recommend embedding the PDF. Make it available as a link is good.

Font embedding is certainly possible and I think is the way to resolve the HTML version issues here. You can use the `$pagestyle` variable in the PHP header to define a ttf font embed with `@font-face`.

MakaraSok:
Updated, but `image009.png` doesn't show up correctly. 

image

Here is code block of it.
```
<p class=MsoNoSpacing><span style='position:relative;z-index:-1894560256'><span
style='position:absolute;left:1590px;top:-100px;width:135px;height:0px'><img
width="300" height="100" src="images/image009.png"></span></span></p>
```

MakaraSok:
Please drop a line or two on how to use the webfonts. Thanks, Makara.

នៅថ្ងៃទី សុក្រ 27 កក្កដា 2018 នៅម៉ោង 2:49 AM Lorna Evans <
notifications@github.com> បានសរសេរ៖

mcdurdin:
In order to use RibengUni as a webfont, you should change the start of easy_chakma.php as follows:

```php
<?php
  $pagename = "Easy Chakma Keyboard Help";
  $pagetitle = "Easy Chakma Keyboard Help";
  $pagestyle = <<<END
@font-face {
	font-family: RibengUni;
	src: url("RibengUni-Regular.ttf") format("truetype");
}
```
append the rest of the document from the `$pagestyle = <<<END` line.

Then copy RibengUni-Regular.ttf into the help folder so that it will be deployed to the help.keyman.com site along with the updated php. Given that this is a change only to the help documents, you should be able to implement this without doing a version number increment.

PR #405, opened 2018-07-24
[aramaic_hebrew] Move keyboard from legacy to release

LornaSIL:
1. 
This keyboard is being moved from the legacy "aramaic" to "release". "aramaic" has 3 keyboards. We can get this keyboard checked, but I'd like to wait to merge all 3 at about the same time.

This keyboard includes the "meltho" fonts in the PR (in shared) which the other keyboards will also require, so it should be the first keyboard to be merged.


LornaSIL:
This branch got confused with pulling the meltho fonts out. I'm closing and starting a new PR.

PR #424, opened 2018-08-15
[sil_philippines] New keyboard for Philippines

LornaSIL:
There is an issue where web keyboard isn't working completely right in Firefox. Keystrokes work in Chrome just fine and in desktop.
>I was testing the Philippines keyboard today.   Keystrokes that work fine on desktop, didnt work on web in Firefox, but worked in Chrome. It uses a lot of deadkeys and it was an underscore deadkey that wouldn't act as a deadkey. Is this a known issue or a new bug in FF? I did reboot and still had the issue.
>A bit of further information is that when I changed the "_" to "\" for the deadkey, then it works in Firefox.
>I don't want to change the keyboard since it's been in use for 12 years.

Other keystrokes that don't work on FireFox web are ```'=``` and ```&:```


DavidLRowe:
A few minor comments about welcome.htm (and presumably also sil_philippines.php)

- in the "Key to Reading the Chart" section, there seems to be an extra space before the "a" in `' a means`
- In the chart itself:
  - the entry with ~`A in Keying Sequence column lacks the U+0300 grave accent in Character column
  - the entry with ~`i in Keying Sequence column lacks the U+0300 grave accent in Character column
  - the entry with ~`U in Keying Sequence column lacks the U+0300 grave accent in Character column
  - the entry with &u in Keying Sequence column has U+03BC (GREEK SMALL LETTER MU) in the Character column, although the keyboard produces identical looking U+00B5 (MICRO SIGN)

PR #426, opened 2018-08-17
[sil_devanagari_phonetic] Souce migration

andjc:
Chech to see if documentation is sufficient


LornaSIL:
Naming of the keyboard: welcome.htm and readme.htm use "Devanagari Phonetic (SIL)" as the name and all other files use "Devanagri Phonetic (SIL)". We should choose one. Personally, I'd prefer "Devanagari Phonetic (SIL)" and I believe that is what was requested. (I realize the original name was "Devanagri".)

* welcome.htm AND sil_devanagari_phonetic.php
     * "To force conjunct formation..." Change xxS to kxS
* welcome.htm, sil_devanagari_phonetic.php AND .kvks
     * The ```@``` sign cannot be accessed (looks like it's intentional from the code). Should/could that be deleted from the SHIFT graphic in both the welcome and help folder? Also, remove from OSK.
* .kps and .kpj
     * Please remove all version numbers since we follow keyboard version (you might have to do this by editing the .kps and .kpj files in a text editor)
* .kps and keyboard_info
     * Please add all the languages that use the Devanagari script.

LornaSIL:
Everything looks good except you could add "mai" to the language list in ```sil_devanagari_phonetic.keyboard_info```

Then I think it's ready to merge.

PR #430, opened 2018-08-24
[galaxie_hebrew_mnemonic] Migrate source to GitHub

andjc:


LornaSIL:
The only change that I see needed is in the .kps file to select readme.htm in the "Details" <b>Readme file</b>

PR #439, opened 2018-08-29
[sil_pan_africa_mnemonic] Moving keyboard from legacy to release

LornaSIL:


DavidLRowe:
In addition to testing with QWERTY keyboard, I tested some sequences with French as the base keyboard. Some of the sequences with `]` (AltGr+-) didn't really work, but it seems that this is not a problem with the keyboard itself.

PR #440, opened 2018-08-29
[sil_pan_africa_positional] Moved keyboard from legacy to release

LornaSIL:


DavidLRowe:
Tested on KeymanWeb. 
- Shift+RAlt+K brings up Keyman on-screen keyboard
- Shift+RAlt+M brings up Keyman configuration menu
- Shift+RAlt+O (as noted in documentation as example) doesn't produce output

Not sure if anything needs to be done about those.

Documentation (welcome.htm and .php help as well) for Diacritics Below, dot under has `RALT }` but it seems that it should be `RALT .` instead.

With that documentation change for the dot under, I think this PR is ready.

PR #441, opened 2018-08-29
[sil_eastern_congo-rev] Added characters to support Isiro languages

LornaSIL:


DavidLRowe:
.kps includes .kmn and .kvks source files, but I don't think they belong in the package

In the .kmn file, in store(dia_chars_U), U+028A and U+01B1 are switched. The result is that ;7 and ;& work properly with no diacritic, but switch upper/lower case when used with a diacritic.

Minor .kmn comments:
- comment above store(dia_chars_K) is inaccurate -- AK (and perhaps others) not in store
- comment above store(basic_diacritics) is inaccurate -- Tt not in store
- line 112 seems to be a duplicate of line 111

Is the goal to make the keyboard produce NFC output? There are a few places where it doesn't:
- @;a produces 00E4 0304 (not 01DF)
- ';c produces 00E7 0301 (not 1E09)
- `;* produces 0049 0300 (not 00CC)
- ^;* produces 0049 0302 (not 00CE)
- !;* produces 0049 030C (not 01CF)
- :;* produces 0049 0308 (not 00CF)
- @;* produces 0049 0304 (not 012A)
- ~;* produces 0049 0303 (not 0128)

LornaSIL:
I think/hope I fixed everything. It's very close to being there; however, please do not approve this yet as the client is still doing some testing.

LornaSIL:
This keyboard is ready for testing. Sorry it's taken so long!

DavidLRowe:
BTW, I made some Keyman regression tests when I first looked at this keyboard, and revised them to match the updates to the keyboard. I could add those to an extras folder if you think it worthwhile.

PR #442, opened 2018-08-31
[sil_uganda_tanzania] Move keyboard from legacy to release

LornaSIL:


DavidLRowe:
README.md needs copyright start date of 2004 to match HISTORY.md and LICENSE.md

sil_uganda_tanzania.keyman-touch-layout
- default layer, first row, longpress `o` has `ɔ` (U+0254) but diplays U+0244.
- shift layer, third row, full stop key produces U+2019 (probably needs a "layer": "default" to override the shift from being on the shift layer)
- shift layer, third row, B longpress can't get B plus U+032F
- numeric layer, second row, key labeled `&` produces `^`

Observation (no action necessarily needed): the number of keys on rows 2 and 3 of the numeric layer is a bit different from the other two layers, so the keys "jump" a bit when switching to the numeric layer. There's also no 'shift' key to take you from the numeric to shift layer.

PR #448, opened 2018-09-10
Adding paratext version of EuroLatin keyboard

sewhite:
Using \ as a deadkey is impossible for Paratext users, Paratext brings up the marker list popup.
This version uses ] where the original uses \.


keyman-server:

LornaSIL:
We'll need to discuss how to handle your request. If we want this keyboard in the repo then we will want it to go under a different name by itself (we don't put two keyboards together anymore). Also, all the documentation and everything needs updating. So, wait until we decide the best way forward.

LornaSIL:
I don't think we actually want to do this.

LornaSIL:
The options dialog in the keyboard allows RAlt or RCtrl so it is not necessary to use ```\```. Thus, we do not plan to incorporate this change.

PR #381, opened 2018-07-04
[tohono_oodham] Initial mobile keyboard for Tohono O'odham

alevy03:
This is a new keyboard to input script for the Tohono O'odham language. Only implemented for mobile platforms, currently.
This is my first time submitting a keyboard. So critiques, suggestions, and pointers are much appreciated!

PR #451, opened 2018-09-14
[kmcomp, sil_yi, yidhga] Update kmcomp to 10.0.1202.0

darcywong00:
The new version of kmcomp is needed  to merge the keyboard_info file for PR #381. The updated compiler also impacts these two keyboards:

### sil_yi ###
* changes language ID from `ysd` to `ysd-Yiii`
* remove spurious web platform

### yidgha ###
* changes language ID from `ydg` to `ydg-Arab`
* since keyboard targets `any` platform, also added the JS keyboard to the keyboard package


DavidLRowe:
What is the current recommendation? If `FollowKeyboardVersion` is set in the .kps file, should there be a version number in the `Keyboards/Keyboard/Version` element? Or does Developer now update that automatically? 

If the version number in the .kps file is not automatically updated, I'd recommend removing it from the .kps file, so it doesn't need to be manually updated.

darcywong00:
as I understand it, `FollowKeyboardVersion` only applies to the keyboard package version. So the keyboard version still need to be set.

Issue #422, opened 2018-08-15
[LIBTRALO] Add language code for Dan


darcywong00:
User request from 
https://community.software.sil.org/t/libtralo-keyboard/1004/20

Add dnj-Latn-LR to list of supported languages. Orthography is different than the dnj-Latn-CI orthography.

LornaSIL:
Fixed by #455

PR #455, opened 2018-09-14
[libtralo] Added language support for dnj

LornaSIL:
Updated keyboard to support dnj language from Liberia (not Cote d'Ivoire)


LornaSIL:
Fixes #422

PR #457, opened 2018-09-17
Fix Bug `Incorrect Vattu` in Firefox

beniza:


keyman-server:

PR #460, opened 2018-09-18
[sil_indic_roman] Move kbd from legacy to release

LornaSIL:


DavidLRowe:
.kpj file has reference to gff_amharic.kvks

welcome.htm (and possibly also .php file)
- there's a * a the end of the second line of the title. Not sure of its function. Is that a reference to a footnote?
- `+ 'semicolon' means press then release then press next key` -- I had to read it a few times before I got the meaning. What about something like `"x + y" means press and release "x" then press and release "y"` ?

.kmn file
- I expected `' + S` to produce S with acute accent, but only lowercase works. Is this worth noting in documentation?

DavidLRowe:
Documentation changes look good.

I'd still like to see the reference to gff_amharic.kvks removed from the .kpj file.

PR #462, opened 2018-09-19
[sil_mali_azerty] Move kbd from legacy to release

LornaSIL:


DavidLRowe:
Some keying sequences in the documentation don't match what the keyboard does (as tested with Keyman Web).

Characters:
- `b-` (rather than `b(`) for `ɑ̆`
- `B-` (rather than `B(`) for `Ă`
- `kkn` (rather than `,nmm`) for `ŋ͡m`
- `,|` (rather than `!f`) for `Ø`
- `£` (since it has its own key on the AZERTY keyboard, rather than `!l`) for `£`

Diacritics:
- acute is typed with the `'` key (on the key with `4`)
- grave is typed with the `(` key (on the key with `5`)
- macron is typed with the `_` key (on the key with `8`)
- circumflex is typed with the `ç` key (on the key with '9')
- caron is typed with the `à` key (on the key with '0')

PR #463, opened 2018-09-19
[sil_mali_qwertz] Move kbd from legacy to release

LornaSIL:


DavidLRowe:
As tested on Keyman Web, there are a couple of places where the keyboard diagram doesn't correspond with what the keyboard actually does:
- the rightmost key in the second row is labeled "#" but produces "+"
- the rightmost key in the fourth row (to the left of the right shift key) is labeled "#" but produces "-"
- additionally, the # key is used to produce a dot below, but there doesn't seem to be any way to type that character

I also couldn't access the Ctrl+Shift layer, but perhaps that's because I tested on Keyman Web.

PR #466, opened 2018-09-27
[sil_tawallammat] Add new keyboard to repo

LornaSIL:


DavidLRowe:
The following seem to be due to a different font on the OSK. Not sure if anything needs to be done about them:
- symbol on OSK for 'f' key is somewhat different from the character produced by the key.
- symbol on OSK for 'q' key is somewhat different from the character produced by the key.

Keyman Web issue
- Keyman Web screen display has ~ on shift layer for first key on first row, but the character produced is U+200D. 

sil_tawallammat.php and welcome.htm files reference image_xxx.png files. Some of those .png files have an initial capital letter "I" and some have lower case. In the .php and welcome.htm files, the references don't always match the filename. This isn't a problem under Windows, but will be for Linux (and maybe Mac). It would be preferable to make all the filenames and all the references to them lowercase.

LornaSIL:
There are significant differences in shapes of characters depending on the font and regional use. So, we may not worry too much about differences in shapes of characters. I agree we should address the other issues.

DavidLRowe:
Desktop layout, leftmost key on top row: OSK has superscript 2, but keyboard produces ZWNJ. With this issue corrected, I think the keyboard is ready for approval.

Very minor issue on mobile: "zwnj" doesn't fit on key top for iphone4s, but is certainly still usable.

Had OSK display issues, but probably not directly related to the keyboard. It may be that I should have used a different option when installing the keyboard?
- Needed Shift + *Left* Alt + K to bring up OSK (apparently because Right Alt used in keyboard)
- When I held down the Right Alt key, I could type the characters on that layer, but the OSK indicated both the Right Alt and the Left Ctrl, so the display on the OSK didn't show the characters for that layer.

PR #468, opened 2018-10-02
[canadian_aboriginal] Update to latest version of fonts

LornaSIL:
Preparing to work on "fv" keyboards so I'm first updating the fonts.


LornaSIL:
No license info was available in the font download, nor in the original font folder.

PR #474, opened 2018-10-14
[fv_shashishalhem] merge desktop and web and delete kmw

LornaSIL:


DavidLRowe:
It seems as though the file `fv_shashishalhem_kmw.kmn` should not be part of this PR.

PR #475, opened 2018-10-14
[fv_sencoten] merge desktop and web and delete kmw

LornaSIL:


DavidLRowe:
Should there be a readme.htm file?

Do you want to convert the .kvk file to .kvks?

help/ folder has several files not referenced by .php file.

.kps has reference to .msi and Tavultesoft image.

PR #476, opened 2018-10-14
[fv_secwepemctsin] merge desktop and web and delete kmw

LornaSIL:


DavidLRowe:
.kps file has .msi reference

PR #477, opened 2018-10-14
[fv_nuucaanul] merge desktop and web and delete kmw

LornaSIL:


DavidLRowe:
.kps file has reference to .msi file

.kmn has two `begin Unicode > use(main)` statements (I guess it compiled fine, though)

PR #478, opened 2018-10-14
[fv_dakelh] merge desktop and web and delete kmw

LornaSIL:


DavidLRowe:
.kps file has reference to .msi

PR #479, opened 2018-10-16
[fv_denesuline_epsilon] merge keyboard and delete kmw

LornaSIL:


DavidLRowe:
.kps file has .msi reference

PR #489, opened 2018-10-18
[fv_han] merge keyboard and delete kmw

LornaSIL:
* Added welcome.htm file
* Changed ALT key to RALT key (for one key)
* Added Todo note in README.md that keyboard needs canonical ordering adjusted


DavidLRowe:
Minor point: welcome.htm has "keys with a red background are combining accents" just as the fv_han.php file does, but I'm not seeing any red background when the welcome.htm is displayed, whereas the red background is there for the help file. Is this worth fixing?

PR #509, opened 2018-10-26
[fv_tlingit] Removing _kmw from files, changing license, renaming .js…

LornaSIL:
… to keyman-touch-layout

* Changed license
* Changed version
* renamed layout.js to .keyman-touch-layout
* Added readme.htm which is not actually used, but whenever we add desktop in, it will be available
* Added .ico which is not actually used, but whenever we add desktop in, it will be available
* in .php changed "osk" to "osk-tablet"

Those are the main changes. I made no changes to the rules in the .kmn
Since this is a mobile layout, the .kps is apparently not required as no .kmp is created.


DavidLRowe:
On the default and shift layers, there are two keys with period and longpress options. The longpress options are different on the two keys (one has just comma, the other has comma plus other punctuation). Everything works fine, but it seems odd to have those two keys. 

I'm fine to approve it as is, but wondered if that should be changed.

LornaSIL:
I've updated the README.md to include this information under "Todo". I do not recommend we try to fix keyboards while doing this other work right now.

DavidLRowe:
Makes sense.

PR #512, opened 2018-10-29
[fv_kmw_2] Delete kmw from filenames, update metadata

LornaSIL:
41 keyboards:
-  Changed license
-  Changed version
-  renamed layout.js to .keyman-touch-layout
-  in .php changed "osk" to "osk-tablet"

Those are the main changes. I made no changes to the rules in the .kmn
Since this is a mobile layout, the .kps is apparently not required as no .kmp is created.


DavidLRowe:
No readme.htm for these keyboards. Is that a problem?

Some keyboards have identical language tags (for example, fv_dakota_sk and fv_dakota_mb both have `dak-Latn`). I'm assuming that this is fine.

For fv_lakota, the language tag in .keyboard_info file changed from `lkt` -> `ikt-Latn`. It seems as though this is a typo (`lkt` vs `ikt`). If it is not a typo, then the .kmn file should be changed, since it still has "lkt" in the commented out ethnologuecode store line.

darcywong00:
I missed what the direction was re: removing `_kmw` from the filenames.

I had this question while going through 
```
fv_anicinapemi8in_kmw/HISTORY.md → release/fv/fv_anicinapemi8in/HISTORY.md
```

Previously, when keyboards were deprecated, their source folders remained with an added `DEPRECATED.md` file.

This PR seems to rename `_kmw` keyboard folders.

Are we not wanting to keep the `_kmw` directories around?

darcywong00:
In the php help files where you added:
```
<div id='osk-tablet'>
```

The OSK would show the **Default** and **Shift** layers. In order to get other layers, I had to modify it
```
```

LornaSIL:
> Are we not wanting to keep the `_kmw` directories around?

Marc said to delete the folders (rather, rename them), but we can still deprecate them because they remain on the server.

LornaSIL:
> iirc, @mcdurdin would like the mobile apps to get keyboards from .kmp files.
> At some point, these .kps files will need to be made.

I had forgotten that. Can you tell me what files need to be included in the .kps? Just the .js? I don't think we would distribute a .kmx, right?

darcywong00:
> Can you tell me what files need to be included in the .kps? I don't think we would distribute a .kmx, right?

Right, distributing a keyboard package is preferable to distributing a .kmx on its own. There's a summary of what files to include in the **Package file contents** section of https://help.keyman.com/developer/10.0/guides/distribute/packages

darcywong00:
sounds good. I won't be available to do another check, so go ahead and merge

PR #516, opened 2018-11-09
[basic_kbdlisus] MS Lisu Standard Basic keyboard to Keyman using kmco…

LornaSIL:
…nvert

* Added help and enhanced welcome.htm


DavidLRowe:
I realise this has already been merged, but want to record this somewhere:

desktop:
- some keys that are blank on OSK produce underlying QWERTY key (for example, Shift+G)

mobile:
- shift layer, some keys that are blank on touch layout produce characters (ERYG)

PR #517, opened 2018-11-09
[basic_kbdbug] MS Bugis Basic keyboard to Keyman using kmconvert

LornaSIL:


DavidLRowe:
desktop:
- some keys that are blank on OSK produce underlying QWERTY key

mobile:
- some keys that are blank on touch layout produce characters (QWERTY etc)
- shift (uparrow) key doesn't take you out of shift layer, need unmarked key in lower left
- pressing key in lower left on default layer highlights it and adjacent shift button but doesn't do anything

DavidLRowe:
OSK has characters on keys on shift level (QWERTYUIOPSDFGHJKLZXCVB) that now produce no output.

LornaSIL:
I think this is a KeymanWeb bug, not OSK.

PR #518, opened 2018-11-09
[basic_kbdcher] MS Cherokee Basic keyboard to Keyman using kmconvert

LornaSIL:


DavidLRowe:
.keyboard_info and .kps:
- "chr-Latn-US" (is -US needed?)

desktop:
- icon has "CHR" but sides are clipped
- On OSK, key caps for bkquote (1st row, leftmost key) and ' (3rd row, to left of Enter) seem to be swapped
- same for shift level, key for shift+bkquote seems to be swapped with key for shift+'
- same for RALT and Shift+RALT levels, although only the bkquote key produces output

- mobile:
- default layer, top row, left three keys: chars don't fit on key cap
- def layer, 3rd row, 1st & 12th keys swapped
- shift layer, 3rd row, 1st & 12th keys swapped
- rightalt layer, 3rd row, 1st & 12th keys swapped
- shift-alt layer, 3rd row, 1st & 12th keys swapped
- in shift layer, pressing Alt takes you to default. would expect rightalt (or perhaps shift-alt layer)
- all layers, 4th row, key to right of shift is empty. Is this key needed?

PR #520, opened 2018-11-09
[basic_kbdnko] MS NKo Basic keyboard to Keyman using kmconvert

LornaSIL:


DavidLRowe:
help file basic_kbdnko.php
- `</p>` after `Segoe UI` should be `</b>`

desktop:
- some keys that are blank on OSK produce output of underlying key, for example, ~@#$^&ERTU
- some forms on desktop shift layer seem different than welcome.htm: QWYPGH (but could just be a font difference)

mobile:
- touch layout, ctrl layer seems empty (except space bar); seems as though it could be removed
- on shift layer, shift key doesn't take you back to default layer, but 'ctrl' key in lower left does
- some keys that are blank on touch layout, shift layer produce output (as with desktop layout)

DavidLRowe:
On shift layer, OSK still has characters on keys (~@#$^&ERTU) that now produce no output.

touch layout: there doesn't seem to be a way to get out of the shift layer.

LornaSIL:
I fixed the shift issue. The other issue with OSK/Web seems to be a problem with KeymanWeb, not OSK. When I install the OSK on desktop it appropriately doesn't display the non-defined keys.

PR #521, opened 2018-11-09
[basic_kbdntl] MS New Tai Lue Basic keyboard to Keyman using kmconvert

LornaSIL:


DavidLRowe:
both desktop and mobile layouts:
- some keys on default layer, top row, produce U+FFxx chars; is this intended?
- quote key produces U+2019 U+2018, shift+quote produces U+201C U+201D; are these intended?

mobile:
- from shift layer, shift key doesn't take you back to default; need to use Alt key to get out of shift layer, back to default
- from rightalt layer, both shift and def take you to default layer

LornaSIL:
>     * some keys on default layer, top row, produce U+FFxx chars; is this intended?
Yes, I think since this is an East Asian script they probably are wanting the half-forms.

> quote key produces U+2019 U+2018, shift+quote produces U+201C U+201D; are these intended?
I think we should leave this as the MS keyboard does seem to produce this.

PR #522, opened 2018-11-09
[basic_kbdsora] MS Sora Basic keyboard to Keyman using kmconvert

LornaSIL:


DavidLRowe:
.kmn doesn't agree with README.md in terms of coverage
- .kmn has only win,mac,lin,web, README.md has more

welcome.htm:
- heading seems to have a typo: "Shift CTRL)"

desktop:
- q blank on OSK, but produces q
- shift+ctrl+4 doesn't seem to produce the character according to the welcome.htm documentation; OSK has nothing for shift+ctrl layer

mobile:
- not yet tested

DavidLRowe:
On shift layer, OSK still has characters on keys (q and A-Z) that now produce no output.

LornaSIL:
> On shift layer, OSK still has characters on keys (q and A-Z) that now produce no output.

I don't see the characters displayed on the OSK (when installed on my system). I do see them on the Web link, but I don't know how to change what is displayed for "Web". Is that a bug in KeymanWeb?

LornaSIL:
I am feeling this is a KeymanWeb issue.

PR #523, opened 2018-11-09
[basic_kbdtifi] MS Tifinagh Basic keyboard to Keyman using kmconvert

LornaSIL:


DavidLRowe:
desktop:
- store(&NAME) 'Tifinagh (Basic) Basic' -- is this the intended name?
- TZM on icon clipped
- some keys which are blank on OSK produce the output of the QWERTY keyboard, eg shift+E produces 'E', but this could be confused with the Tifinagh character U+2D39 on the shift+D

mobile:
- similar issue with some keys that are blank producing characters from the QWERTY 
- def layer, 3rd row, 12th key, no label, produces '
- need Alt key to leave shift layer
- both shift and def leave rightalt layer

LornaSIL:
> store(&NAME) 'Tifinagh (Basic) Basic' -- is this the intended name?

Yes, this set of keyboards all have "Basic" added to the end...and ```(Basic)``` was part of the original keyboard name.

PR #527, opened 2018-11-09
[basic_kbdwol] MS Wolof Basic keyboard to Keyman using kmconvert

LornaSIL:


DavidLRowe:
for touch:
- shift key doesn't leave shift layer; need "AltCtrl" key to return to default layer. 
- for AltCtrl layer, both shift and "default" return to default layer.

PR #528, opened 2018-11-09
[basic_kbdyak] MS Sakha Basic keyboard to Keyman using kmconvert

LornaSIL:


DavidLRowe:
Should the .php file include a reference to the mobile layout?

PR #529, opened 2018-11-09
[basic_kbdyba] MS Yoruba Basic keyboard to Keyman using kmconvert

LornaSIL:


DavidLRowe:
Desktop RightAlt layer: keys for accented dotted e characters are on H and J on OSK, but on J and K for actual keyboard.

touch layout:
- default layer, row 2, next to last key has `[` but doesn't seem to produce any output.
- default layer, row 3, next to last key has `'` but doesn't seem to produce any output (except for longpress keys which do work)
- default layer, row 3, last key has `\` but doesn't seem to produce any output.

PR #530, opened 2018-11-09
[basic_kbdfthrk] MS Futhark Basic keyboard to Keyman using kmconvert

LornaSIL:


LornaSIL:
@DavidLRowe I changed the documentation, but I did not change the keyboard as we are mostly trying to provide the same keyboard as MS.

PR #531, opened 2018-11-09
[basic_kbdinen] MS India Basic keyboard to Keyman using kmconvert

LornaSIL:


LornaSIL:
Deadkeys need documenting.

DavidLRowe:
touch layout has 'abc' on shift key (where normally there would be an uparrow).

Shift+RAlt+O should give upper case O with macron, but Shift+Alt+O is Keyman default to turn off keyboard, so it doesn't make any character. (workaround: `O RAlt+, -`)

Shift+RAlt+M should give capital M with dot above, but Shift+Alt+M is keyman default for Keyman Menu. (workaround: `M RAlt+, >`)

PR #533, opened 2018-11-12
[lao_phonetic] Version 1.1 - converted from Lao Script with John Durdin

mcdurdin:
This keyboard has been imported from a binary version in Lao Script for Windows by myself and my father @jmdurdin. The intention is to cleanup the .kmn in a future update and restore symbol names and comments (and cleanup the compile optimisations), but this version works correctly and has been requested by some users.


LornaSIL:
Could you consider putting the fonts in ```\release\shared\fonts\laoo\saysettha``` instead of the local ```assets``` folder?

LornaSIL:
Copyright date should be changed from "2015" to "2015-2018" in .kmn, readme.htm, welcome.htm (and subsequently hopefully the .kps and .kpj copyrights will be corrected.

I'm getting warning messages when I try to compile the .kmn file such as:
```lao_phonetic.kmn: Warning: line 2612  warning 209A: The rule will never be matched because its key code is never fired.```
That one is regarding this rule:
```   + "ñ" > "ຍ"  c line(0)```
but there are about 20 warning messages.

Also, in testing against the welcome.htm documentation:
```ko``` is producing ເກາະ and not ໂກະ (as documented) - actually I think the output of ko and ko` is reversed to what it should be.

Otherwise the keyboard seems to work properly, but because of the warnings, the .kmp file is not being created.

File: release/l/lao_phonetic/source/lao_phonetic.kps [hunk 1]
```diff
@@ -0,0 +1,104 @@
...
+      <CopyLocation>0</CopyLocation>
+      <FileType>.txt</FileType>
+    </File>
+  </Files>
+  <Keyboards>
+    <Keyboard>
+      <Name>Lao Phonetic</Name>
+      <ID>lao_phonetic</ID>
+      <Version>1.0</Version>
+      <Languages>
```
darcywong00: This might only pertain to touch keyboards, but is it good practice to set the keyboard font and display font to saysettha?

darcywong00: I just noticed lao_phonetic.js is included in the package, so the fonts will need to be set.

release/l/lao_phonetic/lao_phonetic.kpj
```diff
@@ -0,0 +1,129 @@
+<?xml version="1.0" encoding="utf-8"?>
+<KeymanDeveloperProject>
+  <Options>
+    <BuildPath>$PROJECTPATH\build</BuildPath>
+    <CompilerWarningsAsErrors>False</CompilerWarningsAsErrors>
```
darcywong00: Is this intentional because of the multiple compiler warnings?

`This rule will never be matched because its key code is never fired.`

jmdurdin:
It probably needs to be in assets to enable it to work with the touch keyboard – Marc should be able to confirm this.  Note that I have now found the original keyboard source files so there will probably be an update as soon as Marc can find the time to rebuild the package.

release/l/lao_phonetic/source/lao_phonetic.kps
mcdurdin: The lao_phonetic.js keyboard is desktop only; it shouldn't be in the package (I will remove it).

release/l/lao_phonetic/lao_phonetic.kpj
mcdurdin: Yes.

mcdurdin:
Okay, I've updated the keyboard with the source that @jmdurdin found, which is much better. The `ñ` key will never be matched as the keyboard is setup as positional, but kmcomp does not pick up this issue when the key is in a store (`+ any(C) > index(CLao,1)`, line 237). Now this could be considered a bug or a gap in kmcomp...

However, in this case the `ñ` is being used as a placeholder: there is no single key which generates the `ຍ` consonant, but we still want the consonant in the `CLao` store. Changing this would require a significant number of changes to the keyboard for little benefit. There is no harm done.

If I [update kmcomp to address the missing warning](https://github.com/keymanapp/keyman/issues/1328), I will downgrade the warning to a hint. This would then allow the keyboard to compile even with 'treat warnings as errors' set (so no net change in compile behaviour for any keyboard already in the repo).

I have updated copyright in

release/l/lao_phonetic/source/lao_phonetic.kps
```diff
@@ -0,0 +1,98 @@
...
+      <Description>File Saysettha_README.txt</Description>
+      <CopyLocation>0</CopyLocation>
+      <FileType>.txt</FileType>
+    </File>
+  </Files>
+  <Keyboards>
+    <Keyboard>
+      <Name>Lao Phonetic</Name>
+      <ID>lao_phonetic</ID>
+      <Version>1.0</Version>
```
darcywong00: Keyboard version is now 1.1

mcdurdin: Fixed.

PR #542, opened 2018-11-26
[basic_kbdtaile] MS Tai Le Basic keyboard to Keyman using kmconvert

LornaSIL:


DavidLRowe:
From shift layer, both shift and rightalt take you back to default layer. From rightalt layer, both shift and rightalt take you back to default layer. If this is how it's supposed to be, that's fine. (I was a bit surprised that rightalt didn't take me to the rightalt layer from the shift layer, and that shift didn't take me to the shift layer from the rightalt layer.) Otherwise LTGTM

LornaSIL:
@DavidLRowe I don't think my "fix" will work. But this is something I've seen on many keyboards. From RALT Or CONTROL layer we aren't able to change to shift, just to default. I think it's a bug in Keyman that we need to document clearly and report to the developers. We could go ahead and approve this as we've done other keyboards, but we should make sure to write up a good issue if we haven't already.

PR #543, opened 2018-11-26
[sil_hmd_plrd] New keyboard for Miao script

LornaSIL:
There aren't many Miao fonts. I'm torn about including a font in this package as the SIL fonts are in flux. The ShiShan fonts will soon be replaced with the Shimenkan font and so I can't include Shimenkan (yet), but the ShiShan ones are not ideal. Recommendations welcomed.


DavidLRowe:
OSK shows `{` and `}` in their normal spots, but these keys produce U+16F91 and U+16F90. Other than that, keys seem to produce what they are supposed to.

PR #558, opened 2018-12-03
[pukapuka] Removed special highlighting on vowels (Android issues)

markpenny:
When testing this on a physical Android phone (and not just on an emulator) I found that the "special" highlighting on the vowel keys and modifier diacritics was completely obscuring the letters (making it impossible to use).
I have now removed the special highlights and updated the documentation to reflect the more standard look (with no special keys). Bumped the version up to 1.3 as well.


keyman-server:

keyman-server:

PR #567, opened 2018-12-06
[sil_euro_latin] Switch abc and currency keys on touch layer

### LINKED ISSUE #562: [sil_euro_latin] (abc) and (pound-euro-yen) keys need to switched on symbol layers
Base Layer:
image

Shifted:
image

Numeric / Symbol 1:
image

Symbol 2:
image

The (abc) and (pound-euro-yen) keys should be swapped on the symbol layers. Pressing the [123] key should open the numeric/symbol 1 layer, but pressing in the same place again should return to the base layer.

As it stands, it is always a bit of a hassle to swap back to the base layer after inserting a digit or symbol.

darcywong00:
Fixes #562 

* Switch `*abc*` and `*Currency*` keys on numeric/symbol layers of touch layout
* Fix `K_NUMERALS` key label on Symbol layer
* Added mobile layout OSK to help


release/sil/sil_euro_latin/source/sil_euro_latin.keyman-touch-layout
```diff
@@ -2330,11 +2330,11 @@
             "id": 3,
             "key": [
               {
-                "id": "K_LOWER",
-                "text": "*abc*",
+                "id": "K_NUMERALS",
+                "text": "*123*",
```
darcywong00: This fixes the previous inconsistency of `K_CURRENCIES` switching to `numeric`.

LornaSIL:
LGTM, but I'll let Marc check it too.

mcdurdin:
LGTM too.

PR #571, opened 2018-12-08
[sil_tawallammat] case in welcome.htm for ralt image

glasseyes:
The ralt image in welcome.htm isn't visible on case-sensitive systems

---
This change is [Reviewable](https://reviewable.io/reviews/keymanapp/keyboards/571)


keyman-server:

PR #572, opened 2018-12-08
[rac_brahui] change PNG to png

glasseyes:
touch images not found in welcome.htm on case-sensitive systems

---
This change is [Reviewable](https://reviewable.io/reviews/keymanapp/keyboards/572)


keyman-server:

rachitralli:
@glasseyes thanks

LornaSIL:
If we want the welcome.htm to show up in the packages then the versioning all needs to be bumped too.

darcywong00:
I've updated the keyboard version on behalf of @glasseyes

keyman-server:

DavidLRowe:
These changes seem reasonable, but there is a problem with the example in the .keyboard_info file. The fourth character in the text is U+06C1, but the fourth key is "h", which produces U+062D. The key "o" produces U+06C1. It seems that either the key should change from "h" to "o", or the text should change to match the keys. @rachitralli Can you verify what is correct?

rachitralli:
@DavidLRowe you are right the [h] is wrongly typed by me. The correct is as under

Example: To enter براہوئی (Name of language), 
type  braowUi

keyman-server:

rachitralli:
@darcywong00 thanks

PR #556, opened 2018-12-02
GFF Amharic Keyboard v1.7

dyacob:
Applying the v1.6 enhancements to lone vowels.

PR #576, opened 2018-12-11
gff_amharic v1.7: Setting lastModifiedDate to commit time.

dyacob:
Resetting lastModifiedDate under @LornaSIL 's recommendation: https://github.com/keymanapp/keyboards/pull/556#issuecomment-444532426


keyman-server:

PR #585, opened 2019-01-04
[rac_gawri] submission of rac_gawri keyboard on behalf of author

DavidLRowe:


rachitralli:
@DavidLRowe 
 please check this the y key and i key both producing one Arabic zer, please change one into Arabic zabar, if possible please add English typing in longpress or shift or ctrl alt layer like GBoard of google. If the community wants to type gawari as well as English letters then there is no option available, therefore I want to add English typing like GBoard. if possible please add. thanks

LornaSIL:
@rachitralli if users wish to type English they should change keyboards. We should not attempt to include both of these languages in one keyboard. They are too different.

DavidLRowe:
@rachitralli Please specify exactly what changes are needed to the keyboard. I think you mean the Shift+Y and Shift+I which currently both generate U+0650. (This is also the case for the rac_sindhi keyboard.)

DavidLRowe:
The terms 'Arabic zer' and 'Arabic zabar' are not familiar to me. For the rac_uyghur keyboard, Shift+I generates U+0650 and Shift+Y generates U+064E. Are these the characters you mean? Would you want to do the same for rac_gawri?

rachitralli:
@DavidLRowe sir you are right that in the rac_uyghur keyboard, Shift+I generates U+0650 and Shift+Y generates U+064E. These the characters are correct. I would want to do the same for rac_gawri, please change as desired.

@LornaSIL Madam, got your point. thanks

LornaSIL:
* OSK needs the shift Y key to display U+064E (fatha) and not U+0650 (kasra). It is outputting the right character, the display is wrong
* The graphics in the welcome folder need to be regenerated. They look like tablet graphics and not desktop graphics.
* These characters are on the desktop keyboard, but they are not on the mobile layout: U+0621 (u) U+0670 (E), U+06C3 (O), [, ], |, _, U+064B (B), copyright and register marks. They should be added to the table layout.         
     * For the square brackets, you could add them to the parenthesis keys as longpress.
     * For the other symbols (|,_, copyright mark, register mark) you could add them as longpress to the ```@``` sign.
     * U+0621 - add as longpress to hamza above/U+0654?
     * U+0670 - add as longpress to alef/U+0627?
     * U+06C3 - add as longpress to U+06C1?
     * U+064B - add as longpress to fatha/U+064E?

@rachitralli what do you think of these suggestions for positioning of these missing characters?

I think everything else looks good.

DavidLRowe:
- OSK corrected and graphics redone
- additional keys added to touch layout following @LornaSIL suggestions, except existing longpress keys for `< > { } « »` also moved to parentheses keys, along with new `[ ]`

DavidLRowe:
@rachitralli If anything needs to change, we can do that before the keyboard gets merged.

rachitralli:
@LornaSIL your suggestions are good, @DavidLRowe it is ok. thanks

PR #586, opened 2019-01-04
[rac_uyghur] New keyboard submitted on behalf of author

LornaSIL:
@rachitralli this keyboard is based on the files you sent me. I'll email you a link to the .kmp file for you to test the package.

I did make a change from your keyboard. The "o" key was outputting U+06C1 and I changed that to U+06D5. I did this because your example of the language name (in keyboard_info) used U+06D5 and the language name in the welcome.htm file uses U+06D5 (not U+06C1). Let me know if that's not what you want.

I also changed the touch layout to output the same characters as are used in the desktop version.


DavidLRowe:
touch layout, shift layer, third row, next to last key (to left of 'delete'):
- key produces U+201D instead of U+06C7
- longpress key labeled `\` produces U+06c6
- longpress key labeled `?` produces U+061F
- longpress key labeled `,` produces U+060C

rachitralli:
@LornaSIL @DavidLRowe thank you very much for this act of kindness

PR #588, opened 2019-01-07
[sil_arabic_phonetic] Move existing keyboard from legacy to release

LornaSIL:
* Added README.md, HISTORY.md, LICENSE.md, readme.htm
* Updated keyboard_info
* Added web and touch layout
* Complete new .php file
* Made sure blank keys do not output anything


DavidLRowe:
There seems to be some differences in dates. One place has -2018, another has -2019. What is the date of version 1.0? Aug2018? Jan2019?

Help (.php) file has a portion of the release history, but not the latest. Seems better to omit altogether (or else include complete history).

Touch layout, shift layer, "x" produces U+0652, though that key doesn't seem to have any text on it (as it does in the OSK). In the .kmn file there are two entries for "X", one that produces U+0652 and one that goes to a deadkey.

The .kmn file seems to have duplicate processing of digit keys: once using stores and again with an individual line for each digit.

PR #568, opened 2018-12-07
[rac_pashto] Pashto keyboard

rachitralli:
Pashto keyboard by Rehmat Aziz Chitrali

PR #589, opened 2019-01-07
[rac_pashto] New keyboard submission on behalf of author

LornaSIL:
This keyboard should replace PR #568


LornaSIL:
Try this link: https://keymanweb.com/#ps-Arab,Keyboard_rac_pashto

rachitralli:
@LornaSIL thanks for the link, but the Pashto is still missing in the list please check 
pashto keyboard is missing in the list

LornaSIL:
@rachitralli because of the way you've chosen to name your keyboards, they are all listed under "R"
image

But, since 'pus' is a macrolanguage, I'm wondering if we should have associated the keyboard with something else (that is Central Pashto [pst], Northern Pashto [pbu], and/or Southern Pashto [pbt]). Maybe our keyboard support isn't expecting macrolanguages. We can update the keyboard. Which of these languages are you actually intending to support? pbu, pbt or pst?

rachitralli:
@LornaSIL got your point. thanks

PR #593, opened 2019-01-09
[rac_uyghur][rac_pashto] fixing up help file and home link

LornaSIL:
* help files were not displaying layouts. I think they needed to be in a container
* home link for Uyghur was wrong

No changes were made to keyboard or files used in the package, so I did not bump the version.


DavidLRowe:
rac_uyghur README.md has ref to `rac_pashto`

Don't know about the .php file changes, but they seem OK. Just curious why the change from `<h3>` to `<h2>`.

PR #594, opened 2019-01-10
[rac_help-osk] Fix links and layer keyname

LornaSIL:
Fixes to three help files. No changes to keyboards themselves.


rachitralli:
@LornaSIL @darcywong00 thanks for fixing rac_help-osk

PR #595, opened 2019-01-10
[rac_help_ug_ps2] Change symbol to shift

LornaSIL:
I finally understood what @darcywong00 was telling me.


rachitralli:
@LornaSIL @DavidLRowe thanks

PR #597, opened 2019-01-11
[sil_cherokee_nation] Move keyboard from legacy to release

LornaSIL:
This keyboard is based on one from legacy. Source files taken from internal repo.


DavidLRowe:
The BKQUOTE key on Desktop produces nothing. (KeymanWeb OSK displays the character, however, which might be confusing to a user.) The BKQUOTE key is present on the touch layout. It seems that the easiest way to get them in sync is to enable the key on the desktop layout.

On the touch layout, the shift key takes you to the shift layer, but there's no way to get out of the shift layer, since the shift key doesn't take you out of the shift layer.

DavidLRowe:
The BKQUOTE key on Desktop produces nothing. (KeymanWeb OSK displays the character, however, which might be confusing to a user.) The BKQUOTE key is present on the touch layout. It seems that the easiest way to get them in sync is to enable the key on the desktop layout.

On the touch layout, the shift key takes you to the shift layer, but there's no way to get out of the shift layer, since the shift key doesn't take you out of the shift layer.

release/sil/sil_cherokee_nation/sil_cherokee_nation.keyboard_info
```diff
@@ -0,0 +1,23 @@
+{
+    "license": "mit",
+    "description": "A Cherokee Unicode keyboard following the official Cherokee Nation layout.",
+    "languages": {
+        "chr-Cher": {
+            "font": {
+                "family": "cherokee_web",
+                "source": "digohweli_1_7.ttf"
```
darcywong00: I see the font is available at http://s.keyman.com/font/deploy/digohweli_1_7.ttf
but not in the keyboard repo. 

Should we include a copy and add it to the keyboard package?

LornaSIL: I know modern Windows and OSX would have fonts available. Would Linux? If not, it would be a good idea. If you think I should add it, is the license acceptable for distribution?

LornaSIL: BTW, I do not have access to that repo. It was quite hard to find the font online which makes me think the author removed it. However, I did eventually find the font and the license says:

>LICENSE AGREEMENT

>This typeface is the property of Chris Harvey and its use by you is covered under the terms of a license agreement. You have obtained this typeface software either directly from Chris Harvey or together with software distributed by one of the products at languagegeek.com.

>This font took a lot of work to make, so please, unless you have entered into a specific license agreement granting you additional rights, your use of this software is limited to your workstation for your own publishing use. You may not copy or distribute this software. Aboriginal community members and projects relating to Aboriginal languages may distribute the font freely. Contact Chris Harvey at languagegeek.com to get this free distribution permission.

So, unless we have permission, I don't think we can distribute the font.

I do see that we have Aboriginal Serif in our shared fonts already and it has the Cherokee in it. Shall I just add that font to the package?

darcywong00: I tried using AboriginalSerifREGULAR.ttf in the Android emulator (Android 4.4 Jelly Bean is now the minimum version for Keyman)

### Default Layer
default

### Shift layer
shift

If that looks suitable for you, then I agree about adding it to the package (and update .keyboard_info)

LornaSIL:
@DavidLRowe Darcy just merged the font into the proper repo, so feel free to approve this if you feel it's ready.

PR #602, opened 2019-01-16
[rac_kashmir_shina] Submitting new keyboard on behalf of author

LornaSIL:


rachitralli:
@LornaSIL please change the forllowing

*rac_kashmir_shini to rac_kashmiri_shina

*scl-Arab-Pk to scl-Arab-IN

*please fix this waw وٗٗ as the upper mark is not on the waw

LornaSIL:
> *rac_kashmir_shini to rac_kashmiri_shina

My understanding is that you agreed to the name change ```Rachitrali-Kashmir-Shina```. That means we keep the filename to what I've set it at. We must make it clear that this is not for the Kashmiri language, but it is for Shina in the Kashmir region.

>*scl-Arab-Pk to scl-Arab-IN

Where are you seeing that? Both the keyboard_info file and the .kps file have it set to scl-Arab-IN

>*please fix this waw وٗٗ as the upper mark is not on the waw

I will try to do this. However, I feel quite strongly that this is a poor orthography. Rendering engines are going to have problems with putting two vowels (especially the same vowel!) on top of the consonant. If you have any influence with this community you should help them change to use characters that are already in Unicode. Using a base character plus a vowel to represent one character goes against Unicode encoding. Additionally, I'll link to a page here from Microsoft here: https://docs.microsoft.com/en-us/typography/script-development/arabic Click on "Handling Invalid Combining Marks". Then notice where it says 

>If an invalid combination is found, like two fathas on the same base character, the diacritic that causes the invalid state is placed on a dotted circle to indicate to the user the invalid combination. 

This is a perfect description of what is happening with the two U+0657 characters on the waw: "the diacritic that causes the invalid state is placed on a dotted circle to indicate to the user the invalid combination"

I don't see anything to fix in your first two points, but I'll try to fix the graphic.

LornaSIL:
This is an example of how it appears in Word:
image
image

I've also included examples of the other characters that use vowels or sukun (with an additional vowel added). They just aren't going to shape properly with Microsoft applications.

The same example renders somewhat better in LibreOffice, but notice that the vowel now actually comes between the base character and the sukun, which would not be acceptable if the base+sukun is considered one character.
image

Before we proceed further with this keyboard would you be able to send me a scan of a document that has been published illustrating this orthography? I'd like to see their list of characters.

rachitralli:
Sent from Yahoo Mail on Android 
 
  On Fri, Jan 18, 2019 at 1:25 AM, Lorna Evans<notifications@github.com> wrote:   
This is an example of how it appears in Word:



I've also included examples of the other characters that use vowels or sukun (with an additional vowel added). They just aren't going to shape properly with Microsoft applications.

The same example renders somewhat better in LibreOffice, but notice that the vowel now actually comes between the base character and the sukun, which would not be acceptable if the base+sukun is considered one character.


Before we proceed further with this keyboard would you be able to send me a scan of a document that has been published illustrating this orthography? I'd like to see their list of characters.

LornaSIL:
@rachitralli I don't see your message. Can you comment in the web interface rather than by email?

rachitralli:
screenshot_2019-01-18-17-07-38-92
screenshot_2019-01-18-17-07-43-95
screenshot_2019-01-18-17-08-57-12
@LornaSIL excerpts from Shina Primer published in Indian State Board of Education, India

LornaSIL:
@rachitralli Thanks a lot for this information. We should be able to approve the keyboard, but I still think people will be very unhappy with computer support for their orthography.

rachitralli:
@LornaSIL Thanks

rachitralli:
@DavidLRowe please see comments of @LornaSIL thanks

rachitralli:
@LornaSIL @DavidLRowe thank you very much

PR #604, opened 2019-01-23
Update Galaxie Greek (Phonetic) help

nikita-moor:
Keyboard: [Galaxie Greek (Phonetic)](https://keyman.com/keyboards/galaxie_greek_positional)
System: Arch Linux, Keyman 11.0.105.

The help file explains typing "ὧμέγα" as `w"[[me/ga` which in fact produces "ὧ῾μέγα", i.e. second "[" is excessive. Could it be a typo or specific behavior of the linux port?

P.S. Searching though the whole repository, finds "ὧμέγα" in several other keyboard layouts probably needed to be fixed too.


keyman-server:

PR #615, opened 2019-02-14
[nailangs] Initial submission for Nailangs keyboard

mcdurdin:
This keyboard is moving to open source per discussion with the author Ade Oseni.

Documentation is minimal but hoping it's enough to get over the line.


LornaSIL:
* ɲ and Ɲ are only available on keyboards with 102nd key. I'd like to see those added somewhere else as well (for availability for people who don't have that kind of keyboard), perhaps on the ralt and ralt-shift layer.
* I haven't tested the touch layout yet, but I notice that some special keys do not have any text so you don't know what they do:
     * default layer: T_NUMERIC
     * shift layer: T_NUMERIC
     * numeric layer: T_LOWER, T_SYMBOL
     * symbol layer: T_LOWER, T_SYMBOL
* There is no php help file. Do you want to create that or do you want us to?

LornaSIL:
The touch layout works fine. So, the things above need some work, but mostly it looks good.

release/n/nailangs/source/nailangs.kps
```diff
@@ -0,0 +1,140 @@
...
+      <Description>File Android-promo-graphic-for-desktop-Futago-copy.png</Description>
+      <CopyLocation>0</CopyLocation>
+      <FileType>.png</FileType>
+    </File>
+  </Files>
+  <Keyboards>
+    <Keyboard>
+      <Name>Nailangs</Name>
+      <ID>nailangs</ID>
+      <Version>1.2</Version>
```
darcywong00: This is version 1.3

mcdurdin: This will be resolved automatically in build due to `FollowKeyboardVersion` but I've fixed it anyway.

mcdurdin:
> * ɲ and Ɲ are only available on keyboards with 102nd key. I'd like to see those added somewhere else as well (for availability for people who don't have that kind of keyboard), perhaps on the ralt and ralt-shift layer.

Added (RALT+N, RALT+SHIFT+N).

> * I haven't tested the touch layout yet, but I notice that some special keys do not have any text so you don't know what they do:
>   
>   * default layer: T_NUMERIC
>   * shift layer: T_NUMERIC
>   * numeric layer: T_LOWER, T_SYMBOL
>   * symbol layer: T_LOWER, T_SYMBOL

Fixed.

> * There is no php help file. Do you want to create that or do you want us to?

Added.

PR #617, opened 2019-02-19
[coptic_qwerty] New keyboard submitted on behalf of author

LornaSIL:
Implemented on behalf of Michael Everson based on his "Coptic_English keylayout 2.1.pdf"


DavidLRowe:
Copyright dates and holder in LICENSE.md don't match those in README.md.

Is the "All rights reserved." phrase found in readme.htm and coptic_qwerty.php compatible with MIT license?

Touch layout, shift layer, `.` produces U+FE25 not period. Longpress on that key: keys labeled `,` and `'` produce other characters.

LornaSIL:
I made this keyboard layout exactly match the documentation for v2.1 which is what the author wanted. This required documenting the need to remove Keyman hotkeys.

LornaSIL:
@mcdurdin I'm not sure what the problem with the build is because I've been able to build the package locally without errors. This keyboard does use the RALT SHIFT K, M, and O. I'm not sure where to look now.

DavidLRowe:
copyright date in LICENSE.md doesn't match other dates

OSK still has entries for `SHIFT RCTRL RALT K_BKQUOTE` and `RCTRL RALT K_BKQUOTE` but those are commented out in the .kmn file.

Likewise the .php has data-states='default shift rightalt rightalt-shift rightctrl-rightalt rightctrl-rightalt-shift' but the last two states aren't present.

Otherwise it looks good, as far as I can tell!

DavidLRowe:
@mcdurdin FYI: keyboard built on CI after removing `displayUnderlying` from .keyman-touch-layout file. (It was building locally in Developer, but not on TC.)

mcdurdin:
See #672 for tests on the breaking CI build. If this recurs, it will probably be sufficient to merge the latest upstream/master into the PR branch to ensure latest compiler version. #673 details a long-term fix which also simplifies deployment of compiler updates.

PR #619, opened 2019-02-20
[coptic_greek] New keyboard added on behalf of author

LornaSIL:
This keyboard just has a few changes from the coptic_qwerty keyboard. The changes are the following keys:
* Q, Y, U, H, and C
* There is no touch layout since I thought the Coptic (Qwerty) one would be sufficient


LornaSIL:
I changed this keyboard back to using the reserved hotkeys and added a description of how to turn off the hotkeys.

Changed BCP47 code to use Copt not Greek

Added authors pdf as a link in the welcome.htm file

PR #621, opened 2019-02-25
[rac_ormuri] New keyboard on behalf of author

LornaSIL:
The files from the author were fairly clean. I did add most of the characters on the RALT layer since they are used in the mobile layouts.


rachitralli:
@LornaSIL thanks

DavidLRowe:
copyright date in README.md should change to 2019 to match other files (HISTORY.md, LICENSE.md, readme.htm, etc). Or I suppose you could change them all to 2018-2019, but that's a lot of work!

The touch layout file has two layouts, for tablet and phone, which appear to be identical. One could be deleted.

In my testing, there was one key that didn't work for the touch (phone) layout, shift layer, third row, second key. This could be of a bug in Keyman Developer / Keyman Web 11. Could you double-check that key on your device?

Otherwise LGTM.

DavidLRowe:
No need to test. I reverted to Developer 10 and the key works fine.

PR #626, opened 2019-02-27
[sanskrit_unicode] Created a php file and point to another transliter…

LornaSIL:
…ation kbd since we cannot update this one

With the death of John Richards we will never be able to get the license approval to change from freeware to MIT.

### DIFF
--- a/legacy/s/sanskrit unicode/sanskrit unicode.keyboard_info
+++ b/legacy/s/sanskrit unicode/sanskrit unicode.keyboard_info
@@ -1,12 +1,15 @@
 {
     "id": "sanskrit unicode",
     "name": "Sanskrit Unicode",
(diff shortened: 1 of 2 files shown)


LornaSIL:
@mcdurdin  I added the "related" field. No, the keystrokes are not at all the same as in sil_indic_roman.

LornaSIL:
@mcdurdin 
Even tho there were no errors in the build, the home page hasn't regenerated and there is no help file still.

https://keyman.com/keyboards/sanskrit%20unicode
https://help.keyman.com/keyboard/sanskrit%20unicode

mcdurdin:
> Even tho there were no errors in the build, the home page hasn't regenerated and there is no help file still.

The downloads server ran out of space which killed the build. Have freed up space and watching the build again now.

PR #627, opened 2019-02-28
[rac_dogri] Keyboard submitted on behalf of author

DavidLRowe:
There are a number of questions about this keyboard

Keyboard has 'doi' for language code, but this is macrolanguage of India. Is there a better code than 'doi-Arab'?

In the .keyboard_info file, the example has `DwGri` for keys, but third character in text doesn't correspond. Using `G` for key gives U+063A, using `g` for key gives U+06AF. Either text or keys need to change.

@rachitralli 
- Please check `h` and `H` on both desktop and touch. They have the characters U+06BE and U+062D, but the original images for the welcome.htm had them switched.
- Also check `u` and `U`. The touch layout had them switched, so I changed it to match the desktop.
- Original .kmn file had `\` on Shift+B and U+0655 on `\`. I switched the two, so that U+0655 is on Shift+B and `\` is on the `\` key.
- I made other changes in order to synchronize the desktop and touch layouts, so that each produces all the characters of the other. Check that I haven't left out anything you need.
- Currently the touch layout has `123` to go to numeric layer and `abc` to return to default layer. Is this suitable (even though these are Latin characters)?

@LornaSIL What is the current recommendation for the header lines in .kmn file for `&version`, `&ethnologuecode`, `&language`, `&windowslanguages`?

I put a copy of the .kmp file in your google folder at: https://drive.google.com/drive/folders/1PwKgR4vWTY78177x159vjJna16WC3UIQ

Please test it and let me know what needs to change.


LornaSIL:
>What is the current recommendation for the header lines in .kmn file for `&version`, `&ethnologuecode`, `&language`, `&windowslanguages`?

We no longer recommend the use of those headers. I would comment them out. Sometimes it's useful information to retain.

LornaSIL:
Something has happened to the welcome.htm file. The Arabic text is displaying as ANSI characters. Need to add this line right after ```<head>```
```
<META http-equiv=Content-Type content="text/html; charset=UTF-8">
```

README.md: you may want to add a home link which would be https://keyman.com/keyboards/rac_dogri

php file: change ctrlalt to ctrl-alt

If language code needs to change, you'll need to modify keyboard_info, kps, kmn and all the places where there is a link: welcome.htm and php.

Touch layout, shift layer, bottom right key that has a single quote on it is outputting a double straight quote instead of single straight quote.

keyboard_info file either has wrong sequence for keys or for text. 
DwGri produces ڈوغری
Dwgri produces ڈوگری

@rachitralli you'll need to tell us which is correct. We also need to know if the macro language code doi is correct. Normally we don't use macro language codes.

rachitralli:
@LornaSIL please change the following 
please remove language code doi into dgo, and the other issues may kindly be changed, please
Dwgri , ڈوگری this is correct

PR #629, opened 2019-02-28
[sil_ipa-options] Modified kbd to support options and match current s…

LornaSIL:
…tructure


LornaSIL:
The only thing I changed on the touch layout was the diacritic key text.

I changed the readme.htm file to describe the options choice.

I changed the welcome.htm file to include which tables supported the options "Before" layout. I split the out of the "Diacritics & Suprasegmentals" (then the sil_ipa.php file was just a replacement of the text from welcome.htm)

LornaSIL:
Keyboard package using options has been checked by Andy Black and he says it is fine.

PR #639, opened 2019-03-07
[mongolian_cyrillic_qwerty] Mongolian Cyrillic QWERTY v 1.0

MakaraSok:
This phonetic keyboard layout is designed for Khalkha Mongolian Cyrillic.
It is based on the English QWERTY keyboard and also aligns with Chinese Pinyin.


MakaraSok:
@DavidLRowe Could you please review this PR as Lorna is tied up with something else in the means time?

DavidLRowe:
Keyboard (both desktop and touch) look great.

- It would be preferable to move the welcome.htm and the .png files it references to a `source/welcome` folder. (The .kps file would need to change to reference the new location.)
- Please add a `source/help` folder with a `mongolian_cyrillic_qwerty.php` file. This can be based on the welcome.htm file. It is possible to use the same .png files that the welcome.htm references, or else replace them in the .php file with references to the OSK, something like: 
```
<h3>Desktop Keyboard Layout</h3>
</div>

<h3>Mobile/Tablet Keyboard Layout</h3>
<div id='osk-phone' data-states='default shift symbol'>
</div>
```

MakaraSok:
@DavidLRowe Please check again and see is everything is OK. :)

Some changes have been made to the layout per keyboard author's request.

DavidLRowe:
The change for the desktop layout is fine, but for the touch layout, the keys for y/Y and ;/: are switched. Those keys were changed in the .kmn file, but then you doubled the change in the touch layout. I think you can switch `K_Y` and `K_COLON` in the `id` fields in the touch layout (both default and shift layers), but leave the `text` fields as they now are, and that should correct the problem. 

The .png files referenced in the .php file need to be duplicated in the `help` folder (that is, copied from the `welcome` folder). I believe that all the .png files are correctly showing the new layout.

MakaraSok:
@DavidLRowe 
* the touch layout has been fixed
* the images have been added to the help folder
* switch to Default layer once a key on Shift is pressed

PR #637, opened 2019-03-06
gff_geez v1.2

dyacob:
This update addresses a special case input sequence.  Non-functional changes are the use of localized store names in place of transliterated names.  Also the addition of the Brana font (under the OFL license).

This is a trial update for one keyboard in the GFF collections, 4 other updates will follow after the successful completion of this PR.

PR #641, opened 2019-03-08
gff_blin v1.2

dyacob:
This update for the Blin keyboard is analogous to the #637 update -and should go much smoother following the #637 experience.  Two documentation files from the gff_geez keyboard are included where a typo is fixed and text is synched with Blin.


keyman-server:

keyman-server:

keyman-server:

DavidLRowe:
This pull request looks pretty good. I don't have much familiarity with Ethiopic scripts, so some of my comments will likely reveal my ignorance!

- gff_blin.keyboard_info
   - The `byn` needs a Script code, which I think should be `Ethi`, making `byn-Ethi` (see also comments below on .kps file)
   - I don't think the `<p>` opening and closing tags are needed. I think they should be removed.
   - Is there a reason to deprecate both "gff-byn-7" and "gff_byn_7"? I see the second at `legacy/gff/` but not the first, unless you mean "gff-byn-powerpack-7"?
   - Typically a file is also added to the legacy folder being deprecated. In this case I think you want to add a file: `legacy/gff/gff_byn_7/DEPRECATED.md` containing the text "This keyboard has been deprecated and replaced by release/gff/gff_blin". 
- .kps file
   - I think the entry under Keyboards should be `byn-Ethi` (rather than `-Latn`). When I built and installed the package, Windows indicates that it is `byn-Latn-ER`. 
- .kvks file?
   - Typing Shift+A gives U+12D0, whereas the onscreen keyboard (OSK) indicates U+12A3.

Often we put extra files (such as the .docx and .pdf files in this PR) into a folder called `extras` which is at the same level as `source` (so `gff_blin/extras`). The .pdf can still be referenced from this folder by the .kps file. But if you already have a precedent of putting them in the source folder, then leave them as they are.

dyacob:
@DavidLRowe thanks for the review.  I'll address the feedback shortly and report back here.  Thanks again!

keyman-server:

DavidLRowe:
I meant to suggest looking at PR #613 for an example of a pull request that adds a DEPRECATED.md to the folder where the deprecated keyboard resides.

keyman-server:

dyacob:
@DavidLRowe I've made the recommended changes for the 3 files and tested and build and install.  All seems well, thanks for catching these defects.

One thing I skipped was creation of the DEPRECATED.md file, but only because it appears the DEPRECATED.md files are in place for the legacy byn folders since last year, see:

* https://github.com/keymanapp/keyboards/blob/master/legacy/gff/gff_byn_7/DEPRECATED.md
* https://github.com/keymanapp/keyboards/blob/master/legacy/gff/gff-byn-powerpack-7/DEPRECATED.md

I've also skipped creation of the "extras" folder, due to precedence with the other gff_* keyboards.  I'll make a change to this convention in a future revision.

DavidLRowe:
Ha ha! The joke's on me. It never occurred to me that the DEPRECATED.md files might already exist, so I never thought to look! 

All the other changes look good.

PR #642, opened 2019-03-12
Create lazuri.js

rlqmadkqk:
It seems that this file is needed for mobiles.


keyman-server:

DavidLRowe:
The .js file is created as part of the build process. The current .kps file correctly references it in the build folder and adds it to the keyboard package (.kmp). 

The .js file should not need to be added to the source folder. Let's close this PR and pursue the problem on issue #643.

PR #644, opened 2019-03-14
[lazuri] Update lazuri.kps

rlqmadkqk:
For language, Laz (Latin, Turkey) seems to much. Lazuri would be the best since it is the language name.  All Laz speakers in Turkey use Latin script

### DIFF
--- a/release/l/lazuri/source/lazuri.kps
+++ b/release/l/lazuri/source/lazuri.kps
@@ -84,7 +84,7 @@
       <ID>lazuri</ID>
       <Version>2.3</Version>
       <Languages>
-        <Language ID="lzz-Latn-TR">Lazuri</Language>
+        <Language ID="lzz-Latn">Lazuri</Language>
       </Languages>
     </Keyboard>
   </Keyboards>
--- a/release/l/lazuri/source/lazuri.kps
+++ b/release/l/lazuri/source/lazuri.kps
@@ -84,7 +84,7 @@
       <ID>lazuri</ID>
       <Version>2.3</Version>
       <Languages>
-        <Language ID="lzz-Latn-TR">Laz (Latin, Turkey)</Language>
+        <Language ID="lzz-Latn-TR">Lazuri</Language>
       </Languages>
     </Keyboard>
   </Keyboards>


keyman-server:

DavidLRowe:
The .keyboard_info file has `lzz-Latn` but the .kps file has `lzz-Latn-TR`. It would be better if they matched.

Using `lzz-Latn` in the Keyman package source (.kps) file (instead of the existing `lzz-Latn-TR`) generates `Laz (Latin)`, which you can then edit to `Lazuli` if you wish.

So the change I'm suggesting is removing the `-TR` from the language code in the .kps file so that it matches that in the .keyboard_info file. (You already have the language name as Lazuli.)

rlqmadkqk:
> The .keyboard_info file has `lzz-Latn` but the .kps file has `lzz-Latn-TR`. It would be better if they matched.
> 
> Using `lzz-Latn` in the Keyman package source (.kps) file (instead of the existing `lzz-Latn-TR`) generates `Laz (Latin)`, which you can then edit to `Lazuli` if you wish.
> 
> So the change I'm suggesting is removing the `-TR` from the language code in the .kps file so that it matches that in the .keyboard_info file. (You already have the language name as Lazuli.)

It sounds good. Both .kps and .keyboard_info have lzz-Latn.
The reason for suggesting this change was that Laz people doesn't like Laz (Latin) in mobile screen like below. As long as it says Lazuri, I'm OK with other necessary changes. 

Screenshot_20190315-102136

keyman-server:

DavidLRowe:
I think that we need to increment the version number so that these changes will be built and made available. (Unfortunately, @LornaSIL is away. She knows more about this than I do.)

You'll need to increment the version number from 2.3. It could go to 2.4, or another option is to use 2.3.1, if you want to indicate this is a minor change. Files that need to change:
- HISTORY.md - with the version number and a description like "update language tag and description"
- README.md
  - version number
  - also in the text for `* BCP-47 codes:`, I think that `tr-Latn-TR` should change to `lzz-Latn`
- lazuli.kmn
  - the &KEYBOARDVERSION needs to have the new version number
  - optional, but useful: comment out (or delete) the &ETHNOLOGUECODE line. This is no longer used and eventually will become a warning, so dealing with it now might save time later.
  - optional, but helpful: change the &TARGETS line to only have 'any' (that is `store(&TARGETS) 'any'`). Since 'any' implies all the rest, listing them separately is not necessary.
- lazuli.kps
  - The "Package version follows keyboard version" box is ticked, but the old "2.3" is still there. It seems that you need to untick the box, delete the version number, then tick the box again. Then the package will pick up the version number from the .kmn file.

Hope I remembered everything!

LornaSIL:
This change would require a bump in the version number in all places.

PR #645, opened 2019-03-14
gff_amharic v1.8

dyacob:
This update for the Amharic keyboard is analogous to the #637 & #641 updates -and should go much smoother following their experience.

PR #646, opened 2019-03-15
gff_tigrinya_ethiopia v1.2

dyacob:
This update for the Tigrinya-ET keyboard is analogous to the #637, #641, and #645 updates -and should go much smoother following their experience.


keyman-server:

DavidLRowe:
The onscreen keyboard (OSK) seems to be blank.

keyman-server:

dyacob:
@DavidLRowe the on-screen keyboard is populated now, thank you.

keyman-server:

DavidLRowe:
Language tag (BCP 47 code) in .keyboard_info file is incorrect: Rather than "ti-et-Ethi" it should be "ti-Ethi-ET" or simply "ti-Ethi" if the language (with the same script) doesn't have a different form in another country/region.

Will review keyboard later.

keyman-server:

DavidLRowe:
Minor point on OSK: some key tops on the OSK are blank (or only showing underlying default key) but still produce that underlying key. On default layer: 
```
`=[]\./
```
On shift layer:
```
~!@$%^&*(){}|"?
```

It seems that it would be better to fill the OSK completely.

Another minor point: The gff_tigrinya_ethiopia.php and welcome.htm files have:
```
This is a Tigrinya (tir-ET, ትግርኛ-ኢትዮጵያ) language mnemonic 
input method that applies Ethiopian writing conventions.
```
I wonder if "tir-ET" is the best code to use. The BCP 47 code uses "ti" (since two-letter language codes are preferred over three-letter codes).

keyman-server:

dyacob:
I've populated the punctuation in the OSK, I'll need to check this on other GFF keyboards.  Thanks for catching this.

keyman-server:

DavidLRowe:
The &ethnologuecode line is no longer used (and will become a warning/error in future versions of Keyman). I would suggest either commenting out or deleting that line in the .kmn file.

I think that using the BCP 47 code of "ti" is best.

keyman-server:

dyacob:
I've switched over to "ti" from "tir" everywhere and have commented out the 3 language stores now marked as "Deprecated" in the [Keyman Keyboard Language Reference](https://help.keyman.com/developer/language/reference/).  Effectively, the .kmn file no longer identifies any language for the keyboard, that appears to be a job left to the .kps file.  I learn a lot from these pull requests! :-)

DavidLRowe:
@LornaSIL pointed out that you have two keyboards for Tigrinya, one for Ethiopia and one for Eritrea. Are there differences in the language between the two countries (as between US and UK English)? or just in how the characters are typed? If the former, then the use of `ti-ET` is indicated, but if the latter, then you might be better off using `ti` for both keyboards. But I'm far from being an expert on BCP 47 codes!

dyacob:
There is some difference like US vs UK English for accent and word choice.  But then they also use different subsets of the the Ethiopic syllabary.  Where Eritrea fully eradicated the historic redundant characters inherited from Ge'ez for Tigrinya (and other languages like Blin).  While Ethiopian Tigrinya held on to the historic letters (though thats been gradually eroding).  The Eritrean and Ethiopian keyboards for Tigrinya reflect these different letter inventories.

DavidLRowe:
The value for Shift+4 on the OSK should be `$` (not `#`). With that change, this should be ready to approve and merge.

keyman-server:

PR #647, opened 2019-03-20
[gff_tigrinya_eritrea] gff_tigrinya_eritrea v1.2

dyacob:
This update for the Tigrinya-ER keyboard is analogous to the #646 updates -and should go much smoother following the prior experience.


keyman-server:

DavidLRowe:
@dyacob The gff/shared folder was causing the build to fail (so no new keyboards were actually being deployed). PR #648 is my attempt to switch to using shared/gff folder instead to make the build system happy. 

For this PR, you'll need to wait until 648 is merged, then update your copy, then change all references to the shared folder. (Perhaps just the readme.htm file in a number of places?)

dyacob:
ok, understood, i'll watch for PR #648 to complete.  The data savings of a shared folder are not particularly high in this case.  Since creating the shared folder, I began creating individual side images for the keyboards, so now only two keyboards actually share a common side image.  If the shared folder seems to be more trouble than its worth, the contents can be copied back into the respective keyboard folders as appropriate.

keyman-server:

keyman-server:

dyacob:
@DavidLRowe , done on this end, thanks.

LornaSIL:
All the files appear to be correct.

I had trouble with the installation though. First of all while installing (and before the welcome.htm file popped up) I got this message.

ti-Ethi-ER

Then it installed it and now I have two Tigrigna keyboards. One is the system Windows Tigrigna keyboard (with the white on black fidel) and the other is @dyacob keyboard (with the red fidel) but now associated with geez.

ti-Ethi-ER-ታስክባር

@mcdurdin Is there anything we can do about this or should we just document it or ignore it? Since I suspect this is the issue @DavidLRowe ran into with the gff_tigrinya_ethiopia keyboard I don't think we should just ignore it.

BCP47 code is:
> ti-Ethi-ER

The "ER" is needed because there is another keyboard for "ET"

Or, is it just a problem on my computer?

dyacob:
Hi @LornaSIL , welcome back! I think the dialog is part of the work around for issue https://github.com/keymanapp/keyman/issues/1285 , it also happens with Amharic.  The appearance of the system keyboard and the name change to Geez is unfortunate, I've seen it confuse users. I think its the state of things until Microsoft addresses the issue, some other non-Ethiopian languages were also impacted if I recall correctly.

LornaSIL:
Great! I somehow missed that while I was ignoring emails.

mcdurdin:
@dyacob has the story right. It's another one to follow up with Microsoft again when we get a chance (it may well be fixed in a recent build -- just haven't had time to install and verify).

PR #651, opened 2019-03-24
[infrastructure] Always use wine when not on Windows

rimas-kudelis:
The `build.sh` script as it is now failed to build the layouts on Linux for me. This small patch changes how `KMCOMP_LAUNCHER` is defined: instead of only setting the value to `wine` on darwin, it will only be left empty on Windows and set to `wine` on all other platforms.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -22,11 +22,11 @@ function display_usage {
 KEYBOARDROOT="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 KMCOMP="$KEYBOARDROOT/tools/kmcomp.exe"
 
-if [[ "${OSTYPE}" != "darwin"* ]]; then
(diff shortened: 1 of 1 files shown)


keyman-server:

LornaSIL:
@darcywong00 this is not a keyboard PR. Does this belong in https://github.com/keymanapp/keyman?

rimas-kudelis:
Huh, how can a PR which changes files in this repository belong to another repository? :)

DavidLRowe:
@mcdurdin plans to look at this later this week.

mcdurdin:
> @darcywong00 this is not a keyboard PR. Does this belong in https://github.com/keymanapp/keyman?

This is a pull request for the build infrastructure. It does belong here :) On the surface it all looks fine, just checking it now.

PR #652, opened 2019-03-24
[lithuanian] Recreated and slightly improved legacy Lithuanian layout

rimas-kudelis:
I have a link to Keyman on one of the websites I manage, and the current `lithuanian` layout, which is present in `legacy\l\lithuanian`, has been bothering me a little by not offering a few typographical characters (namely, curly quotes and an en-dash). Although, that's no wonder since it has been and still is that way on Windows.

Attached here is a re-creation of the same layout, but with a few extras. This version of the layout is used by default on Linux, BSD etc., and there are also third-party (well, mine) layout files available for Windows and macOS.

Here's a graphical summary of the changes in the layout:
paveikslas
In the picture, the two characters in orange are additions to the current layout, and two characters in red are changes. Note that even though the changes occur in the default (unshifted) layer of the keyboard, they are unlikely to cause any trouble, as they redefine the only two digits that were left in there (all digits are available in `CTRL ALT` layer).

I tried using `RALT` instead of `CTRL ALT` at first, but it seems KeymanWeb still doesn't support that, so I left that for future.

Since `build.sh` emitted an error when two layouts had the same name (`legacy\l\lithuanian` and `release\l\lithuanian`, I removed the former in order for the build to pass.

Also, it seems that such simple layouts are out of focus for Keyman at the moment, so I won't be surprised or insulted if this PR does not get accepted. Also, I could just edit the legacy layout instead, if that's OK.

Additionally, I'm pretty certain that this also deprecates the layout defined in `legacy\kbd\kbdlt1`, so I marked that layout accordingly.


keyman-server:

LornaSIL:
We are planning to update all the Win keyboards in the near future. I would like to do that and incorporate your improvements at that time if you are willing for us to handle it that way. We don't want to delete the legacy Lithuanian keyboard, just mark it as deprecated. I can move this to the top of my list of Windows keyboards but it would still likely be a few weeks until I can do it.

rimas-kudelis:
@LornaSIL sure, this can wait, no problem.
Like I said though, the reason why I decided to remove the original layout was because of the name collision. Simply renaming the folder worked too, but that meant the layout would no longer be built anyway, or the file too had to be renamed and probably edited. I figured there wasn't much use in preserving it, especially considering that the layout I did works pretty much the same. But it's your call.

I'm a bit curious about your plans WRT Windows keyboards though. What happens once you import them? Will they be offered/used anywhere? What do you plan to do with the duplicates (or near-duplicates), when Windows layout matches a non-Windows one?

mcdurdin:
A few thoughts here:

@rimas-kudelis wrote:
> I tried using `RALT` instead of `CTRL ALT` at first, but it seems KeymanWeb still doesn't support that, so I left that for future.

KeymanWeb does support `RALT` in version 10. If you were unable to use Right Alt / AltGr with KeymanWeb then we do need to figure out why.

@rimas-kudelis wrote:
> Additionally, I'm pretty certain that this also deprecates the layout defined in `legacy\kbd\kbdlt1`, so I marked that layout accordingly.

I think that is correct. The naming scheme for the replication of Windows keyboard layouts is moving to using `basic_kbdxxx`, e.g. `basic_kbdlt1`. So it is worth considering whether this keyboard should be called `basic_kbdlt1` rather than `lithuanian`, just for consistency with the naming scheme.

There are some differences between the existing legacy `lithuanian` and `kbdlt1` keyboards which can be eliminated now, all around handling of Right Alt and Caps Lock. We should be able to handle rules for both of those identically with Keyman 10/11 on Windows, macOS, Linux and Web, so the simplifications we had in the past for web should no longer be required. This means that actually the kbdlt1 keyboard is a better basis than lithuanian.

@LornaSIL wrote:
> We don't want to delete the legacy Lithuanian keyboard, just mark it as deprecated. 

As the legacy Lithuanian keyboard can be replaced by this one as it is backwardly compatible in layout, then the correct approach is to move it from `legacy` to `release` as has been done in this PR. That doesn't delete the `lithuanian` keyboard, just changes its location in the source repository.

@rimas-kudelis wrote:
> I'm a bit curious about your plans WRT Windows keyboards though. What happens once you import them? Will they be offered/used anywhere? What do you plan to do with the duplicates (or near-duplicates), when Windows layout matches a non-Windows one? Or even more broadly, do you tend to avoid duplicate layouts or not really?

One reason for creating duplicates of all the Windows keyboards is to provide a base for other minority language layouts that are closely related to a country-wide layout. We try to avoid duplication -- most of the duplication we see at present in the legacy area, e.g. `kbdlt1` and `lithuanian`, arises from the era when KeymanWeb could not do Caps Lock or Right Alt. Now that this is no longer the case, we are progressively removing that duplication.

keyman-server:

rimas-kudelis:
@mcdurdin 
> >   I tried using RALT instead of CTRL ALT at first, but it seems KeymanWeb still doesn't support that, so I left that for future.
>
> KeymanWeb does support RALT in version 10. If you were unable to use Right Alt / AltGr with KeymanWeb then we do need to figure out why.

<....>

> I think that is correct. The naming scheme for the replication of Windows keyboard layouts is moving to using basic_kbdxxx, e.g. basic_kbdlt1. So it is worth considering whether this keyboard should be called basic_kbdlt1 rather than lithuanian, just for consistency with the naming scheme.

WRT naming, I won't strongly object, but I'm not in favor of that either. Even though this is purely a technical aspect, I don't think tying generic layout names to Windows dll names (which always start with kbd and never have more than 8 characters, which means you end up with only 5 chars for the actual name) makes much sense, does it?

> There are some differences between the existing legacy lithuanian and kbdlt1 keyboards which can be eliminated now, all around handling of Right Alt and Caps Lock. We should be able to handle rules for both of those identically with Keyman 10/11 on Windows, macOS, Linux and Web, so the simplifications we had in the past for web should no longer be required. This means that actually the kbdlt1 keyboard is a better basis than lithuanian.

I'm all in favor of eliminating the duplicates. There is no source for kbdlt1 though, so I still had to recreate the layout from scratch, right? Or will you actually have source when you convert Windows 10 layouts? What's the timeframe for that, BTW?

Either way, I have rewritten the layout to use RALT again and added support for CAPS too. However, it seems I've now lost the opportunity to edit the layout visually, since it is now considered "too complex". Also, I stumbled upon some glitches while testing the non-web layout in Keyman Developer on Windows, most notably, the Right-alt state was not visible in the on-screen keyboard (because depressing AltGr also triggered LCtrl), and initially didn't even seem to work at all (until I switched the layout in Windows keyboard switcher a couple times). Could it be that the layout that Keyman creates actually depends on the layout that was active before? 

...and then the web layout basically didn't work at all in Keyboard Test Host. To explain the latter in more detail, I switched my Windows layout to English US, opened the Test Host, and:

1. RALT was not indicated differently from LALT in the virtual keyboard
2. pressing the top row keys on the keyboard actually produced US output (numbers) instead of Lithuanian (letters with diacritics) – and this was the case even when clicking the JS-rendered virtual keyboard with the mouse! And even when I switched Windows layout to Lithuanian!
3. CAPS state, while in itself appeared to be indicated correctly by the virtual keyboard,  still showed all letters in lowercase
4. CAPS+E produced € (a character from the AltGr layer)

With the web keyboard, I had similarly disappointing results in both Firefox and Edge on latest Windows 10.

Also, the compiler warning me about Extended shift flags being unsupported in KeymanWeb doesn't sound encouraging either. :)

All in all, it seems this PR is far from ready for merging yet, so please don't do that for now. Also, I have a couple questions:

1. again, what's the ETA for import of Win10 layouts, and will they contain layout sources? With my layout as is now, it seems probably more reasonable to just close this PR altogether and instead modify the updated kbdlt1 instead, assuming there will be sources.
2. in kmn files, it seems I can define either all keys or just those that are different from the base. What are cons and pros of each approach? Which one is preferred? What is used as the base layout in case I go with the second approach and only define differences?
3. in kmn files, does the order of rules within a group matter? From my experiments, it seems like it does, and if so, what order should the rules be written in? Can Keyman Developer order them correctly for me?
4. in kmn and kvks, what should I use as name and kbdname, respectively? Should the name be in Lithuanian or English?
5. Developer has changed file id's in .kpj files for me. Aren't these random? Should I commit these changes? Also, it seems to have silently removed the .kmx file from the project. Should I commit that change too?
6. Do I really have to provide an icon for the layout? I kinda assumed Keyman would generate something automatically when it's not provided, but that doesn't seem to be the case.

That's all for now, bye.

mcdurdin:
> WRT naming, I won't strongly object, but I'm not in favor of that either. Even though this is purely a technical aspect, I don't think tying generic layout names to Windows dll names (which always start with kbd and never have more than 8 characters, which means you end up with only 5 chars for the actual name) makes much sense, does it?

Given this is not a pure clone of the Windows keyboard, you are probably right. In which case, there is an argument for giving your keyboard a name to distinguish it from other Lithuanian keyboards that may be in the repo (or added at a future time). We would then deprecate the existing kbdlt1 and lithuanian keyboards in favour of this one.

> I'm all in favor of eliminating the duplicates. There is no source for kbdlt1 though, so I still had to recreate the layout from scratch, right? Or will you actually have source when you convert Windows 10 layouts? What's the timeframe for that, BTW?

Source is not available for kbdlt1 in the repo because we are replacing it. Using the Import Windows Keyboard project tool in Keyman Developer 11 provides an exact replacement. We don't have a timeframe on converting all the layouts; it happens as we have the volunteer time available to work on it.

> Either way, I have rewritten the layout to use RALT again and added support for CAPS too. However, it seems I've now lost the opportunity to edit the layout visually, since it is now considered "too complex". 

Yes, we don't currently support `CAPS` rules in the visual editor. This is a feature that wouldn't be too hard to add (unlike deadkeys, which would be painful!). However, I am hesitant to add it as a feature request at this time because of the number of features and issues that we are already working on. If you feel like this would be super valuable, then feel free to [add a feature request](https://github.com/keymanapp/keyman/issues/new?assignees=&labels=enhancement&template=feature_request.md&title=) on the Keyman repo!

> Also, I stumbled upon some glitches while testing the non-web layout in Keyman Developer on Windows, most notably, the Right-alt state was not visible in the on-screen keyboard (because depressing AltGr also triggered LCtrl), and initially didn't even seem to work at all (until I switched the layout in Windows keyboard switcher a couple times). Could it be that the layout that Keyman creates actually depends on the layout that was active before?

This sounds like it is related to the Windows base keyboard. Can you tell me which base keyboard you have (both your default keyboard in Windows, and the one specified in Keyman Configuration / Options / Base Layout)?

> 
> ...and then the web layout basically didn't work at all in Keyboard Test Host. To explain the latter in more detail, I switched my Windows layout to English US, opened the Test Host, and:
> 
> 1. RALT was not indicated differently from LALT in the virtual keyboard
> 2. pressing the top row keys on the keyboard actually produced US output (numbers) instead of Lithuanian (letters with diacritics) – and this was the case even when clicking the JS-rendered virtual keyboard with the mouse! And even when I switched Windows layout to Lithuanian!
> 3. CAPS state, while in itself appeared to be indicated correctly by the virtual keyboard,  still showed all letters in lowercase
> 4. CAPS+E produced € (a character from the AltGr layer)

These issues all arise from having the wrong target version for your keyboard -- see below.

> 
> With the web keyboard, I had similarly disappointing results in both Firefox and Edge on latest Windows 10.
> 
> Also, the compiler warning me about Extended shift flags being unsupported in KeymanWeb doesn't sound encouraging either. :)

You need to update your target version to KeymanWeb 10.0 with `store(&version) '10.0'`. Earlier versions of KeymanWeb don't support the extended shift flags. Once you do that, you'll get a different set of warnings around deprecated language code data -- instead put that language data into the .kps file.

Another note regarding use of the `CAPS` modifier. If neither the `CAPS` nor the `NCAPS` modifier is used in a rule, then the state of the Caps Lock key is ignored -- so then some of the rules become ambiguous. The appropriate fix is to add `NCAPS` modifiers to all the rules where you want to distinguish Caps Lock. That should improve 

> 
> All in all, it seems this PR is far from ready for merging yet, so please don't do that for now. Also, I have a couple questions:
> 
> 1. again, what's the ETA for import of Win10 layouts, and will they contain layout sources? With my layout as is now, it seems probably more reasonable to just close this PR altogether and instead modify the updated kbdlt1 instead, assuming there will be sources.

We don't have an ETA as explained above; you can import a Windows 10 keyboard in Keyman Developer through the New Project dialog, if you want to do one yourself.

> 2. in kmn files, it seems I can define either all keys or just those that are different from the base. What are cons and pros of each approach? Which one is preferred? What is used as the base layout in case I go with the second approach and only define differences?

In kmn files, you have two options:

1. Use a positional layout -- in which case any undefined keys use US English as the base layout (as a lowest common denominator).
2. Use a [mnemonic layout](https://help.keyman.com/developer/language/reference/mnemoniclayout) -- undefined keys retain the behaviour of the base layout, but rule definitions also need to take the base layout into account, especially with punctuation keys which move around the most.

> 3. in kmn files, does the order of rules within a group matter, and if so, what order should the rules be written in? Can Keyman Developer order them correctly for me? I wanted to order rules in a somewhat sane order (unshifted at the top, then shift, altgr, both, caps), but that didn't seem to work well...

Order is important, but only for rules which would otherwise have ambiguous precedence: e.g. two rules that reference the same keystroke. When compiling, Keyman Developer orders the rules by longest context to shortest, and then uses file order to determine precedence after that. Rules without context retain file order.

> 4. in kmn and kvks, what should I use as name and kbdname, respectively? Should the name be in Lithuanian or English?

The name can be in Lithuanian or English; some people use the keyboard language followed by the English name in parentheses. The name does not have to be identical in the online keyboard catalogue, which can help for finding a keyboard; once installed it is usually pretty obvious which keyboard is which.

> 5. Developer has changed file id's in .kpj files for me. Aren't these random? Should I commit these changes? Also, it seems to have silently removed the .kmx file from the project. Should I commit that change too?

The file ids can change when the filename changes (or between some versions of Keyman Developer). The changes are not important but there is no harm in committing them. .kmx files shouldn't be in the project; is it possible that the file was referenced in a .kps file?

> 6. Do I really have to provide an icon for the layout? I kinda assumed Keyman would generate something automatically when it's not provided (especially since it does that for stock Windows layouts already, as is evident in the Keyman OSK), but that doesn't seem to be the case.

Keyman doesn't generate icons for keyboards. If you just want a stock icon, it's only a few seconds to drop some letters on the icon editor :)

> That's all for now, bye.

Hope I've covered everything here! The main takeaways are:

1. Update the version of the keyboard: `store(&version) '10.0'`
2. Use `NCAPS` for rules that are Caps Lock off.

rimas-kudelis:
Thanks @mcdurdin, I'll apply the suggested changes.

> Given this is not a pure clone of the Windows keyboard, you are probably right. In which case, there is an argument for giving your keyboard a name to distinguish it from other Lithuanian keyboards that may be in the repo (or added at a future time). We would then deprecate the existing kbdlt1 and lithuanian keyboards in favour of this one.

I actually avoided the "it's not a pure clone" rationale, because I think my arguments against windows naming apply even without it. :) I still prefer to use the short name "Lithuanian" for this layout: variations of it are default for the language and are thus called simply "Lithuanian" in all popular operating systems, so I believe it makes most sense to call it by the same name here.

> > Also, I stumbled upon some glitches while testing the non-web layout in Keyman Developer on Windows, most notably, the Right-alt state was not visible in the on-screen keyboard (because depressing AltGr also triggered LCtrl), and initially didn't even seem to work at all (until I switched the layout in Windows keyboard switcher a couple times). Could it be that the layout that Keyman creates actually depends on the layout that was active before?
> 
> This sounds like it is related to the Windows base keyboard. Can you tell me which base keyboard you have (both your default keyboard in Windows, and the one specified in Keyman Configuration / Options / Base Layout)?

I have multiple different keyboard layouts installed on that machine: a bunch of Lithuanian ones as well as a US-English one (I set it up specifically for playing with keyboard layouts, so it's kinda natural). So you're confirming that I should stay aware of which keyboard layout is active before switching to the Keyman Debugger (or whatever that layout is called)? I assume this rule won't apply to distributed positional layouts, is that correct? :)

> > ...and then the web layout basically didn't work at all in Keyboard Test Host. To explain the latter in more detail, I switched my Windows layout to English US, opened the Test Host, and:
<...>
> 
> These issues all arise from having the wrong target version for your keyboard -- see below.
> 
> > Also, the compiler warning me about Extended shift flags being unsupported in KeymanWeb doesn't sound encouraging either. :)
> 
> You need to update your target version to KeymanWeb 10.0 with `store(&version) '10.0'`. Earlier versions of KeymanWeb don't support the extended shift flags. Once you do that, you'll get a different set of warnings around deprecated language code data -- instead put that language data into the .kps file.

Ah, I see. Thanks! I'll bump the target version for sure.

> Another note regarding use of the `CAPS` modifier. If neither the `CAPS` nor the `NCAPS` modifier is used in a rule, then the state of the Caps Lock key is ignored -- so then some of the rules become ambiguous. The appropriate fix is to add `NCAPS` modifiers to all the rules where you want to distinguish Caps Lock. That should improve

:+1: 

> We don't have an ETA as explained above; you can import a Windows 10 keyboard in Keyman Developer through the New Project dialog, if you want to do one yourself.

Ah, got it. I'll have a look. Does the tool import all states correctly, or should I watch out for issues?

> > 1. in kmn files, it seems I can define either all keys or just those that are different from the base. What are cons and pros of each approach? Which one is preferred? What is used as the base layout in case I go with the second approach and only define differences?
> 
> In kmn files, you have two options:
> 
>     1. Use a positional layout -- in which case any undefined keys use US English as the base layout (as a lowest common denominator).
> 
>     2. Use a [mnemonic layout](https://help.keyman.com/developer/language/reference/mnemoniclayout) -- undefined keys retain the behaviour of the base layout, but rule definitions also need to take the base layout into account, especially with punctuation keys which move around the most.

Yeah, I'm using the positional layout. The question was more about which approach you prefer in the repository: whether I should define everything explicitly, or leave unchanged bits from QWERTY implicit. In other words, what is the convention you follow in your repo?

> > 1. in kmn files, does the order of rules within a group matter, and if so, what order should the rules be written in? Can Keyman Developer order them correctly for me? I wanted to order rules in a somewhat sane order (unshifted at the top, then shift, altgr, both, caps), but that didn't seem to work well...
> 
> Order is important, but only for rules which would otherwise have ambiguous precedence: e.g. two rules that reference the same keystroke. When compiling, Keyman Developer orders the rules by longest context to shortest, and then uses file order to determine precedence after that. Rules without context retain file order.

What about cases when e.g. there's a rule with SHIFT+key and another with RALT+key, and then I try to press AltGr+Shift+key? Would any of the two rules apply in such case?

> > 1. in kmn and kvks, what should I use as name and kbdname, respectively? Should the name be in Lithuanian or English?
> 
> The name can be in Lithuanian or English; some people use the keyboard language followed by the English name in parentheses. The name does not have to be identical in the online keyboard catalogue, which can help for finding a keyboard; once installed it is usually pretty obvious which keyboard is which.

I see. Well this question was actually about your conventions above anything else.

> > 1. Developer has changed file id's in .kpj files for me. Aren't these random? Should I commit these changes? Also, it seems to have silently removed the .kmx file from the project. Should I commit that change too?
> 
> The file ids can change when the filename changes (or between some versions of Keyman Developer). The changes are not important but there is no harm in committing them. .kmx files shouldn't be in the project; is it possible that the file was referenced in a .kps file?

I made my initial commit on a Linux machine (had some trouble setting up Windows at the moment), so I just modelled my project files by looking at `libtralo` and `lazuri` projects, which both reside under `releases/l/`. Both of these projects reference .kmx files, that's why I added mine.

> > 1. Do I really have to provide an icon for the layout? I kinda assumed Keyman would generate something automatically when it's not provided (especially since it does that for stock Windows layouts already, as is evident in the Keyman OSK), but that doesn't seem to be the case.
> 
> Keyman doesn't generate icons for keyboards. If you just want a stock icon, it's only a few seconds to drop some letters on the icon editor :)

Generating icons automatically for Keyman layouts along with native ones would be a nice feature. Worth a feature request maybe?

> Hope I've covered everything here! The main takeaways are:
> 
>     1. Update the version of the keyboard: `store(&version) '10.0'`
> 
>     2. Use `NCAPS` for rules that are Caps Lock off.

Alright, thanks again! I'll prolly commit some new fixes this evening.

keyman-server:

rimas-kudelis:
I've commited an updated layout, it seems to work this time (with an exception of KP_EQUAL on Firefox, as described in keymanapp/keyman#1694.

May I assume there's a guarantee that between [CAPS K_FOO], [NCAPS K_FOO], and [RALT K_FOO], the latter will always take precedence if pressing RALT whether or not CAPS is active at that moment?

mcdurdin:
> 
> I have multiple different keyboard layouts installed on that machine: a bunch of Lithuanian ones as well as a US-English one (I set it up specifically for playing with keyboard layouts, so it's kinda natural). So you're confirming that I should stay aware of which keyboard layout is active before switching to the Keyman Debugger (or whatever that layout is called)? I assume this rule won't apply to distributed positional layouts, is that correct? :)

Correct, a distributed positional layout should work the same, no matter what the user's installed Windows layout is. You can verify this by installing the layout yourself.

> 
> > We don't have an ETA as explained above; you can import a Windows 10 keyboard in Keyman Developer through the New Project dialog, if you want to do one yourself.
> 
> Ah, got it. I'll have a look. Does the tool import all states correctly, or should I watch out for issues?

As far as I know, the import tool imports all states correctly. There may be some edge cases around modifiers that Keyman doesn't support (e.g. Korean, Japanese keyboards have some other modifier keys) and around SGCAPS (Swiss German Caps Lock, which behaves somewhat differently than others).

> Yeah, I'm using the positional layout. The question was more about which approach you prefer in the repository: whether I should define everything explicitly, or leave unchanged bits from QWERTY implicit. In other words, what is the convention you follow in your repo?

I don't think we have a consistent convention in the repo at this time, with keyboards contributed over a long period of time, but my preference would be for minimal definition, only the bits that change from US English QWERTY.

> 
> > > 1. in kmn files, does the order of rules within a group matter, and if so, what order should the rules be written in? Can Keyman Developer order them correctly for me? I wanted to order rules in a somewhat sane order (unshifted at the top, then shift, altgr, both, caps), but that didn't seem to work well...
> > 
> > 
> > Order is important, but only for rules which would otherwise have ambiguous precedence: e.g. two rules that reference the same keystroke. When compiling, Keyman Developer orders the rules by longest context to shortest, and then uses file order to determine precedence after that. Rules without context retain file order.
> 
> What about cases when e.g. there's a rule with SHIFT+key and another with RALT+key, and then I try to press AltGr+Shift+key? Would any of the two rules apply in such case?

Modifier keys Shift, Ctrl, and Alt (and the left/right versions) require an exact match for a rule to fire. State keys (Caps, Num Lock, Scroll Lock -- don't use these last two though!) are ignored if not specified, which is why there are explicit on and off state modifiers for them (`NCAPS` vs `CAPS`).

> > The name can be in Lithuanian or English; some people use the keyboard language followed by the English name in parentheses. The name does not have to be identical in the online keyboard catalogue, which can help for finding a keyboard; once installed it is usually pretty obvious which keyboard is which.
> 
> I see. Well this question was actually about your conventions above anything else.

Our conventions here are also not very well defined. I think I'd like to see us moving towards a separate locale name for keyboards in the future -- but this opens a bit of a can of worms for multi-language keyboards, which is part of the reason we haven't done this yet.

> Generating icons automatically for Keyman layouts along with native ones would be a nice feature. Worth a feature request maybe?

Feel free :)

mcdurdin:
> May I assume there's a guarantee that between [CAPS K_FOO], [NCAPS K_FOO], and [RALT K_FOO], the latter will always take precedence if pressing RALT whether or not CAPS is active at that moment?

Yes. `RALT K_FOO` does not specify the Caps Lock state so Caps Lock is ignored. Because `RALT` is a modifier key, it requires an explicit match, so the first two rules would not match.

rimas-kudelis:
Then I think this is probably ready for merge, although I ended up not removing exact duplicates of QWERTY definitions (I can still do that though).
I'll probably import a few more Lithuanian layouts, but I don't have a timeframe for that, and I can do that in a separate PR.

By the way, I think there is one tiny bug in the importer: it imported RALT states as NCAPS RALT, and I had to remove NCAPS manually.

mcdurdin:
> By the way, I think there is one tiny bug in the importer: it imported RALT states as NCAPS RALT, and I had to remove NCAPS manually.

Okay, would you mind opening an issue on this at https://github.com/keymanapp/keyman/issues/new/choose with full repro details?

@LornaSIL can I hand over the keyboard checks to you?

LornaSIL:
I will try to get to it this week. Unfortunately I'm still parttime for awhile.

LornaSIL:
@mcdurdin 
>Given this is not a pure clone of the Windows keyboard, you are probably right. In which case, there is an argument for giving your keyboard a name to distinguish it from other Lithuanian keyboards that may be in the repo (or added at a future time). We would then deprecate the existing kbdlt1 and lithuanian keyboards in favour of this one.

* Can you verify that you are in agreement that this PR will DEPRECATE ```legacy\kbd\kbdlt1``` and delete ```legacy\l\lithuanian``` and the name is simply "Lithuanian".

* Secondly, I will take Lithuanian (kbdlt1) off my list of Windows keyboards to convert to Keyman?

mcdurdin:
> * Can you verify that you are in agreement that this PR will DEPRECATE `legacy\kbd\kbdlt1` and delete `legacy\l\lithuanian` and the name is simply "Lithuanian".

Yes. The keyboard id is `lithuanian`, and AFAICT it is a superset of the legacy `lithuanian` keyboard and the changes are not going to cause trouble. But see below for one possible wrinkle.

> * Secondly, I will take Lithuanian (kbdlt1) off my list of Windows keyboards to convert to Keyman?

I'm not sure that there isn't a case for creating a `basic_kbdlt1` that deprecates `kbdlt1` as well (note: for deprecation pathway purposes, we need to mark only one of the two keyboards as deprecating `kbdlt1`, and I don't think it matters much which one). Every other layout we are converting over has a 100% clone of the Windows keyboard, so without this, we have divergence just for Lithuanian, which is inconsistent. It may not be a big deal, but worth thinking about. What do you think?

LornaSIL:
FYI: I've had to modify a few of the Windows keyboards already that use RALT SHIFT O, K, N (which, as you know, are Keyman hotkeys) to move to keys that aren't hotkeys. So, some of the keyboards we've moved to Keyman are not exact matches. We (David and I) did document the changes in the history file. 
Anyway, I'll keep Lithuanian on my list and go ahead and review this keyboard.

mcdurdin:
> FYI: I've had to modify a few of the Windows keyboards already that use RALT SHIFT O, K, N (which, as you know, are Keyman hotkeys) to move to keys that aren't hotkeys. So, some of the keyboards we've moved to Keyman are not exact matches. We (David and I) did document the changes in the history file.

Oh, I wasn't aware of that. Given that RAlt+Shift+O, RAlt+Shift+K and RAlt+Shift+N are default hotkeys that can be changed by the user, I think I'd prefer to match the Windows layout and include a note in the readme/welcome to the effect that the user needs to change the default hotkeys in Keyman in order to access those specific key combinations.

rimas-kudelis:
> > * Secondly, I will take Lithuanian (kbdlt1) off my list of Windows keyboards to convert to Keyman?
> 
> I'm not sure that there isn't a case for creating a `basic_kbdlt1` that deprecates `kbdlt1` as well (note: for deprecation pathway purposes, we need to mark only one of the two keyboards as deprecating `kbdlt1`, and I don't think it matters much which one). Every other layout we are converting over has a 100% clone of the Windows keyboard, so without this, we have divergence just for Lithuanian, which is inconsistent. It may not be a big deal, but worth thinking about. What do you think?

Can this keyboard deprecate `basic_kbdlt1` in advance? I don't believe it makes any sense to ship both.

LornaSIL:
@rimas-kudelis This is looking pretty good. Here are some requested changes:
* I would like to see at least a readme.htm and possibly a welcome.htm file added (put in source) and then add them to the .kps file as well. If you don't include screenshots of the layout then you should mention that users should use the OSK to view the keyboard layout.
* lithuanian.keyboard_info and README.md - change Lituanian to Lithuanian in the description field
* lithuanian.kps - please change copyright from 2007-2018 to 2019 or else make sure the copyright in other places is 2007-2019, but I think it is best to just be 2019.

I haven't tested the keyboard yet. If the table and phone layouts are exactly the same I'd like you to delete one of them so we only have to make changes to one. If they are different then it's good to include both, but a quick glance looks like they are intended to be the same. If you keep the phone layout, it looks like the K_PERIOD key on the phone layout, default and shift layers, will need a fullstop added to be displayed.

I will begin testing the keyboard, but I wanted to give you this feedback to start on.

LornaSIL:
I tested the keyboard on the desktop and it looks good. The OSK is fine.
For the table layout there are a few issues with the numeric layer:
* None of the digits are being output. You will need to change the code from ```K_1``` to ```U_0031``` and do that for all the digits using the Unicode codepoint.
* In row 2 of the numeric layer only the backslash is outputting the correct character - so all of those need the code changed to the ```U_xxxx`` kind of format where xxxx is the Unicode value
* In row 3 the (, ), +, -, * are all outputting the wrong character (based on what is on the keycap) - so all of those need the code changed to the ```U_xxxx`` kind of format where xxxx is the Unicode value

The mobile layout doesn't really have the same character repertoire as the desktop keyboard. Is this intentional? There are a lot of characters on the mobile layout that aren't in the Desktop layout.

mcdurdin:
> Can this keyboard deprecate `basic_kbdlt1` in advance? I don't believe it makes any sense to ship both.

No, I don't want this keyboard to deprecate `basic_kbdlt1`. We want a full set of the system-supplied keyboards as a base, and this keyboard does diverge from the Windows one. I'm happy to include a cross reference between the two.

Similarly, I would prefer that this keyboard doesn't just use the name 'lithuanian', but I'm willing to let that one pass. It makes it difficult for others to create keyboard layouts for Lithuanian that are differentiated.

LornaSIL:
@mcdurdin I have created basic_kbdlt1 and deprecated ```lithuanian``` and ```kbdlt1``` which is the process we have discussed on handling bringing in Win10 keyboards.

What are the next steps for this keyboard? 
My ideas:
- rename folder and files to ```lithuanian_ext```
- Rename keyboard ```Lithuanian Extended```
- Make this keyboard "related" to "basic_kbdlt1" (but not deprecated)
- Remove the DEPRECATED.md files from ```legacy/kbd/kbdlt1/DEPRECATED.md```

What do you think?

rimas-kudelis:
I see the new `basic_kbdlt1` layout is already available on keymanweb.com. Congrats!

Regarding this layout, I'm actually not so sure there is a reason to proceed with it further.

Firstly, `basic_kbdlt1` is already an improvement over the legacy Lithuanian layout which was the only one available till recently.

Secondly, I don't think there is a single person in the universe whose habits would be disturbed by any of the four minor changes this layouts presents on top of kbdlt1. Furthermore, I don't think that having two near-identical Lithuanian layouts just for the sake of having them would help anyone or make your product any better. Actually, it seems that even though `basic_kbdlt1` deprecates `kbdlt1`, the latter is still rendered as an option available in KMW (see the image below), so my layout would actually become a third near-identical option. And I'm totally not a fan of this idea.

paveikslas

Marc mentioned that an exception of Lithuanian being the only layout not updated from Windows would be something akin to annoyance. I agree with that, but I don't like that the option of having it ported, but deprecating it in favor of its non-destructively improved version is off the table.

And lastly, Lorna found some valid issues with my submission, so it needs more polishing, but I can't promise to do that soon. You could leave this PR open and I'll eventually get to fixing the issues, or you can close it and I'll probably eventually make a new one with the fixes (or reopen this one, if it will be possible), but again, to me, the importance of this has fallen.

Also, just in case, on Windows I call this layout "Lithuanian (Enhanced)", not Extended. Then again, on Linux, it's just called Lithuanian, and the "non-enhanced" option is simply not available.

LornaSIL:
Thanks for your feedback. I would be happy with Lithuanian (Enhanced). 

I am unhappy if we are still seeing the deprecated keyboards being offered on web and mobile. I thought that had been dealt with. Is there an issue for this @mcdurdin ?

I think since you'll need to rename all your folder/filenames when you do get around to finishing this keyboard, you might as well close this PR and start a new one when you are ready. You can delete the branch name when you close it if you wish. It helps keep things tidy. However, make sure all your files are still available on your computer!

BTW, if there are other Windows keyboards that are a priority for you, please feel free to let me know which ones. I'm not able to spend a lot of time on porting them over, but I like to work on ones that I know will benefit someone.

mcdurdin:
> I am unhappy if we are still seeing the deprecated keyboards being offered on web and mobile. I thought that had been dealt with. Is there an issue for this @mcdurdin ?

There are some issues with deprecated keyboards that we are still dealing with in the keyboard search.  They are on my agenda, along with many many other things:

* https://github.com/keymanapp/api.keyman.com/issues/5
* https://github.com/keymanapp/api.keyman.com/issues/7
* https://github.com/keymanapp/keyman/issues/1713
* https://github.com/keymanapp/keyman.com/issues/8
* https://github.com/keymanapp/help.keyman.com/issues/19

LornaSIL:
@rimas-kudelis I'm going to close this PR, but you can still access it if you need it later.

PR #656, opened 2019-03-26
[nlci_malayalam_winscript] NLCI Malayalam WinScript 2.0.0

binilasanki:
Initial submission of the NLCI Malayalam WinScript Keyboard layout redesigned using Keyman 10.
Mozhi Malayalam is also got included in this PR. Although there are no changes in the contents of the file, the modified date has changed. Hence it is in this PR. Please ignore this file from the `release/m/mozhi_malayalam` folder.


keyman-server:

DavidLRowe:
This PR has a change to the file release/m/mozhi_malayalam/source/mozhi_malayalam.kmn which should not be included. Please replace with the original file.

This PR seems to delete the file release/sil/sil_ipa/source/sil_ipa.ico which should not be deleted. Please replace the original file.

In the README.md please add Linux, since that platform is now supported from the same package that handles Windows and Mac.

The project should be set to treat compiler warnings as errors. In Keyman Developer, Project menu, Project Settings menu item, tick the box "Treat compiler hints and warnings as errors". This is the standard for the Keyman build system.

The .kps file needs an entry for keyboard "ml" (to match .keyboard_info file).

In Developer, with the Keyboard package source (.kps) file open, in the "Details" section, remove the existing "2.0.0" in the "Version" field, then tick the "Package version follows keyboard version" box. This will make it easier to keep the version number of the package co-ordinated with the keyboard version number.

There are several places where copyright notices appear. They should all have the same dates and name. Perhaps standardise on `(c) 2002-2019 NLCI, Bangalore` ?

keyman-server:

keyman-server:

DavidLRowe:
These two files don't belong in this pull request:
- release/m/mozhi_malayalam/source/mozhi_malayalam.kmn
- release/sil/sil_ipa/source/sil_ipa.ico

The .keyboard_info file should just have one entry for "ml" (not the second one for "ml-Mala". Similarly, in the .kps file, there should be no entry for "ml-Mala", just the one for "ml". (Further details: The "Mlym" script code for Malayalam shouldn't be present according to the "Suppress-Script" entry for "ml" in the list at http://www.iana.org/assignments/language-subtag-registry/language-subtag-registry)

.kmn file
- &Copyright string is '© NLCI 2000-2019 Bangalore'. Please change to '© 2000-2019 NLCI Bangalore'.
- &Targets string is 'any', but there's no .keyman-touch-layout file. Do you plan to add support for mobile devices in this PR? If not, change &Targets to 'windows macosx linux web'.

DavidLRowe:
If I knew the script, some of my questions might be obvious, but I found the following items in the welcome.htm file to be confusing since they include a Malayalam character rather than the keystrokes used to create that character:
- "nta"
- "using ^"
- "vowel sign elongation"
- "legacy nta"

In the example under "Chandrakala", I think the keys should be `kXl` rather than `kXL` (that is, lower case `l` rather than upper case `L`)

Under "Backtick for change into English", "\`f" doesn't seem to produce "f". Perhaps this is a change needed in the .kmn file.

I think the nlci_malayalam_winscript.php file uses similar wording so it should be updated as well.

LornaSIL:
I would prefer the keyboard names to have "(NLCI)" at the end of the name and not the beginning. If you did that, this one would be "Malayalam WinScript (NLCI)" (the folder/file names are correct, they wouldn't need changing, just the name).

That way, if the keyboards are sorted alphabetically, like here: https://help.keyman.com/keyboard/ it would be sorted with other Malayalam keyboards and not with "NLCI". Search for "SIL" and "Rachitrali" to compare how they get sorted.

binilasanki:
> If I knew the script, some of my questions might be obvious, but I found the following items in the welcome.htm file to be confusing since they include a Malayalam character rather than the keystrokes used to create that character:
> 
> * "nta"
> * "using ^"
> * "vowel sign elongation"
> * "legacy nta"
> 
> In the example under "Chandrakala", I think the keys should be `kXl` rather than `kXL` (that is, lower case `l` rather than upper case `L`)
> 
> Under "Backtick for change into English", "`f" doesn't seem to produce "f". Perhaps this is a change needed in the .kmn file.
> 
> I think the nlci_malayalam_winscript.php file uses similar wording so it should be updated as well.

binilasanki:
> If I knew the script, some of my questions might be obvious, but I found the following items in the welcome.htm file to be confusing since they include a Malayalam character rather than the keystrokes used to create that character:
> 
> * "nta"
> * "using ^"
> * "vowel sign elongation"
> * "legacy nta"
> 
> In the example under "Chandrakala", I think the keys should be `kXl` rather than `kXL` (that is, lower case `l` rather than upper case `L`)
> 
> Under "Backtick for change into English", "`f" doesn't seem to produce "f". Perhaps this is a change needed in the .kmn file.
> 
> I think the nlci_malayalam_winscript.php file uses similar wording so it should be updated as well.

Dear brother.

"nta"
"using ^"
"vowel sign elongation"
"legacy nta"

This are the special characters we added. For "f " there is no characters. So i gave null

PR #661, opened 2019-03-30
gff_blin & gff_geez package updates.

dyacob:
This update does _not_ modify the keyboard files, rather its a change to the packaging.  A new side image is introduced for the Blin KPS and the Ge'ez KPS file has the display and keyboard font set.


keyman-server:

keyman-server:

LornaSIL:
Unfortunately packages will not be rebuilt if we don't bump the version (which, as you know, requires a lot of file changes!). Since this feels fairly straightforward I'm willing for you to update both keyboards in this PR.

keyman-server:

dyacob:
Thanks @LornaSIL , in the update just made I've done what I think is the minimum to bump the version.  This impacted the .kmn file, README.md , and HISTORY.md .  During test builds on my end the .kpj and kvks files also updated.

LornaSIL:
@dyacob appreciate it if you will delete the branch when it finishes building.

PR #668, opened 2019-04-12
[yiddish_pasekh] Move keyboard from legacy to release

LornaSIL:
* Added DEPRECATED.md file to legacy\y\yidish2k
* Used kmdecomp to get a source file for v1.7
* Enhanced documentation per feedback from [community site](https://community.software.sil.org/t/how-to-enter-a-yud-khirek-yud-in-the-pasekh-yiddish-keyboard/2024/6)
* Added language tags, etc where needed.
* In general updated things related to current version of keyman and current structure system

**Note**: On OSK keys for i and u do not exactly match the output of that single keystroke. However, when following another character that is what is output and this is the displayed layout that the author had. Since they more commonly follow a consonant I thought it was okay.


DavidLRowe:
There's a Right Alt image missing that is referenced by welcome.htm (or else, if there's no Right Alt layer, welcome.htm should be changed to remove the heading). [Later] Based on the help file, it seems there is no Right Alt layer, so this heading should come out of welcome.htm.

Other than that, everything LGTM.

PR #669, opened 2019-04-15
[basic_us] Import Windows 10 kbd to Keyman

LornaSIL:
Did an import with Keyman 11. I also changed the mobile layout slightly (from the import) to match the old mobile layout.
I associated a lot of languages with this keyboard based on Issue #611


DavidLRowe:
`K_oE2` is in .kmn code, but not on OSK. It's also not included in touch layout, but maybe that's a design decision.

For touch layout, default layer, `K_LCONTROL` key (fifth row, 1st key) doesn't seem to have any action. 

For touch layout, shift layer, `K_SHIFT` doesn't take you back to default layer, instead you need to use the unlabeled K_LCONTROL key.

LornaSIL:
I did change the OSK to support the 102nd key. It is already supported on another key, so I didn't feel the mobile layout need it duplicated.
I made K_LCONTROL a spacer since I don't want the key.

mcdurdin:
Note: the language tags should not have `-Latn` unless the language is normally written in another script or in multiple scripts. I feel like I need another round of consolidation and clarification on how we determine BCP 47 tags because there are conflicting requirements (e.g. Windows is more persnickety than other platforms) that I don't think I fully understand yet.

LornaSIL:
@mcdurdin do you want me to revise this keyboard and bump the version or are you giving feedback for future keyboards?

Also, are there any guidelines you can give me on when to include the script and when not? We seem to have multiple issues with Windows not understanding the lang tags and so I thought the more information the better.

I suppose for the 2 letter ISO codes we can presume the system will understand the default script? But, Uzbek is written with 3 different scripts and it would take me an extra few minutes to research to know which is default. Most of the others would have been fairly clear that Latn is the default.

mcdurdin:
@LornaSIL, yes, it might be good to revise the keyboard. The [suppress-script field](https://tools.ietf.org/html/bcp47#section-3.1.9) in the [BCP 47 registry](https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry) denotes which languages we should not assign a script to, unless the keyboard diverges from that script.

Keyman Developer does let you override the suppression of the script but by default it should follow the convention (if you find instances in which it doesn't, let me know).

It's quite easy to crash Windows Explorer if you assign a BCP 47 code to a keyboard layout that it doesn't fully grok. One day I need to investigate this further and generate a Windows Feedback report.

PR #672, opened 2019-04-18
[meta] Testing displayUnderlying flag which was breaking CI for unknown reason

mcdurdin:
This pull request is just to test the issue reported in https://github.com/keymanapp/keyboards/pull/617#issuecomment-483876195 regarding `displayUnderlying` breaking CI build.
Do not merge; I will be investigating further.


mcdurdin:
I believe this happened because the coptic_qwerty keyboard had an older master branch it was based off. See #673 for a long term proposed fix.

PR #678, opened 2019-04-22
[basic_kbdurdu] Import Win10 kbd to github and deprecate old ones

LornaSIL:


DavidLRowe:
Needs RTL indication (I think).

On the shift layer: Z, X, V, M produce invisible characters (zero width joiner, etc). Should this be in the documentation somewhere?

PR #680, opened 2019-04-24
[rac_oadke] New keyboard on behalf of author

LornaSIL:


DavidLRowe:
Desktop, default layer `n` has U+06BA on OSK, but key produces U+0646.

Desktop, shift layer, `~` has U+007C on OSK, but key produces U+0687.

Touch layout: from shift layer, both uparrow and `123` keys take you back to the default layer. From ctrl-alt layer, both uparrow and `123` keys take you back to the default layer. If that's the intent, then it's OK, but some might expect to be able to go from shift layer to ctrl-alt layer directly.

LornaSIL:
> Desktop, default layer n has U+06BA on OSK, but key produces U+0646.

> Desktop, shift layer, ~ has U+007C on OSK, but key produces U+0687.

I don't understand either of these comments. They look okay to me. Which keys are you looking at?

DavidLRowe:
I'm mistaken about the 'n' key on the OSK.

rachitralli:
@LornaSIL @DavidLRowe thank you very much

PR #683, opened 2019-04-24
[basic_kbdlt1] Lithuanian kbd imported from Win10

LornaSIL:


DavidLRowe:
Touch layout, numeric layer (as tested in Chrome on iphone using Developer's "Test keyboard on web": 
- `+` key produces hyphen
- blank keys in row 2 (to right of `*`), row 3, and row 4 produce `op[]` etc according to underlying key

DavidLRowe:
Really minor comment: OSK has Ctrl layer with just SPACE.

PR #690, opened 2019-04-29
[list_mismatch_1] Fixing BCP47 codes mostly in keyboard_info

LornaSIL:
The only one where the version needed bumping was the bj keyboard


DavidLRowe:
In bj_cree_east.kmn, `&TARGETS` was changed to 'any', but there is no touch layout.  Will that be OK? (That is, will it use the keyboard layout for mobile devices?)

LornaSIL:
The bj_cree_east OSK is strange in that it only reflects the underlying Latn layout, not the Cans symbols, because people want to see the Latn and not the Cans on their kbd. According to the author.

PR #698, opened 2019-05-05
[sil_el_ethiopian_latin] version 1.4.0

rik-shaw:
The main update is to add Nuer compatibility. Hopefully everything is in order but please let me know if any adjustments are needed!


keyman-server:

keyman-server:

LornaSIL:
Several of the language tags in the .kps file are saying "Ethi" instead of "Latn". Can you adjust those to use "Latn" instead?

LornaSIL:
>Several of the language tags in the .kps file are saying "Ethi" instead of "Latn". Can you adjust those to use "Latn" instead?

My mistake. That is not the case.

However, we would like the .kps and the keyboard_info files to exactly match which languages are being supported.

Right now they do not match :
kps has:
* eth (so either remove it or also add it to keyboard_info)

keyboard_info has:
 * arv, dim, doz, dtn, gdi, gza, hdy, hoz  (so either remove them or also add them to .kps)

Also, for Somali "so" we are supposed to suppress the script, so in kps it should be 
```
        <Language ID="so">Somali</Language>
```
and in keyboard_info it should be:
```
        "so",
```

DavidLRowe:
@rik-shaw According to Marc, you'll need to update your `sil-el-1-4-0` branch to get a copy of the latest compiler from the "upstream" repository: `https://github.com/keymanapp/keyboards.git`

keyman-server:

rik-shaw:
@DavidLRowe I hope I did it correctly, please let me know if there is any issue.
@LornaSIL I sync'd the language list. There may be additional languages added in the future as I confirm which languages have non-english letters for their latin scripts.

DavidLRowe:
The .keyboard_info file is not a valid JSON file. You need to remove the comma after `"zay-Latn"`. (I use https://jsonformatter.curiousconcept.com/ to check .keyboard_info and .keyman-touch-layout files.)

DavidLRowe:
In the .kmn file, the &ethnologue store no longer used (and likely will be removed in future version). You could comment out the line (and leave it for reference) or else just delete it.

Normally the `Image file` specified in the package (.kps) is a 140x250 pixel image that is displayed (in place of the default Keyman one) when the user initiates the install process. If you want to use the 48x48 image, you could pad it with white to get to the 140x250 size.

Optionally, you could make keyboard available for web use:
- In .kmn file, &targets would include `web`
- In .kps file, Files would include `..\build\sil_el_ethiopian_latin.js`

DavidLRowe:
@rik-shaw Do you have the information you need to continue? No pressure! Just want to make sure I'm not blocking your progress.

rik-shaw:
@DavidLRowe sorry for the long delay, I was tied up with other things and wasn't able to follow up on this. I have made a basic "sideimage" and applied the other items, including building for web. I also merged upstream/master into the branch so hopefully it is looking decent. Let me know :-)

keyman-server:

rik-shaw:
@DavidLRowe uh oh now I see the merge of upstream/master build is failing. Let me know if there is anything needed to clean up.

rik-shaw:
@LornaSIL @DavidLRowe Thanks again for the help and for the merge

PR #700, opened 2019-05-07
[multi_pak_phonetic] Move keyboard to release with permission of author

LornaSIL:


DavidLRowe:
There's a mismatch on the BCP 47 codes:

.kps has: 
```
"pbt-Arab"
"pnb-Arab"
```
.keyboard_info has:
```
"pnb-Arab",
"pst-Arab",
```

That is, one has `pbt` and the other `pst`.

PR #702, opened 2019-05-07
[urdu_phonetic] Move keyboard from legacy to release with permission …

LornaSIL:
…of author


DavidLRowe:
In the .keyboard_info file, the "description" uses <b> etc. Are HTML tags supported? I tried a quick search of the .keyboard_info files in the keyboards repository, and didn't find any that had "<".

Does the &version need to set to 11 (because of the "displayUnderlying" in the touch layout)? I don't really know. Developer compiles it with &version 10.0, so perhaps not.

DavidLRowe:
In the .keyboard_info file, the "description" uses `<b>` etc. Are HTML tags supported? I tried a quick search of the .keyboard_info files in the keyboards repository, and didn't find any that had `<`.

Does the &version need to set to 11 (because of the "displayUnderlying" in the touch layout)? I don't really know. Developer compiles it with &version 10.0, so perhaps not.

PR #703, opened 2019-05-07
[meta] Update to kmcomp 11.0.1356.0

mcdurdin:
* Bug fix: Compiler could occasionally fail with undefined results when using Unicode store names (keymanapp/keyman#1736)


mcdurdin:
> The only Files changed are binary. Is there a standard procedure for reviewing this type of PR?

Pull the branch, and launch the executable (if you trust me!). Or you could compare it to the binary files from the download server https://downloads.keyman.com/developer/stable/ ...

PR #709, opened 2019-05-09
[sil_khmer] Move keyboard to github from SSO distribution

LornaSIL:


DavidLRowe:
Would it be better to put "km" first in the list (both in the .keyboard_info file and the .kps file)? As it is now, the keyboard gets installed as `brb-Latn-VN` on Windows, because (I'm assuming) it uses the first code in the list.

DavidLRowe:
From welcome.htm, 
```
"The symbols of the moon calendar (with 2 symbols on top of each other) 
can also be typed by typing the numbers plus type ">" two times. 
E.g. typing "0", "1", ">", and ">" will give the same symbol as Ctrl-Alt-`.
```
But this doesn't seem to work. Possible replacements:
```
E.g. typing "0", "8", ">", and ">" will give the same symbol as Ctrl-Alt-Shift-`.
E.g. typing "1", "0", ">", and ">" will give the same symbol as Ctrl-Alt-Shift-0.
```

Other than that change to the welcome.htm (and the corresponding one in sil_khmer.php), it looks good (though I'm hardly an expert on Khmer!).

PR #710, opened 2019-05-09
[fv_kps_fixes] half of the fv files that need kps files

LornaSIL:


DavidLRowe:
fv_australian.kmn needs change to copyright date.

Curious why fv_goyogohono.kpj  has a reference to itself. I don't recall seeing that before. Probably doesn't make a difference.

Issue #694, opened 2019-05-02
Modify BCP47 codes for fv and inuktitut keyboards


LornaSIL:
Error with fv_australian.kps
Error with fv_blackfoot.kps
Error with fv_dexwlesucid.kps
Error with fv_goyogohono.kps
Error with fv_hailzaqvla.kps
Error with fv_klahoose.kps
Error with fv_naskapi.kps
Error with fv_nexwslayemucen.kps
Error with fv_ojibwa.kps
Error with fv_onayotaaka.kps
Error with fv_onodagega_onondagega.kps
Error with fv_onodowaga.kps
Error with fv_severn_ojibwa.kps
Error with fv_skaru_re.kps
Error with fv_skwxwumesh_snichim.kps
Error with fv_tlingit.kps

Error with inuktitut_latin.kps
Error with inuktitut_naqittaut.kps
Error with inuktitut_pirurvik.kps 

For many (most?) of these there is no .kps file. I don't understand how the mobile layout knows the language. Should we add a .kps file for all of these?

LornaSIL:
@mcdurdin many of these keyboards have no .kps file (I think they are primarily mobile). However, I'm not seeing them in my Andoid Keyman app. Should I be adding a .kps file to these keyboards so that a language will be associated, or how is the language being associated without a .kps file?

mcdurdin:
> @mcdurdin many of these keyboards have no .kps file (I think they are primarily mobile). However, I'm not seeing them in my Andoid Keyman app. Should I be adding a .kps file to these keyboards so that a language will be associated, or how is the language being associated without a .kps file?

Yes, these are legacy keyboards, previously web/mobile only, so in the old model they were distributed as .js files. Now that we are moving to 100% .kmp distribution, we should **definitely be adding .kps files** for the keyboards and moving them to version 10 format, which will then mean removing language ids from the .kmn.

LornaSIL:
#710 
#711

LornaSIL:
Closed by above issues.

PR #712, opened 2019-05-09
[inuktitut_kps_fixes] adding kps to keyboard

LornaSIL:
These 3 keyboards are still under the Apache license. @mcdurdin has written asking for permission to change the license, but I went ahead and added the .kps files so that the keyboards will be available on mobile.


DavidLRowe:
inuktitut_latin.keyboard_info, inuktitut_naqittaut.keyboard_info and inuktitut_pirurvik.keyboard_info have `"license": "mit",`. Is that a problem?

LornaSIL:
I guess we should change it? Or wait until we hear back from the author. I'm not sure what to do.

DavidLRowe:
It seems as though we need to change it to apache until we get the author's approval.

DavidLRowe:
along with #710 and #711 this fixes #694

Issue #695, opened 2019-05-02
Modify BCP47 codes for fv keyboards: list mismatch


LornaSIL:
List mismatch: fv_anishinaabemowin
List mismatch: fv_cree_latin
List mismatch: fv_dine_bizaad
List mismatch: fv_hlgaagilda_xaayda_kil
List mismatch: fv_maori
List mismatch: fv_shashishalhem
List mismatch: fv_statimcets
List mismatch: fv_xaislakala

LornaSIL:
#713

DavidLRowe:
Fixed by #713

PR #713, opened 2019-05-09
[fv_mismatch] fixed bcp47 tags in keyboard_info no ver bump

LornaSIL:


DavidLRowe:
In fv_anishinaabemowin.keyboard_info, `"oj-Can"` should be `"oj-Cans"`

DavidLRowe:
fixes #695

PR #714, opened 2019-05-11
[tohono_oodham] Add breve vowels to Tohono O’odham keyboard

alevy03:
After meeting with native language speakers, we realize that we did not include all vowels with breve marks. This update corrects that oversight, and updates the naming of key codes a bit.


keyman-server:

keyman-server:

alevy03:
@LornaSIL, I made the changes you suggested. It builds fine on my machine, but it looks like there may be a problem in the build on the CI server. I don't have visibility into that, so I'm not sure what the problem is. Please advise.

PR #716, opened 2019-05-13
Third submission Soqotri keyboard

dhenchoz:
I don't know how to build the readme.htm and welcome.htm files.
Could you please send me that information? Thanks


keyman-server:

DavidLRowe:
@dhenchoz I think that more detail is needed in the documentation. 

Three files (plus any supporting files they reference) are needed:
- The readme.htm is a (usually small) HTML file that is displayed when the user begins installation of the desktop keyboard. The goal is to give the user enough feedback about the keyboard to decide whether or not to install this keyboard.
- The welcome.htm file generally has more description about using the keyboard. Often it references graphic images of the different keyboard states (which can be generated from Keyman Developer). The goal is that all the functionality of the keyboard is documented in such a way that the user can figure out how to use the keyboard.
- The online help file (soqotri_arabic.php in this case) often contains the same text as the welcome.htm file, though it can reference the keyboard layouts directly. This is parallel to the welcome.htm file, but used in the online version of the keyboard.

I would suggest looking at https://github.com/keymanapp/keyboards/tree/master/release/sil/sil_arabic_phonetic for an example of these three files.
- The readme.htm file is in the source/ folder
- The welcome.htm file and the .png images it references are in the source/welcome/ folder
- The sil_arabic_phonetic.php file is in the source/help/ folder

But if this is not sufficient information, please let us know.

DavidLRowe:
@dhenchoz Was the information above sufficient? I want to be sure that we're not stopping you from making progress.

dhenchoz:
Dear David,
Thanks for asking. You may find in the attached files what I have until now.
I'm not sure if this is enough or not and whether this should be in English
or Soqotri.
Do you think we could publish the keyboard with those files and update them
later if needed?
Thanks,

DavidLRowe:
@dhenchoz I'm not seeing the attachments you mention.

Yes, we could publish the keyboard now with whatever you have, and add / modify the files later. (An update will require a version number change, and that means modifying several files.) I suggest that you include the files in the pull request.

Some keyboards have a help file in English only, some in English plus another language. In order for Keyman team personnel to be in a position to help users, the help files (welcome.htm and soqotri.php) need to have basic information in English (or perhaps another language of wider communication known by the team). Using an image of the keyboard states is helpful in this regard.

keyman-server:

DavidLRowe:
I'm afraid that some files you added ended up in the root of the repository, rather than in the  release/s/soqotri_arabic/ folder (or one of its subfolders). Specifically: 
- readme.htm 
- welcome.htm 
- soqotriU_.png (a copy of the file in release/s/soqotri_arabic/source/welcome/)
- soqotriU_S.png (a copy of the file in release/s/soqotri_arabic/source/welcome/) 
- soqotri_arabic.kps (an older copy of which exists in release/s/soqotri_arabic/source/)

You'll need to remove these files from the root and put them in the appropriate places.

Your added readme.htm and welcome.htm seem fine (although, as mentioned above, they are not currently in the right location). The soqotriU_.png and soqotriU_S.png images don't seem to reflect the actual keyboard. (For example, the key to the left of the numeral "1" generates a different character than in that image.) You will need to recreate these files to reflect changes you have made to the keyboard.

In soqotri_arabic.keyboard_info and soqotri_arabic.kps files, change language code to sqt-Arab (not sqt-Arab-YE or sqt-YE).

The following change is optional, but would make things clearer for anyone looking at the soqotri_arabic.kmn file in the future.Since you have provided both .kmn and .keyboard-touch-layout files, Keyman can support any platform, so you could change the `store(&TARGETS) 'windows macosx linux mobile'` to `store(&TARGETS) 'any'` (either by editing the .kmn file directly, or using Keyman Developer to tick the "any" box and untick all the others).

So I think you have all the necessary files now. Once you update a few files and get them in the right places, this keyboard should be ready for approval.

dhenchoz:
Dear David,
Thanks for your watchful eye. I have updated the keyboard's images.
It seems to me I didn't have the choice to upload files to a specific
folder.
I'll close the current submission and open a new one while taking your
recommendations into account.
Blessings,
Dominique

Le ven. 21 juin 2019 à 20:39, DavidLRowe <notifications@github.com> a
écrit :

Issue #696, opened 2019-05-02
Modify BCP47 list mismatch


LornaSIL:
List mismatch: galaxie_greek_hebrew_mnemonic
List mismatch: galaxie_greek_hebrew_positional
List mismatch: khmer_angkor
List mismatch: lao_2008_basic
List mismatch: lao_2008_rapid
List mismatch: rohingya_arab
List mismatch: sil_eastern_congo
List mismatch: sil_ethiopic
List mismatch: sil_nigeria_dot 

List mismatch: sil_ipa
- I think this is because we have to define the tag differently in .kps and keyboard_info

LornaSIL:
These are all fixed except for sil_ipa which at this point is deliberately different. I think it's okay to close this.

PR #685, opened 2019-04-26
[bcp_errors_sil] Update BCP47 codes for SIL kbds

LornaSIL:
I made changes to these keyboards:
Error with sil_arabic_phonetic.kps
Error with sil_areare.kps
Error with sil_el_ethiopian_latin.kps
Error with sil_ethiopic.kps
Error with sil_ethiopic_power_g.kps

PR #718, opened 2019-05-13
[kps_fix3] fixes to BCP47 mismatches

LornaSIL:
* Fixed the BCP 47 codes which were mostly problems in kps files so the version had to be bumped.
* I didn't have to bump sil_nigeria_dot as the problem was in the keyboard_info file.
* The Greek/Hebrew packages are different because the "packages" have the version bumped, but the keyboards themselves do not have the version bumped since the keyboards had the right language codes.


LornaSIL:
Fixes #696 
I think I fixed  sil_ethiopic in #685

PR #720, opened 2019-05-14
[rac_aer] fix ctrl-a key on shift layer to be special

### LINKED ISSUE #401: [rac_aer] An action button on the SHIFT layer doesn't work as expected
In touch layout, shift layer, row 5: the "ctrl-a" button takes you to default layer, despite the
"nextlayer": "ctrl-alt"

This happens in `rac_aer` keyboard PR #358.

LornaSIL:
Fixes #401


DavidLRowe:
In the .kmn file, `store(&WINDOWSLANGUAGES) 'x0401'` is commented out, but then that line is added back in a few lines later.

On shift layer, "ctrl-a" key takes you to the ctrl-a layer, and that key is then darkened to indicate it is active (so that's good), but that key also changes from displaying the text "ctrl-a" to displaying an up-arrow. Could that be confusing? (I'm willing to approve this as is, but wondered if another option was desired.)

LornaSIL:
I think I'm going to leave the up arrow since that's what the author used before.

PR #723, opened 2019-05-16
Revert "[nlci_malayalam_winscript] "

mcdurdin:
Reverts keymanapp/keyboards#707

TODO:
- [ ] restore the nlci_malayalam_winscript files to this branch, so that this PR recovers the files that should not have been deleted.


keyman-server:

Issue #488, opened 2018-10-17
[lahu] User request for touch layout


darcywong00:
@sdysart has a user requesting a touch keyboard for Lahu.

We could update an existing [https://keyman.com/keyboards/lahu](Lahu) keyboard  and add the touch layout.

### TODO ###
- [x] Get permission from the original keyboard author

LornaSIL:
@sdysart we've added a touch layout with permission of the author. As soon as it builds it should be available here: https://keyman.com/keyboards/lahu

PR #724, opened 2019-05-17
[lahu] Move Lahu from legacy to release and deleted legacy one

LornaSIL:
* Changed license from freeware to mit with permission of author
* Added mobile layout per PR #488


DavidLRowe:
fixes #488

PR #727, opened 2019-05-19
[itrans_gujarati] New keyboard submission

Shreeshrii:


keyman-server:

LornaSIL:
* The .kps file lists "hi" as the language instead of "gu"
* Also, please add the readme.htm in the "Details" section of the .kps file
* I would like to see the following keys on the OSK: [ ] ; , . ! @ ( ) { } | : " ?
     * I would even like the the ```_``` key to have ZWNJ on the keycap if at all possible. I did this for an Arabic script keyboard and used smallcap characters so they would display better. You can copy and paste the following into the OSK keycap for _ if you want: ᴢᴡɴᴊ
     * Then you would need to regenerate the graphics.
* In the welcome.htm file for the second to last line, I think the key should be "`" rather than ’. If that's the case, you should make the change in the help file as well.

I think that's it. It looks good and will be nice to have this keyboard in the repo.

keyman-server:

keyman-server:

LornaSIL:
@Shreeshrii can you delete your branch where it offers you to do so just above or below this message? Thanks.

LornaSIL:
@Shreeshrii can you delete your branch where it offers you to do so just above or below this message? Thanks.

Shreeshrii:
Thanks!

Shreeshrii:
Thanks!

PR #729, opened 2019-05-20
[sil_limbu_typewriter] Move keyboard from MSKLC to Keyman

LornaSIL:


DavidLRowe:
Copyright (start) date in README.md doesn't match that in LICENSE.md, HISTORY.md and .kmn file. Other than that everything seems fine.

PR #733, opened 2019-05-23
[sil_cipher_music] New keyman keyboard converted from Inkey

LornaSIL:
This was an InKey keyboard that we've converted to Keyman. Documentation was taken from the InKey documentation.


DavidLRowe:
I realize that the documentation came from the original InKey version, but it seems slightly inaccurate in several places:
- there doesn't seem to be a rota on `9`, so "Each of the number keys from 1 to 9..." could be changed to reflect that, or even to "The number keys from 1 to 8, plus five punctuation keys are each set up as a "rota"...
- "quickly pressing the key" to get the rota action is not necessary. Whether you press the key slowly or quickly the rota still works. Perhaps, in the first paragraph under "Keyboard Instructions":
  - remove "quickly" twice
  - remove the sentence: "Pressing it slowly twice produces "33" since the rota is not activated."

Not sure what to do about the Indonesian version! Perhaps put a note in README.md that the Indonesian text in welcome.htm (and the .php file) was not updated when the English was changed??

OR leave the documentation the way it is and note that in README.md?

LornaSIL:
I used Google translate and attempted to make changes to the Indonesian text. I also put a disclaimer in the README.md file. I had planned to change that text earlier, but obviously forgot. Thanks.

DavidLRowe:
Looks good. All that's needed is to change "The English and Indonesian documentation has been slightly" to "The English and Indonesian documentation has been changed slightly" in README.md

PR #738, opened 2019-05-29
[nlci_tamil_winscript] Tamil Winscript keyboard [NLCI]

binilasanki:


keyman-server:

DavidLRowe:
The "build" folder (and all references to it in the package file) should be "build" not "BUILD". This may make no difference on Windows, but other operating systems use case-sensitive folder and file names.

In the "HISTORY.md" file, there are two entries for "2.0". Only one is needed. 

The on screen keyboard file (.kvks) appears to be referencing the Annapurna SIL font. Is this intended? (I'm not very familiar with these scripts.)

The "`" before keystrokes to get English letters doesn't seem to work for f, F, S and M.

In the welcome.htm (and I think also the .php help file), please check the entry for "nn". What I'm seeing is that "nn" gives U+0BA9, and "nnn" gives U+0BA3 (and "N" gives U+0BA3).

Also in welcome.htm, I wonder if the example under "Conjuncts" should be "kxL" (instead of the existing "kxl".

Thanks!

keyman-server:

keyman-server:

DavidLRowe:
The changes look good. I think the only thing needed is to update the Package source to use the "build" folder name (rather than "BUILD"). In Keyman Developer, click on the nlci_tamil_winscript.kpj file, then on the Packaging tab at the bottom, then on the nlci_tamil_winscript.kps file. At that point you can remove the .kvk, .kmx and .js files, then add them back in so that the correct "build" folder name is included.

keyman-server:

DavidLRowe:
You successfully removed the three files with the BUILD reference, but when you added the files back in, you got the wrong ones: they are from the nlci_devananagari_winscript keyboard, not the nlci_tamil_winscript keyboard. 

Also, the language information (`<Language ID="ta">Tamil</Language>`) from the package source (.kps) file was lost.

keyman-server:

PR #743, opened 2019-06-04
[itrans_roman] New keyboard submission

Shreeshrii:
@LornaSIL Please review. 

I have duplicated the rules to allow for UPPERCASE output . Is this the best way to do this?


keyman-server:

keyman-server:

LornaSIL:
* In the keyboard_info file, please delete the comma at the end of the "description" line.
* In the README.md, please change "Fo example," to "For example," (add the "r").
* You could delete the itrans_roman.ico file from the .kps file. It does not need to be distributed in the keyboard package.

LornaSIL:
Here's the results of my testing the keyboard:
* Changes to documentation needed (welcome and help):
     * Instead of ```-kh```, it should be ```_kh```
     * Instead of ```-r```, it should be ```_r```
     * Instead of ```-n```, it should be ```_n```
     * Instead of ```-l```, it should be ```_l```
* Keyboard changes:
     * ```_t``` does not produces the t with macron below. I think you want a rule for ```_t``` in the "c underbar with underscore" section
     * ```..rh``` puts the two dots below the h, not the r (either the keyboard or the documentation needs changed.
     * ```..dh``` puts the two dots below the h, not the d (either the keyboard or the documentation needs changed.
     * ```^r``` is not producing what documentation says. Probably need to add to ISOchandra stores.
     * If you happen to type ```s``` followed by ```h``` you get the results of ```sh``` (ś). If they actually want an ```sh``` should there be a way to type that? 
     * The results of ```'``` is U+2019 which is a punctuation quote mark. Is that what the IAST/ISO scheme want? I would tend to think you would want U+02BC which is not punctuation. It looks the same, but it is used for ejectives and glottals. I will accept the keyboard if you want to use U+2019, I just wanted to raise the issue.

Otherwise it looks good.

LornaSIL:
Regarding uppercasing. Your approach sounds fine. I noticed this rule which probably needs changing to uppercase:
```
"_" + "H" > "ẖ"
```
There does not appear to be an uppercase to that character so it would need to be:

```
"_" + "H" > U+0048 U+0331
```
Also, remember to implement any changes from my previous comments into the uppercase rules.

LornaSIL:
Also, the name needs changing to something like: ```Indic Romanization (ITRANS)``` in the .kmn file. Or, you've called it ```Indic Romanization from ITRANS``` everywhere except the .kmn file so you could keep that name.

keyman-server:

keyman-server:

LornaSIL:
The keyboard name in the .kmn file is still "itrans_roman".
With that change it looks ready to approve.

keyman-server:

Shreeshrii:
@LornaSIL Thank you so much for your detailed review and feedback. All required changes should be in now.

Shreeshrii:
@LornaSIL Thank you!

PR #744, opened 2019-06-11
[basic_kbdazst] Azerbaijani (Standard) Basic

MakaraSok:

*The branch should be named "basic_kbdazst," but I'll leave it with this for now as it doesn't seem to hurt.*


DavidLRowe:
In the .keyboard_info and .kps files, the BCP 47 code is "azj-Cyrl-AM".
- az is Macrolanguage
- azj is "Azerbaijani, North" (Ethnologue: Arabic script, Naskh variant [Arab], used until 1920s. Braille script [Brai]. Cyrillic script [Cyrl], widely used in Armenia. Latin script [Latn], official script in Azerbaijan since 1992.)
- azb is "Azerbaijani, South" (Ethnologue: Arabic script, Naskh variant [Arab], used in Iraq and Iran. Arabic script, Nastaliq variant [Aran], used in Iran.)
- I wonder if the script code should be Latn. This keyboard is certainly producing Latin output.
- Then, after changing the script from Cyrl to Latn, the region code AM for Armenia seems incorrect
- Probably no region code is needed.
- So, I think the BCP 47 code should be "azj-Latn" in both files.

The &targets store in the .kmn file should be changed to 'any' since you have a touch layout. (This will then match the README.md file.)

The .ico file should probably change since the script is not Cyrl.

In the .php file, can you verify the syntax for `<div id='osk'>` and `<div id='osk-tablet'>`? I thought that you needed to list the desired layouts. If you don't specify any layouts, will it list them all?

For the touch layout, are you trying to include all the keys that are on the desktop layout? If so, # and ~ are missing. Similarly there are a number of keys on the touch layout that don't appear on the desktop layout ( ' ! \ « » < > { }  ), but I think that since you are importing the desktop layout, you don't have the liberty to change it.

And I'm curious about the following. For the touch layout, currently the longpress of ö (on the default layer) and Ö (on the shift layer) include both upper and lower case z. I also find it curious that x, X, j, j, ğ and Ğ are on the numeric layer, and especially that some punctuation characters are on longpress of x and j.

LornaSIL:
> In the .php file, can you verify the syntax for <div id='osk'> and <div id='osk-tablet'>? I thought that you needed to list the desired layouts. If you don't specify any layouts, will it list them all?

If you don't specify layouts I think the desktop one will just show default and shift.

This would be the syntax for one desktop scenario
```

```

And this would be the syntax for one mobile scenario
```
<div id='osk-tablet' data-states='default shift rightalt shift-ralt'>

```

Notice the ```shift-ralt``` is different than the ```rightalt-shift``` for desktop!

This would be for numeric:
```
```

I think David is right that the script code is azj-Latn AND you could probably add az-Latn. The script isn't suppressed in this case.

DavidLRowe:
Also in welcome.htm "hiden" is misspelled and should be changed to "hidden".

MakaraSok:
> In the .keyboard_info and .kps files, the BCP 47 code is "azj-Cyrl-AM".
> 
> * az is Macrolanguage
> * azj is "Azerbaijani, North" (Ethnologue: Arabic script, Naskh variant [Arab], used until 1920s. Braille script [Brai]. Cyrillic script [Cyrl], widely used in Armenia. Latin script [Latn], official script in Azerbaijan since 1992.)
> * azb is "Azerbaijani, South" (Ethnologue: Arabic script, Naskh variant [Arab], used in Iraq and Iran. Arabic script, Nastaliq variant [Aran], used in Iran.)
> * I wonder if the script code should be Latn. This keyboard is certainly producing Latin output.
> * Then, after changing the script from Cyrl to Latn, the region code AM for Armenia seems incorrect
> * Probably no region code is needed.
> * So, I think the BCP 47 code should be "azj-Latn" in both files.
> 

I will update this accordingly. By the way, how can one obtain all these info? The details in the .keyboard_info are left as is from the `ImportKeyboard`. 

> The &targets store in the .kmn file should be changed to 'any' since you have a touch layout. (This will then match the README.md file.)
> 

OK

> The .ico file should probably change since the script is not Cyrl.
> 

What should it be changed to? The current one was auto-generated by the `ImportKeyboard` tool.

> In the .php file, can you verify the syntax for `<div id='osk'>` and `<div id='osk-tablet'>`? I thought that you needed to list the desired layouts. If you don't specify any layouts, will it list them all?
> 

Please drop me a link to a doc that instructs how this can be done.

> For the touch layout, are you trying to include all the keys that are on the desktop layout? If so, # and ~ are missing. 

Yes, everything in desktop layout should be in the touch layout.

> Similarly there are a number of keys on the touch layout that don't appear on the desktop layout ( ' ! \ « » < > { } ), but I think that since you are importing the desktop layout, you don't have the liberty to change it.
> 

The touch layout should be updated and made usable. If kept as is, I'm afraid it's not user-friendly. @LornaSIL What is your take on this?

> And I'm curious about the following. For the touch layout, currently the longpress of ö (on the default layer) and Ö (on the shift layer) include both upper and lower case z. I also find it curious that x, X, j, j, ğ and Ğ are on the numeric layer, and especially that some punctuation characters are on longpress of x and j.

The touch layout was recreated using the `basic template`. Minimal changes have been made. Please advise me on what should be done with this. @DavidLRowe @LornaSIL

MakaraSok:
> For the touch layout, are you trying to include all the keys that are on the desktop layout? If so, # and ~ are missing. Similarly there are a number of keys on the touch layout that don't appear on the desktop layout ( ' ! \ « » < > { } ), but I think that since you are importing the desktop layout, you don't have the liberty to change it.

So, should we import the OSK into the Touch Layout and keep it as is? If so, I doubt whether the touch layout is usable. @DavidLRowe

DavidLRowe:
In .kmn, `store(&TARGETS) 'any'` would be better than `store(&TARGETS) 'any windows macosx linux web'`

After that change, all that is left are changes to the touch layout.

DavidLRowe:
Here are some thoughts on a mobile layout:
(A) Add an eleventh key to each row, then put all the alphabetic characters on the default layer
11 keys: püşudbmyohx (with j longpress on x)
11 keys: gısarnəlicğ
11 keys: [shift]çfvtkeqzö<bksp>
5 keys: [num][opt][space][full stop] (with other punctuation as longpress on full stop)[enter]
(B) Put the corresponding upper case alphabetic characters on the shift layer. (The full stop and longpress keys could be repeated for shift layer.)
(C) Numeric layer: Assuming that you keep the 11 keys per row configuration, you'll have extra keys on the numeric layer, since ğ will move off (as will x and j, but they currently have longpress keys). # and ~ (which are on the desktop layout, but missing from the touch layout) will take two, @ can move off the ₼ key. And I'd suggest repeating the comma (,) key, so that someone typing numbers on the numeric layer will have both comma and full stop, but I'd be open to adding some other key.

Underlying rationale:
- keep alphabetic characters together on default and shift layers
- keep the number of characters in a row to a minimum (10 is better, but here 11 seems to be necessary)
- don't mix alphabetic, punctuation and symbol characters on the same key

But maybe @LornaSIL has other ideas!

MakaraSok:
Thank you, @DavidLRowe. Updates have been made. Please take a look at it again.

PR #745, opened 2019-06-11
[basic_kbdbgph1] Bulgarian (Phonetic Traditional) Basic

MakaraSok:

**Known Issue?** *The emulator for some devices (i.e. iPhone) on Chrome doesn't show SHIFT key on Numeric Layer.*
no shift in numeric layer


DavidLRowe:
In the information in `https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry`, "bg" is listed with `Suppress-Script: Cyrl`, therefore the tag should be "bg" rather than "bg-Cyrl" in both .keyboard_info and .kps files.

In the .php file, the id='osk' and id='osk-tablet' need data-states=... to specify which layouts to display. I think you need `id='osk' data-states='default shift rightalt-shift'` and `id='osk-tablet' data-states='default shift numeric'`

.kmn file needs &targets 'any' (and windows, macosx, linux, web should be removed).

touch layout issues:
- `\` on longpress `.` on default layer produces U+044E
- '\' on longpress `>` on shift layer produces U+044E
- `ч` (U+0447) and `Ч` (U+0427) on desktop, but not on touch
- `{` on longpress `ш` on numeric layer produces U+0428 
- `}` on longpress `щ` on numeric layer produces U+0429

touch layer questions/suggestions (optional changes you may want to consider):
- currently there are three letter pairs on the numeric layer, plus `ч` and `Ч` that are missing from the touch layout. Could they be moved to the default and shift layers? Perhaps add an additional key in the second row, move the period key to the right of the space to make room for an extra key, then double up some characters (`щ` a longpress on `ш` for example).
- why not have `.` on shift layer instead of `>`?

MakaraSok:
@DavidLRowe Everything should be fixed now. Please check again.

DavidLRowe:
Changes look very good. One thing to change: The default layer has U+0448 and U+0449 as longpress on U+044E key (just to the left of backspace key), however on shift layer, the upper case characters (U+0428 and U+0429) need to be added as longpress keys on U+042E key.

MakaraSok:
@DavidLRowe Added

PR #746, opened 2019-06-12
[basic_kbdcan] Canadian Multilingual Standard Basic

MakaraSok:

@DavidLRowe @LornaSIL What is the right script tag for this keyboard?


LornaSIL:
* In the .kps
     * I'm surprised the import didn't also include fr-CA and/or fr. You should add fr and fr-CA to both keyboard_info and the .kps BCP47 on the details tab.
          * I don't know what other languages it might support, but I happen to know it supports both en-CA and fr-CA
     * The .kps does not need to include the .php help file. You can delete that from the list of files in the .kps.
* In the .kmn
     * Since you have "any", you can uncheck: Windows, macosx, linux, web
     * For the mobile layout, it would be my preference that the lowercase keys on longpress just contain the lowercase characters. Move the uppercase characters from longpress to the shift layer on the appropriate longpress keys.
* In the keyboard_info file:
     * Add fr-CA and fr to the languages
     * When I asked you to do this keyboard I forgot that there is an old keyboard that this should deprecate. Can you add these lines at the end?
```
  "related": {
    "canadian_ms": { "deprecates": true }
  }
```

* Unfortunately, that means you should add a DEPRECATED.md file to the ```keyboards\legacy\c\canadian_ms``` folder. You can find an example in ```keyboards\legacy\c\cameroon azerty unicode```. Just copy that file into canadian_ms and then change ```release/sil/sil_cameroon_azerty``` to ```release/basic/basic__kbdcan```. 

* What this means is that we need to change the version of this keyboard from 1.0 to 1.1 in every place.
That also means the HISTORY.md file needs updating to:
```
1.1 (2019-06-10)
----------------
* Created by ImportKeyboard from Windows 10 

1.0 (2014-03-21)
----------------
* Initial web and mobile layout
```

Sorry about that. I didn't intend to give you any deprecated keyboards to do at this time as they get more complicated.

The rest of the keyboard seems fine.

DavidLRowe:
@LornaSIL Does this also affect the copyright dates (which occur in several places: LICENSE.md, README.md, .kmn, and perhaps .kps and .kpj as well)? That is, should they be "2014-2019" rather than "2019"?

LornaSIL:
Yes, unfortunately they should be adjusted too.

MakaraSok:
Thank you, @LornaSIL, for the feedback. Please take a look at what have just been updated, and also consider my concern below regarding the difference between `canadian_ms` and `basic_kbdcan`.

> * Unfortunately, that means you should add a DEPRECATED.md file to the `keyboards\legacy\c\canadian_ms` folder. You can find an example in `keyboards\legacy\c\cameroon azerty unicode`. Just copy that file into canadian_ms and then change `release/sil/sil_cameroon_azerty` to `release/basic/basic__kbdcan`.

But characters on Control layer are different from those we have in this keyboard. (See https://help.keyman.com/keyboard/canadian_ms/1.0/canadian_ms.php). The `basic_kbdcan` has less of these characters. 
image

LornaSIL:
I do understand your concern, and I noticed the difference, but forgot to comment on it. The canadian_ms one was imported from Windows Vista I think, and as I've done some of these conversions I've noticed that some of the MS keyboards have been modified since then. I still want it deprecated, but we have two options:

1. Leave the basic_kbdcan as is, deprecate the canadian_ms keyboard and people can still get the deprecated mobile keyboard if they really want it (it's still available for download).

2. Update the touch layout (NOT the Desktop) to include a layer that matches what canadian_ms had. I don't want the Desktop to be updated because we are truly wanting to match what Win10 keyboards are. (BTW, Windows keyboards don't have mobile layouts, so we can't match their mobile layouts.)

My own preference is to do #1. If you do #1 it might be good to add to the HISTORY.md a line that says ```Mobile layout does not match the v 1.0 layout```. But, I'm willing for you to do #2 if that is your preference.

MakaraSok:
Let's go with #1. I will add the note.

LornaSIL:
@MakaraSok it looks like everything was done except to add a note in HISTORY.md file. Did you still want to do that?

PR #747, opened 2019-06-12
[itrans_roman] Version 1.1.0 Add IPA vowels based on user feedback

Shreeshrii:


keyman-server:

keyman-server:

keyman-server:

Shreeshrii:
>By the way, in my opinion since we have a HISTORY.md file where we track changes there is no need to also add the changes into the .kmn file as well. However, that's up to you. It doesn't hurt anything, it just takes extra time.

OK. I did not know what was the recommended way to track changes, so I was noting them in both.

Thanks.

PR #748, opened 2019-06-13
[tibetan_ewts] Convert legacy kbd to release

LornaSIL:
This keyboard was converted from legacy to release. I removed Latin and Cyrillic support as well as hex support. Everything seems to be working as documented.

Since I didn't have the source for the mobile layouts, I just downloaded the mobile keyboard on Keyman and tried to match it.


DavidLRowe:
A couple of things to check in welcome.htm

In the "Beginning Combinations" section
- rw (followed by a) -> U+0F62 U+0FAD, whereas welcome.htm has U+0F4F U+0FAD
- sn (followed by a) -> U+0F66 U+0FA3, whereas welcome.htm has U+0F66 U+0FA8

A very minor point: menus are in a different order than \MENU gives. Any reason not to move the "Half-Digits" menu up between the ASTR and CANT menus?

PR #750, opened 2019-06-14
[basic_kbdkurd] Central Kurdish Basic

MakaraSok:


LornaSIL:
This looks really good. I'd like to see one change. On the OSK, shift layer, there is no way to tell what the character for "H" will produce. It produces U+200C which is a Zero-Width Non Joiner and is not visible. I'd like you to consider adding this to the keycap for the OSK H key: ᴢᴡɴᴊ

Most Arabic script users will be familiar with that acronym. Then recreate the shift graphic so the user will know. These are smallcaps so that they will show up on the keycap. If you just use "ZWNJ" which are normal uppercase letters I think they will be too large.

PR #753, opened 2019-06-16
[itrans_roman] Version 1.1.1 add IPA vowels - map to lowercase characters

Shreeshrii:
also removes the font-family in help files, page was not displaying OK on Mac OS.

### DIFF
--- a/release/itrans/itrans_roman/HISTORY.md
+++ b/release/itrans/itrans_roman/HISTORY.md
@@ -1,7 +1,7 @@
 Indic Romanization from ITRANS - Version History
 ===========================
 
-1.1.1 (2019-06-15)
+1.1.1 (2019-06-17)
 ----------------
 * Add alternative maps for some IPA vowels using lowercase punctuation
 
--- a/release/itrans/itrans_roman/HISTORY.md
+++ b/release/itrans/itrans_roman/HISTORY.md
@@ -1,6 +1,14 @@
 Indic Romanization from ITRANS - Version History
 ===========================
 
+1.1.1 (2019-06-15)
+----------------
+* Add alternative maps for some IPA vowels using lowercase punctuation
+
+1.1.0 (2019-06-12)
+----------------
+* Add support for IPA (vowels only) output, useful for words pronunciation key
+
 1.0.0 (2019-06-01)
 ----------------
 * Indic Romanization in various schemes such as 
@@ -11,6 +19,4 @@ Indic Romanization from ITRANS - Version History
 * from ITRANS and 7 bit ISO transliteration 
 * for easy typing of accents.
 
-1.1.0 (2019-06-12)
-----------------
-* Add support for IPA (vowels only) output, useful for words pronunciation key
+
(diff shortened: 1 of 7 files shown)


keyman-server:

keyman-server:

Shreeshrii:
@LornaSIL Thank you!

I have a few questions:

1. I have set the languages for the keyboard to be sa-Latn, hi-Latn etc since this is for Romanization of Indic scripts. However, when I install it, it shows up with language `sa-Sinh-IN` which is for Sinhala script. Why is it so?  What can I do to change it?

2. I also noticed that when I install a new Keyman keyboard on Windows 10, another corresponding Windows keyboard is also installed. eg. with my ITRANS Hindi keyboard, the Hindi INSCRIPT keyboard is also getting installed. With the ITRANS Roman keyboard, Sanskrit Sinhala keyboard is getting installed.
All these show up in my language choice. I would only like to keep the keyman keyboards in my list. Is there any setting I need to change for this?

Thanks!

DavidLRowe:
@mcdurdin Do you have any suggestions?

Shreeshrii:
Another question:

I am working next on the keyboard for Panjabi language written in Gurmukhi script. I have named it as `itrans_gurmukhi` and `Panjabi/Gurmukhi Phonetic (ITRANS)`. Is that ok? 

Should it be called `itrans_panjabi`?

LornaSIL:
@Shreeshrii if the keyboard will support all of Gurmukhi then I'd be okay with it being called itrans_gurmukhi. Otherwise, I think maybe itrans_panjabi_gurmukhi? Or itrans_gurmukhi_panjabi. What do you think? Is it also called Panjabi on the PK side? If so, we shouldn't just call it itrans_panjabi because that could also be using the Arabic script.

LornaSIL:
I think @mcdurdin will have to respond to the previous questions. I suspect that Windows doesn't know you will use Roman text with Sanskrit so it chooses the closest tag that it knows about to associate it with. I wasn't aware that it also installs the keyboard that it already knows about. You can always uninstall it after it installs it and it will leave your keyboard alone. But, of course that is not ideal.

Shreeshrii:
@LornaSIL Thanks! 

>You can always uninstall it after it installs it and it will leave your keyboard alone. But, of course that is not ideal.

I tried that. Now the Windows keyboard is gone. ITRANS_ROMAN shows up with `Unknown Locale (qaa-Latn)`

LornaSIL:
Ok, I realized this was brought up on the community site and it appears to be a bug that has been reported, but probably not fixed yet: 
https://community.software.sil.org/t/input-method-code-in-windows-differs-from-the-code-chosen-in-the-keyboard-layout-in-keyman-developer/2053
If you want to contribute more information to the problem you could add a comment here:
https://github.com/keymanapp/keyman/issues/1719

We shouldn't use this PR thread anymore since the PR has been merged.

Shreeshrii:
Thanks for link to related issue. I will post there.

PR #755, opened 2019-06-19
[basic_kbddzo] Dzongkha Basic

MakaraSok:
- added touch layout using Keyman Dev, import from OSK, as there are too many keys for the Basic Layout Template. 

*Known Issue: the touch layout is very crowded and the key caps are too small for some to type. If using long press for each sub-character that may help reduce the number of keys by almost half. Most if not all consonants have a subjoined version, i.e. ྐ [U+0F90] is a subjoined version of ཀ [U+0F40].*

*Also, there is an existing Dzongkha keyboard on keyman.com, https://keyman.com/keyboards/sil_dzongkha*


DavidLRowe:
The .php help file has 
`$pagetitle = 'Uyghur (Legacy) Basic Keyboard Help';`
I'm not sure it is visible to the user, but it seems as though it should change to 
`$pagetitle = 'Dzongkha Basic Keyboard Help';`

Because Keyman Desktop, by default, uses Alt+Shift+O, Alt+Shift+K and Alt+Shift+M, the user will not be able to use these keys unless the default is changed. It would be helpful to have a comment in the welcome.htm (and probably the .php help file as well), and perhaps even the readme.htm file, to alert the user for the need to change the default keys. Reference release/c/coptic_greek/source/help/coptic_greek.php for a possible wording.

PR #756, opened 2019-06-20
[itrans_odia] New keyboard submission

Shreeshrii:


keyman-server:

LornaSIL:
Looks good! I have a few comments/questions.
* The keyboard has option keys which seems to work, but I don't see that documented anywhere. Could you document it? I think it would be good to put it in readme.htm, welcome.htm and the help file.
* The ```wa``` keys produce different things depending on context. Is that intentional? So, if I type ```wa```  with space before it, it produces what the OSK indicates (ୱ), as well as what is listed under "Consonants" in the welcome.htm file. If I type ```dawa``` (which maybe couldn't happen?), I get ଦୗଅ.
* Still with the ```wa``` under "Consonants with nuktas", under ```.Da/wa``` I think there is a mistake or something missing because the wa doesn't produce the character indicated.
* Please document that ```LL``` produces U+0B0C
* Please document that ```@R^``` produces U+0B44
* Please document that ```@L``` produces U+0B62
* Please document that ```@L^``` produces U+0B63
* Please document that ```RR^``` produces U+0B60
* Please document that ```LL^``` produces U+0B61

It looks to me like you are supporting almost the whole Oriya/Odia block (exceptions U+0B5C, U+0B5D, U+0B5F (nukta characters which you support through the combining nukta), U+0B72..U+0B77 (fraction signs). Since that is the case, would you consider adding the following languages to the keyboard_info and the .kps file: bdv-Orya, bfw-Orya, dso-Orya, gbj-Orya, gdb-Orya, hoc_Orya, jun-Orya, kff-Orya, kxu-Orya, kxv-Orya, kyw-Orya, pci-Orya, peg-Orya, sat-Orya, spv-Orya, srb-Orya, unr-Orya, xis-Orya, zrg-Orya. These are languages we have been told use the Oriya/Odia script. Adding these language associations is by no means required, but it would really be nice to say we support those languages in this keyboard.

Shreeshrii:
Thanks @LornaSIL. You have identified a bug related to wa. I will make changes and upload files.

keyman-server:

Shreeshrii:
@LornaSIL I have made a number of changes in the help files and fixed `wa` related bug. Hopefully it is all OK now.

LornaSIL:
Very nice. Unfortunately we still have one change that needs to be made. Since you changed the graphics for the welcome file, you need to open the .kps file. 
* Delete the reference to itrans_odiaU_.png and itrans_odiaU_S.png
* Add the file itrans_odia.png

I think it will build with this one change.

keyman-server:

Shreeshrii:
Thank you!

PR #757, opened 2019-06-20
[basic_kbdgeooa] Georgian (Old Alphabets) Basic

MakaraSok:
- added touch layout using Keyman Dev, import from OSK, as there are too many keys for the Basic Layout Template.

 Known Issues
 -------------
 
 * `[SHIFT K_Z]` and `[SHIFT K_K]` produce the same output, `U+10B1`. It could potentially be a bug. `[SHIFT K_Z]` should give `U+10BF` (`GEORGIAN_CAPITAL_LETTER_JHAN`) which is the capital letter of `U+2D1F` (`GEORGIAN_SMALL_LETTER_JHAN`). (The touch layout has this fix.)
 * A number of keys are duplicate in the desktop layout, i.e. 
     * `U+10FB GEORGIAN_PARAGRAPH_SEPARATOR` is found on both default and right alt layers.
	 * `U+0028 LEFT_PARENTHESIS` and `U+0029 RIGHT_PARENTHESIS` are found on both default and right alt layers.
	 * `U+003D EQUALS_SIGN` is found in both shift and right alt layers.
	 * `U+0021 EXCLAMATION_MARK` is found in both default and right alt layers.
	 * `U+00A7 SECTION_SIGN` is found in both shift and shift right alt layers.


DavidLRowe:
Because the SHIFT+RALT keys don't include the ones used by Keyman (O, K, M), you don't need the warning about SHIFT+RALT keys in welcome.htm and basic_kbdgeooa.php (and it should be removed to avoid confusion).

I don't know this script at all, but I'm just raising a question. On the touch layout there are more longpress keys on the default layer than on the shifted layer. Are those all characters that have no corresponding upper case or might there be upper case characters missing from the shift layer?

MakaraSok:
> Because the SHIFT+RALT keys don't include the ones used by Keyman (O, K, M), you don't need the warning about SHIFT+RALT keys in welcome.htm and basic_kbdgeooa.php (and it should be removed to avoid confusion).

OK, noted.

> I don't know this script at all, but I'm just raising a question. On the touch layout there are more longpress keys on the default layer than on the shifted layer. Are those all characters that have no corresponding upper case or might there be upper case characters missing from the shift layer?

The desktop version doesn't list those.

PR #761, opened 2019-06-24
[basic_kbdgthc] 

MakaraSok:

*The Icon is changed from EN (created upon import) to GOT.*


DavidLRowe:
.keyboard_info file needs `got-Goth` to match .kps file. 

For .php file, you don't need the reference to Shift+RAlt keys, since the keyboard has none of these keys.

For the touch layout, I would prefer to see the U+10341 and U+1034A characters moved from the numeric layer (where they have punctuation on longpress keys) to the default layer. (I expect they are where they are because they are on the `[` and `]` keys on the desktop layout.)
- If possible, it's preferable to have all the characters of the script on the default layer (along with the shift layer, if applicable). 
- Putting punctuation on the longpress of character keys makes them hard to find.

MakaraSok:
> .keyboard_info file needs `got-Goth` to match .kps file.
> 
> For .php file, you don't need the reference to Shift+RAlt keys, since the keyboard has none of these keys.
> 


Fixed. I shouldn't have missed these. I've gotta be more mindful next time when copying stuff over. :)

> For the touch layout, I would prefer to see the U+10341 and U+1034A characters moved from the numeric layer (where they have punctuation on longpress keys) to the default layer. (I expect they are where they are because they are on the `[` and `]` keys on the desktop layout.)

The basic template suggested this. It looks like any key to the right of "p" and "l" are pushed to the numeric layer. When seeing the name of the characters `U+10341 GOTHIC_LETTER_NINETY` and `U+1034A GOTHIC_LETTER_NINE_HUNDRED`, I thought they are numbers and it's reasonable to have them in the `numeric layer`. Anyway, they are in the default layer now, longpress of the top right key cap.

PR #763, opened 2019-06-24
[basic_kbdgn] Guarani Basic

MakaraSok:


LornaSIL:
This keyboard uses some of the reserved keys (RALT SHIFT O) so you will need to add the same wording you did to the Dzhongkha keyboard:

>This keyboard makes use of SHIFT+RALT keys which are set by default when Keyman is installed. To use this keyboard, you must go to Configuration.../Hotkeys and select the hotkey for "Switch Keyman Desktop Off". Either change the hotkey or turn it off. Do the same for "Open Keyboard Menu" and for "Open Language Switcher".

This keyboard also has some deadkeys in it so that will need documenting. 
* On the OSK I would add
     * U+0301 to the key to the right of "p" (where the [ is on the physical keyboard)
     * U+0308 to the SHIFT layer on the key to the right of "P" (where the { is on the physical keyboard)
     * U+0303 to RALT layer two keys to the right of the o/tilde (where the [ is on the physical keyboard)
     * U+0300 to RALT layer where the ' is on the physical keyboard)
I think I'll upload my version of .kvks in case that proves helpful to you. You'll need to output new graphics to show these keys.

Then in the welcome.htm you'll need something like:
* This keyboard utilizes deadkeys. 
   * Type ◌́ followed by any vowel (aeiouyAEIOUY) to output an acute accent on the vowel.
   * Type ◌̈ followed by any vowel (aeiouyAEIOU) to output a diaeresis/umlaut accent on the vowel.
   * Type ◌̂ followed by any vowel (aeiouAEIOU) to output a circumflex accent on the vowel.
   * Type ◌̃ followed by any vowel (aeiouyAEIOUY) to output a tilde accent on the vowel.
   * Type ◌̀ followed by any vowel (aeiouAEIOU) to output a tilde accent on the vowel.

Then, since this keyboard does have all of these characters, you'll need to add them to the touch layout. You can find them at the bottom of the .kmn file under the deadkey section.

Since Guarani is used in South America, I'd recommend adding the upside down question mark (U+00BF) and upside down exclamation mark (U+00A1) to the longpress of the period/fullstop key. Those are used just as often as the regular question mark and exclamation mark. Then, you can remove them from the numeric layout. You'll have to move one of the keys that is on the longpress to the main layer. Like move the < and > to the main keycap if you want.

On the numeric layer, the _ key is outputting a normal hyphen. Since you already have a hyphen below it, I would recommend you output U+005F (I think if you change the "modifier" for that key from "default" to "shift" it will work.

LornaSIL:
[basic_kbdgn.zip](https://github.com/keymanapp/keyboards/files/3322440/basic_kbdgn.zip)

LornaSIL:
You'll also need a help file, but let's get the welcome.htm correct first.

MakaraSok:
The `.kvks` you sent helps saving some time. :) BTW, the key for ◌̃ is not in the right location, it should be on `K_RBRKT` rather than `K_LBRKT` according to the imported `.kmn`.

I have updated both welcome and help pages. ptal

LornaSIL:
Looks good. There are two issues. On my Android I just see a square box with the combining marks. Maybe we just leave it and hope that mobile phones improve font support? I'm not sure what to do. I can sort of see the combining mark with the box. And the help file will help with documenting it.

The 2nd issue is something I wasn't clear in explaining. In languages in South America they use BOTH the upside down AND the rightside up question and exclamation marks. They start the question with upside down and end it with rightside up. Same with exclamation. So, you might have something like this ```¿Is this a question? ¡No!```

So, they need all 4 marks on the mobile layout on the longpress of the fullstop/period. So, add those back in and unless you can think of a good way to handle the boxes I think that's the only change needed.

MakaraSok:
Thank you for the comment and suggestion.

> On my Android I just see a square box with the combining marks. Maybe we just leave it and hope that mobile phones improve font support? I'm not sure what to do. I can sort of see the combining mark with the box. And the help file will help with documenting it.

If we use `U+00b4` instead of the combining acute accent `U+0301`, `U+00a8` instead of `U+0308`, `U+007e` instead of `U+0303`, `U+0060` instead of `U+0300` and `U+005e` instead of `U+0302`; they should look more visible. Like so on Android 9:
image

> The 2nd issue is something I wasn't clear in explaining. In languages in South America they use BOTH the upside down AND the rightside up question and exclamation marks. They start the question with upside down and end it with rightside up. Same with exclamation. So, you might have something like this ¿Is this a question? ¡No!

The longpress list gets too long and it would become two lines rather than one nice single line.

LornaSIL:
I'm happy with this:

>If we use U+00b4 instead of the combining acute accent U+0301, U+00a8 instead of U+0308, U+007e instead of U+0303, U+0060 instead of U+0300 and U+005e instead of U+0302; they should look more visible. Like so on Android 9:


I didn't realize the closing "?" and "!" were on the numeric layer so I didn't think there was a way to type them. I think you should move the opening ones from the longpress to the numeric layer and put them on the matching ? and ! keys. That way it will be less confusing to users.

MakaraSok:
Updates have been made to aid with the visibility of the accent marks and the upside down question and exclamation point were moved to the numeric layer as longpress of their respective closing marks.

PR #765, opened 2019-06-24
Aksara Bali Unicode keyboard

kamholz:
Release keyboard for typing Aksara Bali (Balinese script)


keyman-server:

DavidLRowe:
Thanks for submitting this keyboard. 

A couple of optional items:
- Would the keyboard support other languages that use the `Bali` script (kaw-Bali and sas-Bali, for example)? If so, these can be added to the .keyboard_info file and to the .kps (Keyman package source).
- Similarly, you could add 'web' to the targets and this keyboard could be used on the web.

In the Keyman project file (.kpj), we need to have the `<CompilerWarningsAsErrors>` flag set `True`. That can be changed in Keyman Developer using the Project | Project Settings... dialog.

The bali_simbar_panduan.pdf file contains a lot of information beyond what is needed to use the keyboard (and some of it seems to be a bit dated). I wonder if you have access to the source and could make a welcome.htm file with the pages from the PDF document numbered 8-22, and 26-29. 

It's great that the explanations are in (I assume) Indonesian, but it is also helpful for Keyman support if the explanations are in English. I don't know whether or not that is possible in this case. 

The aksarabali_panlex.php needs to have more information about using the keyboard. Often the welcome.htm is taken as the source for the .php file. In this case (if it is not feasible to make a welcome.htm file and you keep the PDF), I suppose you could reference the PDF file and include a copy of the PDF file in the source/help/ folder.

@LornaSIL Do you have any additional thoughts?

kamholz:
Hi David, thanks for your feedback. See below.

> Would the keyboard support other languages that use the Bali script (kaw-Bali and sas-Bali, for example)?

It should work for Kawi, so I'll add that. It does not support the Sasak consonants, but there are alternative ways to write them with the standard Balinese consonants plus rerekan (nukta). Should I add sas-Bali or not?

> Similarly, you could add 'web' to the targets and this keyboard could be used on the web.

Ah, so no other changes are needed other than that? I can definitely add it then.

> The bali_simbar_panduan.pdf file contains a lot of information beyond what is needed to use the keyboard (and some of it seems to be a bit dated). I wonder if you have access to the source and could make a welcome.htm file with the pages from the PDF document numbered 8-22, and 26-29.

> It's great that the explanations are in (I assume) Indonesian, but it is also helpful for Keyman support if the explanations are in English. I don't know whether or not that is possible in this case.

This keyboard is a reimplementation of the Bali Simbar keyboard. The Bali Simbar keyboard is the de facto standard keyboard in Bali for typing in Balinese script. It's meant to be used with the Bali Simbar font, which is not Unicode. This keyboard is the first keyboard I'm aware of that supports typing Balinese in Unicode in the way people in Bali are accustomed. 

The PDF is the guide that comes with the Bali Simbar keyboard. There are a lot of details and special cases, nearly all of which are supported in this keyboard. I don't have the source for the original PDF. In any case, I am not the author of the PDF, just of the keyboard. I have permission from the author to include the PDF with this keyboard, and I don't really want to modify his work. I agree that it would be good to have a new guide, but it would involve more time than I can justify at this point (would have to convert all of the examples to Unicode, translate to English, etc.).

I would be happy to make clearer what the relationship of this keyboard is to Bali Simbar, and that the PDF file is the one distributed with Bali Simbar.

I will address several of the matters you've raised in subsequent commits.

DavidLRowe:
I don't know this script at all, so based on your comments, I'd suggest including kaw-Bali, but not sas-Bali (both in the .keyboard_info file and in the .kps (package source) file).

What would you think of the following for the opening paragraph in welcome.htm? (This is a suggestion only, based on what you wrote above.)
"This keyboard lets you type Askara Bali (Balinese script) in Unicode. It is a Unicode reimplementation of the Bali Simbar keyboard (which is the de facto standard keyboard in Bali for typing in Balinese script)."

It would be good to indicate that the PDF is included with the permission of the author.

When you have finalized the text for welcome.htm, please include that (including the reference to the .pdf file) in the source/help/aksarabali_panlex.php file as well, and place a copy of the .pdf file in the source/help/ folder. This way the online help will have the same information as the installed keyboard.

LornaSIL:
I think @DavidLRowe 's suggestion is good. I like the idea of having the English paragraph he suggested and then linking to the pdf is fine in this case. If you don't feel up to creating the php file, we can do that in a separate PR after this PR is merged. It'll be nice to have a Balinese keyboard available.

keyman-server:

kamholz:
OK, as far as I know I've now addressed all of the comments. Please let me know if anything else is required before we can merge this. Thanks!!

keyman-server:

keyman-server:

DavidLRowe:
In the package source (.kps), the aksarabali_panlex.js file from the build/ folder needs to be added ("Files" tab), so that the web support gets into the package (.kmp) file.

Also under "Files", you can remove the reference to the aksarabali_panlex.ico file. (That file will have been compiled into the .kmx file, so it doesn't need to appear separately.)

Also in the .kps file, "Keyboards" tab, you'll need to use the "Add..." button to add entries for "ban-Bali" and "kaw-Bali". (The entries here need to parallel what is in the .keyboard_info file.)

Since you need to make those changes to the .kps file, could you also add "web" to the supported platforms in the README.md file?

kamholz:
Done.

keyman-server:

DavidLRowe:
The changes to the PR look fine, but I'm seeing "This branch is out-of-date with the base branch". I think you can click on the "Update branch" button to fix that.

kamholz:
Fixed now?

DavidLRowe:
OK, looks good!

keyman-server:

LornaSIL:
Congrats on our first Balinese keyboard!

PR #767, opened 2019-06-25
[basic_kbdhau] Hausa Basic

MakaraSok:


LornaSIL:
This keyboard will also need the statement in welcome.htm and help that you added to the Guarani one:
```
<p>This keyboard makes use of SHIFT+RALT keys which are set by default when Keyman is installed. To use this keyboard, you must go to <b>Configuration.../Hotkeys</b> and select the hotkey for "Switch Keyman Desktop Off". Either change the hotkey or turn it off. Do the same for "Open Keyboard Menu" and for "Open Language Switcher".</p>
```
Any keyboard that uses ALT SHIFT K, ALT SHIFT M, or ALT SHIFT O (or RALT) will need this statement because Keyman makes use of those keys. I've got a Post It note on my monitor to remind me which keys are problematical :)

Both the .kps file and the keyboard_info file have ```ha-Latn-SD``` (which I'm sure came from the keyboard import). Since the main country for Hausa is Nigeria I think we should change it to ```ha-Latn```.

The .ico file looks like an RA. Is it possible to change it to HA?

Everything else looks good.

MakaraSok:
@LornaSIL Thank you for the catch. :)

PR #768, opened 2019-06-25
[basic_kbdhaw] Hawaiian Basic

MakaraSok:


LornaSIL:
This keyboard will also need the statement in welcome.htm and help that you added to the Guarani one because it uses RALT SHIFT O which is one of the reserved keys:
```
<p>This keyboard makes use of SHIFT+RALT keys which are set by default when Keyman is installed. To use this keyboard, you must go to <b>Configuration.../Hotkeys</b> and select the hotkey for "Switch Keyman Desktop Off". Either change the hotkey or turn it off. Do the same for "Open Keyboard Menu" and for "Open Language Switcher".</p>
```
There's just one other change that I think should be made. U+02BB (which is on the OSK at the ' position) is currently on the numeric layer of the mobile layout (on the _ key). You wouldn't know this since it looks like punctuation, but that is an important orthographic character for Hawaiian. I think it would be best to move it to a separate key on both the default and shift layers. If it were me I think I would shorten the space key and put the U+02BB key to the right of the space key. Maybe you have a better idea. Another option would be to put it to the right of the el (l) key.

MakaraSok:
Thank you, @LornaSIL for catching that. Fixes have been made. ptal

Issue #485, opened 2018-10-16
fv_hailzaqvla and fv_yukon plus possibly others. Canonical ordering of diacritic marks


alex-fpcc:
In the FirstVoices keyboard app on iPhone. After upgrading to ios 11 the are symbols that are now in the wrong place.  This is for the  Haíɫzaqvḷa keyboard. 

For example - ṃ́ ḷ́ ṇ́ ṇ̓ ṃ̓ ḷ̓ 
the second symbol under the “.” Is supposed to be over the letters. 
Feedback From Rory Housty <rhousty@gmail.com>
This seems to be a font issue in iOS 11&12 as my iPad running ios 9 is still working fine.
Characters are displayed correctly in the keyboard button but when typed one of the marks moves bellow the other one.

From Marc:

> I've tested this here and it appears to be a renderer change in iOS 11 with multiple marks: the keyboard is incorrect because the dot-below should be before the mark above in the text store for canonical ordering. If you manually insert them in the right order they display correctly.
> 
> Earlier versions of iOS were less pedantic about this but it the keyboard as it stands is technically incorrect.
> 
> It is not a complex change to the keyboard but would involve putting out an update to the FV app. On the Keyman side, the process is now to submit a pull-request to the Keyman keyboards repository at https://github.com/keymanapp/keyboards with a corrected keyboard. The fix goes out to the Keyman app automatically when the change is approved.
>

alex-fpcc:
Is there an easy way for me to tell all of the keyboards that are affected by this?

So far I know it is the haizla and all of the Yukon keyboards.

DavidLRowe:
Here's an attempt to look for U+0323 (combining dot below) in the FirstVoices keyboards in the Keyman keyboards repository under `release`, using `grep` from command line in Ubuntu. This is not exhaustive (and may not even be the best way to search!). Result is `<filename>:<number of occurrences>`

Look for U+0301 U+0323 sequence (wrong canonical ordering):
```
  grep -r -c -H  $'\u0301\u0323' fv*/source/*.kmn
fv_hailzaqvla_kmw/source/fv_hailzaqvla_kmw.kmn:6
fv_nlekepmxcin_kmw/source/fv_nlekepmxcin_kmw.kmn:6
fv_stlatlimxec_kmw/source/fv_stlatlimxec_kmw.kmn:40
```
Look for U+0313 U+0323 sequence (wrong canonical ordering):
```
[...]

PR #770, opened 2019-06-27
[fv_hailzaqvla] use canonical ordering below mark then above mark

LornaSIL:


LornaSIL:
Relates to Issue #485

PR #771, opened 2019-06-27
[fv_gwichin] use canonical ordering below mark then above mark (ogonek and grave)

LornaSIL:
I also added ` and ^ to the mobile layout so they could type grave and circumflex on mobile since they can type them on desktop.


LornaSIL:
Relates to Issue #485

DavidLRowe:
For touch layout, on shift layer, "." produces ">"

For touch layout, on both default and shift layers, key with U+0027 produces U+2019, (Is this the intended behavior?)

The welcome.htm file says "keys with a red background are combining accents" but I don't see any keys with a red background.

PR #773, opened 2019-06-28
[sil_eastern_congo] Add small cap I back into keyboard AND add mobile…

LornaSIL:
… layout


DavidLRowe:
One item to change and some questions:

Longpress "s" has "s" in longpress list but produces "ß"

Longpress accent (to right of space) shows dotted circle under dieresis, but none of the others. This must be just a font / display issue, since I don't see anything different on that key in the file.

The keys seem to jump a bit when moving from default to shift layer. (Numeric layer, of course, has one fewer key on middle row.) Is it worth adjusting padding?

Is it worth changing the comment(s) in the .kmn file to include "6"?

LornaSIL:
I tried to fix the "jumping" of the layout, but as far as I can tell, the keys have the same widths in default and shift layers.

Makara taught me to use spacing diacritics for the display (but not the output) because I was seeing square boxes and couldn't even see what the diacritic was, but I couldn't find the spacing diaeresis yesterday. I found it today.

Since we can't paste into the touch layout except when we look at code, I was putting some text in each key (like the s) and then going to code and pasting in the actual esh. I missed that one :( Thanks for catching it.

PR #783, opened 2019-07-02
[basic_kbdhebl3] Hebrew (Standard) Basic keyboard

MakaraSok:

*This is a RTL keyboard, but by default didn't set that. Should we not set RTL? The SHIFT layer has all English alphabet.*


LornaSIL:
Yes, you are right that right-to-left needed enabling. I'll check the keyboard.

PR #786, opened 2019-07-04
[basic_kbdibo] Igbo Basic

MakaraSok:
- update icon

*`RALT + SHIFT` and `deadkey` are utilized in this keyboard.*


LornaSIL:
* I think the BCP-47 tag could be ```ig-Latn``` and remove the ```-NG``` in both the .kps and keyboard_info
* I can't find the forward slash ```/``` on the mobile layout. Is it there? 
   * Since you have the backslash on the longpress for ```.``` you could put the forward slash where the backslash is on the numeric layout.

Otherwise, everything looks good.

LornaSIL:
I don't see that you added the forward slash to the mobile layout. Is it already there and I just don't see it?

PR #789, opened 2019-07-08
[basic_kbdjav] Javanese Basic

MakaraSok:
- update icon

*The touch layout is tricky to create. Any advice on it is welcome.*


LornaSIL:
* Regarding the OSK, could you add in the display keys for ```- = _ + ``` and ```ꧏ``` (@ key) then recreate your graphics. (Don't try to reimport the OSK because of the issue with the RALT SHIFT keeps which seems to hang importing the OSK.)
* I would suggest you write rules to prevent the output of ```V``` and ```>```.
* It looks like you've done a good job of matching the right characters to go on the numeric layout, and they all seem to work.
* Because there are not many fonts supporting this script, I suggest adding a sentence to the readme.htm (or welcome.htm) file saying something like: 
```
<p>A font must be installed to support this keyboard. Windows includes the <b>Javanese Text</b> font which supports the <b>Javanese</b> script. A freely available font for this script is <A href="https://www.google.com/get/noto/#sans-java" target=_blank>Noto Sans Javanese</A> </p>
```

PR #792, opened 2019-07-08
[basic_kbdturme] Turkmen Basic: Import from Windows and merge in mobile

LornaSIL:


DavidLRowe:
.php file has "Numeric" instead of "numeric" (which causes the display to show a gray background with no keys).

The OSK has an entry on the Left+Ctrl for space. Is that significant?

Everything else looks fine.

PR #794, opened 2019-07-10
[basic_kbdtt102] Tatar Basic import

LornaSIL:
Import, create mobile, deprecate 2 old keyboards.


DavidLRowe:
Everything looks fine, although the OSK has an entry for LCtrl+Space, which probably doesn't hurt anything,  though it might be confusing. Would you want to change that? I'll approve it either way.

PR #796, opened 2019-07-12
[basic_kbdkyr] Kyrgyz Cyrillic Basic import

LornaSIL:
Improved mobile layout from previous.


DavidLRowe:
I don't see that Shift+Alt+K (or O or M) are used by this keyboard, so the note in readme.htm, welcome.htm and the help .php about changing the Keyman defaults doesn't seem to be needed.

LornaSIL:
Oh, yes! I somehow thought it was on the "K" key.

PR #797, opened 2019-07-12
[basic_kbdru1] Russian (Typewriter) Basic import

LornaSIL:
Minor enhancement to mobile layout.


DavidLRowe:
Touch layout seems to have an empty "ctrl" layer (and no way to get to it).

Touch layout lacks a way to type ₽ (RAlt+8 on desktop layout)

LornaSIL:
I was going to tell you where it was. Is there another key you would have checked first?

DavidLRowe:
No, that location is fine. I didn't notice the longpress when I first looked. (Then I saw it on another keyboard and came back and found it on this one.)

PR #798, opened 2019-07-12
[basic_kbdru] Russian Basic import

LornaSIL:


DavidLRowe:
Touch layout, shift layer: shift button doesn't take you out of shift layer. Other than that, everything looks fine.

PR #799, opened 2019-07-12
[basic_kbdbhc] Bosnian (Cyrillic) Basic import

LornaSIL:
This one only has one deprecated keyboard. The "bosnian" mobile layout is Latin, not Cyrillic.


DavidLRowe:
desktop layout, default layer, hyphen key doesn't seem to produce anything although ' is on the key.

LornaSIL:
Oops. That's a deadkey. This keyboard will require more work.

DavidLRowe:
It seems like the upper and lower case acute accented characters added as longpress keys on g and k are reversed (that is, not the character displayed in the longpress menu, but the actual character produced).

PR #804, opened 2019-07-16
[krung] Krung (transliteration) Keyboard

MakaraSok:
This is a transliteration keyboard developed based on an existing keyboard known as [Clever Rabbit](https://code.google.com/archive/p/cleverrabbit/downloads), which uses English characters to write Krung words.


LornaSIL:
* You'll have to put the pdf in the help folder as well if you want people to have access to that document.
* The welcome.htm and krung.php both have "Numerals and Signs" header at the beginning of the file that can be deleted (you have that section at the end).
* The "Consonant" section of the welcome.htm and krung.php files are confusing to me. You say that the autoconverted counterpart of ```q  អ``` is ```អ៊```, but there is no documentation showing how to get that character. You could add a sentence that says: "Certain consonants are mapped, while others are automatically converted to their corresponding counterparts **depending on the context. For example, typing ```qay``` produces the first form ```អៃ``` and typing ```qunh``` produces the second form ```អ៊ុញ```. How to type these contexual forms are described in the documentation below.** 
* Please doublecheck that ```z``` is producing the right thing. I'm getting ```អគ់``` not ```គ``` as the documentation indicates.
* Is it your desire for people to be able to type the uppercase Latin keys? 
  * If not, you might want to write a rule to exclude the ability to type them. 
  * Then, you could/should delete the shift layer in the tablet keyboard. This will mean you to remove the shift key and all those related tasks.
  * Please delete the phone layout. I didn't test it and it appears to be identical to the tablet layout. There is no need to maintain both if they are identical.
* You could make the BCP47 code in both the keyboard_info and the .kps be ```krr-Khmr```. There is no need for the ```-KH```. However, since you are consistent, I would not require you to change it. It's just that the shorter tags are preferred.

I will be traveling and will be back in the office the 24th. I'm not sure that I'll be able to do much more keyboard reviewing during this time.

LornaSIL:
The tablet layout still has a "Shift" layer which I think should be removed. Then you will need to remove the "Shift" key on each layer.

MakaraSok:
It magically reappears after getting deleted. An [issue](https://github.com/keymanapp/keyman/issues/1898) has been reported on this.

PR #805, opened 2019-07-16
Balochi persian

rmlockwood:
New keyboard for Balochi based on the Windows Persian keyboard.


keyman-server:

LornaSIL:
Very nice. It looks quite good although I do have some feedback!
* In the .kps file you can remove:
  * README.md
  * LICENSE.md
  * HISTORY.md
  * These files are just needed for the github repo
* Does this keyboard have any changes to it from the Standard Persian keyboard? If so, it would be nice to say that in the HISTORY.md file, "modified to include characters required for Balochi" or something like that. (BTW, we do plan to pull in the two Persian Windows keyboards, it's on our tasks to do.)
* In the OSK the `` key produces U+200D, but there is no visible way to know that. Did you intend to use a graphic? You could just type in zwj on the keycap. Alternatively, you could add text to the welcome.htm saying that the ` produces U+200D.
* On the tablet layout: 
  * The U+200D can't be seen. Could you type in ```zwj``` just to give a visual indication of what is expected? Capitals seem too big for small devices.
  * Also a visual indication for U+200C for the space on the shift layer ```zwnj```?
  * The two keys to the right of the shift on both the default and shift layers (\ and |) do not produce anything. Since you have them elsewhere you could delete those keys.
* On the phone layout: 
  * the exclamation is producing U+061B, but the character next to it is also producing U+061B. There is no other way to type exclamation.
  * I don't see the multiplication sign on the phone layout, but I could have missed it. You could add it to the + sign as a longpress.
  * I don't think zwj or zwnj are available on the phone layout, should they be?
* In the php file, instead of this:
```
<h2>Keyboard Layout</h2>
<div id='osk'>
</div>
```
you could copy this in there:
```
<div id='osk' data-states='default shift ctrl'>
</div>

</div>

</div>
```
That will give the base layout for all the devices (but it doesn't give the longpress keys)

keyman-server:

rmlockwood:
I fixed these issues. Please take a look.

 

Thanks,

Ron

keyman-server:

PR #806, opened 2019-07-16
[basic_kbdkni] Khmer (NIDA) Basic keyboard

MakaraSok:


LornaSIL:
Everything looks good except the .kmn file needs to say "any" instead of just the selected choices. I'm glad you work on this script! It's hard to tell the difference in all the characters to my untutored eyes.

PR #808, opened 2019-07-16
[basic_kbdmacst] Macedonian (FYROM) - Standard Basic kbd from Win10

LornaSIL:
Mobile layout is based on previous keyboard


DavidLRowe:
Touch layout, default layer, 1st row, 11th key, label is "/", but produces "-".
shift layer, 1st row, 11th key, label is "?", but produces "-".

Everything else looks fine

PR #810, opened 2019-07-16
[basic_kbdbu] Bulgarian Typewriter Basic kbd from Win10

LornaSIL:
Mobile layout is new.


DavidLRowe:
Touch layout, shift layer, third row, key to left of backspace. Most longpress keys (`,?'"\:;`) on this key don't produce the displayed character.

PR #811, opened 2019-07-16
[basic_kbdbgph] Bulgarian Phonetic Basic kbd from Win10

LornaSIL:
Mobile layout is new.


DavidLRowe:
Touch layout, shift layer, key to right of space bar: A number of keys don't produce what is displayed (on base or longpress): `.,';`

PR #812, opened 2019-07-16
[basic_kbdbash] Bashkir Basic kbd from Win10

LornaSIL:
Mobile layout is new.


DavidLRowe:
.keyboard_info has "ba-Cyrl" whereas .kps has "ba-Cyrl-RU". Is that a problem?

Touch layout, alt layer, top row, third key doesn't seem to produce any output. Same for fifth key and last key on that row. Same for second row, first and third keys.

DavidLRowe:
Touch layout, alt layer, top row, last key on the row doesn't produce any output. (I think you need to add `"layer": "default"` to override the "alt" from the layer, or switch to U_ format.)

PR #816, opened 2019-07-18
[basic_kbdlvst] Latvian (Standard) Basic imported from Win10

MakaraSok:


DavidLRowe:
Since the keyboard uses Shift+Alt+O and Shift+Alt+K, you should add to the welcome.htm and .php files a warning about needing to change these default keys.

Just some minor changes to the documentation (welcome.htm and presumably *.php also):
- ' plus 6 produces U+2019 (not U+0027 as in the documentation)
- ' plus - produces U+2013 (not U+002D as in the documentation)
- The entry for Shift+' and s shows `sŠ` but it should be just `Š`
- Is there some way to indicate that ' plus ` produces U+00AD? RIght now the documentation appears empty. One possibility would be to remove that line from the table and add a sentence like: "Typing ' followed by ` will produce U+00AD, the soft hyphen character."

MakaraSok:
> Since the keyboard uses Shift+Alt+O and Shift+Alt+K, you should add to the welcome.htm and .php files a warning about needing to change these default keys.

I think you refer to the Shift AltGr. Now the warning message is in the two pages.

Thank you for the all the catch.

PR #818, opened 2019-07-18
[basic_kbdmyan] Myanmar (Phonetic order) Basic imported from Win10

MakaraSok:
- all keys in both desktop and touch layouts output the expected character (the location of keys on each layer are closed to those of the Desktop except the Numeric Layer which is placed randomly.)


DavidLRowe:
Since the keyboard uses Shift+Alt+O, Shift+Alt+K and Shift+Alt+M, you should add to the welcome.htm and .php files a warning about needing to change these default keys.

Touch layout, shift layer, first row, rightmost key doesn't produce the character displayed on the key top.

Optional: Touch layout, numeric layer: Would you want to drop the period key and its longpress keys? All those keys are on the numeric layer as keys of their own. Up to you!

MakaraSok:
> Since the keyboard uses Shift+Alt+O, Shift+Alt+K and Shift+Alt+M, you should add to the welcome.htm and .php files a warning about needing to change these default keys.

I do not see them. I found only four layers: Default, Shift, Right Alt and Shift Right alt, none of which is a Shift Alt.

> Touch layout, shift layer, first row, rightmost key doesn't produce the character displayed on the key top.

Thank you for catching that. It's a BKSLASH not a SLASH, that's why it didn't output the right character.

> Optional: Touch layout, numeric layer: Would you want to drop the period key and its longpress keys? All those keys are on the numeric layer as keys of their own. Up to you!

That makes sense. I was thinking of consistency more than plausibility. :)

DavidLRowe:
Maybe it is because my US English version of Windows doesn't distinguish between Alt and AltGr, but Shift plus (*either* LAlt or RAlt) plus (K or M or O) invokes a Keyman hotkey.

PR #819, opened 2019-07-22
[basic_kbdmyan] remove the English alphabet from desktop layout

MakaraSok:
The English alphabet is not out of the desktop layout. This is done in accordance with the decision made when talking with Marc.


DavidLRowe:
I believe that this will need a version number increase. This will mean changes to README.md, HISTORY.md, basic_kbdmyan.kmn (and possibly basic_kbdmyan.kpj and basic_kbdmyan.kps, if the `FollowKeyboardVersion` flag is not used). If any other files (welcome.htm, etc.) include the version number, they would need to be changed as well.

DavidLRowe:
No longer need warning about Shift+Alt+K,O,M in welcome.htm and *.php

welcome.htm has 1.0 version number. Either change to 1.1 or remove the version number.

PR #820, opened 2019-07-22
[nrc_crk_cans] First draft of the Plains Cree syllabics keyboard

eddieantonio:
This is a new keyboard for `crk-Cans` or Plains Cree, written in Canadian Aboriginal Syllabics. There is already a layout for Cree in this repo, namely `fv_plains_cree`, but its design is a bit too clunky for my taste. This is an attempt to make a "build-a-syllable" keyboard for writing syllabics efficiently on touch devices.


keyman-server:

keyman-server:

DavidLRowe:
The version number doesn't match our normal scheme (1.0 or 1.0.0 and up). Normally 1.0 or 1.0.0 is used for the first public release.

We normally ask that CompilerWarningsAsErrors be set True. There don't seem to be any warnings during compilation, so it seems that it would be fine to set it True.

The README.htm file should be renamed to readme.htm. (Filenames on some platforms are case sensitive.) In that file, I think "releast" should be changed to "release".

In order to be included in the release section, we will need a "welcome" subfolder in the "source" folder. That "welcome" subfolder should contain "welcome.htm" with help information about the keyboard. It can refer to *.png images of the keyboard. This is used to generate help information for the keyboard that is available off-line. It should include enough information that a user can learn the functionality of the keyboard. Since this keyboard is taking a different approach, it would be very helpful to include some examples.

There should also be a "help" subfolder in the "source" folder. That "help" subfolder should contain "nrc_crk_cans.php", with help information similar to that in "welcome.htm". This is used to generate help information that is available on-line. The .php file, however, can, if you choose, generate keyboard images from the .js file, if the correct commands are given, so it isn't necessary to include *.png images. But since you have so many layers, perhaps this would not be the clearest way to help people use the keyboard.

@mcdurdin is reluctant to have multiple scripts per keyboard. Would you consider removing the `latin` and `shift` layers from your keyboard? What about `numeric`? Would those characters be used with syllabics in running text? 

If you aren't in a position to supply help information now, you could include this keyboard in the "experimental" section for now, then move it to "release" (and bump the version) at a later date.

DavidLRowe:
Also you can just use `mobile` in the `store(&TARGETS)` statement in Keyman.

eddieantonio:
Hey @DavidLRowe, thanks for the feedback!

> The version number doesn't match our normal scheme (1.0 or 1.0.0 and up). Normally 1.0 or 1.0.0 is used for the first public release.

Will do! I was using CalVer for pre-release versions internally, but I can change this to a conventional version number.

> We normally ask that CompilerWarningsAsErrors be set True. There don't seem to be any warnings during compilation, so it seems that it would be fine to set it True.

👍 

> The README.htm file should be renamed to readme.htm. (Filenames on some platforms are case sensitive.) In that file, I think "releast" should be changed to "release".
>
> In order to be included in the release section, we will need a "welcome" subfolder in the "source" folder. That "welcome" subfolder should contain "welcome.htm" with help information about the keyboard. It can refer to *.png images of the keyboard. This is used to generate help information for the keyboard that is available off-line. It should include enough information that a user can learn the functionality of the keyboard. Since this keyboard is taking a different approach, it would be very helpful to include some examples.
>
> There should also be a "help" subfolder in the "source" folder. That "help" subfolder should contain "nrc_crk_cans.php", with help information similar to that in "welcome.htm". This is used to generate help information that is available on-line. The .php file, however, can, if you choose, generate keyboard images from the .js file, if the correct commands are given, so it isn't necessary to include *.png images. But since you have so many layers, perhaps this would not be the clearest way to help people use the keyboard.

Okay! Any exemplary keyboard I should use as an example to base my documentation on? I was going to base my documentation on [SIL EuroLatin](https://github.com/keymanapp/keyboards/tree/master/release/sil/sil_euro_latin/source)

> @mcdurdin is reluctant to have multiple scripts per keyboard. Would you consider removing the latin and shift layers from your keyboard? What about numeric? Would those characters be used with syllabics in running text?

In informal trials, people got _annoyed_ when they couldn't write a word or two in English, even when writing in syllabics. So I'd really like to have one quick button to switch into a QWERTY keyboard, and a way to input common punctuation and Arabic numerals, then go back to syllabics. I may tweak the numeric layout a bit further, though.

> Also you can just use mobile in the store(&TARGETS) statement in Keyman.

👍

mcdurdin:
> In informal trials, people got _annoyed_ when they couldn't write a word or two in English, even when writing in syllabics. So I'd really like to have one quick button to switch into a QWERTY keyboard, and a way to input common punctuation and Arabic numerals, then go back to syllabics. I may tweak the numeric layout a bit further, though.

Yeah, but that's what the globe key is for. If you have just two keyboards installed (common situation), then the globe key should swap between them. Android's setup here is significantly uglier than iOS's, which does a much better job of switching between input methods. Is there a reason why they can't use the Globe key to switch for quick entry?

Two reasons for not including multiple scripts within a keyboard:
* No predictions for the second script
* Text gets tagged with wrong language (this is starting to become important, I believe, for some apps in the mobile sphere; certainly it is critical on desktop systems)

eddieantonio:
> Is there a reason why they can't use the Globe key to switch for quick entry?

I'd have to do some additional user testing, but I think they just don't know what the globe key _does_. Also, on my device (Android 9.0), pressing the globe key is a hinderance for a few reasons:

 1. Opening the modal dialog closes the keyboard (at least in Textra, my texting app, and a few other apps I've tested), thus meaning I have to reselect the text entry I was using whenever I switch the keyboard.
 2. The modal dialog it brings up occasionally locks up, and tries to download the keyboard index (I think), making it unresponsive to inputs for a second or two.
 3. It requires three touches to get to the Latin QWERTY keyboard: one to press the globe; one to select the alternate language. On my device, the languages are way above where my hands were resting to type in my language, so I have to readjust my grip to select either one of the two of my installed languages; finally, one more press to reselect the text area I was typing in. When I'm done writing my word or so in the Latin script, I must press the globe again, then select the language, having to readjust my grip in the process, and reselect the text input. Contrast this with one keypress to switch to Latin, I type my word, and one keypress to switch back.

A feature request to make this a quicker transition _without_ using the globe key and without adding another script to the existing keyboard is to make it so that a very deliberate swipe on the spacebar key to the left swaps to next installed keyboard. A deliberate swipe to the right swaps to the previous installed keyboard. That way, there is a very easy way to swap between installed languages to prevent breaking the flow of the typist.

> No predictions for the second script

I had thought about this, and I think it's reasonable to switch "prediction contexts" when certain layers are selected (e.g., numerals layer, currency layer). The lexical models can predict based on the current context, like "default", "number", "punctuation", "uppercase", and even switch to a different installed lexical model. This is _not_ a feature I'd propose in 12.0!

> Text gets tagged with wrong language (this is starting to become important, I believe, for some apps in the mobile sphere; certainly it is critical on desktop systems)

I'm not sure I understand this issue. Could you please elaborate?

DavidLRowe:
You might end up with English text identified as crk-Cans when it should be en-CA.

mcdurdin:
Globe key: I completely agree on the Globe key experience on Android. It sucks. So let's fix the Globe key :) But I know that it will be a while before we can do that... There is an issue now, at least!

> You might end up with English text identified as crk-Cans when it should be en-CA.

This can be an issue for apps that do language-specific work such as word processing with spell checkers, word breakers, hyphenation, even rendering. It's a design decision we made early on in the piece to try and head off problems before they arise (in the past on Windows this has been a _huge_ issue). Currently on Android, most apps are still too dumb to care about language, but that is changing as the platform and apps mature.

LornaSIL:
I'm not reviewing this keyboard, but if you haven't already, it might be good to take the word "Keyboard" out of the name. It is unnecessary and adds to the length of the name.

DavidLRowe:
If I understand @LornaSIL correctly, that would mean, in the .kmn file, 
change: `store(&NAME) 'Plains Cree Syllabics Keyboard'` 
to `store(&NAME) 'Plains Cree Syllabics'`

keyman-server:

eddieantonio:
Thanks @LornaSIL and @DavidLRowe; I'll remove the word "keyboard" from the name. I'm noticing that  [fv_plains_cree])(https://github.com/keymanapp/keyboards/blob/master/release/fv/fv_plains_cree/source/fv_plains_cree.kmn) calls itself "ᓀᐦᐃᔭᐍᐏᐣ" (nêhiyawêwin, the endonym of Plains Cree). Is it best to use an endonym, or should I use the English name?

LornaSIL:
I don't have an opinion on whether to use the endonym or English name...except you might ask the people which they prefer to use and then hope that there's a font on their system that will appropriately display the name if it is in Cans.

DavidLRowe:
Are there implications when searching for the keyboard on the Keyman site? If the keyboard were only listed in Cans script, would users already have to have a way to type the script, in which case they wouldn't need the keyboard! Or does the Keyman site also use the language tag and get the language name in English from the tag?

eddieantonio:
> Are there implications when searching for the keyboard on the Keyman site? If the keyboard were only listed in Cans script, would users already have to have a way to type the script, in which case they wouldn't need the keyboard! Or does the Keyman site also use the language tag and get the language name in English from the tag?

@LornaSIL I am interested in this as well. If name it "ᓀᐦᐃᔭᐍᐏᐣ" everywhere, will it still be searchable? Or is there somewhere I can also call it "Plains Cree" and "Syllabics"?

LornaSIL:
I think you search by language on Keyman at the minute. Eventually I think Marc wants to be able to search by script, but I don't even know if that is on the current plan. If someone wants to use a Google search then you should make sure your keyboard_info file has the words in the description that you think people will search by.

keyman-server:

keyman-server:

mcdurdin:
The whole "search for a keyboard" issue raises its ugly head again ;-) So... the name of the keyboard should be what you want to see in the UI of Keyman. For now, including all the endonyms that you could want in the `description` field of the .keyboard_info would be sensible. I think we should plan to add a `keywords` field to .keyboard_info that should be searchable as well ... because language names are fraught!

keyman-server:

eddieantonio:
Finally, this keyboard is ready for review! @DavidLRowe, @LornaSIL, please let me know if I'm handling the welcome/help situation correctly.

LornaSIL:
Everything looks good except it would be nice to document U+202F (nnbsp) in the welcome.htm and help file.

eddieantonio:
@LornaSIL Done! The NNBSP is a tricky one!

LornaSIL:
You can delete your branch if you wish.

PR #822, opened 2019-07-25
[basic_kbdbe] Import of Windows kbd with mobile layout

LornaSIL:
Made mobile layout match the two deprecated ones.


DavidLRowe:
On touch layout, I'm not seeing how to get to the symbol layer.

LornaSIL:
There's a longpress on the shift key on the default key. It's weird, but that's how the original layout had it. I could change it if you feel I should.

DavidLRowe:
I guess it's fine as is.

PR #825, opened 2019-07-26
[basic_kbdolch] Ol Chiki Basic imported from Win10

MakaraSok:


LornaSIL:
The keyboard is also outputting these keys which are not on the OSK: 
```
`-=qwrt\!@#$%^&*()+QWERTYIOP| ASDFGHJKL" XVNM?
```
You should either write a rule that blocks them from being output or regenerate the OSK (and graphics) to demonstrate they are available.

The mobile layout works and has all the characters that are currently listed on the OSK. If you decide to add the ones above, we may need to revisit the mobile layout. 

Personally, on the mobile layout I would find it easier to find matched pairs if they are on a paired key. So, I would rather [{( would all be on something like ```(``` rather than ```@```, and the same for the closing quotes. However, I won't let that stop us from approving the PR. I'd just like you to think about ease of use for future keyboards.
So, the only thing that needs doing is deciding how to handle the OSK, whether to write a rule to remove those keys or regenerate the OSK and graphics.

MakaraSok:
Thank you for catching the unexpected output characters. A rule has been written to nullify those. OSK and touch layout have been updated accordingly.

PR #827, opened 2019-07-26
[basic_kbdfi1] Finnish-Swedish with Sami Basic: Import from Win10 and include mobile layout

LornaSIL:
Deprecates 3 keyboards.


DavidLRowe:
Documentation in welcome.htm
- for acute accent it reads: `'a will produce á`, but that first character in the documentation is a U+0027, which is also on the keyboard, whereas the acute deadkey is the U+00B4 on the key to the left of the backspace. So the documentation would be clearer if that first character could be changed from U+0027 to U+00B4
- for acute accent includes `ghjGHJ` which don't combine with acute
- for acute accent doesn't include all the characters that do combine with acute: `áćéíĺńóŕśúẃýźǿǽǻ` (and upper case)
- for dieresis doesn't include w and W that do combine with dieresis

Touch layout includes an apparently empty `ctrl` layer. Is this intentional?

Touch layout doesn't seem to include a tilde deadkey.

LornaSIL:
Not all of the composed characters are available on longpress keys, but I did add the deadkey for tilde so they should now all be available with deadkeys.

DavidLRowe:
desktop layout, shift layer, 1st row, 1st key on left: shows `}` but produces `½`.

touch layout, default layer, longpress `w` labeled w circumflex, but prduces w dieresis.

touch layout, shift layer, longpress `W` labeled W circumflex, but produces W dieresis.

PR #831, opened 2019-07-29
[basic_kbdcr] Croatian/Slovenian Basic import from Win10

LornaSIL:
This keyboard deprecates 3 others.

I also did a fix of the header on the basic_kbdbe.php file altho the keyboards are unrelated (no version bump needed)


DavidLRowe:
Should Bosnian be bs-Latn or bs? The subtag list has suppress Latn script for bs. (Minor point: the tags are in a different order in .kps and .keyboard_info files.)

PR #835, opened 2019-07-30
[basic_kbdfar] Persian (Standard) Basic Imported from Win10

MakaraSok:

What is the data-states of "Shift Ctrl" in the  `.php`? I don't have the right set up to view the `.php`.
Please double check the icon, usually I stick with what's shown in Win10, but this one is tricky because it's in Arabic and I don't know which character that is due to the font variation.


LornaSIL:
* In the .kps file the BCP-47 tag should just be "fa" (suppress script). It is correct in keyboard_info
* I don't see a way to access ZWJ, ZWNJ, LTR, RTL marks on the mobile layout. If it's not there, could you add those? (U+200C is on the shift space, but there is no indication. Maybe put zwnj as text on shift+space.)

The help file is displaying correctly.

MakaraSok:
I have added the RTL and LTR to the touch layout, but I was not able to find a way to put the label to tell which is which. Please let me know how to do that.

LornaSIL:
I just go to the touch layout and on the keycap I type in zwj and zwnj. It may not be the most beautiful, but at least it shows what it is. If I use capitals it is too big. If I use small cap characters, fonts don't always support the small cap codepoints.
image

MakaraSok:
@LornaSIL The abbreviation for LTR and RTL have been updated. Please advise me on how to label the key caps on touch layout for these two keys. `U_200E` and `U_200F` are used instead of the key mapping "Ctrl + Shift" on num key 3 and 4 as there is no ctrl-shift in the Modifier drop-down menu.

LornaSIL:
@MakaraSok do you want me to merge this, or are we waiting for a bug fix? Or you can merge it.

MakaraSok:
The bug was referred to automatically because it happened when I was testing the keyboard on Chrome simulator. It does not affect the keyboard performance in anyway though, but do you see anything considerably weird when looking into this keyboard?

LornaSIL:
When I test using my Android phone (not the Chrome simulator) the longpress displays properly and behaves properly after selecting a longpress key.

PR #836, opened 2019-07-30
[nlci_ipa] IPA [NLCI]

binilasanki:


keyman-server:

DavidLRowe:
Thanks for this keyboard submission. Here are a few comments:

The language tag for IPA is odd. We have to use "und" (undetermined) for the language, and (contrary to what is usually the case) the .keyboard_info file needs `-fonipa` and the .kps (package source) needs `-Latn`. So, "und-fonipa" in .keyboard_info and "und-Latn" in .kps.

In the welcome.htm (and the .php file), there's a "please contact:" with no contact information.

The touch layout file is currently named `nlci_ipa-layout.js` but the preferred naming convention would be `nlci_ipa.keyman-touch-layout`. But regardless of which name you use, it won't be included unless you link it to the desktop keyboard by putting a statement like the following into the .kmn file:
```
store(&LAYOUTFILE) 'nlci_ipa.keyman-touch-layout'
```

Currently you have included Charis SIL fonts in the source folder. This will work, but since we have included the fonts in the `release/shared` folder (at `release/shared/fonts/sil/charis_sil`), we would prefer that you remove these fonts from the `nlci_ipa/source` folder and reference the above location for including the fonts. 

Along with those font (.ttf) files, we would suggest including the following files from the same folder:
- Charis_SIL_README.txt
- FONTLOG.txt
- OFL.txt
- OFL-FAQ.txt

Currently the `nlci_ipa.ico` file is a blank white icon. Were you planning on putting some symbol there?

The copyright dates in the .kmn file don't agree with those in the README.md file.

In the welcome.htm document, there are some PUA (Private Use Area) characters for the upstep and downstep characters U+F19C and U+F19D. Since these characters are now in Unicode, version 5.000 of the Charis SIL font shows them with a black background to alert you to the fact that you should use the real Unicode values. The keyboard produces the correct characters: U+A71B and U+A71C. You'll probably also need to correct the .php help file.

There were a few key combinations in the welcome.htm file that didn't seem to work for me. This might be because they weren't implemented and the so the documentation needs to change, or it might be a bug in the .kmn file (or it could be my mistake!): 
```
!= !^ t> t= 0= 0> s> @@
```

Let us know when this is ready to review again. Thanks!

keyman-server:

DavidLRowe:
Thanks for your work on this keyboard. There are a few issues.

The .kps is in the wrong folder. It needs to be in the source/ folder. (Moving the file will invalidate all the paths to files currently in the .kps file, so they will need to be removed and reinserted to get the correct path.)

The .kps file needs language tag of "und-Latn". (NB: This is an exceptional case in that the language tag in the .kps file and the language tag in the .keyboard_info file are different.)

It looks as though you have decided not to include the Charis fonts in this keyboard package (since you have removed them from the repository, but have not added references to the files in release/shared/fonts/sil/charis_sil). It is perfectly fine to omit the fonts from the keyboard package.

You have added contact information to the welcome.htm file, but not to the nlci_ipa.php help file.

The .kmn file has &TARGETS set to 'any', but there is no nlci_ipa.keyman-touch-layout file, and the README.md only lists Windows, Mac, Web and Linux. It seems that the .kmn file needs to change. Also note that mnemonic layouts are only supported on Windows and Mac (not sure about Linux) and only partially for web use. See https://help.keyman.com/developer/language/reference/mnemoniclayout for details.

In the welcome.htm document, the characters for the upstep and downstep characters are actually U+F19C and U+F19D (in the PUA -- Private Use Area), even though the chart lists the correct characters: U+A71B and U+A71C. (Since these characters are now in Unicode, version 5.000 of the Charis SIL font shows them with a black background to alert you to the fact that you should use the real Unicode values.) The ncli_ipa.php help file also uses the PUA characters and these should also be changed to use the correct characters.

Once the .kps file is corrected, I can install and review the keyboard.

DavidLRowe:
In welcome.htm, there are two places where the title text includes the old PUA value. The title text displays when you hover your mouse over the IPA symbol (up or down arrow), then it shows "USV:F19C" or "USV:F19D" which are incorrect values. 
```
<a href="#ipa517" title="IPA:517, USV:F19D (PUA)">&#42780;</a>
```
should change to: 
```
<a href="#ipa517" title="IPA:517, USV:A71C">&#42780;</a>
```
and
```
<a href="#ipa518" title="IPA:518, USV:F19C (PUA)">&#42779</a>
```
should change to:
```
<a href="#ipa518" title="IPA:518, USV:A71B">&#42779</a>
```



In the .kmn file, the `store(&TARGETS)` statement needs `linux` added in order to match the README.md. (This can be done by editing the .kmn file directly or by ticking the appropriate box in Keyman Developer.)

The BCP 47 code in the .kps file is correct ("und-Latn"), but the code in the .keyboard_info file needs to change to "und-fonipa". This is one of the rare cases where the codes in these two places are different.

In Keyman Developer, for the .kmn file, "Details" tab, under the "Features" section, click "Add" and add "Desktop On Screen Keyboard". At that point you can fill the OSK by using the "Fill from layout" button.

You'll need to compile the keyboard to get the files into the "nlci_ipa/build" directory.

Because the .kps file was moved, some of the paths are now incorrect. In Keyman Developer, open the package source (.kps file). On the Files tab:
- Click on "Remove" for each of the five existing entries to remove them, because they are pointing to the wrong location. (NB: This will delete the BCP 47 code and the "Readme file" link, so these will need to be redone.)
- Click on "Add", navigate to the nlci_ipa\build folder and select all three files (.js, .kmx, .kvk)
- Click on "Add", navigate to the nlci_ipa\source folder and select readme.htm and welcome.htm.

Now the BCP 47 code and the "Readme file" link need to be redone. On the Keyboards tab:
- Click on Add at the bottom to add the "und-Latn" BCP 47 code

On the Details tab:
- Opposite the "Readme file:" label, select "readme.htm" from the list

DavidLRowe:
The BCP 47 code in the .keyboard_info file needs to change from "und-Latn" to "und-fonipa". (The BCP 47 code in the .kps file is correct as "und-Latn". This is one of the rare cases where the codes in these two places are different.)

You correctly changed the USV values in the title text in the welcome.htm file, but we also need to remove the "(PUA)" from both those strings. ("PUA" is the "Private Use Area", the U+E000 to U+F8FF range where these two characters were initially assigned by SIL for its fonts until they were assigned an official Unicode value.)

I see that, when using the "Fill from layout" to create the on screen keyboard, you first need to click the "Auto-fill underlying layout" (at the bottom of the dialog box) in order to get all the characters that haven't been changed.

The nlci_ipa.kps file seems to be empty. You'll need to use "Packaging" in Keyman Developer to access the package options: "Files", "Keyboards", "Details". Then "Compile" to compile the package and make sure it works.

DavidLRowe:
Unfortunately, the .kps file was added in the wrong place. It needs to be in the nlci_ipa/source folder (not in the nlci_ipa folder).

LornaSIL:
Closing since I submitted a new PR.

PR #840, opened 2019-08-01
[haroi] Haroi Keyboard Adapted from VNI.

MakaraSok:
This keyboard was built for an ethnic minority language in Vietnam. It should be able to output all characters needed for the language as well as Vietnamese language.


LornaSIL:
Looks really good...and complicated!

* In the .kpj can you set this to true? ```<CheckFilenameConventions>False</CheckFilenameConventions>```

* Since you say it also supports Vietnamese would you want to add ```vi``` to the list of languages it supports? You would need to add it to keyboard_info and .kps if you wanted to do that, but you don't have to.

MakaraSok:
Thank you @LornaSIL for the review. :)

Updates have been made, ptal.

LornaSIL:
I'm really sorry Makara! It should be ```vi``` and not ```vi-Latn``` because https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry says the script should be suppressed for vi. Change needed for both keyboard_info and for .kps.

PR #841, opened 2019-08-01
[aksarabali_panlex] implement alt + dependent vowel combos; update docs

kamholz:
Adds support for typing ulu plus suku after a consonant. This is an illicit combination of vowel marks, but it is actually attested in palm leaf manuscripts as a way to indicate that an aksara was a mistake and should be ignored. So the keyboard needs to support it!


keyman-server:

PR #843, opened 2019-08-05
[basic_kbdinben] Bengali Basic Win10 keyboard and deprecations

LornaSIL:
I pulled in the welcome.htm and help file from the bengali mobile layout, but I did have to make a few adjustments to the Ctrl and Right-Alt chars.


DavidLRowe:
Touch layout, shift layer: key with period produces "{". Seems as though either the keycap needs to change, or what the key produces. (Desktop has "{" on the shifted ".")
numeric layer: "(" has longpress "{", but ")" doesn't have longpress for "}".

I didn't check every combination, but what I did looks fine.

PR #844, opened 2019-08-09
[balochi_phonetic] Balochi phonetic revisions

rmlockwood:
updates and fixes


keyman-server:

LornaSIL:
The keyboard_info file needs an ending double quote and a comma at the end of the description line:
```
    "description": "This keyboard is laid out phonetically (for example, b=beh).",
```
In the future if you want to test the keyboard_info file ahead of time you can test it here: https://jsonformatter.curiousconcept.com/

Your keyboard rules are for CTRL+SHIFT (not CTRL) so the OSK needs modifying for the keys to be on the CTRL+SHIFT layer, not just CTRL. You'll need to change the welcome file to reflect CTRL+SHIFT (or else change the keyboard to use CTRL only)

While you are at it, it would be nice to type in =/!@#$%&* on the OSK and reproduce the graphics that are used in the welcome file.

The SHIFT Y produces a Y. If you want to prevent that, you could add a rule to your .kmn file.
```
+ "Y" > nul
```

The tablet keyboard has keys for Ctrl (2 keys on right and left), Alt, and AltCtrl on both the default and shift layer. You might want to remove the text on the keycaps so they don't look like they are intended to do anything. Or delete the keys.

If you want the help file to display all of the layouts, you could change your current balochi_phonetic.php file from:
```
<h2>Keyboard Layout</h2>
<div id='osk'>
</div>
```
to
```
<div id='osk' data-states='default shift shift-ctrl'>
</div>

</div>

</div>

<p>For Android and iOS, you may use long press key combinations.</p>
```

rmlockwood:
I’ve done these changes. I also added =/!@#$%&* to the tablet and phone keyboards.

 

I’ve noticed that built-in Android Arabic script keyboards have a nice symbol for zwnj and zwj. A line with arrows going either out or in. Is there a Unicode codepoint for this? If not, how can I get an image to show on my mobile keyboards. I think these symbols would be much better than ‘zwj’ and ‘zwnj’. Nobody will know what they mean.

 

Thanks,

Ron

keyman-server:

rmlockwood:
I don’t have it handy, but if you install an Arabic keyboard on your Android phone you’ll see it.

PR #845, opened 2019-08-09
[balochi_inpage] Change keyboard_info description

rmlockwood:
I didn't think it necessary to create a new version just for changing the description.


keyman-server:

PR #847, opened 2019-08-09
[balochi_urdu] Added the keyboard_info description. Removed other uneeded stuff from…

rmlockwood:
… keyboard_info. It should now support all platforms. Linux wasn't listed before.


keyman-server:

PR #849, opened 2019-08-10
[basic_kbdphags]

MakaraSok:

All spaces and joiners are included in touch layout as longpress on the spacebar. One character does not have a unicode point, [U+FE00], on Shift + BKSLASH. It is placed in the numeric layer in the touch layout, but the visual outlook is different from that of what is shown on the desktop layout.


LornaSIL:
* I would like a different icon if possible? Perhaps use one of the Phags-pa characters instead of EN?

* On the touch layout:
  * I don't see equal (=)
  * Regarding the FE00, I guess you could type "VSI" on the touch layout keycap? I don't have any other good ideas.

* I think it would be useful for you to include the text you have in the welcome.htm file in the .php file as well (everything between "Keyboard Layout" and "Desktop Layout").

MakaraSok:
> * I would like a different icon if possible? Perhaps use one of the Phags-pa characters instead of EN?

I cannot find an existing one. Usually it's up in the system keyboard if one installs it on Windows, but not for this particular keyboard.

MakaraSok:
TODO: 

- [ ] include the text in the welcome.htm file in the .php file (everything between "Keyboard Layout" and "Desktop Layout").

LornaSIL:
I've sent you an .ico via email as I'm not allowed to upload that type here.

PR #852, opened 2019-08-12
[basic_kbdughr1] Uyghur Basic import and deprecations

LornaSIL:


DavidLRowe:
When using the keyboard in a web browser, or when displaying the .php file, keys on the shift layer which are omitted from the OSK (because they are blocked from creating output) display the key from the en-US keyboard. It seems from https://github.com/keymanapp/keyman/issues/1460 that the solution is to assign a no-break space to the keycap.

Other than that problem, everything looks fine with this keyboard. I'll approve it as is, and if you want to change the OSK, dismiss my review.

PR #853, opened 2019-08-12
[basic_kbda1] Import Arabic (101) Basic and deprecations

LornaSIL:


DavidLRowe:
The image for welcome.htm shows ZWJ, etc on RightAlt layer, but they seem to be on ShiftCtrl layer in .kmn file. OSK has them on ShiftCtrl layer, but with no visible indication of what the keys produce.

PR #855, opened 2019-08-13
[basic_kbda3] Arabic (102) AZERTY Basic import and deprecations

LornaSIL:


DavidLRowe:
On desktop layout, Shift+S produces "S" whereas on OSK this show as as a blank key. It seems as though either the keyboard or the OSK should change.

PR #858, opened 2019-08-14
[basic_kbdrum] Russian - Mnemonic Basic import from Win10

MakaraSok:

The .php may have to have a style sheet code block for `<kbd>`.


LornaSIL:
The keyboard looks good!
* For the welcome.htm and basic_kbdrum.php file please change ```This keyboard use five deakeys``` to ```This keyboard uses five deadkeys```.
* For the touch layout, numeric layer, the shift key needs "Next layer" changed to "shift"

Everything else looks good.

LornaSIL:
Please also change the .php file. And "deakeys" to "deadkeys" in the .htm file as well.

PR #859, opened 2019-08-19
[basic_kbdnso] Setswana Basic imported from Win10

MakaraSok:

_The .php may have to have a style sheet code block for `<kbd>` and `table`._


LornaSIL:
* This keyboard will need that statement about resetting the hotkey for RALT+SHIFT+O. 
* The only character missing from the touch layout is U+00A7 §. You could add it to the longpress "S" key.

Otherwise, it looks great.

PR #864, opened 2019-08-21
[basic_kbdtam99] Tamil 99 Basic keyboard imported from Win10

MakaraSok:
- all keys in both desktop and touch layouts output the expected characters


LornaSIL:
Can you write a rule to prevent [SHIFT K_I] from producing anything?

It looks like the square brackets are missing from the touch layout. [ and ]

Everything else looks very good.

PR #865, opened 2019-08-21
Fix typing mistakes in welcome and help pages

MakaraSok:


LornaSIL:
Since this affects the welcome.htm file which is part of a keyboard package we have to bump the version number in the keyboard (or it will not rebuild).

We can either just fix the .php file right now or we'll need to bump the version number in each keyboard.

#866

MakaraSok:
There is nothing substantial to bump the version number, but if that's the only way. We can do it that way. I'm sorry for my tardiness.

DavidLRowe:
@MakaraSok Any action planned for this pull request (for basic_kbdrum)?

LornaSIL:
Documented in #1081

PR #868, opened 2019-08-22
[basic_kbdintel] Import Telegu Basic from Win10 and deprecations

LornaSIL:


DavidLRowe:
touch layout, shift layer, key marked "\" produces "|". I wonder if the keycap should be "|".

PR #875, opened 2019-08-29
[Enga] Initial release

mcdurdin:
First version of Enga keyboard for Enga language in PNG. Note that this keyboard includes the ē ŋ Ē Ŋ characters; it appears that some orthographies do not use ē Ē.


DavidLRowe:
Copyright statement in LICENSE.md (and .kps) needs to match that in .kmn and README.md.

Is there a reason that CompilerWarningsAsErrors is set to False?

Is the style of upper case eng used in the engaU_S.png screenshot (used by welcome.htm and the .php help file) the preferred style for users of this language? (Or maybe it doesn't matter?)

.kvks has entries for RC and SRC even though these keys aren't included in the .kmn file. Is this intentional?

Touch layout, shift layer, "." key produces ">" instead of "."

DavidLRowe:
For BCP 47 code (in .keyboard_info and .kps files), should it be just "enq-Latn" instead of "enq-Latn-PG"?

mcdurdin:
> - [x] Copyright statement in LICENSE.md (and .kps) needs to match that in .kmn and README.md.

Done.

> - [x] Is there a reason that CompilerWarningsAsErrors is set to False?

Accidental!

> - [x] Is the style of upper case eng used in the engaU_S.png screenshot (used by welcome.htm and the .php help file) the preferred style for users of this language? (Or maybe it doesn't matter?)

Good catch. Fixed by switching to Calibri font (which has the N-with-tail rather than big n style).

> - [x] .kvks has entries for RC and SRC even though these keys aren't included in the .kmn file. Is this intentional?

Removed.

> - [x] Touch layout, shift layer, "." key produces ">" instead of "."

Fixed.

> - [x] For BCP 47 code (in .keyboard_info and .kps files), should it be just "enq-Latn" instead of "enq-Latn-PG"?

Yes. Fixed.

PR #879, opened 2019-09-06
[sil_greek_politonic] change tonos to oxia characters

LornaSIL:
Christopher Samuel reported that our keyboard was not polytonic compliant.


DavidLRowe:
It seems as though the README.md should include "web", since the .kmn has it and the .js is now included in the .kps. (Only minimal review of character changes.)

PR #880, opened 2019-09-08
[pid_piaroa] New keyboard for Piaroa (ISO 639-3: pid)

eddieantonio:
This is a keyboard for the Piaroa language, a low-resource language spoken around the border of Colombia and Venezuela.

These keyboard augments a regular regular Latin keyboard with the characters ä/ö and nasalized vowels. These are accessed by using the <kbd>AltGr</kbd> key.

There are a few things I'm unsure about:

 - [x] help PHP file: what should I put here that's not already in the welcome file?
 - [x] Bilingual documentation: I've translated `welcome.htm` to Spanish as `welcome-es.htm`, however it does not appear in Keyman for macOS, even when my system is set in the `es` locale.

Any help with this would be appreciated.

**EDIT**:

Should this be in the namespace of `release/eddieantonio`? I doubt this will be the last keyboard I'll submit ;)


keyman-server:

DavidLRowe:
The .php help file is what is used on the keyman.com site. The welcome.htm file is what is included in the package (and so is available offline). They often have the same content. Eventually it would be nice to be able to have a common source for both.

Often in the welcome.htm file, people will reference .png graphic images of the keyboard layers (default, shift, etc), which can be generated from the On-Screen tab of Keyman Developer. In the .php file it is possible to reference the on-screen keyboard information directly to display the keyboard layers. (See \release\sil\sil_arabic_phonetic in the keyboards repository for an example.)

eddieantonio:
> The .php help file is what is used on the keyman.com site. The welcome.htm file is what is included in the package (and so is available offline). They often have the same content. Eventually it would be nice to be able to have a common source for both.

Okay! I copy-pasted the great big table of special characters from welcome.htm.

> Often in the welcome.htm file, people will reference .png graphic images of the keyboard layers (default, shift, etc), which can be generated from the On-Screen tab of Keyman Developer. In the .php file it is possible to reference the on-screen keyboard information directly to display the keyboard layers. (See \release\sil\sil_arabic_phonetic in the keyboards repository for an example.)

Thanks! I don't think this will be useful for this keyboard, as the table in welcome.htm and the .php file list all key sequences that differentiate this keyboard from a standard QWERTY keyboard.

---

I'm concerned about bilingual documentation; is there an established way for me to provide bilingual documentation for the .php file?

LornaSIL:
sil_cameroon_qwerty has an example of switching between French and English. Xinaliq just does it by alternating languages in each paragraph. The rac_ keyboards all use English followed by Arabic texts. Those are the only examples I can think of off the top of my head.

DavidLRowe:
.gitattributes and .gitignore should not be included in the pull request

eddieantonio:
Is `welcome-{BCP47}.htm` not supported anymore? @mcdurdin (for when he gets back).

LornaSIL:
>Is welcome-{BCP47}.htm not supported anymore? @mcdurdin (for when he gets back).

I don't remember it ever being used. As far as I know we don't have any welcome-{BCP47}.htm files in legacy or in release.

LornaSIL:
@eddieantonio I would prefer your keyboard to go under "p" and call the folder and files to be ```pid_piaroa.``` Including your name in the file structure seems unnecessary since your name will be listed as author. It makes for an unnecessarily long url.

Alternatively, if you think you'll be submitting quite a few keyboards under your name as copyright holder I'd be okay with you creating a folder called ```eas``` and then putting the keyboard in that folder as ```eas_pid_piaroa``` (the same way you did the keyboard for nrc).

eddieantonio:
Okay! I've moved bilingual documentation to both welcome.html and the PHP help file. As well, I've moved the keyboard to `p/pid_piaroa`. I have also deleted extraneous files.

eddieantonio:
@LornaSIL @DavidLRowe what is required for this keyboard?

LornaSIL:
It's definitely a problem on Windows. If it's not a problem elsewhere, feel free to add  ```...that are set by default (on Windows) as "hotkeys" when...```

LornaSIL:
One more typo introduced in readme...rapdio

eddieantonio:
Thanks, @LornaSIL! Done!

mcdurdin:
> It's definitely a problem on Windows. If it's not a problem elsewhere, feel free to add `...that are set by default (on Windows) as "hotkeys" when...`

See https://github.com/keymanapp/keyman/issues/1044 for future resolution to this.

PR #881, opened 2019-09-09
[nlci_kannada_winscript] keyboard info fix

ramesh-k:
This Pull Request is to rectify wrong language name put in description section of keyboard info file.

CC: @binilasanki


keyman-server:

PR #882, opened 2019-09-11
[basic_kbdgeo] Move Georgian Basic to release

LornaSIL:
Deprecate one keyboard.
Mobile matches existing mobile.


DavidLRowe:
Touch layout, ctrl layer, "d" key doesn't produce character. It seems that the layer Modifier is set to rightalt-shift rather than rightalt.

.kps file references source/welcome.htm, but it needs to be source/welcome/welcome.htm

PR #887, opened 2019-09-16
[basic_kbdindev] Devanagari - INSCRIPT Basic import

LornaSIL:
* The mobile layout matches the "legacy" mobile layout.
* Deprecates 3 other keyboards


DavidLRowe:
.php file: div for osk-tablet has `rightalt` and `rightalt-shift`, but the keyboard images are blank. I think they should be `ralt` and `shift-ralt`

PR #888, opened 2019-09-16
[basic_kbdinhin] Hindi Traditional Basic import

LornaSIL:
* Mobile layout closely matches legacy mobile layout
* Deprecates 3 keyboards.


DavidLRowe:
Touch layout, numeric layer, blank keys (ASDFXCVBNM) produce output (apparently from default layer).

PR #889, opened 2019-09-17
[basic_kbdnepr] Nepali Basic import

LornaSIL:
New mobile layout

Deprecate one keyboard.


DavidLRowe:
Desktop layout, Ctrl+Shift layer: I couldn't get Ctrl+Shift+0 to produce anything. Is this a Windows hotkey?

Touch layout, default layer, 3rd row, 9th key from left (just to the right of the "m" key): the two longpress  keys on that key seem to be switched.  

Other than that all looks fine.

PR #891, opened 2019-09-18
[basic_kbdinori] Oriya Basic import

LornaSIL:
Mobile layout is new

Deprecate one keyboard.


DavidLRowe:
Everything looks fine, except the shift-ctrl layer on the touch layout. It only has one key that already appears on the numeric layer. I don't see any way to get to the shift-ctrl layer, nor is it referenced in the .php help file. It seems as though it should be removed.

PR #894, opened 2019-09-20
[piruvik] Change piruvik keyboards to use MIT license with permission…

LornaSIL:
… of author

Version bump required on all 3 keyboards.


DavidLRowe:
Comments based on manual inspection of changed files:

- inuktitut_latin/HISTORY.md needs version changed to "1.3"
- inuktitut_pirurvik.kpj needs version changed to "1.3"
- inuktitut_naqittaut.kpj doesn't have file version in the same places the other .kpj files do (that is, version for .kps file and in the Details section is empty). Is that a problem?
- Does inuktitut_naqittaut.kps need version number change? Currently has 1.4, but also has FollowKeyboardVersion. But you changed version numbers in inuktitut_latin.kps (which also has FollowKeyboardVersion set.

PR #898, opened 2019-09-25
Added regular layout to onscreen keyboard

ind-nt:
Slight fix for on-screen keyboard to ensure regular keys show up.


keyman-server:

LornaSIL:
I'm sorry to have to tell you this, but any enhancement to the keyboard package requires a version bump to the keyboard.

This means changing the version in the HISTORY.md, README.md, welcome.htm, readme.htm, php and of course the .kmn file.
* This is why I usually remove the version number from README.md, welcome.htm, readme.htm and php file and just keep it in the HISTORY.md and .kmn file. It reduces the places it needs changing.

ind-nt:
Thanks again Lorna. I hope this checking doesn't bother you too much.

I've removed the versioning from the other files, and updated the History.md and indigenous_nt.kmn files.

PR #903, opened 2019-10-02
[basic_kbdinpun] Punjabi Basic Win10 kbd import

LornaSIL:
* Mobile layout matches existing layout
* Deprecated 2 keyboards


DavidLRowe:
Touch layout, numeric layer: many unlabeled keys produce output.

PR #904, opened 2019-10-02
[basic_kbdinkan] Kannada Basic Win10 import

LornaSIL:
* Mobile layout matches old one
* Deprecated 2 keyboards.


DavidLRowe:
Touch layout, numeric layer: many unlabeled keys produce output.

Touch layout, shift-ctrl layer: this layer seems superfluous (and there doesn't seem to be any way to get to it).

PR #907, opened 2019-10-07
NLCI Oriya keyboard

binilasanki:
NLCI Oriya keyboard is ready for review.


keyman-server:

LornaSIL:
Copyright dates are inconsistent. Some places have 2000-2019 and others say 2002-2019. It would be good to make those consistent. The HISTORY.md file would indicate that 2002-2019 is correct.

In the .kmn and .kps: instead of the keyboard being named ```Oriya WinScript Keyboard [NLCI]``` I would like it to be ```Oriya WinScript (NLCI)```. That removes the unnecessary word "Keyboard" and we've been using parentheses rather than square brackets.

In the welcome.htm file you have "Vowel Sign Elongation". I wasn't sure how to type the first character. Could you explain more or give an example?

Otherwise, the keyboard seems to be in good shape.

DavidLRowe:
Thanks for those corrections. 

Unfortunately, the on-screen keyboard data seems to have been corrupted. Using Keyman Developer, you can go to the On-Screen tab and use the "Fill from layout" button to update the on-screen keyboard data.

According to the welcome.htm document, typing `rr` should produce U+0B5C, but I'm not seeing that. I don't know if there's a problem with the keyboard or the documentation.

You added a clarifying example for the Vowel Sign Elongation section. Could you add that to the .php help file as well.

Everything else looks good.

PR #908, opened 2019-10-07
Gujarati Winscript Keyboard [NLCI]

binilasanki:
Gujarati Winscript Keyboard is ready for review


keyman-server:

LornaSIL:
It would be great if the keyboard could be named ```Gujarati WinScript (NLCI)``` instead of ```NLCI Gujarati WinScript```. That way the Gujarati is the focus rather than the organization. You can leave all the file folders and names as is, but in the .kmn, .kps, README.md, HISTORY.md, readme.htm, welcome.htm etc it would need to change to ```Gujarati WinScript (NLCI)```. 

* In the .kmn file
   * Please do not select the target of "any". You can have Windows, macosx, linux, web selected.
   * Select the On-Screen tab and then select "fill from layout". Some of the characters were not showing up on the On-Screen keyboard.

* In the welcome.htm file, it is not clear to me how to type the first character in the sequence under "Vowel elongation". That character does not appear to be on the default or shift layers. Could you add how to get that first character?

* In the .kps file, on the details tab, please select the "Readme file" to readme.htm.
* In the .kps file you have the wrong place for the compiled files. You should remove the .js. kvk and kmx files from the .kps and then add them back in. The proper path would be ```\nlci_gujarati_winscript\build\nlci_gujarati_winscript.kmx```

* Please delete from the Pull Request the compiled files that are in the source folder: .kmp and .kmx

* The copyright in LICENSE.md and README.md is "2002-2019 NLCI". In the .kmn it is "2000-2019 NLCI, Bangalore". In the .kps it is "NLCI 2000-2018". Based on the HISTORY.md I believe "2002-2019 NLCI" is correct and you should change the copyright in the .kmn and .kps to match that.


* Please double-check the language(s) this keyboard is intended to support. In the keyboard_info file you list ```"languages": ["hi","mai"],```. In the .kps you list Oriya. I'm thinking none of those are correct and the language should be "guj" maybe?


That's probably enough feedback for now. I'm glad to see this keyboard being made available!

DavidLRowe:
Just a couple of minor things left: 
- The copyright in the .kps (Package Details tab of Packaging) still has 2000 (rather than 2002 as in other places)
- You added a note about vowel elongation to the welcome.htm file. Could you put that in the .php help file as well?

Thanks!

PR #909, opened 2019-10-08
Telugu Winscript Keyboard

binilasanki:
Telugu Winscript keyboard is ready for review.


DavidLRowe:
The copyright date in the .kmn file has 2000 (rather than 2002 as in other places). The copyright notice should be the same in all places.

In the welcome.htm (and also in the .php help file), the documentation says that `=z` should produce U+0C36, whereas it seems to produce U+0C34. Could you check to see what needs to change?

Other than those items, the keyboard looks to be in good shape.

keyman-server:

DavidLRowe:
I think everything for the keyboard itself is fine. The package looks good, except:
- on the "Files" tab, you can remove the "nlci_telugu_winscript.ico" file (since that has already been compiled into the .kmx file)
- copyright date on "Details" tab needs to be "2002-2019" (rather than "2000-2019") to match date range used in other places

PR #912, opened 2019-10-14
[ntl_onekey] add documentation for NTL OneKey keyboard

ericjding:
This submission is intended to bring the documentation of NTL OneKey up to par for moving it from the experimental to the release repository. It includes basic documentation for online help, readme.htm for the Install dialog, and welcome pages (inc. simplified Chinese translation) as well.


keyman-server:

ericjding:
@LornaSIL or @DavidLRowe could you review this pull request for me? Thanks!

DavidLRowe:
@LornaSIL is out of the office, so I'll see what I can do. 

I've asked about how to move this keyboard to `release` (from `experimental`): whether that should be done in this pull request or a separate one. 

It's preferable to have the version number specified as '1.4' (rather than '1.004'), that is, with no leading zeroes.

The copyright dates in the LICENSE.md file should match those used elsewhere (2015-2019).

We need to have an On Screen Keyboard. (I know you're more concerned with touch layout than with desktop, but a frequent request for mobile users is to be able to type on an external keyboard.) In Keyman Developer, for the .kmn file, on the Details tab, scroll down to the "Features" heading at the bottom, select Add, the select "Desktop On Screen Keyboard". Using the "Fill from layout" button should bring in the keys to fill the OSK.

The following changes are not absolutely required, but will bring this keyboard up to current standards:
- The preferred name for the touch layout file is ntl_onekey.keyman-touch-layout (rather than ntl_onekey-layout.js). This change would involve changing the name of the file as well as where it is referenced in the .kmn file (the &Layoutfile store).
- The &Ethnologuecode store in the .kmn file is deprecated. You could remove the line or comment it out.
- I don't see any reason you can't specify 'any' for &Targets (rather than the list you currently have). This would allow use of the keyboard on Linux and web. You'll need to add Linux to the list in README.md. (Web is already there.)
- The ntl_onekey.bmp file doesn't seem to be used and should be removed from the source folder. (The .kmn file references the .ico file. The .bmp file is not necessary.)
- The license information is not absolutely needed in the readme.htm file. If you decide to keep it, however, please change the copyright dates to match those used elsewhere (2015-2019). The same is true for the welcome.htm file.

I'll report here what I find out about moving the keyboard from `experimental` to `release`.

ericjding:
@DavidLRowe thanks! I don't use git often enough these days to remember how to do what, so I've been breaking/fixing things along those lines as we go. I've now submitted what I think should be all the changes you've suggested, though now it says TeamCity build failed. Not sure what I might've broken, but I'll let you figure it out and let me know. :-)

DavidLRowe:
The build failed because submissions from outside our team have to be manually launched.

Because you're wanting to move this keyboard to the `release` section (at least I think that's your intention), you'll need to move the keyboard files. I asked about this and @mcdurdin says: "...commit the files to git, then move them, then commit the move, and finally push"

You've already made the changes and committed them. I think you need to move the entire `ntl_onekey` folder out of `experimental/n` to `release/n`, then make a new commit with those changes. Once those changes are in the PR, I can launch the build and do a final check.

Thanks for all your work!

ericjding:
@DavidLRowe ok, I moved ntl_onekey over to release. Thanks for your support.

DavidLRowe:
Everything looks good except for (what looks to me to be) one minor typo in `welcome.htm` and also in `ntl_onekey.php`: In the chart for `z` the USV is given as `U+1989`, but it seems it should be `U+1988` (which is what the keyboard actually produces, so it's only a documentation issue).

ericjding:
You are correct. Thanks for catching that!

Issue #337, opened 2018-06-13
[infrastructure] Add all .js keyboards to package


darcywong00:
# legacy keyboards
- [ ] georgian

# release keyboards
- [x] ~arabic_izza~ (done)
- [x] balochi_phonetic
- [x] bj_cree_east
- [x] bj_oji_cree
- [x] easy_chakma // (to add on PR #167)
- [x] ekwtamil99uni

// fv append _kmw for .js files
fv_dakelh
fv_denesuline_epsilon
fv_diitiidatx
fv_gitsenimx
fv_gwichin
fv_halqemeylem
fv_han
fv_henqeminem
fv_hlgaagilda_xaayda_kil
fv_kanienkeha_e
fv_ktunaxa
fv_kwakwala
fv_kwakwala_liqwala
fv_migmaq
fv_natwits
fv_nisgaa
fv_nlekepmxcin
fv_nortern_tutchone
fv_nsilxcen
fv_nuucaanul
fv_secwepemctsin
fv_sencoten
fv_shashishalhem
fv_smalgyax
fv_southern_tuchone
fv_statimcets
fv_statlimxec
fv_tagizi_dene
fv_taltan
fv_tsekehne
fv_tsilhqotin
fv_uwikala
fv_xaislakala

- [x] ghana
- [ ] gff_ethiopic_7
- [x] jorai (use jorai.js instead of -layout.js)
- [x] kayan
- [x] kbdsn1
- [x] nw_iranian_latin
- [x] persian_phonetic
- [x] rac_kalasha
- [x] shan
- [x] ~sil_ipa~ (done)
- [x] sil_nubian
- [x] sil_torwali
- [x] tainua
- [x] tem_kdh

LornaSIL:
@darcywong00 do you want us to manually do these or do you have a plan for a way to do this with a script?

darcywong00:
Hmm, this might be too complicated to do with a script.
In addition to adding the JS file to the package, we'd want to make sure applicable font and language tags are associated.

And since we're already editing package files, do we want to add `<FollowKeyboardVersion/>`?  #246

DavidLRowe:
I suspect that the version number would need to be bumped as well? That shows up in README.md, HISTORY.md, xyz.kpj, xyz.kps and perhaps other places as well.

mcdurdin:
Note, my audit just now showed the following .js keyboards in legacy that are not yet in packages:

- [ ] a/armenian_east
- [ ] a/armenian_west
[...]

PR #924, opened 2019-10-28
[basic_kbdarme] Import Armenian Eastern Basic

LornaSIL:

Deprecating two keyboards.

Relates to Issue #337


DavidLRowe:
OSK has entry for CTRL K_SPACE. Not sure that's necessary?
Desktop:
\` produces U+055D not character on OSK.
' produces U+0583 not character on OSK.
~ produces U+055C not character on OSK.
" produces U+0553 not character on OSK.
(perhaps QUOTE and BKQUOTE are switched for both layers?)
{ produces U+0549 which doesn't seem to be character on OSK.
} produces U+054B which doesn't seem to be character on OSK.

Touch:
Has similar issues.

LornaSIL:
I reimported OSK as it seemed quite messed up. Exported OSK images for welcome file.

I fixed keystrokes (or moved the keys) for the following characters (I think) ` ~ - _ = + { } | [ ] \ ' " / ?

DavidLRowe:
touch layout, please check that output is what is intended:
- 2nd row, last two characters (seemed switched)
- 3rd row, tenth character

PR #927, opened 2019-10-29
fix(fv_dakelh): restore enhanced touch layout from earlier version

### LINKED ISSUE #916: [FirstVoices] Dakelh keyboard touch layout is wrong (default)
I have a report that the Dakelh keyboard has the wrong outputs from the default layer. I have not yet tried to replicate but wanted to record this so I didn't lose the information.

mcdurdin:
fixes #916


DavidLRowe:
README.md, LICENSE.md and .kmn files need copyright year changed to 2008-2019. (I think the .kps should pick that up automatically from the .kmn file, otherwise that file should be changed as well.)

I didn't test the keyboard itself, since this PR is restoring an older file. Let me know if you think I should test the touch keyboard.

DavidLRowe:
fixes #916

mcdurdin:
Okay, I think I've updated the Copyright year range everywhere; it might be good if you take a small poke at the touch layout; I've done so but fresh eyes don't hurt!

DavidLRowe:
Well, there are some (mostly minor) problems with the touch layout. 
- Default layer: type `t` then `_` (top row, leftmost key): I'd expect an underlined `t`, but the keyboard actually gives an underlined `ts`. Maybe this is intentional?
- Shift layer: After typing a key on the shift layer, the user is returned to the default layer, *except* when typing a longpress key, which leaves the user on the shift layer. Seems inconsistent.
- Shift layer: Some longpress keys are not producing what the key shows, for example, on the `T` key:
  - underlined `Ts` produces underlined `ts`
  - underlined `TS` produces underlined `t`
  - underlined `T` produces nothing

@mcdurdin How to proceed? I could approve this PR and leave these issues for the future. Presumably they were there before the keyboard migration. Or someone else could fix them in your branch. (I'm assuming you don't have time for this.)

PR #932, opened 2019-10-30
[basic_kbdintam] Tamil Basic keyboard Win10 import

LornaSIL:
Mobile layout mostly matches previous mobile layouts (enhanced)

Deprecating three keyboards.

Relates to Issue #337


DavidLRowe:
Everything looks fine, although on both default and shift layers there are a number of pairs of keys that produce the same output, but that's how the Windows Tamil layout appears as well.

PR #935, opened 2019-11-01
[basic_kbdazel] Azeri Latin Basic Win10 import

LornaSIL:

Deprecating two keyboards.

Relates to Issue #337


DavidLRowe:
BCP 47 code in .keyboard_info and .kps needs to be corrected. Other than that, all looks fine.

PR #936, opened 2019-11-01
[basic_kbdvntc] Vietnamese Basic Win10 import

LornaSIL:

Deprecating two keyboards.

Relates to Issue #337


DavidLRowe:
welcome.htm and basic_kbdvntc.php: change "should be type" to "should be typed"

Do we need to warn about Shift+Alt+O,K,M?

Otherwise looks fine.

PR #937, opened 2019-11-03
gff_geez v1.4

dyacob:
Fix for number input sequence for orders of 10 -one cycle was missing.  The numeral input statements are synched with gff_amharic.

### DIFF
--- a/release/gff/gff_geez/HISTORY.md
+++ b/release/gff/gff_geez/HISTORY.md
@@ -1,5 +1,8 @@
 # ግዕዝ (Ge'ez) Change History
 
+## 2019-11-03 1.4
+* Fix to number input for orders of 10.
+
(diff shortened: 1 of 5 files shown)


keyman-server:

LornaSIL:
@dyacob would you mind changing the .kps to access the new version of Abyssinica from ```sil\abyssinica``` and not ```sil\abyssinica_sil```. Unfortunately we've changed our file naming scheme for all our fonts and couldn't just update the fonts in the ```old``` folder.

Eventually we can try to update all the packages to use the new version and then get rid of the old version of the font.

dyacob:
Thanks @LornaSIL , an update from last evening has the change to the new font directory.  Following the Ge'ez keyboard update I'll apply the change to the other GFF keyboards.

PR #941, opened 2019-11-06
[basic_kbdheb] Hebrew Basic Win10 import

LornaSIL:

Deprecating two keyboards.

Relates to Issue #337


DavidLRowe:
Touch layout seems to have an empty shift-ctrl layer, which doesn't seem necessary.

PR #950, opened 2019-11-11
[basic_kbdusa] US English Table for IBM Arabic 238_L Basic Win10 import

LornaSIL:
New mobile tablet layout.

Deprecating one keyboard.


DavidLRowe:
touch layout: shift key doesn't take you to shift layer. (Shift layer not tested.)

PR #951, opened 2019-11-11
[basic_kbduk] United Kingdom Basic Win10 import

LornaSIL:

Deprecating two keyboards.

Relates to Issue #337


DavidLRowe:
Desktop layout: Images used in welcome.htm don't match what it produced by keyboard. (OSK is correct, as is the .php help file.) Specific keys:
- row 1, leftmost key: default, shift and right-alt layers. 
- row 2, rightmost key: default, shift and right-alt layers. 
- row 3, key to left of Enter: default, shift and right-alt layers. 

Because the keyboard uses Shift+RALT+O, it seems we should include Hotkey warning for Alt+Shift+O.

PR #952, opened 2019-11-11
[basic_kbdlt] Lithuanian IBM Basic Win10 import

LornaSIL:

Deprecating two keyboards.

Relates to Issue #337


DavidLRowe:
.php file has empty keyboard layouts for `rightalt-shift` and `leftctrl-shift`

Otherwise all looks fine.

PR #953, opened 2019-11-11
[basic_kbdit] Italian Basic Win10 import

LornaSIL:

Deprecating two keyboards.

Relates to Issue #337


DavidLRowe:
Touch layout, shift layer, 3rd row, rightmost key: key displays "|", but produces U+00A7.

Other than that, all looks good.

PR #955, opened 2019-11-11
[basic_kbdmlt47] Maltese 47-Key Basic Win10 import

LornaSIL:

Deprecating two keyboards.

Relates to Issue #337


DavidLRowe:
Because this keyboard uses RALT+Shift+O, it seems that we should have a warning about redefining the Shift+Alt+O hotkey. 

PR #959, opened 2019-11-14
fix(ci): dependencies for release packages

mcdurdin:
Fixes keymanapp/keyman#2338.

The dependencies for release packages were referring to other paths.
The pattern for the packages folder is to include the source kmn files
in the project so that they will be built along with the package.

However, the .kps files themselves still referred to the built kmx/js
in the original folders, rather than the expected ..\build\ path.

Given this change is a compile-phase-only change, there is no need to
update version numbers or make new releases.

This will fix the situation where bundled release package installers
are failing to build.

Note: I originally started off thinking we'd need to add a dependency
file into the keyboard folders that had dependencies and build those 
dependencies first, but this is a cleaner fix that avoids 
infrastructure that would only work with shell scripts and not Keyman 
Developer GUI.


keyman-server:

darcywong00:
fv_all is a different animal since it's
> a convenience for the distribution of the FV keyboards standalone on Windows.

I note if I don't have any fv .kmx built, fv_all.kmp consists of fonts, kmp.json, and a readme.htm.
Do you want any checks added to the build script?

release/packages/gff_amh_powerpack_7/source/gff_amh_powerpack_7.kps
```diff
@@ -1,2 +1,231 @@
...
+      <Version>1.4</Version>
+      <Languages>
+        <Language ID="am">Amharic</Language>
+      </Languages>
+    </Keyboard>
+    <Keyboard>
+      <Name>Ethiopic - (Language Neutral)</Name>
+      <ID>gff_ethiopic_7</ID>
+      <Version>1.0</Version>
+      <Languages/>
```
darcywong00: Is this language list intentionally blank? Seems that will prevent it from installing in the mobile apps

darcywong00: Does font information for the 2 keyboards also need to be specified?

mcdurdin: This is a desktop only keyboard, and is really kinda legacy as well, so font information is not used.

mcdurdin:
> fv_all: Do you want any checks added to the build script?

Can we list this as a separate issue to resolve? It's low priority because I don't think people will be downloading the Keyman Desktop + fv_all installer.

PR #961, opened 2019-11-15
I did it this way since cmd line would not work

leepartridge:
I cloned the master and I cloned this fork and neither would allow me to use git in a terminal. I use the same credentials I use here but it keeps saying I don't have authority to upload. 
I have gone through the git guide and it works but it does everything in the browser, no help with the command line.
So, I did it in the browser, drag, and drop.


keyman-server:

DavidLRowe:
See PR #1001

PR #965, opened 2019-11-18
[basic_kbdth0] Thai Kedmanee Basic Win10 import

LornaSIL:

Deprecating two keyboards.

Relates to Issue #337


DavidLRowe:
touch layout, rightalt layer: seems to be duplicate of some of the characters on ctrl layer. Is it needed?

touch layout, ctrl layer (accessed by "Alt" key on default layer): shift and four labeled keys don't work.

PR #969, opened 2019-11-18
[basic_kbdmonmo] Mongolian (Mongolian Script) Basic Win10 import

LornaSIL:
New mobile layout.

Deprecates 1 keyboard

This keyboard is not as high a priority as the other Win10 keyboards that deprecate 2 keyboards.


DavidLRowe:
Desktop layout, shift layer: Shift+J labeled on OSK as "VowSep" but produces no output.

touch layout, default layer: shift key doesn't take you to shift layer. (Shift layer not tested.)

PR #975, opened 2019-11-20
[basic_kbdhe] Greek Basic Win10 import

LornaSIL:

Deprecating two keyboards.

Relates to Issue #337


DavidLRowe:
OSK shows U+0385 for RAlt+L, but keyboard doesn't seem to type this. It seems that RAlt+; does work, so just the OSK needs correcting.

PR #976, opened 2019-11-20
[basic_kbdhe220] Greek (220) Basic Win10 import

LornaSIL:

Deprecating two keyboards.

Relates to Issue #337


DavidLRowe:
On OSK (and welcome.htm image) for default layer, `;` key seems incorrect. It shows ΅ but acts like ΄ .

Other than that, I think everything looks good.

PR #977, opened 2019-11-20
[basic_kbdhe319] Greek (319) Basic Win10 import

LornaSIL:

Deprecating two keyboards.

Relates to Issue #337


DavidLRowe:
touch layout:
- On default layer: `-` (row 1) swapped with `'` (row 4)
- On shift layer: `_` (row 1) swapped with `°` (row 4)

Everything else looks fine

LornaSIL:
I don't understand why the whole file looks new. I only changed what was requested.

DavidLRowe:
The updated .keyman-touch-layout file (and the .kmn file) were placed in the project folder rather than the source folder. This is why they show up as completely new.

PR #982, opened 2019-11-26
[nrc_crk_cans] 1.0.1

eddieantonio:
* Fixed some issues with backspace key layer switching
 * Fixed some typos in the documentation


keyman-server:

PR #984, opened 2019-11-30
feat: Add kmcomp x64 env + tests

mcdurdin:
Note (in draft): this PR cannot be merged until a released version of kmcomp.x64.exe is available. However, it can be used to test kmcomp x64 compatibility across platforms.

Adds: environment setup for x64 vs x86 versions of kmcomp (on Windows, uses x86, on other platforms, uses x64). Because some users are still running 32-bit versions of Windows, I kept the x86 version of kmcomp. I also felt that while the x64 version passed all tests, this reduces the risk of a big changeover until we are certain of the stability of the x64 update.

Adds: test script to build the entire set of release keyboards with both x86 and x64 compilers and compare the results down to the last byte. See the script resources/tests/test_x64.sh to learn more and understand what differences there will be in two subsequent builds (timestamps), and how those differences are ignored. I anticipate that we may be able to use essentially the same script as an automated regression test against the compiler in future.

Changes: the `build.sh -c` (clean) command was unnecessarily validating .keyboard_info (very slow).


keyman-server:

LornaSIL:
@mcdurdin has this already been done in another PR? I feel like it has been.

mcdurdin:
Yes. The script `resources/tests/test_x64.sh` may be helpful for verification but I can pull that out in the future if I want it.

PR #985, opened 2019-12-01
Add initial version of warang_citi

Singkiring57:


keyman-server:

LornaSIL:
It will be nice to have a Warang Citi keyboard on the Keyman site! Thanks.

I have a few initial comments:

* The LICENSE.md should include the copyright holder after the year. Thus:
```© 2019 Mangu Purty```
This should be changed in all places where there is a copyright statement such as the README.md, the .kmn  and .kps files.

* In the .kmn file I believe you could select "any* as the target and delete "windows iphone androidphone"

* Since your .kps file is pulling the graphic and the welcome.htm files from the "source" folder, I believe the welcome folder and its contents could be deleted.

* In the .kps file you can delete the reference to HISTORY.md, LICENSE.md and README.md. Those are not needed in the keyboard package. (Of course, leave then in the warang_citi folder.)

* In the .kps file:
   * go to the "Details" tab. You should select readme.htm (not README.md)
   * Delete "1.0" in "Version" and select the checkbox below that says "Package version follows keyboard version". This makes it easier to maintain the keyboard in the future.
   * Change the copyright to "© 2019 Mangu Purty"
* The "help" folder is actually intended for the website documentation and it needs to be in php format. An example of the page is here: https://help.keyman.com/keyboard/basic_kbduzb/1.1/basic_kbduzb.php
   * So, in the "help" folder you should delete all the files and drop this file in the help folder instead. Then rename the file from warang_citi.txt to warang_citi.php (this site doesn't allow me to upload php extensions. [warang_citi.txt](https://github.com/keymanapp/keyboards/files/3912348/warang_citi.txt)

One last thing. Do you have a recommended Warang Citi font? I think it would be good to suggest a font for the keyboard and a link to where the font can be downloaded. You could put that in the readme.htm file and/or in the welcome.htm file. Another option, if the font is under a suitable license, is to include the font as part of the keyboard package. If you do that, some people put it in the "warang_citi" folder and add another folder called "assets" and put the font(s) in that folder. (An example is the hieroglyphic keyboard here: https://github.com/keymanapp/keyboards/tree/master/release/h/hieroglyphic). Then, you would need to add the fonts to the .kps file. Then when the user installs the keyboard package it would also install the font. However, you should only do this if the font is under a suitable license that allows for redistribution.

I'll test the keyboard later (when I have a font).

**Comment by LornaSIL** [comment]
Ok, I did find these two fonts which I believe you created and I see they are under SIL's OFL which makes them good options for including in the keyboard package. 
https://github.com/Singkiring57/esel-sadom/blob/master/BoyoGagrai/BoyoGagrai.ttf
https://github.com/Singkiring57/esel-sadom/blob/master/DisumSans.ttf

Disum had U+200D (ZWJ) which your keyboard seems to require (but I'm not sure why...perhaps the documentation could be enhanced to explain when you would use ZWJ).
BoyoGagrai does not have U+200D, but it does have some of the other characters that the keyboard supports (!"'(),-./:;?)

And I see Noto Sans WarangCiti: https://github.com/googlefonts/noto-fonts/blob/master/phaseIII_only/unhinted/ttf/NotoSansWarangCiti/NotoSansWarangCiti-Regular.ttf
It has U+200D in it, but not a lot of the non Warang Citi characters.

So, I'm glad to see that there are some fonts available.

In order for this keyboard to work on mobile devices and the web, I think we need to define which font we'll use (and it will need to be included in the package as I suggested above). You can do that in the keyboard_info file. This is how I suggest you change the keyboard_info file to point to the font you wish to use:
```
{
    "license": "mit",
	  "description": "Phonetic Warang Citi Keyboard",
    "languages": { 
       "hoc-Wara": { 
            "font": { 
            "family": "Disum", 
            "source": "DisumSans.ttf" 
         } 
      }
   }
}
```
Of course, you could choose a different font. This is just an example.

**Comment by Singkiring57** [comment]
A very helpful review, thank you!
I will make the necessary changes.

**Comment by Singkiring57** [comment]
Regarding the ZWJ character:
The ZWJ character is used to stack Warang Citi letters on top of others in  order to create consonant clusters. But such stacked letters do not occur in Ho language, they are mostly used for non-native sounds. In modern writing, such letters are not required since consonant clusters are not written that way anymore. Therefore it would not be a big issue even if the ZWJ is absent form the font.

**Comment by LornaSIL** [comment]
Are you ready for me to re-review this PR or are you still making changes?

**Comment by Singkiring57** [comment]
I wanted to improve the touch layout, that is what is taking much time.

**Comment by LornaSIL** [comment]
That is perfectly fine. Just let me know when it's ready for review again.

**Comment by LornaSIL** [comment]
I've done a minimal review to find out why the build is failing:
* ```warang_citi-help.php``` should be renamed to ```warang_citi.php```
* The .kps file is including ```warang_citi-help.php``` - please delete
* The .kps file is including ```warang_citi-help.htm``` - that file is not in the repo, either add the file or delete it from the .kps.
   * If you delete it from the .kps file, also delete it from the .kmn file
* The .kps file is including ```warang_citi-code.js``` - I don't think that file is needed and should be deleted from the .kps file and the repo
* In the .kmn file, please select "any" as a target rather than Windows, IPhone and androidphone

Thanks for including the font. That is helpful.

**Comment by Singkiring57** [comment]
I noticed that I am not able to install the keyboard in my phone without including the `warang_citi.js` file in the `warang_citi.kps` file. So, what is with that?

**Comment by Singkiring57** [comment]
Anyway, I have made the changes.

**Comment by LornaSIL** [comment]
This looks quite good now. Please delete from the .kps:
* warang_citi.kmn
* warang_citi.kvks
* warang_citi.ico

All of these are source files and are part of the compiled keyboard so are not needed in the package. (Of course, the should remain in the repo as they are needed for compiling.)

Yes, you are right that warang_citi.js needs adding to the list of files in the .kps (but don't add the actual file to the repo, you should just add the reference to it in .kps. It should find it in the build folder the same way the .kmx and .kvk are referenced.

I still haven't tested your mobile layout. But, assuming it works, you should change the php file from:
```
```
to
```
<div id='osk-phone' data-states='default shift numeric lock wara-default wara-shift wara-numeric wara-lock'>
```
Put them in the order you want them displayed in. If you don't want all those layers to be displayed you can just not include them in that line. If you aren't using all those layers you may want to delete them from the touch layout.

I'll review the keyboard next week.

**Comment by LornaSIL** [comment]
* The Desktop keyboard seems to work fine.
* On the mobile layout, shift layer, the comma and fullstop keys are producing < and > rather than , and .
* On the mobile layout, wara-default layer, the 8 and 9 are producing 7 and 8.

I don't really like including both a Latin and Warang Citi keyboard for the mobile layout. It seems like it would be just as easy for people to switch to a Latin-based keyboard than to go through all the layers of this keyboard. My preference would be to delete the Latin-based layers and just include the Warang Citi layers. @mcdurdin any opinion here?

**Comment by LornaSIL** [comment]
Please change the keyboard_info file to have this information:
```
{
    "license": "mit",
    "languages": { 
         "hoc-Wara": { 
              "font": { 
              "family": "BoyoGagrai", 
              "source": "BoyoGagrai.ttf" 
           } 
        }
     },
	  "description": "warang_citi generated from template"
}
```
That will allow the phone and the web to use the font.

mcdurdin:
> I don't really like including both a Latin and Warang Citi keyboard for the mobile layout. It seems like it would be just as easy for people to switch to a Latin-based keyboard than to go through all the layers of this keyboard. My preference would be to delete the Latin-based layers and just include the Warang Citi layers. @mcdurdin any opinion here?

We don't normally include multiple scripts in a single keyboard -- unless the target language is written in multiple scripts. The primary reason is that in the operating system, the keyboard will be associated with a specific language. Many applications will tag the text typed with that language, meaning any text typed in the alternate script has the wrong language tag, leading to spell checker issues, automatic corrections which shouldn't be applied, etc. It's better to switch keyboards to type in another language.

Singkiring57:
In India, most people use two scripts over the internet, their native script and Latin. If we go through WhatsApp chats, Facebook comments, we find that the usage frequency of Latin in way more than that of Indian scripts. Most people prefer writing their languages in Latin over the internet because it is much easier than Indic scripts. It is because of this that Google has made a separate Hinglish keyboard.
Initially I only had Warang Citi layer, but I added the Latin layer because not all Ho people can read Warang Citi.

https://scroll.in/article/748337/for-once-rahul-gandhi-shows-the-way-hindi-needs-to-discard-devanagri-and-adopt-the-roman-script

https://timesofindia.indiatimes.com/blogs/The-underage-optimist/scripting-change-bhasha-bachao-roman-hindi-apnao/

LornaSIL:
I feel like the Latin part of the mobile keyboard needs to come out. It is easy to switch keyboards to a standard English or transcription keyboard (just as easy as changing layers). We don't want one keyboard to become so complex that it is difficult to maintain and troubleshoot.

Singkiring57:
Okay then, I'll remove the Latin layer.

LornaSIL:
This keyboard is now available here: https://keyman.com/keyboards/warang_citi
Would you mind deleting the branch in the button that is probably above this comment? Thanks!

PR #986, opened 2019-12-02
[basic_kbdfr] French Basic Win10 import

LornaSIL:

Deprecating two keyboards.

Relates to Issue #337


DavidLRowe:
desktop layout, shift layer: key to left of 1 produces ~ even though nothing listed for that key.

touch layout, rightalt layer: rightmost key in second row produces ¤ even though key top is blank.

LornaSIL:
@DavidLRowe I'm not seeing the desktop shift layer problem with tilde. I see the tilde on the OSK and in the graphic. Can you share a screenshot? Thanks.

DavidLRowe:
My confusion. I misinterpreted my notes, which were assuming a traditional French keyboard where the shift layer is blank for that key.

PR #987, opened 2019-12-02
[basic_kbdrost] Romanian (Standard) Basic Win10 import

LornaSIL:

Deprecating three keyboards.

Relates to Issue #337


DavidLRowe:
I know you didn't design this keyboard, but on both desktop and touch layouts, for the rightalt layer: 
Is ~ really a dead key? I guess it is, but the only thing that works is typing a space afterwards to get a tilde (and there's already a tilde key on the keyboard). The tilde deadkey is not listed in welcome.htm. 

I'm happy to approve this as is, but thought I'd point out where I was confused when reviewing it.

LornaSIL:
I have to say I was confused too since the rules make it a deadkey, but don't do anything. I'm happy to change it if you think I should.

LornaSIL:
I went ahead and removed the deadkey and added a tilde. It's in there several times, but that way the basic design of the keyboard is the same.

Also, I noticed the documentation for grave with wrong...using an ogonek character.

PR #996, opened 2019-12-10
[basic_kbdcz1] Czech (Qwerty) Basic Win10 import

LornaSIL:

Deprecating two keyboards.

Relates to Issue #337


DavidLRowe:
There seem to be six dead keys (two on default layer, two on shift layer, two on rightalt-shift (desktop) or rightalt (touch) layer, but I don't see documentation in the welcome.htm or the .php file.

On touch layout, shift layer, the shift (arrow) key doesn't take you back to the default layer.

DavidLRowe:
Dead key documentation looks good.

I think you changed the wrong spot in the touch layout. Now the shift (arrow) key on the default layer doesn't take you to the shift layer (and the shift key on the shift layer still doesn't take you back to default layer).

PR #1009, opened 2019-12-24
Format fixes for Abyssinica 2.000

dyacob:
Changes in this branch are for documents only.  The new version of Abyssinica  SIL had different line height properties  which impacted layout and lead  to the first page content flowing over onto an additional page.  Spacing between paragraphs was changed on the first page to prevent a 2nd page from forming.


keyman-server:

PR #1010, opened 2020-01-02
[basic_kbdda] Danish Basic Win10 import

LornaSIL:

Deprecating two keyboards.

Added "example" keys from deprecated "danish" keyboard.

Relates to Issue #337


DavidLRowe:
I think the documentation (welcome.htm and .php) needs to change to remove reference to dieresis over capital Y.

PR #1016, opened 2020-01-13
[basic_kbdgkl] Greek Latin Basic Win10 import

LornaSIL:

Deprecating two keyboards.

No "example" keys to add.

Relates to Issue #337


DavidLRowe:
(1) welcome.htm lists:
```
`a will produce ą (available on aAeEiIoOuU).
´c will produce ç (available on cC).
```
but I'm not seeing that the keyboard actually produces these (ogonek and cedilla) with deadkeys. Upper and lower case c cedilla are on the RAlt and Shift+RAlt layers.

(2) dieresis + Y produces U+009f. Should it be U+0178?

LornaSIL:
>`a will produce ą (available on aAeEiIoOuU).

Deleted line.

>´c will produce ç (available on cC).

This is in the acute deadkey. I just separated it out for documentation.

>(2) dieresis + Y produces U+009f. Should it be U+0178?

Yes, I fixed that in the keyboard.

PR #1018, opened 2020-01-15
Added english_shavian_igc keyboard

Shavian-info:
Added the Shavian alphabet keyboard for English, english_shavian_igc.


keyman-server:

LornaSIL:
Thanks for providing a Shavian keyboard!

Here are a few things I see in the information:

* Please add a copyright date to the LICENSE.md file (and other places):
  * © 2020 Shavian.info
* Please delete `.DS_Store` from the root and from source.

I will keep testing the keyboard, but there is an issue with the font you want to distribute. The Gentium font produced by SIL is released under the OFL with the name "Gentium" as a Reserved Font Name. The OFL DOES allow you to take the Gentium font and modify it, but you MUST rename it and not include the name Gentium in the font name. The font also needs to include the license in the license field. So, I'd recommend the font issue get sorted out first. Is that your font so you can make the change? Or is there another font that doesn't violate any licenses that we could put in the keyboard package?

You can read more about OFL here: https://scripts.sil.org/OFL-FAQ_web and I've pulled out a few of the relevant FAQs:

>Question: 5.2 Why can't I use the Reserved Font Names in my derivative font names? I'd like people to know where the design came from.

>Answer: The best way to acknowledge the source of the design is to thank the original authors and any other contributors in the files that are distributed with your revised font (although no acknowledgement is required). The FONTLOG is a natural place to do this. Reserved Font Names ensure that the only fonts that have the original names are the unmodified Original Versions. This allows designers to maintain artistic integrity while allowing collaboration to happen. It eliminates potential confusion and name conflicts. When choosing a name, be creative and avoid names that reuse almost all the same letters in the same order or sound like the original. It will help everyone if Original Versions and Modified Versions can easily be distinguished from one another and from other derivatives. Any substitution and matching mechanism is outside the scope of the license.

>Question: 3.1 Can I change the fonts? Are there any limitations to what things I can and cannot change?

> Answer: You are allowed to change anything, as long as such changes do not violate the terms of the license. In other words, you are not allowed to remove the copyright statement(s) from the font, but you could put additional information into it that covers your contribution. See the placeholders in the OFL header template for recommendations on where to add your own statements. (Remember that, when authors have reserved names via the RFN mechanism, you need to change the internal names of the font to your own font name when making your modified version even if it is just a small change.)

Shavian-info:
Thanks LornaSIL. I have:

- corrected the copyright notices to include the year
- brought the bundled font into compliance with the SIL Open Font Licence
- delete the .DS_Store files from root and source.

LornaSIL:
I would recommend in the .kps file, Details tab, to select the readme.htm under "Optional Information". That allows that file to be read as soon as someone clicks on the .kmp file.

The Desktop works fine.

The phone layout works fine, but it is missing these characters that are on the Desktop: `¥ µ ° ¶ ≈ § † × ÷ … ¢ ‡ © ™ ® – — ` Do you want to add them or do they not matter?

I will test the tablet layout tomorrow, but I guess it will be the same.

Thanks for changing the font name. Could you please put in the license field that it is based on the Open Font License?

If you could change the keyboard_info file to add the font information, then the font will be available on web and on the phones. I think this will work:
```
{
    "license": "mit",
    "languages": {
      "en-Shaw": {
        "font": {
          "family": "Ormin",
          "source": [
            "Ormin-Regular.ttf"
          ]
        }
      }
    },
	  "description": "Shavian alphabet keyboard for English, inspired by the Shavian Imperial Good Companion typewriter layout."
}
```

I think that is everything.

We will also need an online help file (php) which is generally based on the welcome.htm file. Do you want to attempt to create that, or should I do that after we put the keyboard online?

Shavian-info:
Thanks again for your help LornaSIL. I've:
- added the readme.htm to the optional information in the kps file
- updated the keyboard_info file as you requested
- included the OFL information in the bundled font Ormin-Regular.TTF

The additional characters on the desktop keyboard aren't necessary on the touch keyboard layout. They are just nice to have given there was room on the desktop keyboard layout, but are not essential.

PR #1022, opened 2020-01-17
Burushaski Girminas new keyboard

quwatsunny:
burushaski_girminas


keyman-server:

LornaSIL:
I do have some general feedback on the files. My initial thought was that this would be an Arabic script keyboard. I was surprised to see that it is Latin based. I think it would be helpful to make that clear.

* In the README.md you could put this under links: `https://keyman.com/keyboards/burushaski_girminas` That will be Keyman's home page for this keyboard.
* In HISTORY.md please add your name as the creator.
* In keyboard_info file:
   * I would add the description you have put in the README.md file, but maybe add "Latin-based". The keyboard_info description will display on Keyman's home page.
   * Also, please add the languages this keyboard supports. I'll include a sample keyboard_info file below that might work for you.
```
{
    "license": "mit",
    "languages": [
        "bsk-Latn",
        "khw-Latn",
        "scl-Latn"
    ],
	  "description": "Burushãski Girminas is a Latin-base keyboard developed by Quwat K. Sunny for Burushãski of yasin while working on a large project. It can be used flawlessly in other variants Burushaski, shina and khowar."
}
```
* Please delete the `burushaski_girminas.ico` file that is in the root folder. Leave the one in the `source` folder.
* It would be good to add the same description as above to the readme.htm file.
* In the .kps file, please go to the "Keyboards" tab and add the languages as listed above: bsk-Latn, khw-Latn, scl-Latn. Adding these languages helps the operating system to know what languages the keyboards work with and also helps our Keyman website search mechanism to find appropriate keyboards when people search for a keyboard that supports Burushaski or Shina or whatever.
* For the .kmn itself, here are some comments:
   * jj and Jj produce the same characters. Is that intentional? If you want a capital J with caron you could have it output `U+004A U+030C`
   * gg produces a g with caron, but Gg produces a G with breve. If you want Gg to produce G with caron you should output `U+01E6`.
   * ii produces a i with caron, but Ii produces a I with breve. If you want Ii to produce I with caron you should output `U+01CF`.
* As it currently exists, the welcome.htm file is not helpful. That file should describe how to use the keyboard. I would like to see a list of which keystrokes produce which characters so that users will know how to use the keyboard. ( I couldn't test the keyboard without looking at the code.)

The reason you want matching pairs is sometimes applications will automatically uppercase or lowercase characters and you will get odd results if you don't have matching i+caron and I+caron. 

Everything else looks pretty good.

quwatsunny:
@LornaSIL Thank you very much for your input.  I have changed every single thing that you had suggested icluding the welcome.html file. Please see if you can include it now.

LornaSIL:
It looks much better. Just a few things:
* Please delete `burushaski_girminas\burushaski_girminas.ico`. Leave the one that is in `burushaski_girminas\source\burushaski_girminas.ico`
* You removed the rule for Gg. Was that intentional?
* You now have a welcome.htm (with very little information) and a welcome.html. The welcome.html has the information you want on keyboarding. But, the .kps is calling to include the welcome.htm. The best thing to do would be to update welcome.htm with the correct information and delete welcome.html.
* You still need to add the three languages to the .kps file. Go to "Keyboards" tab in the .kps. Then, where it says "Languages" click on "Add..." and add `bsk-Latn`, then do the same thing to add `khw-Latn` and `scl-Latn`.

Thanks!

quwatsunny:
@LornaSIL 

-  burushaski_girminas\burushaski_girminas.ico has been deleted

- RULE change for Gg was intentional as I realized it would always come at the end  or in the middle and will never start as capital letter. 

- deleted welcome.html and it is now welcome.htm

- I have editred .kps code to include the langauges with <language> tags. 

Thanks.

LornaSIL:
Even though you added the languages to the .kps, it's not showing up in the UI when I open the .kps. When I look at the source I'm not seeing the `<Languages>` wrapper. Did you add those manually or through the UI? If you added it with the UI then there's a bug in the version of Keyman that you are using. I think it needs to be:
```
      <Languages>
        <Language ID="scl-Latn">Shina (Latin)</Language>
        <Language ID="bsk-Latn">Burushaski (Latin)</Language>
        <Language ID="khw-Latn">Khowar (Latin)</Language>
      </Languages>
```
One last thing is that the *readme.htm* just has the same text as README.md and has lost all of its html. So, when I open readme.htm in a browser it looks horrible. I'm afraid I wasn't clear when I said to copy the description from the README.md. I didn't mean to do away with all your html in the file.

These are the last two things. Hopefully we can release it when those are resolved.

mcdurdin:
I think those <Language> tags were added manually — they need to be wrapped in a <Languages> tag

quwatsunny:
@mcdurdin @LornaSIL  Thanks. 

1. languages have been added with language codes (manually) 
2.  new readme file with html code and formated text has been updated. 

I appreciate you time and help.

PR #1025, opened 2020-01-17
[basic_kbdgae] Gaelic Basic Win10 import

LornaSIL:

Deprecating two keyboards. 

Added example to keyboard_info from gaelic deprecated keyboard. 

Relates to Issue #337


DavidLRowe:
I could approve this as is, but there are a couple of things that you might want to change:

In welcome.htm and .php file, you could change:
`This keyboard makes use of the SHIFT+RALT+O keys which are set...` to
`This keyboard makes use of the SHIFT+RALT+O key which is set...`

In welcome.htm and .php file, the instructions for the acute accent dead key use `´` U+00B4, whereas the keyboard and OSK.have `'` U+0027. This might be a bit confusing.

PR #1026, opened 2020-01-20
balochi_phonetic revision

rmlockwood:
…n the Balochi alphabet. In their place put fatah, kasreh and shin.


keyman-server:

DavidLRowe:
Original description: Put feh, hah and tatweel on the shift level since they are not used in the Balochi alphabet. In their place put fatah, kasreh and shin.

rmlockwood:
Hi David,

 

How do I include a font for the keyboard?

PR #1028, opened 2020-01-21
Update burushaski_girminas.kps

quwatsunny:
For some reason, the windows task bar shows Shina as the name of the keyboard while the primary language for the keyboard is inteded to be Burushaski. It possiblly was due the order of the languages and I have reordered the languages.

### DIFF
--- a/release/b/burushaski_girminas/source/burushaski_girminas.kps
+++ b/release/b/burushaski_girminas/source/burushaski_girminas.kps
@@ -17,8 +17,8 @@
   </StartMenu>
   <Info>
     <Name URL="">Burushaski Girminas</Name>
-    <Copyright URL="">© Quwat K. Sunny</Copyright>
-    <Author URL=""></Author>
+    <Copyright URL="">2018 © Quwat K. Sunny</Copyright>
+    <Author URL="https://www.facebook.com/qsunny">Quwat K. Sunny</Author>
   </Info>
   <Files>
     <File>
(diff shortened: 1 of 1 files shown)


keyman-server:

LornaSIL:
I'm not sure if this change will fix it or not. However, if you want to try doing that, you'll also need to:
* make the same change in the keyboard_info file (we like to keep the order in sync in those two files)
* change the version of the keyboard in the .kmn and anywhere else the version is listed (HISTORY.md, README.md, readme.htm) My preference is to only include the version in .kmn and HISTORY.md so I don't have to change all the files each time. The reason the version has to change is that the user system will not prompt a download of a new keyboard unless the version is changed.

quwatsunny:
> I'm not sure if this change will fix it or not. However, if you want to try doing that, you'll also need to:
> 
> make the same change in the keyboard_info file (we like to keep the order in sync in those two files)
> change the version of the keyboard in the .kmn and anywhere else the version is listed (HISTORY.md, README.md, readme.htm) My preference is to only include the version in .kmn and HISTORY.md so I don't have to change all the files each time. The reason the version has to change is that the user system will not prompt a download of a new keyboard unless the version is changed.

yes, I have noticed other things as well. I will take some time to make all the improvements with a new version.

LornaSIL:
Before you update your keyboard, please make sure to update your master because I added a help file for your keyboard based on welcome.htm.

PR #1032, opened 2020-01-21
[basic_kbdtuf] Turkish F Basic Win10 import

LornaSIL:

Deprecating two keyboards.

Relates to Issue #337


DavidLRowe:
Observation: diacritics over dotted and dotless i give the same character. I doubt this is something that needs any changes.

It seems that welcome.htm and the .php help file should warn about Ralt+shift+M (as well as +O which is already there).

PR #1033, opened 2020-01-21
[basic_kbdtuq] Turkish Q Basic Win10 import

LornaSIL:

Deprecating two keyboards.

Relates to Issue #337


DavidLRowe:
dieresis y doesn't work. I expect the documentation should change (two places). Other than that all looks fine.

PR #1034, opened 2020-01-21
[basic_kbdmaori] Maori Basic Win10 import

LornaSIL:

Deprecating two keyboards.

Relates to Issue #337


DavidLRowe:
touch layout has empty ctrl layer with no way to reach it

PR #1035, opened 2020-01-22
Abyssinica SIL 2.000 migration & Doc Fixes

dyacob:
* KPS files point to Abyssinica-Regular.ttf (new) instead of Abyssinica-R.ttf (old)
* Supported OSs that not listed in ReadMe.md files are added.
* Tweak to gff_gurage_amharic package description in keyboard info file

### DIFF
--- a/release/gff/gff_amharic/HISTORY.md
+++ b/release/gff/gff_amharic/HISTORY.md
@@ -1,6 +1,10 @@
 gff_amharic Change History
 ==========================
 
+1.9 (22 Jan 2020)
+-----------------
+* Package migration to Abyssinica SIL 2.000
+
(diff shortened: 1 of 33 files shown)


keyman-server:

LornaSIL:
Unfortunately...when you touch any of the files used to create packages, you have to bump the version number so that people will be prompted to update their keyboards. That means updating the version number in .kmn and at a minimum the HISTORY.md. I won't require you to update all the copyrights.

The only files you've updated that don't touch the keyboard packages are the README.md and keyboard_info files.

dyacob:
I've made the version bumps throughout and went ahead and changed the copyright date to 2020 while I was at it. Thanks!

PR #1036, opened 2020-01-22
[indigenous_nt] Update to 3.4.1

ind-nt:


keyman-server:

LornaSIL:
* The .kvks (on screen keyboard) is now using ALT and not RALT. But the keyboard itself uses RALT. I think you need to make sure the OSK is using RALT.
* Please add the keyboard name in HISTORY.md at the top

I think that's everything. The kvks change is essential to change and the HISTORY.md would be very good to change.

ind-nt:
Hi Lorna,

Ok, I've corrected it.

What I wanted to do is have the alt key on their physical keyboard act in
the same way as their on-screen keyboard in keymanweb.com. This works for
the left-alt key, but you have to manually click the right-alt key on the
screen to get the alt state.

That might just be a keymanweb.com issue though.

Thanks for your help!
- R

LornaSIL:
The OSK is simply a reflection of the underlying keyboard.

You *can* write different rules in the .kmn file, something like this:

```
$KeymanOnly: + [RALT K_A] > 'æ'
$KeymanWeb: + [ALT K_A] > 'æ'
```
You would have to experiment as I've rarely done this.

PR #1039, opened 2020-01-23
[basic_kbdest] Estonian Basic Win10 import

LornaSIL:

Deprecating two keyboards. 

Added example to keyboard_info from estonian deprecated keyboard. 

Relates to Issue #337


DavidLRowe:
In the .keyboard_info file, in the "keys" entry, should "[[" be "["? Otherwise all looks fine.

DavidLRowe:
From @LornaSIL : "the double [[ is required for it to appear once and not think it's code" (The example was taken from legacy/e/estonian

PR #1040, opened 2020-01-24
[basic_kbdgr1] German (IBM) Basic Win10 import

LornaSIL:

Deprecating two keyboards.

Added example to keyboard_info from german_ibm deprecated keyboard.

Relates to Issue #337


DavidLRowe:
some minor issues:
- welcome.htm and .php documentation for acute accent has:
`´a will produce á (available on aAcEiIoOuUyY).`
but it should be "e" not "c".

The keyboard does use Shift+RAlt+M which is one of the default hotkeys, however, the same symbol is available on RAlt+M, so maybe there's no need to mention it?

DavidLRowe:
Could you change the wording "SHIFT+RALT+M keys which are" to "SHIFT+RALT+M key which is" in the readme.htm, the welcome.htm and the .php files?

Also the documentation for acute (in the latter two files) lists "c" rather than "e", that is instead of "aAcEiIoOuUyY" it should read "aAeEiIoOuUyY".

PR #1045, opened 2020-01-26
Migration of Remaining GFF Keyboard Sources to Abyssinica SIL 2.0

dyacob:
This update changes remaining source files to use the 2.0 version of the Abyssinica SIL font.

Some compiled .kmp files in the keyboard repository (particularly the legacy folder) do include the older version, however, the source for the packages do not appear to be in the repository any longer.


keyman-server:

LornaSIL:
The keyboard build is failing. The online build claims it's because of the `gff_amh_powerpack_7.keyboard_info` file, but it seems well formed. Why don't you delete these two lines which are obviously wrong:
`"lastModifiedDate": "2012-03-20T10:58:47Z",`
and 
`"version": "8.0.0",` 
and see if that fixes it.

If that doesn't fix it then, it looks like we may have to modify `gff_ethiopic_7` which doesn't have a language tag. I don't really know if the system will build without that.

dyacob:
Thanks @LornaSIL , I've pushed in an update to the keyboard_info file only.  For the gff_ethiopic_7 keyboard, I could add the list of all applicable languages.  I'm not sure what to make of the warnings, they keyboard mappings worked as expected so I don't know what to address there.

PR #1047, opened 2020-01-27
[basic_kbdusx] United States Int Basic Win10 import

LornaSIL:
Deprecating two keyboards. 
Added example to keyboard_info from us_int deprecated keyboard. 
Relates to Issue #337


DavidLRowe:
Touch layout, shift layer: deadkey + space doesn't work for the 3 deadkeys on this layer.

Otherwise all looks fine.

DavidLRowe:
Correction: typing deadkey on the touch layout and then, while still on the shift layer, typing a space produces a space rather than the character being used for the deadkey. The same is true of the desktop layout if, after typing a deadkey, Shift+space is typed.

PR #1049, opened 2020-01-28
[balochi_latin] Initial version

rmlockwood:


keyman-server:

LornaSIL:
I only found one problem. 
* On the tablet layout:
   * On the default layer, the shift key needs "next layer" set to "shift"
   * On the shift layer, the shift key needs "next layer" set to "default" and you may also want to set Key Type to "Special (Active".
   * On both the alt and shift-alt layers the shift key has a similar problem.

Now, I'd like you to add the help file into this PR if you are willing.

* Add a folder within `source` called `help`
* Download the attached file into the help folder
* Rename the file in that folder to balochi_latin.php
* Commit the file when you make the touch layout changes

[balochi_latin.txt](https://github.com/keymanapp/keyboards/files/4125259/balochi_latin.txt)

rmlockwood:
Ok. Did those changes. Thanks for catching them.

PR #1051, opened 2020-01-29
GFF Typing Manual Bug Fix

dyacob:
The document updates has small updates to the title metadata property to indicate the document language.  Within the documents a number of special labial-velar characters were in the wrong columns and have been swapped into the appropriate column (perplexed on how this happened, but fixed now).


keyman-server:

LornaSIL:
IF you want these to go into the existing keyboard packages, then a version bump is required.

@mcdurdin if he just wants these documents updated for the next time he updates the keyboards, can we go ahead and approve it? Or will that be a problem?

dyacob:
Could I update the version number in the .kps file only, unchecking "Package version follows keyboard version" under "Details" and **_not_** update the version in the .kmn file, README.md and HISTORY.md, etc?

When the keyboard .kmn file changes in the future, I would then go back to normal (recheck "Package version follows...", update HISTORY.md, etc).  My thought here is to trigger the package rebuilding and update notifications, while not implying a change to the input method.

On a related note, chatting with @andjc , we are both OK with the deletion of the AbyssinicaSIL fonts in our respective legacy directories.  If ever there is a need to rebuild the .kmp files in the same directories, the .kps source will need to be located and can then be edited to use the new version.

mcdurdin:
> Could I update the version number in the .kps file only, unchecking "Package version follows keyboard version" under "Details" and **_not_** update the version in the .kmn file, README.md and HISTORY.md, etc? I would bump the .kps file from 1.5 to 1.5.1 for example.

You could, but I think it's probably more hassle than it's worth! I'd just keep it simple and bump the version in the keyboard.

mcdurdin:
> @mcdurdin if he just wants these documents updated for the next time he updates the keyboards, can we go ahead and approve it? Or will that be a problem?

It won't be a problem to leave the version number the same, if that's what you want to do -- the changes will not impact the published package but would be deployed to the help website.

LornaSIL:
@dyacob should I just approve this or do you want to make a version change so that people will pick up the new documentation?

dyacob:
hi, sorry, i got side tracked but have been meaning to bump the versions.  I'll make an effort to do so on Tuesday.

LornaSIL:
I think all these keyboards are now online.

dyacob:
Thanks @LornaSIL !

PR #1056, opened 2020-02-05
chore: add examples to release keyboards from related legacy keyboards

### LINKED ISSUE #1006: New basic_xxxx keyboards are missing language examples
When we deprecated the older KeymanWeb keyboards, e.g. replacing [albanian](https://github.com/keymanapp/keyboards/blob/master/legacy/a/albanian/albanian.keyboard_info) with [basic_kbdal](https://github.com/keymanapp/keyboards/blob/master/release/basic/basic_kbdal/basic_kbdal.keyboard_info), we didn't copy across the language example data. These examples can normally be seen on keymanweb.com, for example, at [keymanweb.com](https://keymanweb.com/#fa,Keyboard_dari_clra).

The examples took quite a lot of work to produce and have been valuable as a way of 'getting started' with a keyboard.

mcdurdin:
Fixes #1006.

This adds examples to .keyboard_info files for release keyboards that replace legacy keyboards. The changes were automatically generated by script, and ignore all subtags of the bcp47 code (apart from the basic language subtag) in making the matches. (No other language tag transform was done -- so if there were e.g. ISO639-3 -> ISO639-1 transforms, they'll have been missed. Also, where a keyboard was associated with the wrong language, no example will be included.)

The code for parsing and making the transform is included in tools/fixups/ in the hope it could be used in future for other such batch jobs.

Note: in reviewing, you might want to select 'Ignore whitespace changes' under the settings cog!

### DIFF
--- a/release/basic/basic_kbdal/basic_kbdal.keyboard_info
+++ b/release/basic/basic_kbdal/basic_kbdal.keyboard_info
@@ -1,10 +1,16 @@
 {
-    "license": "mit",
-    "languages": [
-        "sq",
-        "als-Latn"
-    ],
-	  "description": "This keyboard layout is designed for Albanian.",
+  "license": "mit",
+  "languages": {
+    "sq": {},
+    "als-Latn": {
+      "example": {
+        "keys": "Kor[[;",
+        "text": "Korçë",
+        "note": "A city in southeastern Albania"
+      }
+    }
+  },
+  "description": "This keyboard layout is designed for Albanian.",
   "related": {
     "albanian": {
       "deprecates": true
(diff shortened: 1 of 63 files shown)


mcdurdin:
I opted not to add font data; just the language examples.

mcdurdin:
Those other three files are fine — they are just tooling in case we want to run scripts across multiple keyboards again so we don’t have to reinvent the wheel each time!

DavidLRowe:
Just want to double check the filename: `tools/fixups/fixupExamples.ts` since another file refers to 
`fixupExamples.js`.

mcdurdin:
The `fixups` folder is a [nodejs](https://nodejs.org/) project.

The file `fixupExamples .ts` is the TypeScript source file for the program. Before running the program (with `node fixupExamples.js`), you'd type `npm install && npm run build` to compile the .js from the .ts, in that folder.

This is not necessary for the average user of the repository. Perhaps I should add a README to the folder with this information?

DavidLRowe:
I don't think a README file is needed.

PR #1057, opened 2020-02-06
[saraiki] Initial release

darcywong00:
Keyboard submission on behalf of author https://community.software.sil.org/t/saraiki-keyboard/2995

> Note: The script is Arabic.


DavidLRowe:
Preliminary comments:
README.md 
supported platforms should include macos, linux, web

saraiki.keyboard_info, readme.htm, welcome.htm
need more detail; welcome.htm could include keyboard images

need to add release/s/saraiki/source/help/saraiki.php which can be based on content of welcome.htm

darcywong00:
The author has posted more documentation info so I'll be adding that

darcywong00:
I've fixed the BCP 47 script to "Arab" (Arabic) and added the author's additional info.

DavidLRowe:
@darcywong00 Can you verify the image for the shift layer for Shift+D key? The keyboard generates "ڈ" but the image in the welcome and help folders and the OSK seem to show "د" for Shift+D (as well as for unshifted "d"). I suspect that the image file (in both welcome and help folders) and the OSK are incorrect.

Is there any reason not to include "web" in the &targets? (and in README.md)? We have been suggesting "windows macos linux web" for &targets in the .kmn file and a similar list in the README.md. @mcdurdin suggested "desktop" for &targets, but that doesn't seem to include web, so perhaps "desktop web"? I know that there are a number of combinations that are equivalent, but it would be nice to standardise on one.

So, just a few minor changes and the keyboard should be ready to approve.

darcywong00:
Good catch on the OSK. 

* I've fixed Shift-D in the OSK source and updated the image files.
* Also fixed the typos in the image file names (saraiki vs sairiki)
* Finally, I added `Web` target, and am **not** including the js file in the package

### unshifted
saraiki_unshifted

### shifted
saraiki_shifted

### ctrl-alt
saraiki_ctrl_alt

rachitralli:
@darcywong00 @DavidLRowe @LornaSIL I checked the OSK please see Ctrl + alt 
the Saraiki language uses  ڄ 0684  and  06B3 ڳ which are already available in shift version but on the Ctrl Alt version the following letters are extra 
06B2  ڲ   and ڃ 0683

LornaSIL:
@rachitralli We do appreciate your interest in these keyboards. Do remember that languages can have multiple keyboard layouts (for example English has 5-6 on Keyman already). We are aware that Saraki also has multiple orthographies and sometimes keyboards include characters from words borrowed for another language.

PR #1058, opened 2020-02-11
[saraiki] Fix RTL and distinguish left/right ctrl-alt on OSK

darcywong00:
Follow-on to #1057 

This should address issues found on https://keymanweb.com/#skr-arab,Keyboard_saraiki

### DIFF
--- a/release/s/saraiki/HISTORY.md
+++ b/release/s/saraiki/HISTORY.md
@@ -1,6 +1,10 @@
 Saraiki Change History
 ====================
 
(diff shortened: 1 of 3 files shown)


LornaSIL:
The .kmn didn't get a version bump so I don't think resources were updated.

DavidLRowe:
You're right.

@darcywong00 You'll need to submit another PR that increments the version number in the .kmn file to 1.1.

PR #1059, opened 2020-02-12
[saraiki] Bump version to 1.1

darcywong00:
This was missed in #1058


LornaSIL:
Keyboard seems to be working properly now on Web. I can access ctrl-alt and it is right to left.

PR #1061, opened 2020-02-12
Burushaski Urdu Girminas

quwatsunny:
This is a burushaski keyboard based on urdu script. It has been developed on the basis of request from two variants of butushaski language.


quwatsunny:
new request requires

PR #1066, opened 2020-02-18
Burushaski Girminas Updated

quwatsunny:
Please review.  Also, please update the help page with the new one. 
I have deleted the other keyboard board to avoid conflicts.  It is still giving php file confilct error. I am not sure why.


quwatsunny:
ok i have made the final changes.

LornaSIL:
Since we are having such trouble getting the files in the proper order, would it be okay with you if I submit a Pull Request for this keyboard using the files you've committed? I'll fix up the files giving problems and then commit them if that's okay with you.

Right now you don't have a .kpj file so it won't build. I don't know what else needs fixing, but I can see you've committed compiled files as well as the source, and it might be easiest to have me fix it up rather than me making comments and you trying to fix the files.

quwatsunny:
> Since we are having such trouble getting the files in the proper order, would it be okay with you if I submit a Pull Request for this keyboard using the files you've committed? I'll fix up the files giving problems and then commit them if that's okay with you.
> 
> Right now you don't have a .kpj file so it won't build. I don't know what else needs fixing, but I can see you've committed compiled files as well as the source, and it might be easiest to have me fix it up rather than me making comments and you trying to fix the files.

yes please, go ahead. MY Keyman Developer does not let me do anything. It gives debugging error and shutting itself down. I uninstalled and reinstalled with no luck.  This started after the new update on keyman.  If you can help it will be greatly appreciated.

LornaSIL:
Replaced by #1068

PR #1068, opened 2020-02-19
[burushaski_girminas] Updated keyboard on behalf of author

LornaSIL:
I used the author's files, tested it and fixed a few issues.


quwatsunny:
it looks good.

LornaSIL:
Thanks. We'll still need it checked by someone on the Keyman team, but it's good to know it looks good to you. Now, you had another keyboard you wanted to submit? Would you like to create another PR with just that keyboard or do you want to send me the files for me to submit?

DavidLRowe:
.kps has bsk-Latn-PK for BCP47 code, whereas .keyboard_info file has bsk-Latn. They should agree. (Also, the three BCP47 codes are in a different order in the two files. It would be preferable if they were in the same order.)

.kmn file has different copyright statement than README.md. The copyright statements should all agree.

The documentation in welcome.htm for typing characters not on the keyboard is somewhat confusing. For example, it says:
A + A gives Ã
C + C gives Ć
when actually
a + a gives ã
A + a gives Ã
c + c gives ć
C + C gives Ć
Sometimes the second character has to be lowercase, sometimes it has to be uppercase. (You might consider adding extra rules for the doubled uppercase letters so that the second letter can be uppercase or lowercase.)

S + S appears twice in the documentation, first for šŠ (which is correct), but then for Ș where it seems that it should be "S + Dot". Note however that
s + . gives Ș
S + . gives ș
that is, these two entries seem to be reversed.

The touch layout has both tablet and phone layouts, which are almost but not quite identical. The tablet layout has some extra longpress entries. Is the phone layout needed?

In contradiction to the example in the welcome.htm documentation, there is no longpress on the "J" key that gives "Ǯ", instead the longpress is on the "Z" (or you can get the character by typing J twice). It seems that the documentation could be changed, or that key could be added to the "J" key as well.

Minor point: The longpress keys on "z" are in different order on the default and shift layers. Also for "c" and "n" (and maybe others I didn't notice!).

quwatsunny:
@LornaSIL I have updated the files as suggested by @DavidLRowe  including the welcome page and help page. See how you can further improve it.

LornaSIL:
Where did you make the changes? In your branch? I have to make changes in the files in this PR but I can look at your files to see what changes you made.

quwatsunny:
> Where did you make the changes? In your branch? I have to make changes in the files in this PR but I can look at your files to see what changes you made.

I have made the changes in my branch. I don't have write access to your branch . If you can please copy fr mine including the help and welcome

LornaSIL:
I did delete the phone layout as it seemed to be the same as tablet and it's easier to maintain one layout than two. Tablet works on phone.

DavidLRowe:
Most of my comments have been addressed. There are a few remaining issues with differences between the keyboard and the documentation in welcome.htm. The latter says:
R + H gives Ř
K + K gives Ǩ
but on the keyboard:
r + h gives ř
R + h gives Ř
R + H gives RH
k + k gives ǩh
K + k gives Ǩh
K + K gives KK

I'm curious why the choice of R + H for Ř, when all the other similar diacritics are formed by doubling the letter, that is, I would have expected R + R.

Is the K + K combination supposed to produce the "h" as well? If so, that should be reflected in the documentation.

I'm also curious why the touch layout, shift layer, now has Ǯ as a longpress on the "Q" key. Was that intentional? If so, do you want to add a longpress for the unshifted "q"?

quwatsunny:
> Most of my comments have been addressed. There are a few remaining issues with differences between the keyboard and the documentation in welcome.htm. The latter says:
> R + H gives Ř
> K + K gives Ǩ
> but on the keyboard:
> r + h gives ř
> R + h gives Ř
> R + H gives RH
> k + k gives ǩh
> K + k gives Ǩh
> K + K gives KK
> 
> I'm curious why the choice of R + H for Ř, when all the other similar diacritics are formed by doubling the letter, that is, I would have expected R + R.
> 
> Is the K + K combination supposed to produce the "h" as well? If so, that should be reflected in the documentation.
> 
> I'm also curious why the touch layout, shift layer, now has Ǯ as a longpress on the "Q" key. Was that intentional? If so, do you want to add a longpress for the unshifted "q"?

The reason it is r + h is that people are used to  press r + h in  the absence of keyboard where ř is intended 
ř has no capitalizatiom as it always comes in the middle or at the end of the word so the rule can be omitted.

 k + k gives ǩh
 K + K gives Ǩh is correct and unintentional otherwise.  


Ǯ must be under J as longpress  and is in unintentional otherwise.  @LornaSIL  could you please correct these as observed

DavidLRowe:
I guess I could approve this as is, but there's still some lack of clarity in the documentation. The welcome.htm file says "R + H" when in reality "R + H" doesn't work. You need "R + h" to get "Ř".

PR #1072, opened 2020-02-21
Indonesia suku

bennylin:
Layout for Indonesian keyboard, for Indonesian language and Indonesian local languages.

Based on, and an improvement of US Basic layout.


keyman-server:

LornaSIL:
It looks like you've updated files in basic/basic_kbdus. Using basic_kbdus as a starting point is perfectly fine and that's one reason we've added these keyboards. However...you should really create your new keyboard in the "i" folder called ` indonesia_suku` and all the files should be renamed from `basic_kbdid` to `indonesia_suku`. We are using the "basic" folder for all the keyboards we've imported from Windows and we would like to keep the `basic` folder for just those keyboards. (I've just realized I should put a README.md file in that folder to make it clear.)

bennylin:
closed this, I forgot to change the subdirectory from basic/basic_kbdus. I'll do another pull request after I changed it to basic/basic_kbdid

DavidLRowe:
@bennylin You need to change the folder to "indonesia_suku" and place it in the "i" folder under "release". That is, the path should be "release/i/indonesia_suku".

bennylin:
https://github.com/keymanapp/keyboards/pull/1077

PR #1073, opened 2020-02-21
[basic_readme] Added README.md to basic folder

LornaSIL:
Someone wanted to add a keyboard to basic_kbdus and I realized we needed to document how this folder is used.


DavidLRowe:
In the first line, "The" should be "They". 

And perhaps you want to make the connection more explicit, for example, change:
```
imported from Windows 10. The should match
```
to 
```
imported from Windows 10, so the Keyman desktop keyboard layout should match
```

PR #1078, opened 2020-02-22
US basic - touch layout improvement

bennylin:
This is an improvement of the touch keyboard for US Basic

* The original layout of 14 keys wide was not mobile friendly. It basically copy paste the pc layout.
* Replaced it with layout based on euro-sil touch layout. It's 10 keys wide, has 5 layer to include most symbols, including more rare ones..

Further modify the layout

* Show the ',.?' punctuation marks
* Add a key for the aposthrope (very often used in English language), and shown as '':;' punctuation marks
* Modify the shift layer, the only punctuation shown is '<' (and '>'), corresponding to the '.,' keys on the default layer
* Modify the numeric layer, the second row corresponding the the top row (numbers); the third row are the rest of the symbols.
* Modify the symbol layer, delete symbols already displayed on the previous layers.
* No changes on the currency layer


keyman-server:

LornaSIL:
I would prefer that we not touch the tablet layout since it's been available a long time, but I'm willing for you to add a phone layout with fewer keys on each layer. Keep in mind that the character repertoire should remain the same as the desktop.

PR #1079, opened 2020-02-23
[basic_kbdsg] Swiss German Basic Win10 import

LornaSIL:
Deprecating two keyboards.
Added example to keyboard_info from swiss_german deprecated keyboard.
Relates to Issue #337


DavidLRowe:
grave deadkey + space doesn't seem to produce a grave on the touch layout. Is that because it needs a "layer": "default" to match the rule in the .kmn file?

Other than that, everything seems fine.

PR #1082, opened 2020-02-25
[Udi] Initial version of Udi Keyboard

purplecloudict:
This keyboard layout is designed for Udi language of Azerbaijan. It includes an on-screen keyboard which can be viewed by clicking on the Keyman icon and selecting the On-Screen Keyboard menu item.

Эта раскладка клавиатуры предназначена для языка уди Азербайджана. Он включает в себя экранную клавиатуру, которую можно просмотреть, щелкнув значок Keyman и выбрав пункт меню.


keyman-server:

LornaSIL:
This keyboard looks much better. Here are some further comments:
* Please delete the `build` folder from your Pull Request
* Please delete the udi.kpj and uki.kpj.user file from your Pull Request

On the mobile layout, default layer:
* The longpress on the fullstop key for the single quote `'` isn't producing anything (because it's a deadkey on the desktop layout. Please change **Code:** `K_QUOTE` to `U_0027` (You have the same issue for the same key on the shift layer and you should fix that one too).
* On the shift layer where you have longpresses that use the rules you created that use codes like `T_S_Acute` you should change **Modifier:** to `shift`. (If you used U_015E you should NOT change **Modifier:** to `shift`.).
* On the numeric layer for the `[` and `]` those are both deadkeys. Do you want to leave them as deadkeys for the mobile layout, or do you want them to produce the square brackets? If you want them to produce square brackets you should change `K_LBRKT` to `U_005B` and `K_RBRKT` to `U_005D`.

The help file (udi_keyboard.php) doesn't have the correct syntax. That file is what will be displayed online. I would like to recommend we do something similar to what you did for the readme.htm file where you included both English and Russian. I'm going to upload my suggested file here. It will be called udi_keyboard.txt, but you should download it, rename it to udi_keyboard.php and then put that in the help folder in place of the one you have now. This file is based on both your welcome.htm files.

[udi_keyboard.txt](https://github.com/keymanapp/keyboards/files/4252572/udi_keyboard.txt)

LornaSIL:
Beautiful. I think there's just one thing left. On the shift layer of the touch layout, the longpresses on the `I` key are producing the wrong thing. 
* Change the first one to have a code of U_0049
* Change the second one to have a code of U_01CF

If you felt like it, you could change the description in the keyboard info file from `"Udi Keyboard generated from template"` to `"This keyboard layout is designed for Udi language of Azerbaijan. Эта раскладка клавиатуры предназначена для языка уди Азербайджана."`

All of that needs to be put on one line. This is the text that will display on the home page of the keyboard so it's not to enough enough of a description that people will know they want to use the keyboard. 

If you don't want to change the keyboard_info, I can approve it with the text as is, but you should change the two longpress keys on the `I`.

Then I think it's ready to approve.

By the way, I was informed that the rule I gave you above:

```
if(&platform = 'touch') + [T_Q_Acute] > U+0071 U+02B9
```
could be expressed this way (without the platform statement):
```
+ [T_Q_Acute] > U+0071 U+02B9
```
However, there is no need to change that now unless you feel like it.

purplecloudict:
Thanks Lorna,

I have made the recommended changes.

Mark

LornaSIL:
Remember to commit your changes!

LornaSIL:
The keyboard_info file has a two problems. One is that "description" part is no longer there. The 2nd is that the format for your file doesn't allow for Unicode apparently. So, let's change it back to this: Then I can approve it.
```
{
    "license": "mit",
    "languages": [
        "udi-Latn"
    ],
	  "description": "This keyboard layout is designed for Udi language of Azerbaijan."
}
```

PR #1086, opened 2020-02-26
[basic_kbdsl] Slovak Basic Win10 import

LornaSIL:
Deprecating two keyboards.
Added example to keyboard_info from slovak deprecated keyboard.
Relates to Issue #337


DavidLRowe:
There was no example in the .keyboard_info file, but I didn't see any example in the deprecated keyboards either, so I'm assuming that comment was copied from another keyboard.

Documentation says:
¨a will produce ä (available on aAeEoOuU).
but dieresis doesn't work with e or E.

PR #1087, opened 2020-02-27
Initial version of taigi_poj keyboard

shiami:
My Tâi-gí (Taiwanese) Pe̍h-ōe-jī keyboard as the result of the Keyman workshop.


keyman-server:

DavidLRowe:
The copyright statement occurs in a number of files. It needs to have the year (or a range of years in the case of a revision to an existing keyboard) and the name of either a person or a legally recognized entity. So, assuming that Ngô͘ Ka-bêng is a valid name, you'd need `© 2020 Ngô͘ Ka-bêng` in the README.md file. (And I see that you have the correct copyright information in the LICENSE.md file and the .kmn file, as well as other places.) With that change, we can accept it into the experimental section of the keyboards repository.

Some design questions that you may want to consider, either now or before you finalize the keyboard:
- It looks as though you use some numbers (23579) after a vowel to choose a diacritic mark. Eventually you'll want to include more detailed information in the welcome.htm and taigi_poj.php help files.
- For the desktop layout, it seems that there are a lot of keys repeated on the Alt and Shift+Alt layers. (The o͘, ng, ch, chh, and raised n appear to be the unique ones.) Are the repeated keys necessary? Also note that Shift+Alt+L gives lowercase and Alt+L gives uppercase.
- Perhaps you could indicate the tone function of the numbers by, for example, using "2´" on the key in the OSK. 
- For the touch layout, you currently have kept the number key row (except for the zero key) on the default and shift layers, giving 5 rows on those layers. That is more than is recommended for a phone, so you'll want to check that your users are comfortable with this.
- You might want to put the zero back on the default layer (and move the hyphen elsewhere, though perhaps it is a common character?), then you wouldn't need to include numbers on the current numeric layer, and could perhaps combine the remaining keys on the current numeric and symbol layers into a single symbol layer.

So these are some ideas to think about, but recognize that I don't know anything about your context, so don't think that I'm directing you to make changes!

shiami:
> * Shift+Alt+L gives lowercase and Alt+L gives uppercase

Is that the common design I should follow to?

shiami:
> * For the desktop layout, it seems that there are a lot of keys repeated on the Alt and Shift+Alt layers. (The o͘, ng, ch, chh, and raised n appear to be the unique ones.) Are the repeated keys necessary?

I did it because I designed the Alt, Alt-Shift layers to be an advanced typing keyboard comparing to the Default, Default-Shift layers. I don't know is it a good way to design keyboard, but I think maybe someone would like to use it since the keyboard space is available to do so.

A lots of keys repeated is considered more convenient when using On-screen Keyboard. Not need to switch the Alt all the time, and keep using Alt, Alt-layer as another faster way to type. 

Is that make sense to you? Or I should not add the Alt, Alt-shift layers as an advanced way to type. Should I move the design to a brand new keyboard? Thanks.

shiami:
> * It looks as though you use some numbers (23579) after a vowel to choose a diacritic mark. Eventually you'll want to include more detailed information in the welcome.htm and taigi_poj.php help files.

> * Perhaps you could indicate the tone function of the numbers by, for example, using "2´" on the key in the OSK.

I think that's not gonna be a problem since there are already some other keyboards people are using with this same rules. But good to know!

shiami:
> * For the touch layout, you currently have kept the number key row (except for the zero key) on the default and shift layers, giving 5 rows on those layers. That is more than is recommended for a phone, so you'll want to check that your users are comfortable with this.
> * You might want to put the zero back on the default layer (and move the hyphen elsewhere, though perhaps it is a common character?), then you wouldn't need to include numbers on the current numeric layer, and could perhaps combine the remaining keys on the current numeric and symbol layers into a single symbol layer.

Both number 1 to 9 and hyphen are common characters.
With this design, the users even don't need to change to other layers unless some special symbols need to key in.

I already developed a native keyboard app for Android and iOS with this design, and tested with the community. So I'll keep it that way. Thanks!

shiami:
@DavidLRowe I already update the README as you suggested. And I also reply some design questions you gave. Please take a look. If that's okay for you, I'm ready to merge. Thanks!

shiami:
@DavidLRowe After I installed the app, I got the page without images.
Is there something wrong?
Annotation 2020-02-29 000244

DavidLRowe:
Oops. My mistake: I didn't catch the fact that you hadn't included the 
.png images in you package source file (.kps).

You'll need to go to the Packaging tab in Developer, then add those files.

Unfortunately, you'll also need to increase the version (to 1.0.1 for 
example, or 1.1 if you prefer) so that the Keyman.com site will offer 
the new version. This means:

  * Adding an entry in HISTORY.md with the 1.0.1 correction
  * Changing the version number in README.md
  * Changing the version number in taigi_poj.kmn

You'll need to create a new pull request with the revised files.

Let me know if you have any questions.

Thanks,
David

shiami:
@DavidLRowe Done. https://github.com/keymanapp/keyboards/pull/1090

PR #1089, opened 2020-02-28
GFF Musnad

dyacob:
This is the initial commit of the GFF Musnad keyboard for Old South Arabian.


keyman-server:

LornaSIL:
It will be good to have a keyboard supporting this script!
* It looks like a file to `gff.sgw-ethi.gurage/gff.sgw-ethi.gurage.kpj` crept into this PR. I think it needs deleting.
* The welcome.htm, readme.htm and README.md say ` supports the enty of Musnad`. Can you change `enty` to `entry`?
* The keyboard_info file (what will be displayed on the keyboard home page) has the templated description. It would be good to update that to at least the first sentence of what is in the readme.htm. Remember that in the keyboard_info it must all be on one line or it will fail the json test.
* In the .kps file:
   * Details tab: Please select the keyboard font and display font as I think this will be an unusual script for most people.
* In the .kmn file:
   * I'm seeing many names for this keyboard in different files. The one in the .kmn says "GFF Musnad (𐩣𐩯𐩬𐩵)". In general you've had the practice of this format "𐩣𐩯𐩬𐩵 (Musnad)" in the .kmn file. Amharic keyboard had `store(&NAME) 'አማርኛ (Amharic)'` I think it's best to do it that way in the .kmn so that when you select the keyboard you won't see GFF first. But in the .kps and .kpj I think it's fine to have it with GFF. Anyway, I'm not going to be too fussy on that, but I prefer the GFF to be last and not first for when keyboards are sorted online (and I do realize we were not consistent with the SIL ones either originally).
   * For all the keys that are not in the "noMap" store, I think it would be good to fill them in on the OSK: `-=[]\;',./~!@$%^&*()_+{}|:"<>?
* On the mobile layout, longpress for K_H it seems that the 2nd longpress key will output the same thing as the first longpress key. I think the rule should be changed from `Code: T_HETH` to `Code: U_10A6D`
* The gff_musnad.php file looks like it was copied from someone else's project and never updated. Something like this would work:
```
<?php
$pagename = '𐩣𐩯𐩬𐩵 (Musnad) Keyboard Help';
$pagetitle = $pagename;
// Header
require_once('header.php');
?>

<p>
The GFF Musnad keyboard supports the entry of Musnad Script which is also commonly referred to as “Old South Arabian” and “Sabaean” script by different communities. 
</p>
<h1>Keyboard Layout</h1>
</div>
<h2>Mobile/Tablet Keyboard Layout</h2>
<p>Due to the size and number of keys, some characters are hidden in the long press.
Press and hold on the key with a little dot on the top right to reveal them.</p>
<div id='osk-tablet' data-states='default'>
</div>
```

dyacob:
Thanks for the great review @LornaSIL as always.  I believe I have addressed each topic as advised.

LornaSIL:
I think it's good to go, except it might be good to add the font to the keyboard_info file:
```
{
    "license": "mit",
    "languages": { 
         "xsa-Sarb": { 
              "font": { 
              "family": "Qataban", 
              "source": "Qataban-1_0.ttf" 
           } 
        }
     },
	  "description": "The GFF Musnad (𐩣𐩯𐩬𐩵) keyboard for Old South Arabian script."
}
```

Also, I need to add this font to another repo that is used for the Web version.

LornaSIL:
Keyboard is online: https://keyman.com/keyboards/gff_musnad

I'm checking on why the question marks.

dyacob:
Thanks @LornaSIL !  i see the ? also, I tried from my Windows computer where I know I have the Musnad font.  I copied and pasted into MS Word and they were still ?s , so i think the original UTF-8 characters are lost.

LornaSIL:
The keyboard_info file in the repo still has the characters and the .kmp download still has the characters.

LornaSIL:
I just remembered those are SMP characters and that may be the issue.

dyacob:
I've attached a screenshot of how the keyboard info file appears to via the github web interface.  The characters look ok.  I'd guess the SMP characters are getting lost in translation on the way from github to the keyman web site.

MusnadKeyboardInfoInGithub

LornaSIL:
@mcdurdin fixed this and it looks good. Thanks!

PR #1091, opened 2020-03-02
[basic_kbdyba] added m to characters with accent

LornaSIL:
Updated keyboard to support acute and grave accents on `m`.

Since this is a Windows imported keyboard I discussed with Marc who agreed we could enhance the keyboard this way.


LornaSIL:
Notify here when approved:
https://community.software.sil.org/t/yoruba-basic-keyboard/3082

PR #1092, opened 2020-03-04
[munji] submission on behalf of author

DavidLRowe:
Changes made to author's files (DLR 2020-03-03)
- removed copyright statement from readme.htm and welcome.htm
- added | to on-screen keyboard, shift layer, OEM2 key
- touch layout, symbol layer, the » key had U_006B (and therefore produced 'k'), now changed to U_00BB

Work still needed:
- documentation in welcome.htm to cover the Shift+Ctrl layer (four invisible characters), plus the U+00A0 on the Shift+U
- currently using | as symbol for ZWNJ U+200C, but that could be confused with | U+007C on nearby key
- keyboard offers a way to type < U+003C, but evidently not > U+003E
- do you need U+00D7 twice on the desktop shift layer (Shift+Bkquote and Shift+6)?


LornaSIL:
* Please add this link to the README.md: Home: https://keyman.com/keyboards/gff_musnad

* In the .kps file, please make the copyright 2020 SIL International (not just SIL)

* In the .php file change `<div id='osk-phone' data-states='default shift numeric'>`  to `<div id='osk-phone' data-states='default numeric symbol'>`

* The welcome.htm only documents the ZWNJ. I see that U+200D, U+200E and U+200F are also on the CTLR+SHIFT layer. That would be good to document, but since this is considered experimental it wouldn't be essential.

The only thing that is essential here is the .php file getting fixed. I just listed the others since you do need to make that change.

PR #1093, opened 2020-03-04
[newa_romanized] new layout for Nepal bhasa - Newa romanized

sapradhan:
This is another popular layout in Nepal originally by @anandakm. I am submitting this on his behalf.


keyman-server:

LornaSIL:
The keyboard is in good shape. Here are some things that should be fixed.
* newa_traditional is now working on KeymanWeb (but the help file is still a problem which we still need to resolve), so I think in this keyboard we should also use the .ttf instead of the otf font. So please make these changes:
   * Add the ttf to the Pull Request (and perhaps remove the oft)
   * Use the .ttf in the keyboard_info file
   * kps:
      * In the list of files remove the otf and add the .ttf instead
      * In the Keyboards tab make sure to select the font (ttf)
* I would like the README.md file to have this under links: "Home: https://keyman.com/keyboards/newa_romanized"
* On the touch layout, shift layer, change the K_PERIOD code to say U_1144C.
* The touch layout doesn't support U+11477, U+1143E, U+1143F (Desktop does)
 * On the desktop, shift state, K_4 produces U+1142C U+11442 U+200D but you don't have that on the touch layout. Maybe you should include it? If so, you would need to write a rule similar to how you did T_half_ya_zwnj

sapradhan:
>       
>       * Add the ttf to the Pull Request (and perhaps remove the oft)
>       * Use the .ttf in the keyboard_info file
>       * kps:
>         
>         * In the list of files remove the otf and add the .ttf instead
>         * In the Keyboards tab make sure to select the font (ttf)
> 
>     * I would like the README.md file to have this under links: "Home: https://keyman.com/keyboards/newa_romanized"
> 
>     * On the touch layout, shift layer, change the K_PERIOD code to say U_1144C.
done 

> 
>     * The touch layout doesn't support U+11477, U+1143E, U+1143F (Desktop does)
I think you mean U+11447, added that to symbols layer, 
U+1143E and U+1143F are already there in E and shift E respectively 

> 
>     * On the desktop, shift state, K_4 produces U+1142C U+11442 U+200D but you don't have that on the touch layout. Maybe you should include it? If so, you would need to write a rule similar to how you did T_half_ya_zwnj
Added to default layer R

PR #1094, opened 2020-03-04
[finongan] submission on behalf of author

DavidLRowe:
Changes made to author's files (DLR 2020-03-03)
- copied the description from readme.htm for the .keyboard_info file
- in the .php help file, changed "symbols" to "symbol" for OSK display of phone and of tablet

Work still needed:
- perhaps only a bit of editing in the documentation files


LornaSIL:
* The php header isn't quite right:
```
<?php 
  $pagename = 'Finongan Keyboard Help';
  $pagetitle = $pagename;
  require_once('header.php');
?>

```
A quote got deleted somewhere in there.

* In the README.md please add this link: Home: https://keyman.com/keyboards/finongan
* Touch layout for phone:
   * Shift layer on the longpress N there is a tilde N, but that is not on the default layer on the "n" key, nor on the tablet layout, nor on the desktop keyboard. I think it could be deleted.
* Touch layout for tablet:
   * Shift layer on the longpress fullstop, the middle one shows a colon, but produces a semicolon. The keycap needs to be semicolon

I think that's it for now.

DavidLRowe:
@LornaSIL I think I've corrected the issues you identified.

PR #1095, opened 2020-03-04
[newa_traditional] use ttf instead of otf in package, fix some issues in tablet layout

sapradhan:
for
https://github.com/keymanapp/keymanweb.com/issues/10

Some issues in tablet layout were also reported and I have made the changes. Do I need to bump version for these ?


keyman-server:

LornaSIL:
Yes, anytime you touch something that affects the keyboard package, you
need to bump the keyboard version. At a minimum the HISTORY.md, and .kmn
files need a version bump. If you include the version number elsewhere it
should be changed as well.

>

PR #1096, opened 2020-03-05
[pingelap] submission on behalf of author

DavidLRowe:
Documentation may need some tweaks before it's ready for release.


LornaSIL:
Since it's experimental I could approve it as is. If you want to fix the following you can:

* README.md add under Links "Home: https://keyman.com/keyboards/pingelap"
* The keyboard is called "pingelap" in various places. I would have capitalized it, but maybe it's intentional?
* The description, especially in keyboard_info could be enhanced to say "This keyboard is designed for typing the Pingelap language of Micronesia."

PR #1097, opened 2020-03-05
[basic_kbdlv] Latvian Basic Win10 import

LornaSIL:
Deprecating two keyboards.
Added example to keyboard_info from latvian deprecated keyboard.
Relates to Issue #337


DavidLRowe:
Everything looks fine, except touch layout, alt layer has "~" that acts as a dead key but is not the same color as other deadkeys (on that or other layers).

PR #1098, opened 2020-03-06
⦁	[oua] initial submission

Juanan-Ochoa:
This is a keyboard for the berbers languages Tagargrent [oua] and Tumzabt [mzb]


keyman-server:

LornaSIL:
Juan, since you've made changes already you may want to do a new Pull Request. You should make some changes first:
* The folder it is in "o" should be lowercase, not uppercase "O".
* However, it seems like your files are named "teg2". If that's the case, they should be in a "teg2" folder under "t". And my preference is we should not use digits, so if the keyboard is called "Teg" then you could put it in a "teg" folder under the "t" folder and all the files should be called "teg" (lowercase). The keyboard NAME can be uppercase. 
* Please do not commit any desktop.ini files. It's possible that they are hidden on your system, but still being committed. You may need to go to explorer, click on View and select Hidden items. Screenshot:
image

* Next, you've commit the compiled files .js, kvk, kmx. You should not commit those files, you should commit the source files which would be .keyman-touch-layout, .kvks, .kmn
* Also, you are missing a keyboard_info file and some documentation in the source folder. Do you want us to guide you through all these changes or do you want someone to take your files and submit them on your behalf?

PR #1101, opened 2020-03-06
[haisla]

madebybridget:
initial submission


keyman-server:

LornaSIL:
Thanks for your submission. I have an overall suggestion and then I'll comment on specifics.

The previous practice for FirstVoices keyboards was to put the keyboard in a folder called "fv" and then the keyboard folder and filenames would be something like "fv_haisla". Since this seems to be part of the FirstVoices set of keyboards, I would recommend you do that. Doing that has longterm benefits so you'll know which folder all your keyboards are in. However, it makes it messy right now, because you will need to rename all the files from "haisla" to "fv_haisla" and all the internal points to the keyboard will need fixing. Are you able/willing to do that? It IS best to do it now so that the url will be stable (all the keyboard urls are based on the folder name).

Next are my comments:
* The keyboard_info file has a very generic description. This file is used for the homepage of the keyboard and it would be good to update it to something like "This keyboard is for the Haisla language and has been developed by FirstVoices and updated to suit their orthographic changes." Please note that the description field in this particular file must be all on one line or the build will fail.
* In the .kps file:
   * Files: Delete the keyman-touch-layout file. This is a source file and doesn't need distributing, the .js file is the compiled version and you've included that.
   * Keyboards tab: Please add the BCP47 code of "has-Latn"
   * Details tab: Please add the readme.htm under Readme file.
   * Details tab: Please delete 1.0 under version and select "Package version follows keyboard version"
* In the .kmn file:
   * Please go to the On-Screen tab and select "Auto-Fill underlying layout" (bottom right), then click on "Fill from layout" (top left)
   * The touch layout is in pretty good shape. However, here are some comments:
       * I believe that mobile layouts should mostly have the same character set as the desktop layout. As such, I would recommend you delete all the longpress keys on the K_O and SHIFT K_O.
       * On the shift layer, the K_PERIOD is outputting ">". I think if you change the Modifier from "current layer" to "default" it will produce the fullstop.
* The help filename is misspelled and so will not display online. Change to haisla.php (or fv_haisla.php if you are renaming it).
   * This line needs changing from 
```
```
to
```
<div id='osk-phone' data-states='default shift numeric symbol'>
```
And this line needs deleting since you do not have a table layout:
```
```

If you need help with renaming from haisla to fv_haisla I can help you figure that out.

PR #1105, opened 2020-03-10
Initial version of Gilaki keyboards

rmlockwood:


keyman-server:

LornaSIL:
Thanks for your submission. I have a few comments.

On the gilaki keyboard:
* Could you add a more descriptive sentence in the keyboard_info file? It could just say "Keyboard for the Gilaki language based on the Persian layout." or something like that. This is the information that shows up on the keyboard home page.
* In the .kps file you will need to remove the Lateef font and then add it back in from the "\shared\fonts\sil\lateef" folder. 
* In the php help file could you add this at the bottom?
```
</div>
```

On the gilaki_phonetic keyboard:
* Could you add a more descriptive sentence in the keyboard_info file? It could just say "Keyboard for the Gilaki language based on a phonetic layout." or something like that. This is the information that shows up on the keyboard home page.
* In the .kps file you will need to remove the Lateef font and then add it back in from the "\shared\fonts\sil\lateef" folder. 
* In the welcome (and help) file it would be useful to describe the deadkey behavior.
* The help (php) file seems to still indicate it's based on a Persian layout. Should that part be changed?
* In the php help file could you add this at the bottom?
```
</div>
```

I hope this will work! In the future it would be best to do this in three Pull Requests. Add a font in one Pull Request (we may still have to do that before the keyboard will compile). Then do each keyboard in a separate Pull Request (I find that easier to manage). Try updating these issues I found and we'll see if it compiles without putting the font in master first.

rmlockwood:
Thanks for catching these things. I’ve made the changes you requested.

 

Thanks,

Ron

LornaSIL:
These two keyboards are now online.

https://keyman.com/keyboards/gilaki

https://keyman.com/keyboards/gilaki_phonetic

rmlockwood:
Thanks!

PR #1108, opened 2020-03-10
[basic_kbdhela2] Greek (220) Latin Basic Win10 import

LornaSIL:
Deprecating two keyboards.
Relates to Issue #337


DavidLRowe:
Some minor issues with the touch layout:
- key in lower left is labeled "Alt" in default layer and "rightalt" in shift layer.
- shift key (up arrow) doesn't change color when in shift layer
- shift key doesn't take you out of shift layer (but fortunately the dead keys do!)

PR #1111, opened 2020-03-11
[basic_kbdno1] Norwegian with Sami Basic Win10 import

LornaSIL:
Deprecating two keyboards.
Relates to Issue #337


DavidLRowe:
Keyboard produces Ÿ (as well as ÿ), but only ÿ is listed in documentation.
Keyboard produces ỳỲ, but not listed in documentation.
Keyboard produces ŷŶ, but not listed in documentation.

Touch layout, default layer, 4th row, next to last key: key top shows _ (U+005F) but produces - (U+002D)

PR #1115, opened 2020-03-12
[basic_kbdsn1] Sinhala Basic Win10 import

LornaSIL:
Deprecating one keyboard.


DavidLRowe:
The source folder includes a .keyman-touch-layout file, but it's not referenced in the .kmn file (and the README.md doesn't include touch in supported platforms).

PR #1116, opened 2020-03-12
[basic_kbdiulat] Inuktitut - Latin Basic Win10 import

LornaSIL:
Deprecating one keyboard.

This keyboard supports two scripts, but as the keyboard was designed by MS it has been agreed we will allow it (otherwise we request people not to try to support two scripts in one keyboard).


DavidLRowe:
OSK has problems:
default: `\` on `'` key, but produces `'`
shift: `|` on `'` key, but produces `"`
rightalt: U+0142 on backquote key, but produces nothing; nothing on `'` key, but produces U+0142
shift+rightalt: `|` on `'` key, but produces U+0141

DavidLRowe:
Also OSK has LC layer with only a space. Probably not an issue, but if you're changing the OSK anyway, you could correct that.

PR #1120, opened 2020-03-12
[basic_kbdsorex] Sorbian Extended Basic Win10 import

LornaSIL:
No mobile layout.
Deprecating one keyboard.


DavidLRowe:
There actually is a mobile layout.

PR #1121, opened 2020-03-12
[basic_kbdsors1] Sorbian Standard Basic Win10 import

LornaSIL:
No mobile layout.
Deprecating one keyboard.


DavidLRowe:
In welcome.htm and .php file: change `´a will produce ć` to `´c will produce ć`

PR #1122, opened 2020-03-12
[basic_kbdsw09] Sinhala - Wij 9 Basic Win10 import

LornaSIL:
No mobile layout.
Deprecating one keyboard.


DavidLRowe:
shift+backslash is blank on OSK and welcome.htm image, but produces |

everything else looks fine.

PR #1123, opened 2020-03-12
[cabecar] initial submission on behalf of author

DavidLRowe:
I made one change to the keyboard: `Shift+[` now does the same as unshifted `[` (that is, a deadkey for acute accent). I incorporated that change in the OSK. Before the `Shift+[` was unlabeled on the OSK, but producing `{`.

Eventual changes needed:
- Description of keyboard in
  - cabecar.keyboard_info
  - readme.htm
  - welcome.htm
  - cabecar.php
- .pdf file needs to be updated to reflect the `Shift+[` deadkey


LornaSIL:
The keyboard is not well documented, but as an experimental keyboard we can accept it that way. The description would be good to update when the author has time. Also, the deadkeys are not documented and that should be in the welcome.htm and php file.

On the symbol layer
* the "}" is not outputting what is on the keycap
* the Euro is not outputting a euro
* the upside down exclamation is outputting a euro

I would like to see those 3 things changed and then I can approve it.

PR #1124, opened 2020-03-12
[basic_kbdcherp] Cherokee Nation Phonetic Basic Win10 import

LornaSIL:
New. No deprecations.


DavidLRowe:
When I type the following, I get nothing (rather than what welcome.htm documents):
```
s x
s X
n h
n H
```
Otherwise seems fine.

PR #1125, opened 2020-03-13
[wakhi_cyrillic] submission on behalf of author

DavidLRowe:


LornaSIL:
The keys on the shift layer which output multiple characters aren't working. I think they need the Modifier set to "shift". 
* longpress on K_W
* longpress on K_P
* longpress on U_0425
* longpress on K_D
* longpress on K_L
* longpress on U_0492
* longpress on K_C
* longpress on K_N
* longpress on U_04B6
Otherwise it looks okay.

PR #1126, opened 2020-03-13
[ishkashimi_cyrillic] submission for author

DavidLRowe:


LornaSIL:
I am concerned about the use of two Latin characters in this keyboard. Does the orthography truly use U+00F5 and U+00D5? I would think that they should use Cyrillic characters (not mixing characters from two scripts) which would mean using `U+043E U+0303` and `U+041E U+0303`.
```
+ [K_S] > 'õ'
+ [SHIFT K_S] > 'Õ'
```
to
```
+ [K_S] > U+043E U+0303
+ [SHIFT K_S] > U+041E U+0303
```

PR #1127, opened 2020-03-18
GFF Tigrinya-ET v1.3.9

dyacob:
This version introduces an initial touch layout intended for evaluation prior to finalization.


keyman-server:

LornaSIL:
It's great to see a mobile layout coming in for Ethiopic! I do have a lot of comments on the mobile layout, just a few on other things.

* HISTORY.md needs the version number added besides the date.
* BCP47 tag:
   * keyboard_info gives `ti-Ethi-ET` and kps gives `ti-ET`. I can't remember which is correct, but they should be consistent.
* Version in .kmn is invalid. You cannot use abc, etc.
* Mobile layout:
   * **default** layer
      * 1st row, 7th key from left (K_Y): none of the longpress keys output the correct character
      * 1st row, 9th key from left (shift K_K): none of the longpress keys output the correct character
      * 2nd row, 1st key on left: keycap shows 12D2 but outputs 12D0
      * 2nd row, 2nd key on left: 3rd longpress shows 12A4 but outputs 12A3
      * 2nd row, 8th key on left: longpress keys all outputting wrong character
      * 3rd row, last key on right: many of the longpress keys are not producing what is on the keycap. Some appear to be duplicates as well.
      * 4th row, 1st key: the output seems to be dependent on context. If that's not what you want, you may wish to change them to use the U_12A0 code.
   * **numeric** layer
      * 2nd row, 2nd key from left: first longpress outputs backslash rather than forward slash. Is it even necessary since the forward slash is on the main keycap?
      * 2nd row, 3rd and 4th keys from left: outputting Ethiopic punctuation rather than Latin.
      * 2nd row, 7th key from left ($): There's a $ on the main keycap and also on longpress. Is that necessary?
      * 2nd row, final key on right: keycap is `"` but it produces `*`. Since you have `"` on the longpress this should probably have `*` on the keycap.
      * 3rd row: comma key produces Ethiopic comma.
      * 3rd row: `'` key produces `]`.
   * **numeric2** layer
      * 2nd row, 1st key from left is producing gemination mark and not `_`
      * 2nd row, Yen key is producing wrong character. (00A4)
      * 3rd row: comma key produces Ethiopic comma.
      * 3rd row: `'` key produces `]`.
   * **layer2** layer
      * 1st row, all of the keys have the same thing on the longpress as on the main keycaps (a number of other keys on this layer do that too). I don't think there is a need to duplicate them.
      * 2nd row, 4th key over: all characters (including longpress) are producing something different than is on keycap
      * 2nd row, 5th key over: all longpress keys are producing something different than is on keycap
      * 2nd row, 6th key over: all characters (including longpress) are producing something different than is on keycap
      * 3rd row, last key on right: many of the longpress keys are not producing what is on the keycap. Some appear to be duplicates as well.
      * 4th row, 1st key: the output seems to be dependent on context. If that's not what you want, you may wish to change them to use the U_12A0 code.

dyacob:
Hi @LornaSIL , really sorry to give you so much work there, too much late night work on my end I'm afraid.  Your feedback is terrific, so very thorough, The SIL is really lucky to have you!

I've made the corrections and tested.  A general remark on where a symbol shows up again in a longpress, this is mirroring the iPhone behavior.  The two "numeric" layers should be nearly identical to the iPhone keyboard.

LornaSIL:
Looks good. The BCP47 tag should be changed to match in keyboard_info and .kps. Otherwise I'm happy with approving it. Let me know when you are ready.

dyacob:
I think I got the languages synced now.  Language settings are only in the .kps and .keyboard_info files.  Previous language identifiers in the .kmn file have been removed, since they were set with deprecated variables.

PR #1129, opened 2020-03-23
[me_en] submission on behalf of author

DavidLRowe:
This seems complete enough that it can go directly into release.


LornaSIL:
* I would prefer to see the name of the keyboard changed everywhere from "Me'en" to "Me’en (Latin)" (README.md, HISTORY.md, .kps/Details/Package name, welcome.htm, readme.htm, php title.)

* Please add this in the README.md under Links:
`Home: https://keyman.com/keyboards/me_en`

* It would be nice to have more of a description in the keyboard_info file (at a minimum the same thing the readme.htm file uses).

* The php file indicates there is a tablet layer. That should be removed.

* There are strange longpress keys on g, h, j, l, n, m. I don't know that they are necessary as all those keys are also on either the numeric or the symbol layer. It makes it confusing, but I would not require them to be removed before I approve it.

PR #1130, opened 2020-03-24
[sil_mali_qwerty] Changes made by author

LornaSIL:
I think author wants to check it first.


LornaSIL:
I think the author is happy with the changes made.

DavidLRowe:
There are some combinations (,L ,S ,F ,z ,Z) that are not represented on mali_all.png and that diagram indicates that ,/ should make / but that is not the case. The PDF has more detail and includes the combinations missing from the diagram.

PR #1131, opened 2020-03-25
[clavbur9] Clavier du Burkina v9.1

LornaSIL:
This version adds a mobile layout.


DavidLRowe:
I could approve this "as is", but I have some minor questions:

There's no upper case for ;G ,I ;Z (in either .kmn or touch layout). Any reason not to add them?

OSK has a Ctrl layer with just a space character. It seems that that could be removed.

I expect this is just an issue of font, but I'll note that when I use the accent key (to right of space bar), I see some accents with dotted circles and some without.

The `«` (longpress key on `[` on numeric layer) produces U+00A0 as well, but the `»` doesn't produce U+00A0 before it. It seems like they should be consistent. (One could also argue that ‹ and › should also have no-break spaces, but those aren't used that often.) I wonder if the logic in group(angleit) is faulty.

LornaSIL:
Ok, I think I've fixed these issues.
The issue with the accents is that for some of them I'm using the spacing version on the keycap because they display better when a font doesn't include 25cc, but some of the combining marks do not have spacing versions and so I've used the combining mark on the keycap.

PR #1134, opened 2020-03-30
[nrc_crk_cans] Update to v1.1.0

eddieantonio:
This is a minor release of the keyboard. Since I deleted three keys, I'm not sure whether this should be v2.0.0 or left as v1.1.0. Please see `HISTORY.md` for a list of updates.


keyman-server:

LornaSIL:
If it's a minor update it fine to go from 1.0.1 to 1.1.0.

LornaSIL:
I see two issues:
* On the numeric layer you have:
```
              {
                "id": "K_EQUAL",
                "text": "=",
                "layer": "latin"
              },
```
I don't think you want a "latin" layer which doesn't exist. Maybe "default"?
* The build doesn't work because your .kps file is asking for two files that were .jpg and now are .png. So, remove layout-numeric.jpg and layout-punctuation.jpg and then add the .png versions back in.

eddieantonio:
That was a copy-paste error! I have updated the files specified 👍

PR #1143, opened 2020-04-02
[rac_hazaragi] Rachitrali Hazaragi version 1

LornaSIL:
New keyboard on behalf of author.


DavidLRowe:
README.md
- "Home" link references "rac_pashto" Should it be "rac_hazaragi"?
- Should "Web" link use `#haz-arab` (instead of `#haz`)?
- There's a problem with the example

According to the example in the .keyboard_info file, typing these keys:
`Hzragi`
should produce this text: 
06BE 0632 0627 0631 06AF 06CC 
However, it seems to produce this text:
062D 0632 0631 0627 06AF 06CC 
There's a problem with the first character (`H` or `h`) and with the order of the `r` and `a`.
@rachitralli Can you provide a correction?

The source/help/ folder has unneeded files: Thumbs.db and rac_hazaragi.php.bak

Everything else seems fine.

rachitralli:
@LornaSIL @DavidLRowe thanks

PR #1145, opened 2020-04-03
Fix ? on touch layout shift layer

DavidLRowe:


DavidLRowe:
relates to #1138

DavidLRowe:
Author will fix this in another PR.

Issue #1149, opened 2020-04-08
README.md was deleted by #1055


mcdurdin:
README.md was deleted by PR #1055. It needs to be restored.

When it is restored, add a link to https://help.keyman.com/knowledge-base/95 for mac configuration.

DavidLRowe:
Should be fixed by #1150

PR #1150, opened 2020-04-08
restore README.md and add Mac link

DavidLRowe:


DavidLRowe:
@mcdurdin Check that the added link is what you expected

fixes #1149

mcdurdin:
> @mcdurdin Check that the added link is what you expected

Can you move it up into the mac prerequisite section? (note that the kmcomp.x64 compiler is also needed but I haven't yet had the time to get the latest version into the PR and test then prepare it for release).

PR #1155, opened 2020-04-10
[english_12_keys] Version 1.0

bennylin:
New keyboard based on my previous keyboard's rotation key. Reminded me of the old phone's keypad. 

Benefit: larger buttons helps people with big fingers or visually impaired.

touch


keyman-server:

LornaSIL:
This is quite creative! I'm not sure that there will be a wide user-base for such a keyboard though as flip phones were quite difficult to text with. It might be best if you just distribute this one yourself and not through the Keyman repository.

bennylin:
I'm not sure if number of downloads or number of user base is the determining factor, since this is a Latin script layout, and beside English, the alphabet is used in many other languages. Also there's no metrics of the usage/download of each layout. And as I wrote in the documentation, this layout could be particularly useful for people with big fingers (me included, I personally like this), or sight problem, or old people, or people unfamiliar with qwerty layout. I personally believe there would be more people using this than some of the layout for minority languages with a handful of speakers.

But I take your feedback as the final word, and closing this request.

PR #1156, opened 2020-04-11
[sanjha_punjabi] Sanjha Punjabi keyboard on behalf of author

LornaSIL:


DavidLRowe:
Two minor points:

README.md needs (c) symbol

readme.htm and welcome.htm have "Sanjha Punjabi" whereas .keyboard_info has "Western Punjabi". Is that OK?

PR #1157, opened 2020-04-12
[indonesian_15_keys] Version 1.0

bennylin:
Based on the 12-keys keyboard, and adapted it for Indonesia keyboard.


keyman-server:

LornaSIL:
Along with PR #1155 we will ask you to distribute this keyboard through your own channels rather than the Keyman github repo. It doesn't seem to be a keyboard that would be widely used.

PR #1160, opened 2020-04-15
[sil_yoruba_dot] Yoruba with Dot keyboard

LornaSIL:
Deprecating one keyboard.
Relates to Issue #337


DavidLRowe:
touch layout, numeric layer, `|` is on a longpress of `\` but produces `\`.

Does it matter that ^ ~ ` are on desktop but not touch?

Desktop layout has `c` and `C` on the rightalt and shift-rightalt layers and they work. But was the intent to have `v` and `V` there? (This is also the case for the sil_yoruba_bar keyboard I just merged.)

Issue #1159, opened 2020-04-14
[sil_euro_latin] OSK issues on old Android devices (SDK  19 / KitKat)


darcywong00:
Noted from keymanapp/keyman#2977

The sil_euro_latin keyboard has OSK issues on old Android devices. 

<img src="https://user-images.githubusercontent.com/7358010/78859031-474a8c00-7a58-11ea-8bf6-6b9825f6e40d.png">

The legacy european2 keyboard used DejaVuSans.ttf, but the font isn't included in the sil_euro_latin keyboard package ~or .keyboard_info file~.

darcywong00:
Resolved

PR #1162, opened 2020-04-15
[sil_euro_latin] Updated kbd to include DejaVuSans font in the package

LornaSIL:
Relates to PR #1159 
* Added font to "shared\font" folder
* Added font to package
* Added font to touch layout
* I did not do anything to keyboard_info as that looks correct to me. Please let me know if I'm wrong.


DavidLRowe:
Should this close #1159 ?

PR #1163, opened 2020-04-17
[rawang] Rawang keyboard moved to release

LornaSIL:
I had to decompile the .kmx to get to the source so it's not pretty.
Deprecating two keyboards.
Relates to Issue #337


darcywong00:
I was wondering why rawang.kmn shows as a binary file on this PR.
When I open it in notepad++, it says the file encoding is "UCS-2 LE BOM". I would have expected "UTF-8-BOM". 

Maybe kmdecomp is creating "UCS-2 LE BOM"?

DavidLRowe:
Version number in README.md doesn't match that in HISTORY.md and .kmn files.

Some of the characters used for accents and other things have RAlt keys (<>/[]) and others don't, but everything is documented, so except for the version number everything looks fine.

PR #1166, opened 2020-04-21
[numanggang] on behalf of author

DavidLRowe:


LornaSIL:
* I would like to see this link in the README.md file: `https://keyman.com/keyboards/numanggang`
* I would like to see "Keyboard for typing Numanggang" in the readme.htm.
* The header for the php file needs the name changed
* In the welcome.htm file it says: `Type the | key to get the "Ŋ" character. Type the | key again to get the "Ŋ" character. Type the | key a third time to get "ŊŊ". ` This should be changed to `Type the | key to get the "Ŋ" character. Type the | key again to get the "|" character. Type the | key a third time to get "ŊŊ". ` (That is when it is typed twice it is "|".)
* In the .php file it says `Type the | key again to get the "Ŋ" character.` Again, that should be changed to `Type the | key again to get the "|" character.`
    * Also in the php file the "Touch layout" section needs to say `osk-phone` rather than just `osk`.

I'm having problems testing the touch layout, so do that as soon as I can.

DavidLRowe:
@LornaSIL Thanks for catching those errors. I think I've corrected them all.

PR #1173, opened 2020-04-22
[turkmen_cyrl] Turkmen Cyrillic keyboard

LornaSIL:
Deprecating two keyboards.
Relates to Issue #337


DavidLRowe:
Minor issues on touch layout:
shift layer, shift key has no arrow
numeric layer, `|` produces U+04D8
Otherwise looks good.

PR #1177, opened 2020-04-23
[hindi_modular] Hindi Modular keyboard

LornaSIL:
Used source from internal repo to add source so it can be in release.
Deprecating one keyboard.
Relates to Issue #337


DavidLRowe:
In the source/help folder there's a spurious hindi_modular.php.bak file that should be removed. Other than that, everything looks good to me.

PR #1179, opened 2020-04-24
[bod] Take 3 - Adding packageFilename to keyboard_info

LornaSIL:
Documentation (https://help.keyman.com/developer/cloud/keyboard_info/1.0/) indicates this is necessary for packages in the legacy folder:
> string, filename of the .kmp file that will be distributed, relative to the keyboard base folder, including extension, required for experimental/ and legacy/ folders. If no .kmp exists, then keyboard is js only. For release/ folders, this must be id/build/id.kmp.


LornaSIL:
We can but hope.

PR #1181, opened 2020-04-24
[cjk] Take 2 for Chinese, Japanese and Korean RR - add kmp files

LornaSIL:
Added kmp files to source folder
Added package info to keyboard_info
Relates to Issue #337


LornaSIL:
Something is still wrong with korean_rr

PR #1184, opened 2020-04-24
[korean_rr] Add kmp file

LornaSIL:


LornaSIL:
@mcdurdin I don't know what's wrong with this.

DavidLRowe:
Build log has:
korean_rr.keyboard_info: Error: Invalid .keyboard_info file: version field is "1.1" but should be "1.0"

The kmp.inf file has version 1.0 whereas the .keyboard_info file has version 1.1.

DavidLRowe:
korean_rr.js has `this.KBVER="1.1";` so maybe the .kps/.kmp need to specify 1.1?

LornaSIL:
Thank you. I was trying to figure out if it was the language tag, but that seemed fine. The version in the .js file is 1.1 and we can't change that, so I had to change it everywhere else. I really don't understand the history of the file, so the HISTORY.md file is rather weak.

mcdurdin:
Good news: I have located the source for this keyboard. We own the copyright (developed originally by Tavultesoft): https://github.com/sillsdev/keyman-keyboards-internal/tree/master/Chinese-Japanese-Korean%20KMW/New/Korean. There's a bit there to weed through, but looks like with time we'll be able to move it to release/

LornaSIL:
@mcdurdin I'll need some help with this. I don't know how to work with IMEs in Keyman. 
When I use the korean_rr that is online: https://keymanweb.com/?_ga=2.171456034.325236962.1589815358-2039366057.1561996340#ko,Keyboard_korean_rr

it works.

When I try to type in **Keyman Developer / Test keyboard**, it just outputs the text on the keycap and doesn't offer the other characters in the IME.

Then when I go to test it online (from the Keyman Developer / Build / Open debugger in local browser) I get this message:

image

I also get that message when I try it on my phone.
I'm assuming there is something basic that I'm not aware of needing to set?

LornaSIL:
@mcdurdin 

Ok, I now have part of the problem sorted out. I added all the `hide_ime.call_js` (etc) files to the source folder. (Personally, if a kmn file calls for a file I think it should complain when it builds - which it doesn't).

So, at this point the web version and tablet version are working. I'm not getting the IME to display choices on Desktop. I've tried in the test Window of Developer and I've tried installing the .kmp. I'm not sure what else to try.

I could commit what I have to this PR and have you look at it to see what I'm missing. If I do that, I have another question. It looks like there is an Access db that you may have used to create all the `_js` and/or `korean_ime.js` files. Should I commit that to an "assets" folder? If you want me to commit it, should I commit the .zip with mdb inside, or just commit the .mdb? If it's needed for recreating any of these files we should maybe commit it. But, all the Unicode files in the folder shouldn't need committing. Comments?

mcdurdin:
> (Personally, if a kmn file calls for a file I think it should complain when it builds - which it doesn't).

Can you open an issue on this with a repro?

> I could commit what I have to this PR and have you look at it to see what I'm missing.

Sure -- but I'm not sure when I am going to be able to dig into this; I'm pretty over-committed at present.

> It looks like there is an Access db that you may have used to create all the `_js` and/or `korean_ime.js` files. Should I commit that to an "assets" folder? If you want me to commit it, should I commit the .zip with mdb inside, or just commit the .mdb? If it's needed for recreating any of these files we should maybe commit it. But, all the Unicode files in the folder shouldn't need committing. Comments?

Not really sure -- perhaps commit it all in a zip file except for Unicode data text files which can be retrieved from unicode.org.

LornaSIL:
Wow, it passed! @mcdurdin are you able to review this? My problem is that it now compiles, but I don't see the IME popup and I don't know how to test it.

I put the Access db files in an `extra` folder as you suggested.

mcdurdin:
> My problem is that it now compiles, but I don't see the IME popup and I don't know how to test it.

So I have tested this here now. The web keyboard seems to be working (desktop only, requires On Screen Keyboard to be visible -- doesn't currently support touch devices); if you try typing say, 'ta', you'll see a window popup over the OSK: 
image

On the desktop (Windows only), it will combine the jamo into hangeul when you press spacebar but does not present an IME.

Given this, the only targets selected should be `windows` and `web`.

I have a long-term pathway forward for this; it will depend on the LDML keyboard project being completed, but then we will be able to introduce cross-platform IME support services for this kind of interaction and have it work on all devices, with the ability to customise the logic of the IME for each language. But this concept is a long way off for now.

languagewarrior:
How do you tag Keyman with an existing fontype and language dialect

mcdurdin:
> How do you tag Keyman with an existing fontype and language dialect

This might not be the right place for this question -- unless your question relates to the korean_rr keyboard, can you ask at https://community.software.sil.org/c/keyman?

PR #1185, opened 2020-04-24
[sundanese] Version 1.0.4 

Ilhamkang:
Dear @LornaSIL , I have updated the Sunda Baku keyboard with some fixes. Please check it. 

Regards,


keyman-server:

LornaSIL:
@Ilhamkang please let me know when you are ready for this PR to be reviewed. Thanks.

Ilhamkang:
Now it is ready to be reviewed. 

Actually I am going to make another new PR for new "Sundanese Latin" keyboard after this update. I have made the folder called sundanese_latin with the keyboard layout and other files. Somehow it included into this PR. So I deleted everything on sundanese_latin folder, just to make no confusion.

Ilhamkang:
Thank you @LornaSIL for approving this update. 

One question, why the built result from Keyman's server of this keyboard mobile version looks a bit different than the built result from my local drive? It seems there is no dotted circle U+25CC ◌ on every dependent vowel and final or medial consonant. I'm using Android 5.1.1. 

This is built from my local drive (.kmp imported to the mobile phone from PC) 
local built


This is built from Keyman's server 
from keyman's server

regards,
Ilham

LornaSIL:
I am not sure the complete answer for the dotted circles question, but it is true that different fonts and rendering systems will sometimes display or not display the dotted circles. Sometimes I see square boxes for the dotted circle! So, I don't think we can expect the user experience to always be the same. Hopefully some day...

Ilhamkang:
OK, thank you @LornaSIL .

PR #1152, opened 2020-04-08
[fv_all] Add JS keyboards to FirstVoices keyboard package

darcywong00:
Currently, fv_all/build.sh gathers First Voices .kmx keyboards into a single package.
For Keyman 14, the mobile apps will be dealing strictly with .kmp keyboard packages.

This PR updates the build.sh script to also include touch layout keyboards needed for the mobile First Voices app.

PR #1190, opened 2020-04-29
[fv-all] Version bump to trigger build of fv_all.kmp

darcywong00:
#1152 didn't update the version in fv_all.kps.in so it didn't get built.

Also added DejaVuSans.ttf shared font to go with sil_euro_latin


DavidLRowe:
@madebybridget  FYI: This pull request increments the version number so that the build system will detect that it needs to build a new version.

PR #1193, opened 2020-04-30
[vm_tamil_modular] Move source to release

LornaSIL:
Deprecating one keyboard.
Relates to Issue #337


DavidLRowe:
welcome.htm
- entry ம in leftmost column has keystroke 'k' but should be 'e'
- no keystrokes for 6th and 7th entries in Grantha Consonants column (perhaps `['` and `[^`)
- similarly keystrokes missing for corresponding columns in Grantha section of chart

DavidLRowe:
In Grantha section of chart
- second row, the `{''` entry has an extra `'`
- fourth row, the `[}]` for the consonant should all be `}` -- something I didn't see first time

PR #1194, opened 2020-04-30
[vm_tamil_typewriter] Move source to release

LornaSIL:
Deprecating one keyboard.
Relates to Issue #337


DavidLRowe:
OSK/welcome.htm images have
- blank for Shift+Q, but keyboard produces Q
- blank for Shift+A, but keyboard produces A
- ' for Shift+8, but keyboard produces *

In welcome.htm, under Keystroke Examples, in the Grantha Consonants column, there are no keystrokes for the 6th and 7th entries. It seems as though these might be `$[` and `${`.

In welcome.htm, in the Grantha Consonants chart, there are two columns (corresponding to the above) where there are no keystrokes listed.

In welcome.htm, at the bottom the * is not listed with other punctuation.

PR #1198, opened 2020-05-01
[sil_ipa] Add hotkeys to access options and rename keyboard

### LINKED ISSUE #964: [sil_ipa] Add hotkey so macos and linux users have access to options
Current MacOS and Linux users do not have access to the options dialog to change to "before" deadkeys.

See: https://community.software.sil.org/t/is-it-possible-to-get-an-ipa-keyboard-where-the-dead-keys-come-before-instead-of-after/1872/16

LornaSIL:
Added hotkeys
Renamed keyboard to match other SIL keyboard names with "(SIL)" after the name.
Closes #964


DavidLRowe:
.kps file still lists "SIL IPA" for .kmn and .kps files. Will that be a problem?

extras folders has .doc (and .pdf) with "SIL IPA", version 1.5 and NRSI. Does that need updating?

Eventually I got the RAlt keys to work for changing the order (on Windows; haven't tested any other platform). 

Should I approve this as is, or would you want to change the .doc file and/or the .kpj file?

PR #1204, opened 2020-05-06
[thamizha_bamini] Keyboard moved from legacy to release

LornaSIL:
Deprecating two keyboards.
Relates to Issue #337


DavidLRowe:
some changes needed in welcome.htm
- Quickstart paragraph 1: documentation talks about short and long vowels being near each other, but the two examples are wrong: instead of `q` and `a`, those characters are on `Shift+m` and `m`; instead of `x` and `c`, `Shift+x` and `x`. Not sure I know enough to comment on the rest of the paragraph.
- Quickstart paragraph 2: Example shows `b` and `q`, but it's really `q` and `h`
- Pulli Marks and Vowels paragraph 1: Pulli mark inserted with `;` (not `f`). The rest of the paragraph doesn't seem to be implemented, that is typing `f` repeatedly gives `ககக` rather than `க்கக்` as the text indicates.
- Keystroke Examples section: The keystrokes for the two examples are not correct. Use the ones from the Keyboard Reference Chart.
- Keyboard Reference Chart (items identified by keystrokes)
  - q Shift+F should be q -
  - q Shift+$ should be q - -
  - Shift+N Shift+Q entry has characters U+0B9E U+0BC8, which is a duplicate of the column to the right. It seems that these characters should be U+0B93 U+0BC7
  - n d s (last item in last row before the Grantha Consonants section): there's an extra `s` after the Tamil characters
  - In \ row for Grantha consonants, n ] should be n \ and Shift+N ; should be Shift+N \ 
  - In ]] row, replace ]] with ]
  - Final row of table lacks keystrokes
  - Punctuation: There are some symbols not mentioned: ' (second way to type U+0027) and = then on the shift layer: ~ * _ + { } (Up to you whether to include these or not)

DavidLRowe:
welcome.htm has opening text "ah," -- is that intentional?

In your correction to the Quickstart paragraph you now have: Shift+m above m and Shift+x beside x. (I wouldn't have bothered mentioning it if I had not already made a comment. Up to you whether you want to change it or not.)

LornaSIL:
@DavidLRowe I'm afraid to touch the documentation anymore as I don't understand the Quickstart wording.

PR #1205, opened 2020-05-06
[thamizha_new_typewriter] Keyboard moved from legacy to release

LornaSIL:
Deprecating two keyboards.
Relates to Issue #337


DavidLRowe:
welcome.htm: In "Keystroke examples" section, Grantha column, last entry
should be `b$s` (rather than `$fs`)

In "Keyboard Reference Chart" (cells identified by keystroke letters) - first section
- `kP` entry has extra character (characters for `k;kP` rather than just `kP`)
- `"{` entry seems like it should be `Q` rather than `"{`
- `Q` entry seems like it should be `q` rather than `Q`
- These entries don't seem to produce the corresponding character:
  - `"{`
  - `l{`
  - `Qh`
  - `Jh`
  - `Eh`
  - `k{`
  - `u{`
  - `Yh`
  - `H{`
  - `s{`
  - `Wh`
  - `Dh`

In "Keyboard Reference Chart" (identified by keystroke letters) - Grantha section
- `^` row: seems like `^` should be `&` in entire row (as for the `%` and `^` columns)
- `]` row: seems like `]` should be `+` in entire row (as for the `%` and `^` columns)
- `B` row: seems like `B` should be `\` for first part of row (as for the `%` and `^` columns), but after that neither `B` nor `\` seems to work
- NB: the .kmn file has two entries for `"b"` -- perhaps one should be for `"B"`??

In "Keyboard Reference Chart" (identified by keystroke letters) - Other Characters section
- first digit `0` should be `+`
- `+` should be `|`
- second `.` (that indicates it produces `?`) should be removed
- digit `9` should be `(`
- second digit `0` should be `)`
- Add `;` and `:` ?

OSK: some deadkeys indicated with square brackets (bhnpABPZ) but others aren't (i[]{}%^)

LornaSIL:
I think I made all the documentation changes. I didn't add `[` and `]` around the non-deadkeys. I am vacillating how to handle it because many of them are not deadkeys, but they only produce something in context with other keystrokes. I'm inclined to leave it, but if you have other suggestions feel free to offer them.

DavidLRowe:
My comment above about changing `"{` to `Q` was wrong. I should have been `"[` to `Q`. So now the `Q` should move left one column (from the `C` column to the `c` column), and what is currently `Q` (in the `C` column, should be `Q}`.

`\bs` (final row, final column in Grantha section) should be `\Bs` (rather than `\bs`).

I don't have a strong opinion about marking the keys with [x]. The fact that they are Latin keys on a Tamil script keyboard will probably indicate they do something different.

PR #1206, opened 2020-05-07
[thamizha_anjal_paangu] Keyboard moved from legacy to release

LornaSIL:
Deprecating two keyboards.
Relates to Issue #337


DavidLRowe:
Under "Examples", the second Phrase seems to have an extra "<" at the end.

In the Keyboard Reference Chart, it seems that the key sequences for "o" and "oo" are switched with the ones for "u" and "uu". For example, the character above the "ko" key sequence is actually produced by "ku" (as the column header would indicate).

Entry for `ma` seems like it should have character `ம` rather than `மா` (which is listed for `maa` in the next column).

Under the "Troubleshooting" section, I'm not seeing the "Correct output" in Keyman web:
Assuming this is referring to the `g` and `a` keys
- when I type `gag` I get `கக்` (which is what I'd expect, but not what "Correct output" lists)
- when I type `gggg` I get `க்க்க்க்`
Is this section still relevant?

PR #1207, opened 2020-05-07
[thamizha_tamil99_ext] Keyboard moved from legacy to release

LornaSIL:
Deprecating one keyboard.
Relates to Issue #337


DavidLRowe:
In welcome.htm:
Under Examples, the keys for typing "alphabet" don't seem to be complete.

Under "The SRii Character", the keystrokes don't seem to create the character.

Under "Other Characters", the first set of Character and Keystrokes columns don't line up very well. Eventually I figured out that "debit" was "V" not "B"! But this is minor, so don't go to too much trouble to fix it.

In "Keyboard Reference Chart", the row with key "[" in the first column is missing "[" in all the following columns.

Under "Other Characters", `I` indicates it will insert U+0BCD U+0BB0 U+0BC0, but it seems to insert U+0BB6 U+0BCD U+0BB0 U+0BC0 (that is, an extra U+0BB6 first). 

(Presumably the .php help file needs these changes as well.)

PR #1214, opened 2020-05-13
Initial version of Esperanto keyboard

anvalon:
This is the initial version of a simple Esperanto keyboard for Keymann Desktop


keyman-server:

LornaSIL:
@anvalon Thanks for your keyboard submission. There is one major thing that needs to change and then I see a few things that would enhance the keyboard.
* We already have a keyboard whose ID is `esperanto` in our `legacy` keyboards: https://keyman.com/keyboards/esperanto. Besides having a different layout than your, it is too late to change that ID so you'll need to choose another ID (even though your is in the release folder). You can leave the Keyboard name the same, just the folder and all the filenames need to change. So, maybe `esperanto_x` or `esperanto_uniphone` or `esperanto_standard` or something like that. Unfortunately, that means you have to change the file associations and paths in the .kpj and .kps files.
* Could you add this text `This keyboard is designed to enable simple input of the Esperanto special characters.` to the README.md and keyboard_info files? Although neither of those are used in the downloaded kmp, it would improve the home page for the keyboard.
* Esperanto has a language tag of "eo" and so that is the language tag you should use in keyboard_info and in the .kps. IANA (https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry) indicates the script should be suppressed, so just change "it-IT" to "eo" in both files.
* You have a `esperanto\source\welcome.htm` and `esperanto\source\welcome\welcome.htm`. I think you could delete the welcome folder with the welcome.htm in it since the one in source seems to look very good.
* In the .kmn file you have a target of "any" and "windows". You could uncheck "windows".
* Lastly, we have 3 keyboards in `legacy` for Esperanto. None of them have all the source files or all the files we require to go into `release`. It's on my list to move these keyboards to `release`. One of them matches your keyboard, so it would be great if we could deprecate that one. It means you would need to add a few lines to your keyboard_info file. I hope you don't mind doing this. If you can update keyboard_info (with different language tag, different description, and add the deprecation), this is what I think the whole file should look like:
```
{
    "license": "mit",
    "languages": [
        "eo"
    ],
	"description": "This keyboard is designed to enable simple input of the Esperanto special characters.",
    "related": {
      "esperuni": { "deprecates": true }
    }
}
```
Most of these are easy fixes. The complication comes at the beginning where all the filenames need to change. I'm sorry for you having to do that.

anvalon:
Hello Lorna,

thank you very much for your help. I made the changes that you proposed. I
have now a couple of questions on the keyboard.
1) Though I choosed "any" (and so it should compile also for Android), it
doesn't worrk on it (Huawei P8 Light, Android 6.0), i.e. the long pression
on a char works, but digiting c and x it does't get a ĉ.
2) I choosed the "eo" language, but Installing on Windows, it installs it
with the English keyboard layout and not that of Esperanto. So I had to
manaully remove the English keyboard and add the Esperanto one frome the
Windows 10 language options.
3) Then I didn't find out haw to delete the old esperanto files on the
repository

Could you please tell me, what changes I have to perform, in order to solve
these two issues?
Thank you in advance

*--*
*Andrea Vaccari*


Il giorno mer 13 mag 2020 alle ore 21:48 Lorna Evans <
notifications@github.com> ha scritto:

LornaSIL:
You can close this Pull Request and delete the branch. That will be sufficient to get rid of these files. Your other Pull Request #1217 is now your active version of Esperanto that we are working on.

anvalon:
Ok, thank you, I close then this pull request and continue on the other one.

PR #1217, opened 2020-05-14
[eokeyboard] Esperanto Initial version of eokeyboard

anvalon:
This is the first commit of the eokeyboard files


keyman-server:

LornaSIL:
I didn't really carefully test your keyboard previously. I noticed a few things that need changing. 
* Remove the `store(&MnemonicLayout) "1"` statement (or just comment it out). That will help your mobile layout to work.
* You've added a lot of keys to the On-Screen keyboard (OSK) layout that are not on a standard US keyboard. However, adding them to the keycaps of the OSK does not actually add them to the keyboard. You need rules for that in the layout tab. Adding these two stores and one rule to your keyboard should make those keys work:
```
begin Unicode > use(main)
store(input) "`-=[]\;'/~@#^&*()_+{}|:" '"' "?"
store(output) "\'ìè+ùòà-|" '"' "£&/()=?^é*§ç°_"

group(main) using keys
+ any(input) > index(output,1)
```

Now, you've also added characters to the On-Screen keyboard for CTRL-ALT and CTRL-ALT-SHIFT layers.

If you want access to those you need to add more rules for those. I'll let you figure that out. The rule for the Euro would be something like:
```
+ [CTRL ALT K_5] > "€"
```
Personally, since you have plenty of room on the CTRL ALT layer, I wouldn't add the curly braces to the CTRL ALT SHIFT layer, I would just find a spot on the CTRL ALT layer. That makes it less complicated.

So, now there are a few things that need fixing in the mobile layout. Go to "touch layout"
* On the default layer:
   * Click on the `K_SHIFT` button and where it says "Next layer" select "shift"
   * Click on the `K_LCONTROL` button and select "Next Layer" "ctrl-alt"
      * Then click on "Remove longpress popup"
* On the shift layer:
   * Click on the `K_SHIFT` button and where it says "Next layer" select "default"
   * Click on the `K_LCONTROL` button and select "Next Layer" "ctrl-alt"
      * Then click on "Remove longpress popup"
* Next, select the ctrl layer and select "Del" to delete that layer
* Next, select the shift-ctrl-alt layer and select "Del" to delete that layer (f you put everything you need on the ctrl-alt layer).

I'm not sure if that is everything, but if you do all this, it will be closer to being complete!

anvalon:
Hello Lorna,

I'm new to GitHub, so I'm trying to understand how it works. I closed the other pull request, but I don't know where the release files of the first keyboard ("esperanto") are located, in order to delete them.

> 
> I didn't really carefully test your keyboard previously. I noticed a few things that need changing.
> 
>     * Remove the `store(&MnemonicLayout) "1"` statement (or just comment it out). That will help your mobile layout to work.

Well, I did it, and now it works also on mobile, thank you very much. So I removed the esperanto characters from the longpress popup, as not necessary any more.

>* You've added a lot of keys to the On-Screen keyboard (OSK) layout that are not on a standard US keyboard.

Yes, I added them, because I think I didn't really understand yet, how the Keyman Desktop application works :-). The esperanto keyboard should be indeed very simple. It should contain only the shortcuts for the 6 esperanto characters (ĉ, ĝ, ĥ, ĵ, ŝ, û) that are usually not present on any normal hardware keyboard. The other keys should be the same that are normally present on the user's keyboard. So, if the user has an english keyboard, the other keys should be the same as usual, and for all other countries keyboards as well.

So I deleted all the characters in the onscreen keyboard and reset the touch layout to the "template-latin" template. But now, when I install the eokeyboard on my Windows 10 Italian, an English keyboard is also installed:

eokey_install_02

And the Keyman window looks then like that:

eokey_install

So I have to remove it from the Windows language options, as I don't need it. But I noticed that the eokeyboard doesn't work as it should, in that the shortcuts actually works, but the keyboard is an english keyboard, that is unusable (or very difficult usable) on an italian hardware keyboard.

Then, there is another less important problem, that concerns the languages structure both in the keyman application and in the windows languages table, in that it is too complex with too many keyboards:

eokey_install_03

In the previous release, when I didn't select the "eo" language in the keyman configuration, the Esperanto keyboard appeared under the first menu item "Italiano (Italia)", and that was much simpler. So I think it'd be very good, if the eokeyboard could be automatically installed within the user's default language. But, if this is not possible, or if it is more correct that another language main menu item is created, then the eokeyboard should at least install with the correct user's default system language. Could you please tell me, how to achieve this?

Thank you again for your help

anvalon:
I installed now the EuroLatin (SIL) keyboard, and after adding the italian language and removing the other languages that were installed by default, it correctly installed under the "Italiano (Italia)" main menu item:

immagine

I would like to achieve the same result with the eokeyboard

anvalon:
I solved the issue by simply removing any language from the package, and so it installs into the system language.

LornaSIL:
We have to have the language (eo). We can't accept it without that. I know when I've been installing and uninstalling keyboards, sometimes my Windows keyboard associations get mixed up, so it may just be your computer.

I'm not sure what to tell you about the mnemonic setting. If you want people to just be able to use their default keyboard, then maybe you should change it back to mnemonic. In that case, I think I would prefer that you remove the On-Screen keyboard, because it will be inaccurate.

However, when you had the mnemonic setting, the mobile layout didn't work. 
This document seems to indicate that to be expected:
https://help.keyman.com/developer/language/reference/mnemoniclayout

So...you should not set the keyboard to mnemonic, don't include an On-Screen keyboard, and I think your mobile layout should be simplified to not include all the extra characters your Italian keyboard includes.

Then, you do need to include `eo` as the language.

LornaSIL:
I've been playing around a bit. If you want to have a mnemonic keyboard, you could keep the mnemonic header. Then add these rules to the bottom:

```
c mobile rules
platform('touch') "C" + "x" > "Ĉ"
platform('touch') "c" + "x" > "ĉ"
platform('touch') "G" + "x" > "Ĝ"
platform('touch') "g" + "x" > "ĝ"
platform('touch') "H" + "x" > "Ĥ"
platform('touch') "h" + "x" > "ĥ"
platform('touch') "J" + "x" > "Ĵ"
platform('touch') "j" + "x" > "ĵ"
platform('touch') "S" + "x" > "Ŝ"
platform('touch') "s" + "x" > "ŝ"
platform('touch') "U" + "x" > "Û"
platform('touch') "u" + "x" > "û"

platform('touch') "C" + "X" > "Ĉ"
platform('touch') "G" + "X" > "Ĝ"
platform('touch') "H" + "X" > "Ĥ"
platform('touch') "J" + "X" > "Ĵ"
platform('touch') "S" + "X" > "Ŝ"
platform('touch') "U" + "X" > "Û"
```

So, if you do all that, then go to On-Screen keyboard and click on "Auto-fill underlying layout" and then click on "Fill from layout" I think you'll be in good shape.

anvalon:
I'm studying the Keyman Developer language more deeply and and then I will tell you what I managed to do.

anvalon:
I defined then the keyboard as positionas, as I wanted the user can use his own keyboard keys as usual.  I inserted a custom layer in the touch layout with the special esperanto keys. I also inserted the "eo" language in the package info. Now, it seems working well on desktop and mobile. I submitted the changes in the repository, by dragging the whole "release" folder and pressing "commit changes". I don't know whether it's right, or I had to create a new branch and open a pull request, as I still don't know how git works.

anvalon:
But I continue not to completely understand how the keyman application works: 
1) It seems that the keyboard now works also on mobile devices, but I tried to install the italian language for the keyboard, as I noticed it was available, and the touch layout was not the same that I developed.
2) Then I still don't understand why rules work on mobile devices, only if the keyboard is not positional.

anvalon:
I realized that there are at least two reasons, why it's better to write a positional keyboard for mobile devices: 1) I noticed that writing the special chars on mobile devices, by changing layer each time I have to type a special character, is too slow, and 2) I will write soon a keyboard, where it is absolutly necessary that rules also work on mobile, because there are not only rules for special chars, but also for replacing key sequences. So I tried again to compile the keyboard as positional, and the rules works indeed on the mobile keyboard, but on the desktop keyoboard I didn't succeed to let the punctuation keys of my hardware keyboard work fine, as they type all other chars. I think tried everything, I tried to install the italian keyboard for the esperanto language, both on windows and on keyman desktop, I tried to remove the english keyboard that was installed bt default by keyman - I don't know why -, and so on. And I think it is not a good idea to write code in order to achieve that, because every user should use the eokeyboard with his own hardware keyboard.

So, if it is not possible to let the hardware keyboard work as usual, it seems to me at present, that the best solution - as far as I know the keyman application - is to write two keyboards, one mnemonic for desktop devices and one positional for mobile devices. But I really wonder whether this may be the best solution, as I find it very strange that it is not possible to achieve the same goal with one unique package.

LornaSIL:
One thing that might be helpful for you to know is that you can change your "base keyboard" in **Keyman Desktop / Configuration / Options** and go down to the bottom of the screen and select "base keyboard". Then you can choose Italian. That might help.

You shouldn't have two create a desktop and a mobile keyboard. That's why I gave you the separate rules for the touch layout. But, at any rate. Try that. I've asked @DavidLRowe to also take a look and see if he can advise.

anvalon:
> One thing that might be helpful for you to know is that you can change your "base keyboard" in **Keyman Desktop / Configuration / Options** and go down to the bottom of the screen and select "base keyboard". Then you can choose Italian. That might help.

Thank you, but it's already set to italian.
 
> You shouldn't have two create a desktop and a mobile keyboard. That's why I gave you the separate rules for the touch layout. But, at any rate. Try that. I've asked @DavidLRowe to also take a look and see if he can advise.

I inserted the rules for touch layout, that you wrote, but they don't seem to work on a mnemonic keyboard.

DavidLRowe:
Currently the "mnemonic" keyboard feature really only works on Windows (with limited support on Mac, Linux, Web). If you create a touch layout, you have to specify all the keys, so you really end up choosing a particular system layout as a base, so there's no practical way to make a touch layout that would adapt to the user's normal system layout.

One option is to make your Esperanto keyboard a Windows-only (or at least desktop-only) mnemonic keyboard that supports the six characters you include in this keyboard. You'd probably want to omit the on-screen keyboard since it would be misleading for a user with any keyboard other than the one on which you based it. This is basically what you currently have (if the touch layout and on screen keyboard are removed).

The other option I see is to create an Esperanto keyboard using the Italian layout as the base. Keyman Developer offers the option, by using the Project menu, New Project menu item, Import Windows Keyboard icon, of importing a base keyboard layout. I see "Italian" and "Italian (142)" as two of the options. You would have a touch layout based on the Italian keyboard and could add the  extra Esperanto characters as long press keys.

anvalon:
> Currently the "mnemonic" keyboard feature really only works on Windows (with limited support on Mac, Linux, Web). If you create a touch layout, you have to specify all the keys, so you really end up choosing a particular system layout as a base, so there's no practical way to make a touch layout that would adapt to the user's normal system layout.

Then would it be better to make it positional, for larger compatibility? And if yes, how can the user use its hardware keyboard punctuation keys as usual?

> One option is to make your Esperanto keyboard a Windows-only (or at least desktop-only) mnemonic keyboard that supports the six characters you include in this keyboard. You'd probably want to omit the on-screen keyboard since it would be misleading for a user with any keyboard other than the one on which you based it. This is basically what you currently have (if the touch layout and on screen keyboard are removed).
 
So one dektop-only keyboard and one for mobile devices?

> The other option I see is to create an Esperanto keyboard using the Italian layout as the base. Keyman Developer offers the option, by using the Project menu, New Project menu item, Import Windows Keyboard icon, of importing a base keyboard layout. I see "Italian" and "Italian (142)" as two of the options. You would have a touch layout based on the Italian keyboard and could add the extra Esperanto characters as long press keys.

As I wrote in a previous post, I will write soon a more complex keyboard, with many substitution rules, that cannot be replaced by the choose of a particular char on a long press menu. So, I'd like the rules to work also on mobile devices.

anvalon:
So, I finally split the keyboard into two separate packages, for desktop (mnemonic) and mobile (positional). Now both of them correctly install and works on my test devices. In the mobile package I specified the "eo" language, otherwise it didn't install on Android, while in the desktop package I didn't specify any language, so that it can be installed under the user's system language. I think this can be a good compromise, since the application does not allow to compile the two versions, with different options, into the same package.

anvalon:
Tell me please, if it's ok, or whether I have to change anything, for the keyboard to be published on the keyman repository. Thank you.

LornaSIL:
I'm sorry for not responding earlier. If you are submitting two keyboards we will want them in two different Pull Requests. 

You mention that you use "eo" for the mobile, but not for the Desktop. We must have "eo" associated with the keyboard. We will not be able to approve a keyboard without the language associated with it. Unfortunately, if you feel the package needs distributing without a language association, then you'll need to distribute the package outside of the Keyman repository.

Lastly, I had asked you to put in the lines for deprecating another Esperanto keyboard in your `keyboard_info` file. At this point your keyboard has diverged from the Esperanto keyboards already in our repo, so I'll ask you to remove those deprecation lines from the keyboard_info.

DavidLRowe:
@anvalon Are you planning to continue work on this keyboard? If not, we can close this pull request. You can always reopen it at a later date, or submit a new one. Thanks for your interest.

anvalon:
You can close it for now, thank you


Il giorno lun 4 gen 2021 alle ore 19:04 DavidLRowe <notifications@github.com>
ha scritto:

Issue #1215, opened 2020-05-14
[vm_tamil] standalone build is failing


mcdurdin:
The vm_tamil package is not building because it has dependencies on other packages. We need to add references to the keyboards into the vm_tamil project and build them and add references to the build artifacts in release/packages/vm_tamil/build instead.

To be clear: don't copy the keyboards into the vm_tamil folder; instead add the existing .kmn keyboards to the vm_tamil.kpj.

```
[14:55:27][Step 1/2] vm_tamil.kps: Compiling package vm_tamil.kps...
[14:55:27][Step 1/2] vm_tamil.kps: Warning: File C:\BuildAgent\work\3a080a5cc3da0019\keyboards\release\packages\vm_tamil\source\..\..\..\v\vm_tamil_modular\build\vm_tamil_modular.kvk does not exist.
[14:55:27][Step 1/2] vm_tamil.kps: Warning: File C:\BuildAgent\work\3a080a5cc3da0019\keyboards\release\packages\vm_tamil\source\..\..\..\v\vm_tamil_modular\build\vm_tamil_modular.kmx does not exist.
[14:55:27][Step 1/2] vm_tamil.kps: Warning: File C:\BuildAgent\work\3a080a5cc3da0019\keyboards\release\packages\vm_tamil\source\..\..\..\v\vm_tamil_typewriter\build\vm_tamil_typewriter.kvk does not exist.
[14:55:27][Step 1/2] vm_tamil.kps: Warning: File C:\BuildAgent\work\3a080a5cc3da0019\keyboards\release\packages\vm_tamil\source\..\..\..\v\vm_tamil_typewriter\build\vm_tamil_typewriter.kmx does not exist.
[14:55:27][Step 1/2] vm_tamil.kps: Error: The build was not successful. Some files were skipped.
[14:55:27][Step 1/2] vm_tamil.kps: Failure: 'C:\BuildAgent\work\3a080a5cc3da0019\keyboards\release\packages\vm_tamil\source\vm_tamil.kps' was not compiled successfully.
```

LornaSIL:
Fixed by #1218

PR #1218, opened 2020-05-14
[vm_tamil] Add .kmn files and change location of compiled files

LornaSIL:
Hopefully addresses Issue #1215 

Sorry I'm so dense. I've done the packages so infrequently that I forgot how to do it.


release/packages/vm_tamil/source/vm_tamil.kps
```diff
@@ -105,9 +105,7 @@
       <Name>Tamil Modular (Visual Media)</Name>
       <ID>vm_tamil_modular</ID>
       <Version>1.1</Version>
-      <Languages>
-        <Language ID="ta">Tamil</Language>
-      </Languages>
+      <Languages/>
```
mcdurdin, line 108: Looks like Tamil was removed from the list of languages and missing for both keyboards now?

PR #1225, opened 2020-05-20
[rac_munji] initial submission on behalf of author

DavidLRowe:
Waiting for author's okay before review.


DavidLRowe:
@rachitralli Please check this keyboard submission. After you check it, it can be reviewed. 

Changes I made:
- created new .png keyboard images for use with welcome.htm
- changed the , ; ? on touch layout to display the Arabic punctuation character that the key generates
- corrected a number of keys that were creating the wrong character
- since there were three touch layouts (phone, tablet, desktop) that were all identical, I deleted the tablet and desktop ones and left the phone layout (since Keyman will use it for all three situations)

Questions about the touch layout:
- Right now on the numeric layer, the character to the right of the shift key has U+061B, then `«` and `<` and `{` as longpress keys. Three keys to the right of that is a key with `"`, with `»` and `>` and `}` as longpress keys. Normally those two keys are `[` and `]` so it makes sense to put other paired grouping symbols there as longpress keys. However, on your keyboard you've put other keys there. I can see several options:
  - Replace the U+061B and `"` with `[` and `]` (since U+061B and `"` are available as longpress characters on both the default and shift layers, so there's no real need to have them on the numeric layer as well)
  - Move the paired grouping symbols to be longpress keys on the `(` and `)` keys

Please let me know what you think.

rachitralli:
@DavidLRowe sir thank you very much for this correction. Due to corona virus in Pakistan I am unable to correct the suggested corrections, your all suggestion are right. if you have some spare time please correct the touch layout as desired. I have check the kmp file in my desktop computer, this is OK. thanks

DavidLRowe:
@LornaSIL (hopefully) ready for review

LornaSIL:
A few things you could do. 
* The readme.htm isn't set in the .kps Details tab. You could add that so it shows up on installation.
* The RAC.jpg is fatter than it should be. You could replace it with one you created in the past. `\rac_hazaragi\source\welcome\RAC.jpg` has a thinner one that you created. You could replace the existing one if you wish.
* The longpress keys on the default layer are often on the shift layer. You could remove them if they exist elsewhere, but it's okay to leave them too.

None of these are big issues, so if you don't want to make the changes, that's fine.

DavidLRowe:
I think I've addressed the first two issues, but will leave the third one the way the original author had it.

rachitralli:
@LornaSIL @DavidLRowe thanks for you time and help

PR #1226, opened 2020-05-21
[Kmhmu keyboard] SIL Kmhmu

MakaraSok:
Updates in this version:
* Replaced shift+u and shift+y with appropriate Unicode Characters
* Changed name from kmhmu_2008 to SIL Kmhmu
* Updated touch layout
* Added support for iOS, Android, Web and Linux
* Added the latest udpated Kmhmu MX and Kmhmu OT to the package


LornaSIL:
@MakaraSok 
Since we are deprecating a keyboard that is deprecating another keyboard, there's a bit more work involved.
* Add a DEPRECATED.md file to `release\k\kmhmu_2008`. It should say:
`This keyboard has been deprecated and replaced by release/sil/sil_kmhmu`
* Edit the DEPRECATED.md file in `legacy\k\kmhmu`. It should also say:
`This keyboard has been deprecated and replaced by release/sil/sil_kmhmu`
* Edit the `sil_kmhmu.keyboard_info` file and add `kmhmu` to also be deprecated.

Other things to fix:
* In the `HISTORY.md` file it would be wise to add `* Copyright transferred to SIL International with permission of author` (if that's true).
* `sil_kmhmu.php` still has the copyright for John Durdin. Please change that to SIL International since that seems to be what you are changing it to.
* In the .kmn file you could change "targets" to "any" and uncheck all the other selections.

I think that's it.

MakaraSok:
Thank you very much for the review, Lorna. Updates have been made accordingly.

PR #1227, opened 2020-05-22
[obolo_qwerty] Obolo Qwerty v1

katelem247:
Obolo Qwerty v. 1.o


keyman-server:

LornaSIL:
* Could you change the description in the `keyboard_info` (and readme.htm) file to say: "Obolo Qwerty keyboard, a simple keyboard in QWERTY layout, for typing Obolo language." The keyboard_info file description is what appears on the main download page so it would be nice if it were more meaningful.
* For LICENSE.md, README.md, readme.htm, welcome.htm, obolo_qwerty.kmn and obolo_qwerty.kps could you add the year so it says "© 2020 Rogers Edeh"?
* The welcome.htm file needs to be enhanced to include all the keystrokes for how to type Obolo on this keyboard. The welcome.htm file in the obolo_chwerty keyboard would be a good starting place.

Once you add the documentation then I will test the keyboard!

katelem247:
Done. @LornaSIL

LornaSIL:
Thanks for making those changes. That mostly looks good. 

For the touch layout, it is impossible to get the macron character. You should select the `T_NAS` character and change the **Key Type:** from `Special` to `Default`. Then I think it will work. Do that on both layers.
image

The other thing isn't essential to fix, but it would be nice. I am seeing a square box (on my android) for the dot below, macron above, and acute accent (on the keycap). If you could use U+00B4 on the acute keycap (both layers), it would look nicer. You would see the acute, but not the square box (the same way you used U+0060 for grave on the keycap). For the dot below and macron above, I don't know that I have a good alternate solution. I guess just leave it and hopefully newer Androids will display it better. 

One other tiny change. The readme.htm file now has an extra double quote at the end of the description. Could you delete that?

I think that's it.

katelem247:
In Obolo, macron applies only to letter n. So, I set it to combine with n. With other letters it will "context beep."
For the accent keys, I used dotted circle and combining diacritics for the keycap. But they keep showing boxes in place of the dotted circle. It happens on Android 5 and 8. I've added tahoma font to the .kps file but that died not change the box. I decided to put that circle to differentiate "dot below" from "fullstop" and "macron" from "hypken", etc. But since it's not working, I think I will go with your suggestion here.

katelem247:
I set T_NAS to special key to differentiate it from hyphen key.
I can't find "underdot" as a space modifier in the character map, and using fullstop will cause confusion so I've allowed it with the dotted circle (which displays box).
Please confirm if everything is ok.
@LornaSIL

LornaSIL:
I think you still need to upload the revised `obolo_qwerty.keyman-touch-layout` to use U+00B4 for the acute? Otherwise it's fine.

katelem247:
Done

katelem247:
Thank you

PR #1229, opened 2020-05-25
GFF Tigrinya-ET 1.3.11 (1.4 Beta 3)

dyacob:
* Fixed a missing longpress mapping for K_2 on layer2
* Adjust 0 longress for degrees sign to match iPhone


keyman-server:

PR #1230, opened 2020-05-26
Balochi multiple

rmlockwood:
Modified all 4 keyboards to have non-Balochi letters on non-default layers. Standardized font use and info. files.


keyman-server:

LornaSIL:
Thanks for this submission.

For the balochi_inpage.php file the desktop should not be:
```
<div id='osk' data-states='default shift right-alt right-shift-alt left-ctrl'>
```
It should be:
```
<div id='osk' data-states='default shift rightalt rightalt-shift leftctrl'>
```
I don't know that the syntax is documented anywhere. Sorry.

I'm curious to know about your renaming the keyboards. Will it be sufficient for people to see "B. Inpage" to understand it's Balochi? I guess I can understand if the names are too long and they can't see whether it is Balochi InPage or Balochi Phonetic, but I'm concerned about the name being too short.

rmlockwood:
The keyboard name thing is kind of an experiment. On mobile I get: “Balochi, Southern (Arabic) (Balochi In…” the rest is cut off. I don’t need to have Balochi on there twice. Although in the selection I’ll only see B. Inpage. What would be ideal is to do something different with the ISO name and script, it’s too long. Do we need to have the name there?

 

Ron

rmlockwood:
I fixed the php file.

LornaSIL:
A suggestion was made that you could call each keyboard by the second word followed by Balochi. So "InPage Balochi", etc. Would that be better?

LornaSIL:
Also, this issue mentions a problem with the readme.htm files in some of the balochi keyboards. Would you be able to address that problem before we finalize this update?
https://github.com/keymanapp/keyboards/issues/1237

rmlockwood:
I don’t really like it, because when you have Persian Balochi it’s not clear which language it is. Let’s try what I have right now.

 

Thanks,

Ron

rmlockwood:
done

LornaSIL:
I'm sorry about this, but apparently the Linux issue is more related to the format of the .htm files than about the syntax you had. They are apparently utf-16. Can you save all of them as UTF-8? I'm told you can save them without a BOM. 

Then, verify that it does include `<meta "charset=utf-8" /> 	`

rmlockwood:
Try that.

PR #1232, opened 2020-05-28
[wakhi_global] initial submission

WX-JO:


keyman-server:

LornaSIL:
I will be checking your keyboard. However, I would **really** like to see you change the name to something like "Wakhi Latin" rather than "Wakhi Global". I know people have also experimented with Cyrillic and Arabic script orthographies for Wakhi and it feels like it would be clearer to say "Latin". The word "Global" feels kind of unusual to me. It gives the implication that it will work with any language. However, I won't reject the name, I just hope you will strongly consider "Wakhi Latin".

WX-JO:
> I will be checking your keyboard. However, I would **really** like to see you change the name to something like "Wakhi Latin" rather than "Wakhi Global". I know people have also experimented with Cyrillic and Arabic script orthographies for Wakhi and it feels like it would be clearer to say "Latin". The word "Global" feels kind of unusual to me. It gives the implication that it will work with any language. However, I won't reject the name, I just hope you will strongly consider "Wakhi Latin".

Hi. Thank you for  checking the keyboard and for your comment.  We can call it Wakhi Anglicized (that's how the authors of this alphabet call it because it is based  to a certain extent on English orthography ). Recently, I have made also the version in Cyrillic. And there is one in Arabic script. This one should be something like a common alphabet/keyboard for the Wakhi speakers across the 4 countries. 
I would like to avoid calling it "Wakhi Latin" because there is another alphabet which is used in scientific circles and it is known as "Latin".  
It's a bit complicated, as the Wakhi speakers live in 4 countries and use 3 different scripts.
So if you agree with "Wakhi Anglicized" - do I have to change and re-submit the keyboard?

LornaSIL:
I do prefer "Wakhi Anglicized" to "Wakhi Global".  Thanks for being willing to do that.

WX-JO:
> I do prefer "Wakhi Anglicized" to "Wakhi Global". Thanks for being willing to do that.

Great! That would be good. So do I need to re-submit the whole thing again with the corrected name? (I am sorry, I am new to the procedure, so I am asking maybe unnecessary questions).

DavidLRowe:
You could change the folder name and file names plus all the references inside files like the .kmn file. BUT you might find it easier to make a brand new project with the new name, then copy and paste information from the old project to the new one. There is a feature request for Developer to be able to do this, but it hasn't been included yet.

DavidLRowe:
You'll probably want to close this pull request and delete the Wakhi-Global branch, then create a new branch and make a new pull request with the new files.

WX-JO:
> You'll probably want to close this pull request and delete the Wakhi-Global branch, then create a new branch and make a new pull request with the new files.

Ok, I will do that. It seems the easiest. Thank you for instructions.

PR #1233, opened 2020-05-28
[gff_amharic] Fix welcome.htm for Linux

ermshiperete:
Webkit2 seems to have problems displaying XHTML - or at least this
page. This change allows to show `welcome.htm` on Linux.

### DIFF
--- a/release/gff/gff_amharic/source/welcome.htm
+++ b/release/gff/gff_amharic/source/welcome.htm
@@ -1,8 +1,5 @@
-﻿<?xml version="1.0" encoding="UTF-8"?>
-<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.1//EN" "http://www.w3.org/TR/xhtml11/DTD/xhtml11.dtd">
-<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en">
+﻿<html>
 <head>
-  <!-- meta http-equiv="content-type" content="application/xhtml+xml; charset=utf-8"/ -->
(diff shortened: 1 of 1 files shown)


keyman-server:

LornaSIL:
@ermshiperete Unfortunately we can't just change the welcome.htm file and have it display properly. The version number for the whole keyboard needs bumping in order for the package to rebuild and download for existing users of the keyboard. 
Can we ask @dyacob to make this change (and for any other keyboards where it's needed) when he has time?

ermshiperete:
> Can we ask @dyacob to make this change (and for any other keyboards where it's needed) when he has time?

Sure, that would be great!

ermshiperete:
See also #1237

DavidLRowe:
See #1237 for more details. This issue is present in many keyboards. gff_amharic just happened to be the first one encountered. Unfortunately, fixing the welcome.htm file means that the version number needs to be incremented in several files (and possibly the date and copyright statements as well).

ermshiperete:
> Oops! I see this is closed, am I too late?

No, I closed the PR because it won't be merged without more work and I created #1237 to track this problem (in gff_amharic and other keyboards).

> I can look into the HTML issue.

That's great. Thanks!

> What is a recommended Linux distro and version to test on? I haven't updated my local linux in a few years. thanks.

[Wasta linux](https://wastalinux.org/) has keyman pre-installed IIRC. Otherwise Ubuntu 18.04 or 20.04.

PR #1234, opened 2020-05-28
[ghana] New version to support characters from ghanakeyboards

LornaSIL:
This version adds support for the languages and characters in the ghanakeyboards. I'm not sure if ghanakeyboards should be deprecated or not. The layout isn't exactly the same, but it is close.


DavidLRowe:
Everything looks fine except one small issue on the touch layout, shift layer, the "." key produces ">".

PR #1236, opened 2020-05-29
[wakhi_anglicized] initial submission

WX-JO:


keyman-server:

LornaSIL:
Thanks for the changes! Here are a few things that need to change:

* In order for ~~ to turn into ~, you need to change this rule:
```
"~" + "~" > "`"
```
to
```
U+0303 + [K_BKQUOTE] > "`"
```
This is because K_BKQUOTE has already turned into a tilde so that rule will never match. So, the left of the "+" must be U+0303. Then the right of the "+" must use the virtual key because it doesn't actually think of it as a tilde, it thinks of it as a backquote.

* On the desktop keyboard, I do not see any rules for typing the schwa. Thus, I think you should remove the longpress key on "e" and "E" for the schwa character. Or, if schwa is needed, then it seems you should add a rule to the desktop keyboard to type it.

* In the .kps file, in the "files" tab, you should remove "wakhi_anglicized_help.pdf", "wakhi_anglicizedU_.png" and "wakhi_anglicizedU_S.png"
     * THEN, you need to re-add those 3 files, but this time from the project folder. You had added it from your desktop so it's not finding the files.
 
* In the help (php) file, there is one changed needed.
     * This line `<div id='osk' data-states='default shift ctrl'>` should have the word "ctrl" removed so it should look like this: `<div id='osk' data-states='default shift'>`

* Also, in the help folder you should delete the `desktop.ini` file that crept in.

Everything else looks good.

WX-JO:
> LGTM. I removed the capital schwa from the touch layout and deleted the desktop.ini file from github.

Thank you! I forgot about the capital schwa. 
That's strange the desktop.ini was still there. I removed it. Thank you for removing it.

LornaSIL:
No problem! I'm always a little hesitant to edit someone's files on GitHub, but this time it mostly worked. Your keyboard is now online: https://keyman.com/keyboards/wakhi_anglicized

Issue #1237, opened 2020-05-29
Several keyboards don't show their html pages on Linux


ermshiperete:
Several keyboards don't show their help pages, readme, options etc because they use a flavour of html that the browser component that is used on Linux doesn't support.

## How to fix

This can be fixed by changing the top of the html file to:

```html
<!DOCTYPE html>
<html lang="en">
```

or simply

``` html
<html>
```

and/or saving as UTF-8.

## Affected keyboards

The affected keyboards are:

  - [x] balochi_inpage (readme.htm)
  - [x] balochi_phonetic (readme.htm)
  - [x] balochi_scientific (readme.htm)
  - [x] balochi_urdu (readme.htm)
  - [x] ekwtamil99uni (readme.htm)
  - [x] embera_north (readme.htm, welcome.htm)
  - [ ] fv_denesuline_epsilon (fv_denesuline_epsilon_kmw_help.htm, welcome.htm)
  - [ ] fv_diitiidatx (fv_diitiidatx_kmw_help.htm, welcome.htm)
  - [ ] fv_gitsenimx (fv_gitsenimx_kmw_help.htm, welcome.htm)
  - [ ] fv_halqemeylem (fv_halqemeylem_kmw_help.htm, welcome.htm)
  - [ ] fv_dakelh (fv_dakelh_kmw_help.htm, welcome.htm)
  - [ ] fv_henqeminem (fv_henqeminem_kmw_help.htm, welcome.htm)
  - [ ] fv_kanienkeha_e (fv_kanienkeha_e_kmw_help.htm, welcome.htm)
  - [ ] fv_ktunaxa (welcome.htm)
  - [ ] fv_kwakwala (fv_kwakwala_kmw_help.htm, welcome.htm)
  - [x] fv_kwakwala_liqwala (fv_kwakwala_liqwala_kmw_help.htm, welcome.htm)
  - [ ] fv_migmaq (fv_migmaq_kmw_help.htm, welcome.htm)
  - [ ] fv_natwits (welcome.htm)
  - [ ] fv_nisgaa (fv_nisgaa_kmw_help.htm, welcome.htm)
  - [ ] fv_nlekepmxcin (fv_nlekepmxcin_kmw_help.htm, welcome.htm)
  - [ ] fv_nsilxcen (fv_nsilxcen_kmw_help.htm, welcome.htm)
  - [ ] fv_nuucaanul (fv_nuucaanul_kmw_help.htm, welcome.htm)
[...]

PR #1003, opened 2019-12-16
Mozhi malayalam v2

beniza:
Updated version of Mozhi Malayalam Keyboard.

This is keyboard should _ideally_ replace all the existing Mozhi Keyboard layouts on the server. Please request for the ref. communications regarding this discussion if needed.

PR #1249, opened 2020-06-09
[htm_fix_xhtmla] Fix more .htm xhtml headers

LornaSIL:
Relates to #1237


DavidLRowe:
Minor issues:
embera_north, .kps file:
The copyright statement is the wrong way around (has "©Ethnos360 1999-2020" rather than "©1999-2020 Ethnos360")

mozhi_malayalam, .kps file:
The copyright statement is the wrong way around (has "2006-2020 © Raj Nair" rather than "© 2006-2020 Raj Nair")
Note that there is currently an open PR #1003 for this keyboard. I'll make a note on that PR about this update.

shan, README.md:
add Linux to supported platforms
Also this keyboard doesn't have a readme.htm file. Should one be added?

Issue #1158, opened 2020-04-13
[jawa] unexpected space on the right


bennylin:
On this layer, I added 0-width spacer on each of the 5 rows, and this is the result:

![image](https://user-images.githubusercontent.com/3002108/79083420-672fc780-7d58-11ea-9d70-c9add0eda8ca.png)

This is my code.
https://github.com/bennylin/keyboards/blob/gh-pages/jawa.keyman-touch-layout

I'm not sure if this is intended or a bug.

I tried to remove some of the spacers, but the result was not as expected as well.
![image](https://user-images.githubusercontent.com/3002108/79083475-bfff6000-7d58-11ea-97e1-a6deb9ad8e65.png)

DavidLRowe:
@bennylin Is this still an issue for you? The link to your code doesn't seem to be correct, or else I don't have permission to view it. Can you point me to the keyboard where is your keyboard located?

PR #1251, opened 2020-06-10
htm_jawa] Fixing html files with xhtml headers

LornaSIL:
Relates to #1237

Issue #1158 seems to be related to keyboard changes that were in development rather than in the existing `jawa` keyboard. 

That issue should be closed and the author should recreate an issue in the https://github.com/keymanapp/keyman.com repo.


DavidLRowe:
It's not clear to me why jawa.ico and jawa.jpg are in the .kps file.

Does the jawa-help.htm (&kmw_helpfile) need to be in the .kps file?

PR #1258, opened 2020-06-24
[fv_plains_cree] Change Y final

eddieantonio:
It turns out that ᐩ is prefered among most Syllabics writers I surveyed, instead of this: ᕀ
This changes the key appropriately.


keyman-server:

DavidLRowe:
Thanks for this contribution, @eddieantonio . All changes to the FV keyboards need to be coordinated through @madebybridget , so we'll wait for approval from Bridget before reviewing.

LornaSIL:
Also, at this point I believe the license needs to change back to First Voices as it was initially. We shouldn't have changed the license and copyright when you last updated the keyboard.

madebybridget:
There seem to be three keys that randomly link to invalid or unrelated characters:
+ [CAPS K_8] > ''
+ [CAPS K_7] > U+F898
+ [CAPS K_6] > U+F85E
But it looks like everything else is there. Let’s revert to this so that the community can access their keyboard again, and then I can work to fix some of these weird issues.
Approved!

madebybridget:
Approved!

DavidLRowe:
@LornaSIL , should the copyright statement be:
```
Copyright (c) 2015-2020 FirstVoices, SIL International, 2015 First Peoples' Cultural Foundation
```
everywhere? Because it looks as though the 2015-2018 dates were reverted (and the fv_plains_cree.kmn file never was changed).

If you can confirm that, @eddieantonio can make the changes, which are needed in fv_plains_cree.kpj, README.md, LICENSE.md, fv_plains_cree.kps, fv_plains_cree.kmn

Also it looks as though the version was reverted 9.1 to in README.md and that it should be 10.0.1.

I've not done anything about checking the actual keyboard.

LornaSIL:
Copyright should be `2015-2020 FirstVoices, SIL International, 2015 First Peoples' Cultural Foundation` everywhere.

I don't think the version should revert. It needs incrementing, no?

DavidLRowe:
Yes, the version should be 10.0.1 everywhere. AFAICT, README.md is the only place where the version number was changed to an older value.

eddieantonio:
Okay, @DavidLRowe, @LornaSIL, I've updated all the copyright statements and the version number — please let me know if there's anything left!

PR #1259, opened 2020-06-24
[nrc_crk_cans] fix Y final; add rare nwV syllabics

eddieantonio:
This changes the Y-final, as in #1258, and also adds rare nwV syllabics that were missing from the previous versions of this keyboard.


keyman-server:

PR #1260, opened 2020-06-25
[bod] Move from legacy to release and include source

LornaSIL:
Relates to #337 
* Move keyboard from legacy to release (not a deprecation)
Documentation improved:
* English documentation is a Google translation of the Chinese documentation. It's not perfect, but better than it was.
* Add help file.
* Extra folder just has documents that I used to help me improve the documentation.


DavidLRowe:
LCtrl+9 and LCtrl+0 show up as square boxes for me (U+0FD5 and U+0FD6), but I'm assuming that's just because I'm missing a font that has those characters. They show up in the OSK and in the display of the help file.

In the welcome.htm file, the heading for the Left Control diagram is labeled "Right ALT (unshifted)".

LornaSIL:
@mcdurdin 

I am closing this PR because we do not have permission to put the keyboard source on our site under MIT license. We wrote the author about moving it to MIT, but up till now we have not received a response. One of the email addresses was no longer valid.
The keyboard is still in `legacy` as an online and mobile keyboard.

PR #1261, opened 2020-06-25
[bod_transliterating] Move from legacy to release and include source

LornaSIL:
Relates to #337

* Move keyboard from legacy to release (not a deprecation)
* Documentation improved:
   * English documentation is a Google translation of the Chinese documentation. It's not perfect, but better than it was.
   * Add help file.
   * Extra folder just has documents that I used to help me improve the documentation.


DavidLRowe:
In welcome.htm, the Left Control image is mislabed as Right Alt.

LornaSIL:
@mcdurdin 

I am closing this PR because we do not have permission to put the keyboard source on our site under MIT license. We wrote the author about moving it to MIT, but up till now we have not received a response. One of the email addresses was no longer valid.
The keyboard is still in `legacy` as an online and mobile keyboard.

Issue #1246, opened 2020-06-05
Issue with Kᵂak̓ᵂala FV keyboard


madebybridget:
Hi folks, the Kᵂak̓ᵂala FV keyboard has recently changed and no longer functions in the way that it should. A user shared her screenshots from two devices she has with the keyboard: one with the old layout and one with the new. The first screenshots attached below on the iPad are of the old, proper layout, versus the second set of screenshots on the iPhone are of the new, non-functional layout. It seems like all of the special characters got moved to the number keys, where when you hold down the numbers they type special characters but don't visibly show up as a long-press option.

Please let me know how we can revert these changes back to an older build of the keyboard with the appropriate layout. Thanks so much!

![image3](https://user-images.githubusercontent.com/58344928/83929520-7d4e7880-a748-11ea-9b82-31ad95271115.png)
![image2](https://user-images.githubusercontent.com/58344928/83929530-850e1d00-a748-11ea-8898-e555c7502bc1.png)
![image4](https://user-images.githubusercontent.com/58344928/83929538-8b9c9480-a748-11ea-9163-7d377ff060ef.png)
![image5](https://user-images.githubusercontent.com/58344928/83929542-8d665800-a748-11ea-9097-505dcae30c1f.png)

DavidLRowe:
Are we talking about the release/fv/fv_kwakwala keyboard (in the GitHub repository at: https://github.com/keymanapp/keyboards/tree/master/release/fv/fv_kwakwala). The last change to this keyboard was 18 Oct 2018. At that point three legacy keyboards were deprecated: legacy/k/kwakiutl, legacy/k/kwakwala_u, legacy/u/uwikala_u</del>

Or is this the release/fv/fv_kwakwala_liqwala keyboard (in the GitHub repository at: https://github.com/keymanapp/keyboards/tree/master/release/fv/fv_kwakwala_liqwala). The last change to this keyboard was also 18 Oct 2018.

LornaSIL:
https://github.com/keymanapp/keyboards/tree/master/release/fv/fv_kwakwala).
This one looks correct to me on an android.

https://github.com/keymanapp/keyboards/tree/master/release/fv/fv_kwakwala_liqwala
).
This one seems to have problems.

LornaSIL:
Fixed by #1262 @madebybridget

PR #1262, opened 2020-06-26
[fv_kwakwala_liqwala] Revert mobile layout and fix headers on htm files

LornaSIL:
Relates to #1246 

Relates to #1237

@madebybridget do you want to check this first?


madebybridget:
There seem to be three keys that randomly link to invalid or unrelated characters:
+ [CAPS K_8] > ''
+ [CAPS K_7] > U+F898
+ [CAPS K_6] > U+F85E
But it looks like everything else is there. Let’s revert to this so that the community can access their keyboard again, and then I can work to fix some of these weird issues.
Approved!

PR #1268, opened 2020-07-02
cabecar edited upload

victoriaq22:
This is an updated version of the Cabecar keyboard.  Fixes small errors, needs further user testing before full documentation will be completed.  Uploaded previously (July 2) in the wrong spot, no changes to the keyboard package have been made since then.


keyman-server:

PR #1269, opened 2020-07-03
Katelem247 patch 1

katelem247:
This update replaces the high tone accent (which is not required) by a more frequently used punctuation (the hyphen-dash). It also changes the backspace behaviour for "ch, sh, n̄."


keyman-server:

LornaSIL:
When a keyboard is updated, the version number must also be updated. Also, the copyright date should not be changed to 2020, it should be extended from the initial date..so you should have `2019-2020`. Thus, you will need to change:
* HISTORY.md - please add the new version and date and describe what you changed. Something like this:
```
1.1 (2020-07-07)
----------------
* Removed dot under e (or whatever your change is)
* Changed backspace behavior for ch, sh, n̄
```
* README.md - Change copyright (`2019-2020`) and version (`1.1`)
* LICENSE.md - Change copyright (`2019-2020`) 
* readme.htm - Change version (`1.1`)
* welcome.htm - Do you need to change the documentation for the changes you made for high tone?
* obolo_chwerty.kmn
   * Change copyright (`2019-2020`)
   * Also, the version was never set for the .kmn file before. That was a mistake! Please make sure in the Details tab to put "1.1" in the Keyboard version field.
* obolo_chwerty.kps - In Details tab, change copyright (`2019-2020`) 

I don't really understand these two rules.
```
+ [CTRL SHIFT K_C] > 'cC' c for desktop
+ [CTRL SHIFT K_X] > 'xX'
```
Do you really want them to output both "cC" or should you write two rules, one to output the "c" and another to output "C"?

katelem247:
@LornaSIL, I've addressed all those issues you raised.
Done. Please cross-check.

katelem247:
Thank you @LornaSIL

PR #1272, opened 2020-07-04
cabecar upload july 4

victoriaq22:
minor fixes updated, including capitalization errors.  Needs further user testing before all documentation is finalized.


keyman-server:

DavidLRowe:
For the physical keyboard
(1) You may want to make the { key, that is [SHIFT K_LBRKT], also work as the acute deadkey. In that case, include a rule:
```
+ [SHIFT K_LBRKT] > dk(highTone)
```
In that case, you'll need to update the OSK (see next item).

Alternatively, if you want this key to really produce nothing, include a rule that says:
```
+ [SHIFT K_LBRKT] > nul
```

(2) The on screen keyboard (OSK) needs to be updated (using the "On-Screen" tab in Keyman Developer). Currently:
```
the ` key is blank, but produces "|"
the [ key is blank, but is the acute deadkey
the ' key is blank, but is the dieresis deadkey
the / key is blank, but is the underline deadkey
the { key is blank, but produces "{"
the " key is blank, but is the dieresis plus acute deadkey
the ? key is blank, but is the underline plus acute deadkey
```
Note that what you put on the OSK is for display and doesn't affect what the keyboard actually produces.

For the five deadkeys (or six, if you decide to include SHIFT K_LBRKT), if would be helpful to put something on the OSK. Some people like to include a dotted circle (◌ U+25CC) and then add the diacritic(s).

You may need to update the keyboard images used by the help file.

(3) Currently you have the "@" as the only key on the Ctrl+Alt layer. Would you want to move that to the Alt layer? If you do, be sure to update the OSK as well. If you leave the key on the Ctrl+Alt layer, it would be helpful to document that (eventually).

Touch keyboard:
(1) On the "shift" layer, the key with "." produces ":" -- that will need to be corrected.

(2) On the "numeric" layer, the keys with "." and "¿" have "«", "<", "{" and "»", ">", "}" as longpress keys. Do you need the "«" and "»" keys? If not you could eliminate all these longpress keys, since the "<", "{", ">" and "}" appear on the "symbol" layer. (In the model keyboard that Keyman Developer provides, these grouping symbols are longpress keys on "[" and "]", so people can think to find them. It's a bit odd to have the grouping longpress keys on non-grouping characters, but it's not incorrect.)

(3) On the "symbol" layer, the "}" key produces "¨" and the "€" key produces "¬" -- these will need to be corrected.

Eventually (before moving to "release") you'll want to harmonize the information in the welcome.htm file and the cabecar.php file. (The first is bundled with the keyboard and available offline. The second is the source for the online help for the keyboard.)

DavidLRowe:
Your changes all look good, except the "." on the shift layer of the touch layout and the longpress "," key on that "." key don't output the correct characters. "." produces ":" and "," produces ".".

With those corrections, this pull request should be ready.

DavidLRowe:
@victoriaq22 with those changes this revised keyboard should now be available. It's still in the experimental section. When you finalize the documentation (as well as any changes to the keyboard itself) you can move it to the release section.

PR #1273, opened 2020-07-05
GFF Gurage 0.7 Beta 1

dyacob:
A letter frequency based Keyboard for Gurage orthography.

Note that the Keyman web based touch debugger doesn't compose syllables correctly.  This is the same issue found with recent touch updates for Tigrinya-ET, that this touch layout is derived from.  The problem isn't present once installed on a mobile phone.


keyman-server:

LornaSIL:
@dyacob it looks like you have some conflicting files. Did you do a pull before you made these changes? I see I made changes a month ago when I updated the htm files to work on Linux. In HISTORY.md I did put the wrong date though! You can fix that when you fix the conflicts.

dyacob:
thanks @LornaSIL , I _really_ thought I had synched earlier but apparently it didn't get into this branch.  I _think_ its  all up to date now.

LornaSIL:
* In the README.md can you add web and mobile to platforms?
* In the .kps 
   * you should add this file: `gff_gurage.js`
   * At this point you must include the `-Ethi` in the BCP47 tag. This is because IANA (language-subtag-registry) doesn't have "suppress script"
* I tested most of the main keycaps in the touch layout and some of the longpress keys. They seem to work fine. Unfortunately (for you) we are currently requiring that the globe (*Menu*) key be to the left of the space on each layer (not on a longpress). We don't want people to not be able to switch keyboards which can happen on some systems (like ios maybe?). It would easily fit on the two punctuation layers. For the non-gurage layer you may need to put the K_LOWER on the K_NUMERIC as a longpress to make room for the globe key. (If any of your other layouts have hidden the globe key, we'll require this change when you update the keyboard).

I think that's it.

dyacob:
README.md and .kps changes are made.  The Globe/*Menu* button is now on each layer.  Long presses added to the layer-shifting button.

PR #1274, opened 2020-07-06
[tamajaq_unicode] Tamajaq-Unicode keyboard

MakaraSok:
This package was created based on an existing .kmx devised by Christian Grandouiller. 

Added features:
- support any platforms
- OSK
- touch layout


LornaSIL:
I don't really like having "Unicode" in keyboard names since all keyboards
should be Unicode.  It might be worth including "Latin" in the name since
the language group also uses the Tifinagh script.

Also, the php header says "haroi".
I didn't look at anything else.

MakaraSok:
Thank for your comment. The keyboard name is adopted from its original version. I will ask the author whether or not the word Unicode can be dropped.

The php is now fixed. I didn't replace the name with the right. Thank you for catching that.

LornaSIL:
I consider the words "Keyboard" and "Unicode" to be unacceptable in a keyboard name, so you will need to convince him to drop the word "Unicode" and possibly use a word like "Latin" to distinguish it from other keyboards.

LornaSIL:
There's another keyboard in our repo called "Tawallammat Tifinagh (SIL)". It's for the same language. You might ask if it would be appropriate to call it "Tawallammat Latin". That would make it more clear which Tamajaq language this is for.

For the readme.htm file I think the text you used in README.md, keyboard_info and welcome.htm would be more helpful to people: "This keyboard covers all posible characters needed to type Tamajaq language for Latin orthography. Additional characters have been added to avoid having to switch keyboard to French or English while typing. "

Either the documentation or the keyboard is wrong for the "g" with dot below. It has a dot below in the documentation, but a dot above in the keyboard. The capital "G" has a dot below.

There's a similar issue with the "j".The keyboard puts a dot above (and since the dot on the "j" disappears it looks like a standard j). The capital "J" has a dot below.

On the touch layout, numeric layer, the longpress for ‹ and › you'll need to change the rule for "<", ">", "{" and "}" to output what's on the keycap.

Because this keyboard is so heavily reliant on ```. , / : ; ^ ` ~``` you may wish to consider putting a longpress key to the right of the spacebar that has those keys on it. Put that on both the default and shift layers so that it's always accessible. I found switching layers to find the right key rather irritating.

Another possibility would be to do away with those keys on the mobile layout and put all the special characters on the longpress keys (so K_C would have the longpress produce c with cedilla, and K_J would have the longpress produce the j with dot, etc). It's more work, but it might be most useful to the people.

One last thing, the desktop keyboard has the ability to produce curly quotes ( “ ” ‘ ’) not just the angle quotes (« » ‹ ›). Those could be added to the longpress keys on the numeric layer where you have angle quotes.

I think that's everything.

MakaraSok:
Thank you for the feedback/suggestions, Lorna. It took a while to get through to the author and making changes to this keyboard. Please see my responses underneath each relevant quot below.

> There's another keyboard in our repo called "Tawallammat Tifinagh (SIL)". It's for the same language. You might ask if it would be appropriate to call it "Tawallammat Latin". That would make it more clear which Tamajaq language this is for.
> 

The name has been changed as suggested with the agreement from the keyboard author per email exchange.

> For the readme.htm file I think the text you used in README.md, keyboard_info and welcome.htm would be more helpful to people: "This keyboard covers all posible characters needed to type Tamajaq language for Latin orthography. Additional characters have been added to avoid having to switch keyboard to French or English while typing. "
> 

Done.

> Either the documentation or the keyboard is wrong for the "g" with dot below. It has a dot below in the documentation, but a dot above in the keyboard. The capital "G" has a dot below.
> 
> There's a similar issue with the "j".The keyboard puts a dot above (and since the dot on the "j" disappears it looks like a standard j). The capital "J" has a dot below.
> 

This was done on purpose at the time the keyboard was developed. The author is aware of this inconsistency.

> On the touch layout, numeric layer, the longpress for ‹ and › you'll need to change the rule for "<", ">", "{" and "}" to output what's on the keycap.
> 

Done. 

> Because this keyboard is so heavily reliant on `` . , / : ; ^ ` ~ `` you may wish to consider putting a longpress key to the right of the spacebar that has those keys on it. Put that on both the default and shift layers so that it's always accessible. I found switching layers to find the right key rather irritating.
> 

Done. 

> Another possibility would be to do away with those keys on the mobile layout and put all the special characters on the longpress keys (so K_C would have the longpress produce c with cedilla, and K_J would have the longpress produce the j with dot, etc). It's more work, but it might be most useful to the people.
> 

I do not know how to get a key to output two or three characters on a longpress key.

> One last thing, the desktop keyboard has the ability to produce curly quotes ( “ ” ‘ ’) not just the angle quotes (« » ‹ ›). Those could be added to the longpress keys on the numeric layer where you have angle quotes.

These can be output using a combination of a comma and the angle bracket.
image

LornaSIL:
For future reference, you can add a longpress key, then change the "code" to something like `T_0067_0307`

Then write a rule in your .kmn file such as:
` + [T_0067_0307]` > U+00067 U+0307`
This outputs the g with a dot above.

DavidLRowe:
There's also a feature request (https://github.com/keymanapp/keyman/issues/2858) to allow U_xxxx_yyyy to output two characters, but no immediate plans for implementing that.

PR #1279, opened 2020-07-10
GFF Gurage 0.7 Beta 2

dyacob:
* Fixed some 2nd layer compositions that could not be tested in the web based debugger.


keyman-server:

PR #1280, opened 2020-07-10
[easy_chakma] Update Font

MakaraSok:
@Bivuti has requested to update the font included in this keyboard. Nothing else has been changed in the process, but the font. Hence, version number is not bumped. @LornaSIL Let me know if it's better to bumped it.

> It's last updated on 15 June 2020. Many more signs added in the font. It's now support Stroke (𑄥𑄳𑄑𑄳𑄢𑄮𑄇𑄴), Andrew (𑄆𑄚𑄴𑄳𑄓𑄳𑄢𑄪), Spleen (𑄥𑄳𑄛𑄳𑄣𑄨𑄚𑄴), splashing(𑄥𑄳𑄛𑄳𑄣𑄳𑄠𑄥𑄨𑄁), etc.


LornaSIL:
The package will not be offered to existing keyboard users unless you bump the version.

DavidLRowe:
Changing the version number (from 1.19 to 1.20) will mean changing:
- HISTORY.md
- README.md
- easy_chakma.kmn

The version number also shows up in easy_chakma.kps and easy_chakma.kpj, but I think the setting for package following keyboard version should take care of those.

The version number is also mentioned in the readme.htm file. That may or may not need to be changed.

Bivuti:
@LornaSIL  No need bumped the version number because this is a minor change. @MakaraSok please your comments.

DavidLRowe:
If the version number is not increased, then the updated package (with the updated font) will not be made available to users, because the system will think that they already have the latest version (1.19).

If, however, the version number is increased (to 1.20, for example), then the system will make the updated package available to users.

Bivuti:
@DavidLRowe okay, so please update the version accordingly. 
Thanks.

MakaraSok:
@LornaSIL @DavidLRowe I didn't realize that. Thank you very much for pointing it out.

The version number has been bumped to 1.20.

Issue #1277, opened 2020-07-07
Rename legacy packages with full stop in their names


mcdurdin:
As part of our Keyman 14 work, we need to use full stop to delineate keyboard IDs (filenames) and bcp-47 codes (this is a part of the bundled installer work). There are three keyboards in legacy that use full stop in their IDs. Can we rename these (using the normal deprecation strategy):

* `legacy/a/amazigh_lat.kmn/` (1 download last month), deprecate in favour of `amazigh_lat`
* `legacy/g/gandhari-keyboard-2.7/` (46 downloads last month), deprecate in favour of `gandhari`
* `legacy/v/vai1.1/` (zero downloads last month), deprecate in favour of `vai11` perhaps?

Deprecated keyboards will not be impacted by the changes because they will not be part of the bundling strategy -- they will only be available as standalone .kmp downloads for the truly desperate!

PR #1281, opened 2020-07-10
[1277] Rename folder and files for amazigh_lat and gandhari

LornaSIL:
Relates to PR #1277


DavidLRowe:
It seems as though the text of legacy/g/gandhari-keyboard-2.7/DEPRECATED.md should be:
```
This keyboard has been deprecated and replaced by legacy/g/gandhari
```
rather than:
```
This keyboard has been deprecated and replaced by legacy/g/gandhari-keyboard-2.7
```


The .keyboard_info file for amazigh_lat seems to have kept more metadata (packageIncludes, platformSupport) than the one for gandhari, but I have no idea if that is significant.

DavidLRowe:
fixes #1277

PR #1282, opened 2020-07-10
[athinkra_vai] Move Vai (Athinkra) to release with permission of Athinkra

LornaSIL:
* Athinkra gave permission to move to MIT and the author sent me his source files.
* Deprecated vai1.1
* Added documentation

Relates to PR #1277


DavidLRowe:
welcome.htm (also athinkra_vai.php): In "...accessing histor syllables...", it seems that "histor" should be "historic".

There are six keys on the RAlt layer which have a hyphen on the displayed key. For two of these (f and k), the hyphen is removed when the Vai character is displayed, but for the other four (s, d, c, b) the hyphen remains. Is this intended? For example:
RAlt+s gives -s (002D 0073) and typing e gives -ꔖ (002D A516), but
RAlt+f gives -f (002D 0066) and typing e gives ꘓ (A613).
Did this keyboard work like this in the past? Maybe we just approve it as is, and make a note in the README.md of this behavior.

LornaSIL:
>welcome.htm (also athinkra_vai.php): In "...accessing histor syllables...", it seems that "histor" should be "historic".

Fixed

>There are six keys on the RAlt layer which have a hyphen on the displayed key. For two of these (f and k), the hyphen is removed when the Vai character is displayed, but for the other four (s, d, c, b) the hyphen remains. Is this intended? 

I fixed a few rules that weren't working, but the main issue is that there aren't historic syllables for every vowel so the user must choose the right vowel to follow. I am leaving those as is, but documented that there aren't syllables for every vowel and that the hyphen will remain.

PR #1283, opened 2020-07-11
[qom] Initial submission

MLocria:
This is the first version of the keyboard. We have done some testing with a limited number of users and it seems to be working as intended.  After it is publically available in Keyman's repository/website for download, we expect to continue developing it further by collecting much more feedback.
Please, let me know if I need to introduce any changes. I followed [these instructions](https://github.com/keymanapp/keyboards/pull/1283).
We are really excited to finally get this to the community! :)
Manuel.


keyman-server:

LornaSIL:
Thanks for your keyboard submission! Here is some feedback.

* The folder name must be lowercase. I don't know if you can rename it from "Qom" to "qom" and have github automatically pick up the difference or if you need to do a rename. Something like `git mv Qom qom `
* The copyright isn't consistent. We need it to say `© 2020 Latinoamérica Habla` in all places. So,
   *  the README.md, qom.kmn and welcome.htm needs the year added between the copyright sign and copyright holder.
   *  the readme.htm and qom.kps (under details) needs year and copyright holder 
* The description in `qom.keyboard_info` could be more descriptive. This is what will show up on the keyboard home page. In readme.htm and README.md you have `Qom keyboard based on Latin American Spanish ` which is better than the existing description. I think this might be even more helpful: `Qom keyboard for the Toba language (based on Latin American Spanish).` 
* The keyboard as currently written isn't working quite right. I can type `y[y` and the whole thing turns into y with tilde. I think if you rewrite these line:
```
begin Unicode > use(modifiers)

group(modifiers) 
c Comma Above Right store (If U+02BC is preceeded by character below, they merge with it)
any(CommaAboveRight) U+02BC > index(CommaAboveRight,1) U+0315

c Special Characters (Combined with apostrophe)
U+0079 U+02BC > U+1EF9
U+0059 U+02BC > U+1EF8 

nomatch > use(main)

group(main) using keys
```
to the following you will get the behavior you wish for. It also simplifies the keyboard to do away with one group.
```
begin Unicode > use(main)

group(main) using keys

c Comma Above Right store (If U+02BC is preceeded by character below, they merge with it)
any(CommaAboveRight) + [K_LBRKT] > index(CommaAboveRight,1) U+0315

c Special Characters (Combined with apostrophe)
U+0079 + [K_LBRKT] > U+1EF9
U+0059 + [K_LBRKT] > U+1EF8 
```

For the mobile layout
*  longpress keys have many keys that are not on the desktop layout. Should you remove those?  
* on the numeric layer I don't see a way to access the `[`
* The `'` and `+` have longpress keys that wouldn't make much sense to a user. Normally, I would want to put `[` and `]` on the main keycaps so that the left and right brackets go with what is on the longpress. If you do that, then you'll want to put `'` and `+` on the `]` and `}` keys which are currently at the far right of the 2nd row on the numeric layer.

For the help file:
* Please rename it from `qom.php.txt` to `qom.php`
* Change line 15 from: `<div id='osk' data-states='default shift'>` to `<div id='osk' data-states='default shift rightalt'>`
* Change line 21 from: `<div id='osk-phone' data-states='default shift numeric'>` to `<div id='osk-phone' data-states='default shift numeric symbol'>`

I think that's it!

LornaSIL:
I think you've done all I asked for...except that the folder name didn't get changed from `Qom` to `qom`. Did you try using the `git mv Qom qom` command? If that didn't work, then you may need to open a new PR with the correct folder name before committing it (and closing this one).

MLocria:
Thank you very much, @LornaSIL , for all your great comments. And I feel a quick apology is in order, as some of the points you made were clearly stated in the documentation :)

I think I have now fixed all the issues you highlighted. Regarding the mobile layout, I made some changes but I'm purposefully leaving most of it as it is because we want this keyboard to be usable for both Spanish and Qom at the same time. In future releases, we will fine-tune it more and more.

Please, feel free to let me know if I missed anything!

MLocria:
In case you're interested: I'm working from a Windows machine and the filesystem is not case-sensitive. This means that I couldn't change the folder name right away (I had to change it to Qom1 and then back to qom finally)

MLocria:
I think I have now done everything. Please, let me know if I'm missing anything!!

LornaSIL:
Well, that was creative! I'm also on Windows and just found I had to delete your branch and repull it for me to see the case change.

PR #1292, opened 2020-07-16
[athinkra_vai_typewriter] New keyboard on behalf of Athinkra

LornaSIL:


DavidLRowe:
Shift+RAlt+2 produces U+0A62, but I think it should be U+A622.

Minor: Some of the "-x" keys on the RAlt layer leave the - behind when a non-combining vowel follows. (I think you fixed this on another vai keyboard.) I'm happy to approve it "as is".

PR #1295, opened 2020-07-23
fix: remove obsolete server references in php files

mcdurdin:
Depends on keymanapp/help.keyman.com#200

The primary objective of this PR was to remove references to the legacy `$site_tavultesoft` variable which was causing site errors. The global variable `$keyman_com` will be available for use to refer to https://keyman.com once PR 200 referenced above is merged.

Version numbers do not need to be updated because the php files are updated on help.keyman.com with every deployment.

### DIFF
--- a/legacy/b/bengali_inscript/source/help/chart.php
+++ b/legacy/b/bengali_inscript/source/help/chart.php
@@ -1,30 +1,30 @@
 <?php /*
   Name:             bengali_inscript
   Copyright:        Copyright (C) 2005 Tavultesoft Pty Ltd.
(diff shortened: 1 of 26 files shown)


DavidLRowe:
@mcdurdin IIUC this should not be merged until keymanapp/help.keyman.com#200 is merged. Is that correct?

mcdurdin:
> Is that correct?

Well, the changes won't work until that PR is merged, but the links in question are already broken right now, so it's no big deal to merge as soon as you like. Once the other PR is merged, then the links will start to work :-)

PR #1297, opened 2020-07-24
[sxava] Add help file

LornaSIL:
I didn't ask the author of this keyboard to make a help file.


DavidLRowe:
It seems that the string for $pagename (line 2) needs double quotes rather than single quotes. (See #1296)

LornaSIL:
I think that is because there was a single quote within the keyboard name. I don't think that's a general requirement.

DavidLRowe:
Oh, good point. I hadn't noticed that.

PR #1298, opened 2020-07-24
[athinkra_vai_typewriter] Update version and info in kps

LornaSIL:
kps was missing some information. Even tho the keyboard worked we felt it was better to get it correct.


DavidLRowe:
The .kmn file is UTF-16. I think it would be helpful to convert to UTF-8.

Other than that, everything seems fine.

PR #1300, opened 2020-07-31
[aksarabali_panlex] update documentation and Vimala font

kamholz:


keyman-server:

LornaSIL:
This all looks straightforward. Normally we do copyright dates for the first year of the keyboard and the last update. So, in this case I'd prefer to see `Copyright (c) 2019-2020 The PanLex Project of The Long Now Foundation`

Would you be willing to make that change?

Also, the help file has the same information `help\aksarabali_panlex.php` has the same information in it as welcome.htm. Could you please make similar changes to the help file as you did to the welcome file?

(As an FYI for myself and others, I've already updated the font on `s.keyman.com` which is used for displaying the help file.)

kamholz:
Just updated the years and the `aksarabali_panlex.php` file.

PR #901, opened 2019-09-30
Sahaptin / Ichishkiin Yakima Compact Keyboard

kelivermore:
I have another version of this keyboard (a full version compared to this compact one) to submit, but I wanted to figure out the process first! Please let me know if I'm missing or needing to change anything in this. 

Kw’ałanúushamatash, thank you!
Keegan

PR #1301, opened 2020-08-03
[sahaptin_yakima_full] Initial Submission of Sahaptin_Yakima full keyboard

kelivermore:
This is a re-submission of files from last year with requested changes.


keyman-server:

DavidLRowe:
Hi @kelivermore. Thanks for this keyboard submission.

Because the Keyman system rebuilds all keyboards from the source files, they need to be in a particular format. For this keyboard, the project (.kpj) file is referencing files that aren't in direct subfolders (that is, the .kpj file is not referencing the files in the source folder), which means that the project won't build on anyone else's computer. You'll need to correct that before we can test the keyboard.

If you're in the Portland OR area, then we're in the same time zone. If you'd like to connect over the Internet, you could share your screen and I could try to walk you through the process.

Here are some other things I noticed: 
- the two .png files in the sahaptin_yakima_full folder don't seem to be referenced. Normally they would be in the same folder as the welcome.htm file and referenced from that file.
- there are some places that indicate (c) 2019 and some that indicate (c) 2020. These should all be consistent. (It is possible to specify a range (c) 2019-2020, if you began the work in 2019.)
- There are different spellings of Ichishkiin (with and without a character between the final two "i" characters). 
- There seem to be different characters used: sometimes U+2019 (right single quotation) and sometimes U+00B4 (acute accent). This should be verified.
- There's a reference to "the barred ? letter". I wonder if you have tried to use a character that is not being properly rendered. You may need to specify that the text of the welcome.htm file is UTF-8.
- The description in the .keyboard_info file is rather long. You may want to consider shortening it.
- We will need a "readme.htm" file. This can be very short. You could use the shortened description from the .keyboard_info file as the text.

Let me know if anything needs more explanation!

DavidLRowe:
Related: #901 (closed Sahptin Yakima Compact keyboard)

kelivermore:
Hi David, I really appreciate your email! I’m definitely still trying to figure this out, as I’m a linguist trying to do these computer/coding things. I am in the Portland/Eugene area, and would be most grateful for guidance on correcting my submission and learning how to get these files submitted. This week and next week are not as great for me, so perhaps we could look at sometime after the 25th? I’d be happy to set up a screen share call. Thank you so much for that offer. I’m excited to learn some more about this process! Again, thank you! Keegan  From: DavidLRoweSent: Monday, August 10, 2020 16:58To: keymanapp/keyboardsCc: Keegan Livermore; MentionSubject: Re: [keymanapp/keyboards] [sahaptin_yakima_full] Initial Submission of Sahaptin_Yakima full keyboard (#1301) Hi @kelivermore. Thanks for this keyboard submission.Because the Keyman system rebuilds all keyboards from the source files, they need to be in a particular format. For this keyboard, the project (.kpj) file is referencing files that aren't in direct subfolders (that is, the .kpj file is not referencing the files in the source folder), which means that the project won't build on anyone else's computer. You'll need to correct that before we can test the keyboard.If you're in the Portland OR area, then we're in the same time zone. If you'd like to connect over the Internet, you could share your screen and I could try to walk you through the process.Here are some other things I noticed:the two .png files in the sahaptin_yakima_full folder don't seem to be referenced. Normally they would be in the same folder as the welcome.htm file and referenced from that file.there are some places that indicate (c) 2019 and some that indicate (c) 2020. These should all be consistent. (It is possible to specify a range (c) 2019-2020, if you began the work in 2019.)There are different spellings of Ichishkiin (with and without a character between the final two "i" characters).There seem to be different characters used: sometimes U+2019 (right single quotation) and sometimes U+00B4 (acute accent). This should be verified.There's a reference to "the barred ? letter". I wonder if you have tried to use a character that is not being properly rendered. You may need to specify that the text of the welcome.htm file is UTF-8.The description in the .keyboard_info file is rather long. You may want to consider shortening it.We will need a "readme.htm" file. This can be very short. You could use the shortened description from the .keyboard_info file as the text.Let me know if anything needs more explanation!—You are receiving this because you were mentioned.Reply to this email directly, view it on GitHub, or unsubscribe.

PR #1303, opened 2020-08-11
[sil_makuri] Intial release    

sdysart:


keyman-server:

LornaSIL:
Thanks for the keyboard submission!

Here are a few things you could change or consider changing:

* It would be good to put a link in the README.md file under "links"
```
Links
-----
https://keyman.com/keyboards/sil_makuri
```
* The keyboard name in the .kps should be changed from `SIL Makuri` to `Makuri (SIL)` to match the .kmn name. That's in the details tab
* In the help .php file. This line `<div id='osk-phone' data-states='default shift Num1 Num2'>`  needs `phone` changed to `tablet` (`<div id='osk-tablet' data-states='default shift Num1 Num2'>` ) (Alternatively you can call the "Touch Layout/Platform" a phone layout, not a tablet layout).
* On the OSK I cannot see the diacritics very well. You could click on the key and type two spaces after the combining mark. That moves the combining mark to be more centered. It's a hack, I know, but it helps to see the diacritic. Another option is to put a U+25CC circle before the combining mark. Then re-export your graphics for the welcome folder.
* The Shift-X outputs X but you don't have it on the OSK. Either add the X to the keycap, OR, if you want to prevent outputting an "X" you could add this rule to your keyboard:
```
+ [SHIFT K_X] > nul
```
* In the touch layout, Num2 layer, top row, farthest key to the right, `K_0` should be changed to `U_2206`
* In Num2 Layer, you also have a way to access the Num1 layer, but this is on the keycap `?123`. Were you wanting the question mark on the keycap or just the digits?

Overall it looks very good.

darcywong00:
I've updated documentation and @sdysart will update the keyboard.

PR #1307, opened 2020-08-18
[sil_makuri] Update icon

darcywong00:
Following up on review comment from #1303
> I'm approving it although I dislike the icon file. It looks like it came from some other language.

This update from @sdysart changes the icon to an "M".


keyman-server:

PR #1311, opened 2020-08-28
[sahaptin_yakima_computer] initial submission

kelivermore:
This submission is via GitHub.


keyman-server:

DavidLRowe:
@kelivermore This keyboard project (desktop layout only, no touch layout) seems unnecessary, since both #1312 and #1313 have the same desktop layout. So it seems that we can close this pull request.

PR #1312, opened 2020-08-28
[sahaptin_yakima_mobile_expanded] initial submission

kelivermore:
This submission is via GitHub.


keyman-server:

DavidLRowe:
@kelivermore Hi Keegan, we're suggesting that you make this the main keyboard submission. It has the desktop keyboard layout plus a touch layout. There's no need to include "mobile" in the name of the keyboard. If you make this the standard one, then you can make a "compact" one that is separate (that is, use the name "sahaptin_yakima" for this keyboard (desktop plus mobile) and use the name "sahaptin_yakima_compact" for another keyboard (desktop plus mobile) that includes a smaller touch layout).

Alternatively, you could include both touch layouts in this keyboard, by tagging one as "tablet" and the other (smaller) one as "phone". Then on a phone you would get the smaller layout and on a tablet you would get the larger layout. 

You'll need to include the documentation for the desktop (that was in #1311) in the welcome.htm for this keyboard.

Because the renaming of the keyboard will mean renaming lots of files, you may find it easier to close this PR and open another. 

Feel free to contact me directly if you have questions.

kelivermore:
Kw’aɬanúushamash, thank you so much for your guidance, David! I'll make these changes and re-submit.

PR #1313, opened 2020-08-28
[sahaptin_yakima_mobile_compact] initial submission

kelivermore:
This submission is via GitHub. The arrangement of the keys within the files may change slightly as users get a chance to try using the keyboard for an extended period of time.


keyman-server:

DavidLRowe:
See comments on #1312 

Also, there's no particular reason to put this in "experimental" as opposed to "release".

You might want to wait and see what feedback you get on the other mobile layout before pursuing this one, but you (and your community!) will have to be the judge of that.

Thanks for your work in developing these keyboards. Let me know if you have any questions.

PR #1314, opened 2020-09-02
[bu_phonetic] update the metadata and .kmn

MakaraSok:
The keyboard has been spot-tested and proof to work as expected. 

There are some notes on this version:
* move from \legacy to \release
* apply new file organization
* slightly tweak the .kmn to fix unsupported errors and empty stores
* no changes to the core of the keyboard funcutionalities
* contents of the documentation and help page are kept as they are for now
* contents of welcome.htm and readme.htm are adapted from the legacy welcome.htm
* the .doc cannot be included in the package, only .pdf is included to be referred to in welcome.html and readme.htm
* icon was imported from the old .bmp file
* the keyboard name has been changed to "BU Phonetic" per Lorna's suggestion
* this keyboard currently supports Windows and macOS

Please feel free to give any feedback and suggestions to get the metadata up-to-date.


LornaSIL:
Thanks @MakaraSok I have some comments (of course!)

* I think it would be good for the version to be 2.0. It never had a proper version since it was based on the year, but we could consider that version 1 and bump this to 2.0. That would mean changing the version in the HISTORY.md, README.md, .kmn.
* I see you had to change the minimum Keyman version to 10 in the kmn file. Can you remove all references to Keyman 6.0 as a minimum? Don't worry about saying Keyman 10. Just remove statements like "(for Keyman 6.0 and higher, www.tavultesoft.com)" from the readme.htm, welcome.htm, bu_phonetic.php, and keyboard_info files.
* The language tag in the keyboard_info file is "la", but in the legacy/b/bu keyboard_info file it listed:
```
        "en",
        "fr",
        "de",
        "und-fonipa",
        "es"
```
You probably don't need to list "en", but I think you should include the others and remove "la". Then, of course, make the .kps consistent with those additions/deletions so the two match. I'm not sure what to do about "und-fonipa". Currently, I believe it should be "und-fonipa" in the keyboard_info file and "und_Latn" in the .kps. That will change with version 14.0, but for now I believe it needs to be that way.

* In the kmn file
   * Can you change the name to "Bu Phonetic"
   * Can you delete the "message" header? It is referencing a freeware license.
   * For the .bmp, it would be better to open it in a graphics program and save it as bu_phonetic.ico (that will better match our current naming scheme and reduce chances of problems in the future). Then you would need to change the header statement to `store(&BITMAP) 'bu_phonetic.ico'`
* In the welcome.htm, the "target" for the pdf is `<a href='help\BUkeys.pdf' target='_blank'>PDF format documentation</a>`. Since the pdf will be in the package it won't actually be in the help folder any longer, so you can delete `help\` and it will work.
* Lastly, can you add an extra folder at the root level called "extras", so you have `bu_phonetic\extras` and then put the bukeys.doc file in that folder. That way we still have the source of the pdf when it needs updating.

LornaSIL:
Oh, another thing is we need to deprecate the old keyboard.
So, add a DEPRECATED.md` file to the `legacy\bu` folder. It should have one line of text that says "This keyboard has been deprecated and replaced by release/b/bu_phonetic" You can copy the file and then modify it from the `legacy\b\bangla_joy_kab` folder.

Then, the bu_phonetic.keyboard_info file needs the deprecation info added. Basically, this should be your new keyboard_info file:

```
{
    "license": "mit",
    "languages": [
        "fr",
        "de",
        "und-fonipa",
        "es"
    ],
    "description": "The BU Phonetic Keyboard provides access to a wide range of characters for Latin-based scripts in Unicode 4.1.0 (www.unicode.org).",
    "related": {
      "bu": { "deprecates": true }
    }
}

```

LornaSIL (approved):
LGTM. (I deleted a .tmp file in extra)

PR #1316, opened 2020-09-04
[sahaptin_yakima] simpler resubmission

kelivermore:
This submission was made via GitHub, in a simpler manner based on David's previous comments. 

I am most grateful for the guidance!


keyman-server:

DavidLRowe:
There are differing spellings of Ichishkíin / Ichishkiin (with and without accented "i") in the welcome.htm. Is this intentional?

Eventually this keyboard will need an online help file. It is possible for us to accept the keyboard without the .php help file and add one based on the welcome.htm file.

The "targets" value (checkboxes in Keyman Developer or corresponding list in the .kmn file) can be set to "any". This will include all the ones you checked, plus add "web" which just means the desktop keyboard can be used on the web. (That is, you don't have to do anything extra to support the web option.)

I'll add more comments about the keyboard itself later.

DavidLRowe:
Comments on the keyboard:

Because Keyman defaults to displaying the US English key for any key that isn't redefined, some keys are producing a character even though the key is blank on the on-screen keyboard.

Keys that have blank keys showing, but actually produce characters when typed:
default layer: 1234567890-=[]\;/
shift layer: ~!@#$%^&*()_+QER{}|DFGJ:ZVB<>?

If you want these keys to not produce anything, the easiest solution would be to add these to the `store(english)` you already have:

```
C deadkeys for English-only characters
store(english) "qerdfgjzvb`"
+ any(english) > dk(notused)
```
Alternatively, if you want the key to actually produce the character, then adjust the on-screen keyboard (OSK) and documentation. (You may want include rules in the .kmn file for such keys rather than rely on the Keyman defaults.)

BUT please note that we strongly suggest the desktop and mobile keyboards be able to produce the same characters, and you have digits and punctuation available on the numeric layer. So that would argue for including those characters on the desktop keyboard.

Touch layout: 
- default layer, fourth row, fourth key from left is labeled "x", but produces "x̱".
- shift layer, first row, fourth key from left is labeled " ’ "  but produces nothing.
- shift layer, first row, fifth key from left, longpress is blank, but produces "Ɨ́".
- shift layer, fourth row, fourth key from left is labeled "X", but produces "X̱".
- shift layer, fourth row, fifth key from left is labeled "Xw" but produces nothing.
- shift layer, fifth row, key to the right of space bar is blank, but produces ">". Longpress on that key is labeled "," but produces "<".
- numeric layer, key labeled "&" produces "^".
- numeric layer, on "(" key, longpress key labeled "[" produces "l" (lowercase "L").
- numeric layer, on ")" key, longpress key labeled "]" produces nothing.


In addition to the above corrections, here's something to consider:
Just curious why 'x' with macron below is on the default layer, whereas 'k' with macron below, belted 'l', and vowels with acute accent are on the alt layer. It may well be that the 'x' with macron below is much more common than the simple 'x', but it might also be that users will find it easier to think of "plain" characters being on the non-Alt layers and the characters with "something else" (macron below, or belt, or accent) on the Alt layers.

kelivermore:
Shix̱ páchway, David,
I really appreciate these notes and comments! I've gone in and fixed them, and am ready to re-submit.

On your note about the x̱ character: that character is so much more prevalent than the regular x, and as I was testing the keyboard out, I found it such a hassle to have to hit the ALT every time I needed the x̱. When I explained that I could make that switch to some colleagues, they agreed that it would be a smart move for when people get used to the keyboard. Again, I really appreciate your thinking about the usability of this project!

PR #1319, opened 2020-09-07
[sil_hawaiian] move to release from legacy

MakaraSok:
* Imported only the Unicode keyboard, even though the original package has both Unicode and ANSI keyboards
* Updated metadata


LornaSIL:
Thanks for doing this. We have been in the practice of prefacing the filenames with "sil_" rather than making it the suffix. So, instead of "hawaiian_sil" I think it should be "sil_hawaiian". The actual keyboard name would remain the same as it is "Hawaiian (SIL). 
* That means it should be moved to the "sil" folder
* That means renaming all the files and the internal pointers (like in the kpj, kps, kmn). Sorry :(
* I would like to see the fonts moved from the help folder to `sil_hawaiian\extras`.
* The fonts should be added to the .kps.
* You'll need to add a "related" section to the keyboard_info file:
```
{
    "license": "mit",
    "languages": [
        "haw-Latn"
    ],
	  "description": "Hawaiian (SIL) is a Unicode keyboard with compatible fonts, included with permission from Hale Kuamo'o. \r\n",
  "related": {
    "hi-keyboard": {
      "deprecates": true
    }
  }
}
```

I think that's it.

MakaraSok:
Changes have been made per feedback. If anything is missed, please give me a heads-up.

LornaSIL (approved):
LGTM

PR #1320, opened 2020-09-08
[venetia_et_histria] Move to release from legacy

MakaraSok:
* Removed Tavultesoft from metadata

_Please check the embedded layouts in the .php file. I was not sure the `data-states` of the `osk` are right. If there is a reference on what to call for all available layers, please do forward the link to me._


LornaSIL:
Thanks for this submission.
The Desktop keyboard works fine. Is there a reason you can't say "any" as a target?
* Touch layout
   * does not allow you to access the shift layer. You need to change the "Shift" key to have "Next Layer: shift" and on the shift layer, have the "Shift" key to have "Next Layer: default".
* The copyright statement used to be for Tavultesoft. We should transfer the copyright to "SIL International" and not to the authors. At least that is my understanding of how we are supposed to handle that. That would require changing in README.txt, kmn, kps, readme.htm, welcome.htm, etc.
* A DEPRECATED.md file should be added to `legacy\v\venetian`
* The keyboard_info file needs the "related" or deprecated info like this:
```
{
    "license": "mit",
    "languages": [
        "vec-Latn", "lld-Latn"
    ],
	  "description": "Unicode keyboard for entering Venetian and Ladin languages, covering input of all the necessary Latin characters.",
  "related": {
    "venetian": {
      "deprecates": true
    }
  }
}
```

Regarding the osk data-states. The mobile (tablet or phone) just uses the layer name as you name it. The Desktop is more complicated. I don't think it's documented anywhere at this point except in one of my google docs. These are the ones I know work:
```
            'default':'Unshifted',
            'leftalt':'Left Alt',
            'rightalt':'AltGr (Right Alt)',
            'alt':'Alt',
            'leftctrl':'Left Control',
            'rightctrl':'Right Control',
            'ctrl':'Control',
            'ctrl-alt':'Control+Alt',
            'leftctrl-leftalt':'Left Control+Left Alt',
            'rightctrl-rightalt':'Right Control+Right Alt',
            'leftctrl-leftalt-shift':'Left Control+Left Alt+Shift',
            'rightctrl-rightalt-shift':'Right Control+Right Alt+Shift',
            'shift':'Shifted',
            'shift-alt':'Alt+Shift',
            'shift-ctrl':'Control+Shift',
            'shift-ctrl-alt':'Control+Alt+Shift',
            'leftalt-shift':'Left Alt+Shift',
            'rightalt-shift':'AltGr (Right Alt)+Shift',
            'leftctrl-shift':'Left Control+Shift',
            'rightctrl-shift':'Right Control+Shift'

```

I tried to test your help file and currently it's not displaying the layouts, but I think something must be messed up in my testing because I'm pretty sure you have the correct states.

MakaraSok:
Thank you for the review. Updates have been made in the new commits and my response to one of your comments is shown below.

> Is there a reason you can't say "any" as a target?

The original version does not include 'any' and when taking a closer look at it, it does not have 'linux' in the target.

LornaSIL:
Keyman for Linux did not exist when this keyboard was created. I don't see
a technical reason not to add Linux as a target.

LornaSIL (approved):
LGTM

PR #1323, opened 2020-09-10
[sil_korean_korda] move to release from legacy

MakaraSok:
* Updated metadata

_This keyboard package contains to keyboard layouts, jamo and latin._
_There is an outstanding issue which I do not know how to fix:_
image


LornaSIL:
I'll try to look at it...but I do know we cannot have two keyboards in one folder. You will need to split the latin one out into another keyboard. (I know that won't help this problem.)

The way we've done this in the past is after the two keyboards are approved we make a package and add both keyboards into the package...but I seem to remember @mcdurdin not wanting to keep going with packages. What we *can* do is to make the keyboards related (through the keyboard_info file) but not deprecated. So you do the line about "related" but instead of deprecated true you say deprecated false.

LornaSIL:
Remember the keyboard filename must follow the folder name. I think I would create a "sil_jamo_korda" and a "sil_latin_korda" (or perhaps reverse them to "sil_korda_jamo" and "sil_korda_latin"). Each keyboard in a different folder. Then the kpj and jps must follow the folder/filename convention as well. (If we decide to do a package later than contains both keyboards then we could use sil_korean_korda.

Regarding the error for this rule: `+ ' ' > use(tojamo_jongseong) call(jamo_combine) dk(space)`. You could try to move the "use" statement to the end of the rule and see if it works as intended `+ ' ' > call(jamo_combine) dk(space) use(tojamo_jongseong) `

Does the documentation give you enough information to know what it is doing? If you cannot figure it out, do the jamo_korda one and I'll try to tackle the latin_korda one after.

mcdurdin:
> Regarding the error for this rule: `+ ' ' > use(tojamo_jongseong) call(jamo_combine) dk(space)`. You could try to move the "use" statement to the end of the rule and see if it works as intended `+ ' ' > call(jamo_combine) dk(space) use(tojamo_jongseong) `

That won't work correctly; however, you could add a pair of match/nomatch rules to the `tojamo_jongseong` group to do the same thing:

```
+ ' ' > use(tojamo_jongseong)
```
...
```
group(tojamo_jongseong)
```
...
```
nomatch > call(jamo_combine) dk(space)
match > call(jamo_combine) dk(space)
```

PR #1324, opened 2020-09-10
[sil_moore] move to release

MakaraSok:
* Updated metadata
* Added Linux support

_A similar error to the one in Korean Korda keyobard is experienced here._
image


LornaSIL:
* In the keyboard_info file the `-latn` should be changed to `-Latn`
* Unfortunately, the links in the welcome file aren't really accurate anymore. You could link to https://software.sil.org/doulos/ and https://software.sil.org/charis/
* It might be good to add an on-screen keyboard. That would be helpful to users. 
* You should also add the deadkeys to the OSK keycaps so it wouldn't be so hard to figure out:
   * [SHIFT K_LBRKT] "¨"
   * [RALT K_2] "~"
   * [RALT K_7]  "`"
   * [RALT K_LBRKT] "^"
* You might consider adding the layout to the welcome.htm file as I'm not finding the keyboard as intuitive as it indicates.

So, regarding the bad rules. For a line like this:
`+ any(keys_0u) > use(dkouts) index(outs_0u,1)` I think you just delete `use(dkouts)`
Then delete the group:
```
group(dkouts)
any(deadkeys) > index(deadchars, 1)
```

This rule `any(deadkeys) + ' ' > index(deadchars, 1)` looks like it does what they were wanting (to output the deadkey). The old version of Keyman must have been more forgiving and didn't fail with those rules.

MakaraSok:
> * Unfortunately, the links in the welcome file aren't really accurate anymore. You could link to https://software.sil.org/doulos/ and https://software.sil.org/charis/

This may not be necessary to update because the users will be taken to [this page](https://scripts.sil.org/cms/scripts/page.php?site_id=nrsi&id=) where they can browse for the fonts.

MakaraSok:
Please do not approve this keyboard just yet as some changes are expected per users' feedback.

LornaSIL:
The changes you made look correct. We will await your feedback.

MakaraSok:
> Please do not approve this keyboard just yet as some changes are expected per users' feedback.

This comment was not supposed to be here. Please disregard it and move forward. Thanks.

LornaSIL (approved):
LGTM.

PR #1326, opened 2020-09-11
[sil_korda_jamo] split korda jamo from korda latin

MakaraSok:
* The name was adapted from the existing keyboard name, Korean KORDA Jamo.

_split korda jamo from korda latin_ 

_ref: https://github.com/keymanapp/keyboards/pull/1323_


LornaSIL:
* There are three files were the link is to the "help" folder for about_korda.pdf. It's fine for that file to be in the help folder so that it's not duplicated everywhere. However, when that file is pulled into the package it will just be in the same folder as everything, so the link should not include "help". Instead of `<a href='help/about_korda.pdf'>` it should be `<a href='about_korda.pdf'>` This needs changing in readme.htm, welcome.htm and sil_korda_jamo.php.
* sil_korda_jamo.php needs these lines deleted:
```
﻿<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01//EN" "http://www.w3.org/TR/html4/strict.dtd" />
<html>
<head>
  <meta http-equiv="content-type" content="text/html; charset=utf-8" />
  <title>Korean KORDA Jamo (SIL) Keyboard Help</title>
  <link rel="stylesheet" href="style.css" type="text/css" />
</head>
<body>
   <h1>Korean KORDA Jamo (SIL) Keyboard Help</h1>
```
as well as the last two lines:
```
</body>
</html>
```
I'm not sure how to handle the `"sil_korda_latin": {"deprecates": false}` line since that one isn't approved yet. But the build doesn't seem to fail so it's possibly okay.

Otherwise, I think the keyboard looks good.

MakaraSok:
> * sil_korda_jamo.php needs these lines deleted:
> 
> ```
> ﻿<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01//EN" "http://www.w3.org/TR/html4/strict.dtd" />
> <html>
> <head>
>   <meta http-equiv="content-type" content="text/html; charset=utf-8" />
>   <title>Korean KORDA Jamo (SIL) Keyboard Help</title>
>   <link rel="stylesheet" href="style.css" type="text/css" />
> </head>
> <body>
>    <h1>Korean KORDA Jamo (SIL) Keyboard Help</h1>
> ```
> 
> as well as the last two lines:
> 
> ```
> </body>
> </html>
> ```

@LornaSIL 
 Will the css file work without referring to it in the head?

LornaSIL (approved):
LGTM

PR #1327, opened 2020-09-11
[sil_korda_latin] split korda latin from korda jamo

MakaraSok:
* The name was adapted from the existing keyboard name, Korean KORDA Latin.

_split korda latin from korda jamo_ 

_ref: https://github.com/keymanapp/keyboards/pull/1323_

> Regarding the error for this rule: + ' ' > use(tojamo_jongseong) call(jamo_combine) dk(space). You could try to move the "use" statement to the end of the rule and see if it works as intended + ' ' > call(jamo_combine) dk(space) use(tojamo_jongseong)

This was done, but not tested yet. Please help test this out and let me know if anything slipped through along the way. Thank you, @LornaSIL .


MakaraSok:
fix the error per @mcdurdin's comment at https://github.com/keymanapp/keyboards/pull/1323#issuecomment-691733689.

LornaSIL:
These comments are similar to the ones I made for sil_korda_jamo:

* There are three files were the link is to the "help" folder for about_korda.pdf. It's fine for that file to be in the help folder so that it's not duplicated everywhere. However, when that file is pulled into the package it will just be in the same folder as everything, so the link should not include "help". Instead of `<a href='help/about_korda.pdf'>` it should be `<a href='about_korda.pdf'>` This needs changing in readme.htm, welcome.htm and sil_korda_jamo.php.
* sil_korda_jamo.php needs these lines deleted:
```
﻿<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01//EN" "http://www.w3.org/TR/html4/strict.dtd" />
<html>
<head>
  <meta http-equiv="content-type" content="text/html; charset=utf-8" />
  <title>Korean KORDA Latin (SIL) Keyboard Help</title>
  <link rel="stylesheet" href="style.css" type="text/css" />
</head>
<body>
   <h1>Korean KORDA Latin (SIL) Keyboard Help</h1>
```
as well as the last two lines:
```
</body>
</html>
```

* Also, the help filename does not follow the naming scheme. It should be `sil_korda_latin.php` and not `sil_korean_korda.php`
* Even though "Latin" is in the keyboard name, this keyboard still outputs Korean characters, so the script in the .kps should not be Latin. I think the language tag should just be "ko" as you have it in keyboard_info.
* You should add a DEPRECATED.md file to `legacy\k\korda` either to this PR or the other one. The contents should be something like:
```
This keyboard has been deprecated and replaced by release/sil/sil_korda_jamo and release/sil/sil_korda_latin
```

LornaSIL (approved):
LGTM

PR #1328, opened 2020-09-11
First Update - Qom Keyboard

MLocria:
This update introduces the following changes to the keyboard:
1) Documentation fixed
2) Adjustments and reorganization of the on-screen keyboard

I switched my files to a Linux version of git and it seems that GitHub shows as if the whole file has changed in some cases. However, if you look closely, that's not the case. I'm not too sure about what caused this, but please let me know if there's anything I can do.

Cheers,
Manuel


keyman-server:

MLocria:
In the end, further changes will be made before creating the pull request.

DavidLRowe:
WIndows and LInux use different line ending conventions: CR+LF on Windows vs LF on Linux.

MLocria:
Thanks for clarifying, David!

PR #1333, opened 2020-09-17
[gandhari] Moved from legacy to release

LornaSIL:
**Moved** from legacy to release.
Author gave permission to change license from freeware to MIT.
Keyboard was decomposed from kmx.
Metadata files added.


DavidLRowe:
HISTORY.md: change date from 2019 to 2020

LICENSE.md, README.md, .kmn (and .kps and .kpj): change end date of copyright date range to 2020

Does gandhari-keyboard-2.7/DEPRECATED.md need to change to refer to release/g/gandhari (rather than legacy/g/gandhari)?

The .kmn file has two store(&KEYBOARDVERSION) statements, one with '2.8' and one with '10'. It seems the second should be removed.

Would you want to include "web" as a target?

I'm assuming there is no OSK because of the use of deadkeys.

Trivial: 
The .kmn file is not UTF-8.
There seems to be an unused "group(junk)" at the end of the .kmn file.
There is no icon.

LornaSIL:
Your "trivial" comment about UTF-8 was not so trivial. I hadn't been able to create an OSK, it would hang Keyman and it turned out fixing that made it so I could create an OSK.

There was no icon in the original keyboard and I'm not sure what it should be, so I chose to leave it out.

DavidLRowe:
Good to know that a non-UTF-8 .kmn file can cause trouble for Keyman Developer.

All the changes look good. Only one left is the copyright date in the .kmn file.

DavidLRowe (approved):

PR #1335, opened 2020-09-17
chore: update code signing

mcdurdin:
Removes sha1 signing and hands signing over to agent on Keyman build agents for Windows bundled installers.


darcywong00 (approved):
lgtm

PR #1339, opened 2020-09-21
[sahaptin_yakima] resubmission

kelivermore:
Changes made to better align the desktop keyboard with the touch mobile/tablet keyboard. This submission is via Github.


keyman-server:

DavidLRowe:
This is looking very good! Just a few fairly minor things to check and/or correct:

In the "welcome.htm" file:
- Check the text: "... on only Ichishkiin letters. Only letters shared with the Ichishkíin alphabet ...". It seems to have two different spellings of the language name in adjacent sentences (but perhaps these are different words?). Anyway, please check all your documents for "Ichishkiin" vs "Ichishkíin".
- The HTML text refers to the file "sahaptin_yakima_computer_layout_alt.png", but that file's name actually ends with ".PNG". On Windows that won't matter, but Mac and Linux have case sensitive file names. I suggest that you rename the file so that it ends with ".png". (Windows may not let you make that rename directly. You may have to rename the file to "sahaptin_yakima_computer_layout_alt.xxx" and then rename it to "sahaptin_yakima_computer_layout_alt.png".) Alternatively, you could change the text of welcome.htm to refer the file using the name "sahaptin_yakima_computer_layout_alt.PNG".

sahaptin_yakima.kmn
- The "targets" value should be 'any' (by itself) rather than 'any web'. 
- In order to make the character set for the desktop and touch keyboards similar, it seems that you need to 
  - add "~" to store(english) so that that key won't produce any output
  - add ":" to the On-screen keyboard and other documentation since that key is producing the character (and you want to match the characters produced in the touch layout)

There's a minor issue with the touch layout: changing from the default layer to shift layer causes some keys in the bottom row to "jump". You can fix this by changing the width of the space bar on the shift layer to match that on the default layer.

kelivermore:
Hi David, thank you for all of this feedback! I'm going to resubmit now that I've made these changes.

Thank you for your patience during this process! I am really, really grateful.

PR #1340, opened 2020-09-21
[wancho] Add Wancho keyboard

punchcutter:
Noto Wancho is now on Android 11 and iOS 14, but there are no keyboards yet. I made this one with Banwang, the creator of Wancho script.


keyman-server:

LornaSIL:
Congratulations on submitting the first Wancho keyboard for this repository!

As usual, I do have somethings that should be fixed. I realize you have submitted this into the "experimental" section so I won't require that everything must be changed, but I'll let you know everything I see.

**Required Changes***
* The keyboard_info file needs the language tag changed from `nnp-Latn-IN` to `nnp-Wcho`
* Likewise, the .kps file, should be `nnp-Wcho` (the `-IN` is unnecessary)
* In the .kps you need to remove the file `NotoSansWancho-Regular.ttf`. This is because you selected a location on the desktop. Once you remove that file, then you should add it back in, but you should add the one that is in the source folder.

**Suggested Changes***
* You have as a description for the keyboard "Wancho generated from template". I would like to see something more descriptive like "This keyboard is designed for the Wancho language of India using the Wancho script." You should add this description (or even more information) into these files:
   * wancho.keyboard_info
   * README.md
   * readme.htm
   * welcome.htm
* The wancho.kmn file ... The On-Screen Keyboard has some blank keys that produce characters. I would suggest that you go to the OSK, click on "Auto-fill underlying layout" and then click on "Fill from layout" again. That way people can see all the symbols and brackets. If you don't wish for the keyboard to produce the Latin letters that it currently does, you could add two lines to your .kmn file that will prohibit producing them (WDFXVB)
```
store(nul) "WDFXVB"
+ any(nul) > nul
```
Adding these two lines to your kmn code will prevent outputting those characters.

Regarding the touch layout. Are you wanting to have Latin letters on the keycaps? I know some people do, but it seems like you might want Wancho letters? I guess that's up to you. 

The main thing you should look at is on the numeric layer, the `$` is producing a Wancho character. Is that your desire? If you want `$` output then you should change the "code" from modifier "shift" Code: "K_4" to Modifier: "(current layer)" Code: "U_0024".

Since Wancho script is fairly new and not in many operating systems, there is another change I would like to see you make in the keyboard_info file. The lines after the language code add the font so that the mobile keyboard will have access to the font within Keyman. This is basically what I'd like your keyboard_info file to look like:

```
{
    "license": "mit",
	  "description": "This keyboard is designed for the Wancho language of India using the Wancho script.",
    "languages": { 
         "nnp-Wcho": { 
              "font": { 
              "family": "NotoSansWancho", 
              "source": "NotoSansWancho-Regular.ttf" 
           } 
        }
  }
}
```

Eventually, I'd like to see you expand the welcome.htm file, but for now this should be sufficient.

punchcutter:
Great, thanks for the feedback. One thing I had asked about before was removing Naga from Wancho Naga because they feel this is a derogatory term. I removed it here and I think that's fine. 

I removed the TTF and added back in the .kps, but now I don't see Wancho on screen when I install on Android. Not sure if I did something wrong there.

I auto-filled the layout so the underlying Latin is there. That's fine like that. And yes, we want Wancho letters on the keycaps, not Latin. 

For the `$` slot it is the Wancho currency symbol which is how Banwang decided to add it. I can check if there's any reason not to do it that way, but he seemed to think it was fine to replace `$`.

LornaSIL:
A few things left:
* In the kps file it would be good to go to the "Keyboards" tab and select NotoSansWancho for the font in both locations. You may have had that set before, I can't remember, but when I asked you to remove the font and add it back in then that association got lost.
* You mention wanting Wancho letters on the keycaps. There are Wancho letters on the desktop keyboard, but not on the touch layout. Is that how you want it? If you want Wancho letters on the touch layout you'll need to change them from Latin to Wancho.
* Regarding the language name. Altho you hand edited the name to remove "Naga" I am afraid that when the Keyman system builds the keyboard it's going to use the official name, not the name you hand edited. At this point we don't really have a good solution for you. I know this is an issue for several language groups, but we don't currently have a way to work around it.
* You don't have an online help file, but I can create that for you once we get this keyboard online.

punchcutter:
Yes, I had the fonts set in the .kps and also had set the Touch keyboard, but somehow it all got reverted when I made the other changes. I fixed that now.

 I had asked previously about removing "Naga" and Marc said it relies on the system so that's fine if there's nothing we can do, but it seemed to work here when I remove it and install on Android so I figured it was worth a shot even if it doesn't work everywhere. Anyway, right now it's more important to have a keyboard that can be used.

DavidLRowe:
@LornaSIL @mcdurdin Is it worth including something in keyman.com FAQ (or whatever it's called) that directs people on what to do if they think the language name is wrong?

LornaSIL (approved):
LGTM

LornaSIL:
Your keyboard is now online and hopefully the help file will be there soon.

mcdurdin:
> @LornaSIL @mcdurdin Is it worth including something in keyman.com FAQ (or whatever it's called) that directs people on what to do if they think the language name is wrong?

Yes, I think perhaps a knowledge base article is appropriate -- under /knowledge-base on help.keyman.com. Would you be willing to write something @DavidLRowe? Keyman 14 will improve the situation with the visibility of language names from .keyboard_info files, but of course it is good to feed that information back to ISO639, Ethnologue, langtags, etc. as well.

Issue #1332, opened 2020-09-16
Broken internal links for keyboard help files


darcywong00:
CI is now checking broken internal links on https://github.com/keymanapp/help.keyman.com PRs.

The following keyboard help have missing files:
(if we fix, maybe only fix the most current versions ~and minor bump in versions?~)
```
galaxie_greek_positional/1.0/keyboardchart.htm
galaxie_greek_mnemonic/3.2.1/keyboardchart.htm
indic%20roman%20transliteration/1/welcome_files/filelist.xml
khowar/1.0/Khowar.ico
sxava/1.0/sxava_ekzerco_1.png
tibetan_direct_input/1.4/0f96.png
gff_tir_er_7/1.0/TigrignaErTyping-English.pdf
gff_tir_et_7/1.0/TigrignaEtTyping-English.pdf
thamizha_anjal_paangu/2.1/keyboardchart.htm

lingfilsemitica/3.6.0/welcome_files/filelist.xml:
lingfilsemitica/3.6.0/welcome_files/editdata.mso
lingfilsemitica/3.6.0/welcome_files/item0007.xml
lingfilsemitica/3.6.0/welcome_files/themedata.thmx
lingfilsemitica/3.6.0/welcome_files/colorschememapping.xml
```

gff_amharic needs a lot of AmharicTyping-X.pdf

These are missing keyrenderer.js and welcome.htm/welcome-X.htm:
```
thamizha%20bamini/2.0/keyboardchart.htm
thamizha%20typewriter/2.0/keyboardchart.htm
thamizha%20anjal/2.0/keyboardchart.htm
tamil%20-%20visual%20media/1.2/keyboardchart-modular.htm
tamil%20-%20visual%20media/1.2/keyboardchart-typewriter.htm
```

darcywong00:
low priority

LornaSIL:
If it's just a change in the help file folder of a keyboard we don't need to bump the version number of a keyboard, right? Or does it not get transferred to the help repo if we don't bump the version?

Some of these are now deprecated. Do we need to do anything with the deprecated ones?

darcywong00:
Oh right, we can leave the version intact. Let's wait to see what @mcdurdin wants to do with the help for deprecated keyboards

mcdurdin:
The broken links cause trouble for us -- if we can remove the broken links in the .php files, that's all we need to do.

DavidLRowe:
[...]

PR #1341, opened 2020-09-21
[help_links] Fix broken links in help files

LornaSIL:
Relates to #1332 

I cannot find a problem with gff_amharic. The pdf files seem to be properly named and in the help folder. (Later, I see older versions of the help files are missing these pdfs. Do I need to submit a PR to the help.keyman.com repo?)

I believe the other issues are resolved.


DavidLRowe (approved):
As far as I can tell these changes look fine!

PR #1349, opened 2020-09-29
[anii_2015_fr] move to \release

MakaraSok:
* Added screenshots to welcome.htm


LornaSIL:
* I would prefer that the folder/filename be `anii` and not `anii_2015_fr`. People used to add the year to the keyboard name as a way of versioning a keyboard and that's not really how we do it these days. (Sorry I didn't say anything sooner).
* I think I would like the keyboard name to be "Anii" rather than "Anii_2015_fr".
* In the .php file you have `  <div id='osk-mobile' data-states='default shift numeric'></div>` I think that "mobile" needs to change to "phone" because that's what it is called in .keyman-touch-layout
* In `anii_2015_fr.keyman-touch-layout` tablet layout, there are many, many layers that are blank. You could delete all but the "default" and "shift" layer.
   * You would also need to delete all the Ctrl, Alt, Alt Ctrl keys on the default and shift layers.
* On the phone layout:
   * Default layer, `;` key - doublecheck all the longpress keys. They are producing wrong results based on keycaps
   * Shift layer, `/` key - produces semi-colon and all the longpress keys seem to be producing wrong results based on keycaps
   * Numeric layer, I don't see how to get a `1` or a `7`. They are strangely laid out. Also, some of the longpress keys are producing the wrong characters based on keycaps: <, {, >, }

I haven't reviewed the tablet layout yet.

MakaraSok:
> * In `anii_2015_fr.keyman-touch-layout` tablet layout, there are many, many layers that are blank. You could delete all but the "default" and "shift" layer.

Would these keys "Ctrl, Alt, Ctrl-Alt etc" be used by external keyboards on tablet devices?

MakaraSok:
Touch layout for phone has been redesigned based on the OSK. Other things have been fixed.

LornaSIL (approved):
LGTM

PR #1351, opened 2020-09-30
fix: broken links

mcdurdin:
Relates to keymanapp/help.keyman.com#233.

PR keymanapp/help.keyman.com#237 contains the same fixes, in order to verify prior to commit here. It also fixes older versions of docs that would not be updated by the keyboard documentation push. I will revalidate to ensure I haven't missed anything once this merges.


darcywong00 (approved):
lgtm (compared with keymanapp/help.keyman.com#237)

PR #1353, opened 2020-10-01
[sil_tepehuan] move to release from legacy

MakaraSok:
* Moved to \release from \legacy\s\spantepu
* Added Doulos SIL font to the package
* Deprecated  mextep 2005 and spantepu 2014
* Enhanced documentation

* Created by SIL Mexico 2014


MakaraSok:
There are multiple dates in the keyboards related to this:

- maxtep created in 2005 per .kmp, updated in 2008 per .keyboard_info
- spantepu (no date in .kmp) updated in 2014 per .keyboard_info

Using 2008 for now as it's the earlier recorded date in the .keyboard_info.

LornaSIL:
It looks good to me. The only thing I would suggest is that you add "Web" as a target. It compiles fine if you add Web and I see no reason not to.

LornaSIL (approved):
LGTM

PR #1354, opened 2020-10-02
[sil_ywq_plrd] Hei Yi – Black Yi (SIL) new keyboard

LornaSIL:
I also added two different fonts for Miao. One set is for this keyboard and one set for the sil_hmd_plrd keyboard which I hope to update.


DavidLRowe:
Shift+W is blank on OSK but produces U+16F28.
Shift+X is blank on OSK but produces "X".

I think everything else is fine.

DavidLRowe (approved):

PR #1356, opened 2020-10-05
[sil_lpo_plrd] Add new Lipo (SIL) keyboard

LornaSIL:


DavidLRowe:
Shift+E produces U+16F91 and Shift+C produces U+16F8F though the OSK has blanks for both these keys. Does the OSK or the documentation need to indicate what these keys do?

DavidLRowe (approved):

PR #1358, opened 2020-10-05
[sil_yna_plrd] Add new Gan Yi (SIL) keyboard

LornaSIL:


DavidLRowe:
Minor: welcome.htm (and also .php) have a duplicated sentence: "A font must be installed to support this keyboard."

] and ' show blank on OSK but produce U+16F90 and U+16F91. (Could you use U+2197 north east arrow and U+2191 upwards arrow to represent these on the OSK?)
x - key is blank on OSK but produces "x".

DavidLRowe (approved):

PR #1360, opened 2020-10-08
[klingon-latin] revision of initial commit for klingon-latin

panglott:
The initial commit for Klingon Latin needed some files to be reorganized and other changes to be made, which have been made in this pull request.


keyman-server:

DavidLRowe:
The reorganization of the files looks very good. Just a few more things to change.

klingon_latin.kmn
Sorry I wasn't clearer in my previous comment. Please include "(c)" in the copyright field, so that the whole field is "(c) 2020 Jeremiah Rose"

There are some differences between the On-screen keyboard (OSK) and what the keyboard actually produces. You'll need to change either the OSK (to reflect what the keyboard actually produces) or change the .kmn file so that it outputs what the OSK displays.

(1) Shift+6 is blank on OSK, but produces "^". 
If you want the key to produce nothing, the .kmn file needs a rule such as:
```
+ [SHIFT K_6] > nul
```
to override the Keyman Developer default of displaying the US English key value for any key that isn't redefined.
If you want the key to produce the "^" character, then add that to the OSK.

(2) The same is true for the Shift+- producing the underscore character. 

(3) And also for Alt+Shift+- producing the underscore character.

(4) Shift+t shows "t" on the OSK, but produces "ts".

Touch layout:
Shift layer, "nD" (longpress on "D") doesn't produce anything.
The issue is that, for that longpress key, the Modifier is set to "(current layer)", which in this case is the Shift layer, so the key is passed to the .kmn file as [Shift T_n_D] which fails to match the rule in the .kmn file. The solution is to change the Modifier to "default" so that no "Shift" is attached and just "[T_n_D]" is passed to the .kmn file.

welcome.htm
The link to klingon.wiki seems malformed. Currently it has "link:http://klingon.wiki/En/Fonts" but the link needs to be "http://klingon.wiki/En/Fonts" or something similar, that is without the "link:".)

It would be nice if the welcome.htm had some additional help for users. One fairly easy way to do this is with a graphic file for each the four layers (default, shift, alt, alt + shift). From the OSK tab in Keyman Developer (once you are happy with the OSK changes), you can generate such graphics using the Export button (above the keyboard display).
- Click Export
- Change "Save as type" to "Image (*.png)"
- Navigate to the klingon_latin/source/welcome folder
- Specify a file name (klingon_latin, for example)
- Click Save

In the Output Bitmap Parameters dialog, select:
- One file per shift state
- Output Unicode keyboards

This should create four files:
- klingon_latinU_.png
- klingon_latinU_A.png
- klingon_latinU_S.png
- klingon_latinU_SA.png

which you can reference in the welcome.htm file.

DavidLRowe:
There is also work to be done on the Keyman package source (.kps) file. This is in the Packaging section. (In Keyman Developer, click on the .kpj file at the top, then the "Packaging" tab along the bottom, then on the .kps file that is displayed)

If you like, I think that I have enough information to complete that for you. But perhaps you'd rather figure it out for yourself, in which case I'd be happy to give you any help you need.

release/k/klingon_latin/klingon_latin.keyboard_info
```diff
@@ -0,0 +1,5 @@
+{
+    "license": "mit",
+    "languages": ["thl-Latn"],
```
darcywong00, line 3:
```suggestion
    "languages": ["tlh-Latn"],
```
From [langtags](https://ldml.api.sil.org/langtags.json) .

When you add the keyboard to the .kps package file, use the same language tag.

DavidLRowe: Good catch! "thl" is a totally different language.

DavidLRowe:
@panglott Are you planning to continue work on this keyboard? If not, we can close this pull request. You can always reopen it at a later date, or submit a new one. Thanks for your interest.

DavidLRowe:
@panglott  We'll close this PR for now. Reopen it or submit a fresh one if you'd like to continue this project. Thanks!

PR #1367, opened 2020-10-11
GFF Gurage 0.7 Beta 4

dyacob:
Updates to include a new font library for the new orthography.  Documentation updated accordingly.


keyman-server:

LornaSIL:
@dyacob Sorry to delay on checking this. 
Basically, the changes look fine. However, there is one significant issue.

* You will need to update the `gff_gurage_and_amharic` *package* (go through version change, etc) because *it* calls for `Zebidar-R.ttf` which is no longer available. You'll have to update the version and then the file name for the font in the .kps.

dyacob:
I believe that this has been fixed with the last update.  thanks!

LornaSIL (approved):
LGTM

PR #1368, opened 2020-10-22
add myancode_san (sanskrit-myanmar) keyboard

pndaza:
None of previous Myanmar keyboard support full Sanskrit extensions. They cover only six codepoint. So i created a new one based on myWinExt.


keyman-server:

LornaSIL:
Thanks for your submission. I haven't looked at it yet, but I'm wondering if there would be value in us incorporating your changes into myWinExt? Since it's intended to be "Extended" it seems like that would be appropriate. Then we wouldn't have so many similar keyboards. We could of course acknowledge your work in myWinExt if we do that.
Let me know what you think.

pndaza:
Most of us (myanmar) familiar with legacy win-innwa keyboard layout.Among Existing Keyman myanmar keyboard, myWintExt has similar layout to win-innwa keyboard layout at no-shift state. But in myWinExt, writing for stacked constants ( eg. က္က) somehow different with unicode storage order. No dedicated key exists for virama . Nowadays many of myanmar unicode users used to type stacked constants using virama. SIL Myanmar3 keyboard layout support virama key. But layout is so many different with legacy layout and cannot type ⁠ေ ( sign vowel e ) first.

pndaza:
What i want is common layout for Myanmar script.
Nowaday myanmar keyboard  and other extensons script keyboards (mon, sagaw karen, western pwo and eastern pwo etc) have totally different layout.

Sanskrit have 10 extensions.
Mon have 7 extensions.
Sagaw karen have 4+1 extensions.
Western pwo have 9 extensions.
Eastern pwo have 3 extensions.
Kayah have 3 extensions.

In my new keyboard layout, i place Sanskrit extensions to adjacent keys. So replacing these keys with other script extensions, i think we can have same layout for Myanmar script.

LornaSIL:
Language tag in both keyboard_info `sa-Mymr`, not `san-Mymr` (The kps file is already correct). This is because this particular language does have a two letter language code.

The copyright is for SIL International currently. Do you want that changed to yourself? If you want to change the copyright to yourself you will need to change it in many places. (The license itself (MIT) cannot be changed.)

In the help file (php) you should change the header lines from:
```
  $pagename = 'myWin Extended (SIL)';
  $pagetitle = 'myWin Extended (SIL)';
```
to
```
  $pagename = 'MyanCode-San';
  $pagetitle = 'MyanCode-San';
```

On the keyboard itself, the output of SHIFT + 4 is `၎`, not what is on the keycap. If you want what is on the keycap you'll need to update the rule:
```
+ [SHIFT K_4] > '၎'

```
Everything else looks good. I don't read Burmese (or Sanskrit) so I can't check your documentation.

pndaza:
I fixed what you point. I have no knowledge about license. I just look other keyboard sources and copy license info.

LornaSIL:
Good. I should have mentioned that you'll need to regenerate the graphics for the layouts since you changed the SHIFT layer with the K_4.

Regarding the copyright holder...if you want it to be copyrighted to SIL International, that is fine, and you do not need to do anything futher. But, since you are the author you could change all the copyright strings to have your name instead. That's up to you.

pndaza:
Thank for pointing again. First, i make shift + K+4 output to '၎' to be similar with other layout. But standalone '၎' is rare case. '၎င်း' is short-form of 'လည်းကောင်း'  and can be found mostly. Android Gboard keyboard also output '၎င်း' instead of '၎'. So I change output of SHITF + K_4 to '၎င်း'.

LornaSIL (approved):
LGTM

PR #1369, opened 2020-10-25
GFF Gurage Legacy 0.1 & GFF Gurage 0.7

dyacob:
This PR contains two keyboards; `GFF Gurage` which is in the Keyman repository already and is adjusted here to work compatibly with the new keyboard `GFF Gurage Legacy`.  GFF Gurage Legacy is a keyboard designed to work with older Gurage orthographies from 1966 - 1998 and supported by Unicode 4.1.


keyman-server:

LornaSIL:
I'm going to just start adding comments, because I found a few typos and forgot to write them down!

Since all the filenames for gff_gurage_legacy use an underscore, the folder name for `gff_gurage-legacy` should be changed to `gff_gurage_legacy`.

Mobile gff_gurage_legacy:
* U+2BCD keycap (far top right key on default layer). The first longpress should be `U_2BC8`.
* non-Gurage layer, U+133D keycap, many of the longpress keys have the wrong Code. U+133A glyph should be `U_133A`, etc.

* The kmn, kps, welcome.htm, LICENSE.md and README.md, etc files have inconsistent copyright dates. I suppose the earlier dates refers back to the old deprecated keyboard.

I cannot claim to have tested everything, but this seems to be the things that jumped out at me.

dyacob:
Hi @LornaSIL , I believe the issues that you pointed out have been addressed.  Thanks for catching them.  I have also added more detail to the README.md to better indicate the continuity from the 2008 keyboard (gff-sgw-7) that this one replaces.

LornaSIL (approved):
LGTM

PR #1370, opened 2020-10-26
[clavbur9] Add a new language to keyboard and bump version

LornaSIL:
No changes to the keyboard.


DavidLRowe:
It seems that the date in HISTORY.md should be October 2020.

DavidLRowe (approved):

PR #1372, opened 2020-10-27
[kayan] Add Andika to the package

MakaraSok:
* Add Andika font to the package 
* Update numeric layer


LornaSIL:
@MakaraSok thanks for the update.
* Can you delete the .kvk file in the source folder (you have a kvks)?
* Can you rename `kayan-layout.js` to kayan.keyman-touch-layout` and then update the .kmn to reference the .keyman-touch-layout file rather than layout.js?

I think that's it.

LornaSIL (approved):
LGTM

PR #1379, opened 2020-11-04
[khmer_advanced] All rarely used character in one place

MakaraSok:
* the keyboard is designed specifically for academic researchers or the like who need to conveniently type those without using multiple key combination.
* the keyboard is built for all popular OSes.


LornaSIL:
I haven't checked your keyboard, but the build failed because you have an extra comma in the khmer_advanced.keyboard_info at the end of the description line.

LornaSIL:
I could approve it as is, but I wanted to check on the **phone** layout. It does not include the keys on the default digits row (the tablet layout does). Wouldn't you want to add those? Maybe to the shift layer for the phone layout?

If you do that, you'll need to regenerate your graphics. Also, your graphics in the help file are mislabeled. You have shift labels as default and default labeled as shift.

Lastly, in your project settings (in the kpj) can you please select the checkbox for "Check filename conventions". It's not essential, but it does make a cleaner keyboard.

MakaraSok:
Hi @LornaSIL, thank you for the catch. Yes, the Lek Attaks be in the Shift layer. I lost where I was when getting back to the keyboard after working on something else and the break. 

Please see the updates and fixes in the latest commit. Thank you!

LornaSIL (approved):
LGTM

PR #1383, opened 2020-11-08
Fix Gilaki language problems

LordArma:
We talked about it here: https://github.com/keymanapp/keyman/issues/3748


keyman-server:

DavidLRowe:
Thanks @LordArma for this keyboard submission. Unfortunately, we need to ask you to choose a different name for this keyboard project. It would not be fair to users of the current "gilaki" keyboard to have this very different layout be given to them as an update.

If this new layout is the standard way to type gilaki, would you consider calling the keyboard "gilaki_standard"? Or perhaps you have another idea for naming the keyboard.

This change will require every file that is currently named `gilaki.*` to be renamed to `gilaki_standard.*` **and in addition** all the references inside files (such as in `gilaki_standard.kmn`, for example) need to be changed from `gilaki` to `gilaki_standard`. The keyboard project folder `gilaki` also needs to be renamed to `gilaki_standard`.

You may find it easier to close this pull request and create a new one with the correctly named project folder and files (rather than delete and replace all the files in this existing pull request). Either way is fine.

Thanks again for taking the time to contribute your work!

LordArma:
> Thanks @LordArma for this keyboard submission. Unfortunately, we need to ask you to choose a different name for this keyboard project. It would not be fair to users of the current "gilaki" keyboard to have this very different layout be given to them as an update.
> 
> If this new layout is the standard way to type gilaki, would you consider calling the keyboard "gilaki_standard"? Or perhaps you have another idea for naming the keyboard.
> 
> This change will require every file that is currently named `gilaki.*` to be renamed to `gilaki_standard.*` **and in addition** all the references inside files (such as in `gilaki_standard.kmn`, for example) need to be changed from `gilaki` to `gilaki_standard`. The keyboard project folder `gilaki` also needs to be renamed to `gilaki_standard`.
> 
> You may find it easier to close this pull request and create a new one with the correctly named project folder and files (rather than delete and replace all the files in this existing pull request). Either way is fine.
> 
> Thanks again for taking the time to contribute your work!

Hello. Thank you because of your answer. Do you have any statistics about current Gilaki users? As I know as a Gilak, no one use current layout because it's just a copy of Arabic layout. If you are serious about it, I can change my layout to gilaki_standard and send another pull request.

DavidLRowe:
I don't have statistics about downloads of that keyboard, though I expect someone on the Keyman development team might be able to find that information. 

Regardless of how many users there are, the policy of the keyboards repository is to preserve layout compatibility for a given keyboard name.

So, yes, please submit a PR with a different name (such as gilaki_standard). Thanks for your understanding!

LordArma:
A new pull request has been submitted here:
https://github.com/keymanapp/keyboards/pull/1386

PR #1385, opened 2020-11-13
[bj_naskapi_common] Naskapi (common) keyboard

MakaraSok:
* Split Naskapi (Common) from Naskapi (Classical)
* Added support for any platforms
* Added OSK


LornaSIL (approved):
LGTM

PR #1386, opened 2020-11-17
Add Gilaki (Standard)

LordArma:
According to the conversation that we had here:
https://github.com/keymanapp/keyboards/pull/1383


keyman-server:

LordArma:
A new pull request is submited here:
https://github.com/keymanapp/keyboards/pull/1386

DavidLRowe:
Thanks for this submission. It's in pretty good shape, but there are a few things to change.

For the copyright, you could assign the copyright to SIL International as you have currently done, but you also are welcome to use your own name (or the name of a legally recognized organization), as long as you agree to the MIT license. The copyright appears in several places and needs to be consistent. In particular, it is missing from the gilaki_standard.kmn file and the gilaki_standard.kps file.

The BCP 47 code (in both the gilaki_standard.keyboard_info file and in the gilaki_standard.kps file) should be glk-Arab rather than glk-IR.

In the gilaki_standard.kmn file
- the copyright needs to be consistent with the README.md and LICENSE.md files (as mentioned above)
- the store(&VERSION) refers to the version of Keyman needed to compile the keyboard file. It should be 9.0 or higher. I needed to change this in order to compile the keyboard.

Similarly, in the gilaki_standard.kvks file, I needed to change `<version>1.0</version>` to `<version>10.0</version>` in order for the keyboard to compile. 

The desktop layout seems fine. There are some problems with the touch layout:
- You cannot get to the shift layer. On the default layer, for the Shift key, set "Next Layer" to "shift". (Similarly, on the shift layer, the Shift key needs to have "Next Layer" set to "default" so that the user can get back to the default layer.)
- The two keys to the right of the Shift key don't produce the correct output. This is because they are named K_Z and K_X, and are therefore linking to what the .kmn file produces for those characters. If you use U_0638 (instead of the K_Z) and U_0637 (instead of the K_X), you should get the behavior you want.
- I haven't yet tested the shift layer

We will need some detail in the welcome.htm file. This is the documentation that is packaged with the keyboard. It should be sufficient to include images of the default and shift layers of the keyboard. These can be generated from Keyman Developer, On-Screen tab, Export button. When the "Save As" dialog comes up, select "Save as type" "Image (*.png), provide a base name ("gilaki_standard" for example) and navigate to the "gilaki_standard/source" folder, then select "One file per shift state" and OK. This should create gilaki_standardU_.png and gilaki_standardU_S.png files for the default and shift layers, and you can reference these files from welcome.htm.

The source/help/gilaki_standard.php file provides online documentation similar to that in welcome.htm, but the .php file is able to generate the images directly. You already have the code to do that, but a few adjustments are needed:
- Because your desktop keyboard has no shift-ctrl layer, you should remove "shift-ctrl" from the data-states string.
- Because your keyboard has no phone layout, you should remove the "div" element containing `id='osk-phone'`

What follows is a suggestion of something you might want to consider, but it is strictly up to you whether or not you want to change it. 
- move the backspace key down to the third row, replacing the Enter key. Make the backspace twice as wide as a normal key.
- move the Enter key down to the fifth row, to the right of the space. Make the Enter key twice as wide as a normal key.
- reduce the size of the Shift key to the size of a normal key.

This would give each of the rows 13 keys (or 11 plus a double-wide key) and might make a neater appearance. 

Thanks again for your contribution. Let me know if anything I've written is not clear.

DavidLRowe:
@LordArma  Are you planning to continue work on this keyboard? If not, we can close this pull request. You can always reopen it at a later date, or submit a new one. Thanks for your interest.

PR #1387, opened 2020-11-18
[bj_naskapi_common] fix the broken link

MakaraSok:
* A link to the PDF file on welcome page was broken and this PR is to fix it.


DavidLRowe:
In the .kps file, the version number was listed as 1.3 (changed from 1.2). Rather than changing that to 1.4, I removed the explicit version number so the version number be picked up from the keyboard. (I think the version of the package would still have followed the keyboard version even with the old version number there, but having this old version in the file is confusing.)

BTW, I think a version number change from 1.3 to 1.3.1 would have been fine here, since this really is fixing a minor issue, but 1.4 works!

DavidLRowe (approved):

PR #1390, opened 2020-11-20
[sil_kmhmu] half space after the special two

MakaraSok:
This version updates to include the two special characters in the Lao consonant  store (`store(Consonant)`) so that the rules for Thin Space and Full Space are triggered and work as expected. 

The previous version would not add a half space after the two characters ໞ (on Shift + Y) and ໟ (on Shift + U) when the spacebar is pressed.

We would like to ensure as well whether the two fonts get installed when the .kmp is either installed via the Keyman Desktop configuration or via keyman.com.


LornaSIL (approved):
LGTM

PR #1391, opened 2020-11-21
GFF Gurage Legacy Keyboard v0.11

dyacob:
Reset language ID to sgw-Ethi-XL to synch with https://github.com/keymanapp/lexical-models/pull/105 and avoid the issue reported in https://github.com/keymanapp/keyman/issues/3806 .

### DIFF
--- a/release/gff/gff_gurage_legacy/README.md
+++ b/release/gff/gff_gurage_legacy/README.md
@@ -5,7 +5,7 @@ Copyright (C) 2008-2020 Ge'ez Frontier Foundation
 
 Version 0.11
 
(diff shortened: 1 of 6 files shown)


keyman-server:

LornaSIL:
I was going to accept this after I changed swg to sgw in 3 files. But, now I realize you didn't update the version in the .kmn file. Could you do that? Then I think I can accept it.

dyacob:
@LornaSIL , thanks for catching the swg mixup and missing version bump.  The version should be set right now.

LornaSIL (approved):
LGTM

PR #1397, opened 2020-11-26
chore: cleanup old isis php files

mcdurdin:
isisinfo.php is not present, so this generates warnings on the site. Not needed.


LornaSIL (approved):
LGTM

PR #1401, opened 2020-12-01
GFF Amharic v2.0

dyacob:
The package in this pull request provides two keyman keyboards (.kmn) and two touch layouts (.keyman-touch-layout).  The **_wish_** with the package is to provide a single KMN file (the newer one) and two touch layouts (the old and the new).  However, since the KMN file references a touch layout file internally, there does not appear to be a way to associate more than one layout file with a single KMN file.  A thought for the future would be to allow the touch layout file to reference the .kmn file which would avoid the present complication.

The older KMN file sets only phone and tablet targets, yet the package installer will still install the older keyboard under Windows.  Is this a bug?  Is there a way to configure a package to only install one of the two keyboards on desktop systems?  I would like to address this issue before the final merge.  Alternatives would be to remove the older touch layout entirely or create a new package for it with a name that indicates it is deprecated.

Noteworthy also is that when the two keyboards are installed, they get installed with the language associations 
"English (United States)" (new) and  "Geez (geez-Ethi-ET)" (old).  I believe this is because the older keyboard is still using the deprecated language identifiers and the new keyboard does not.

**Notes on the New Keyboard:**

* Uses Version 10 of the Keyman keyboard language.
* Functionally equivalent to the older keyboard.
* The touch specific statements are revised to align with the version 10 language and latest capabilities.
* The primary difference is in the touch layout, which uses a new arrangement similar to work done recently for Gurage and Tigrinya-ET touch layouts.


keyman-server:

LornaSIL:
I'm not completely sure I understand what you are trying to accomplish...but you cannot include two .kmn and two keyman-touch-layout files in one keyboard. The individual keyboards are no longer called a package, it is when you include two keyboards that it needs to go in the packages folder.

You CAN create another `gff_amharic_new` (or whatever you want to call it) KEYBOARD with the new keyboard: `keyboards\release\gff\gff_amharic_new`

THEN, you would need to create a PACKAGE: `keyboards\release\packages\gff_amharic_multiple` (or whatever) and then include the two keyboards (`gff_amharic` and `gff_amharic_new`) in that **package**. 

I think the steps would be to first get the new keyboard submitted and approved and then to submit the package after the new one is merged.

dyacob:
Thanks @LornaSIL , the desire was to have just one .kmn file, but two touch layout options for it.  I don't know if there is much value in preserving the older touch layout though.  I will update the package to just have the new one only.  If there is feedback from people who still want the old touch layout, then I'll make an entirely separate package for it.

dyacob:
@LornaSIL , I've now committed the updates so that there is only a single .kmn and layout in this package.   thanks!

LornaSIL:
It looks like you've removed the tablet layout. Is that correct?

If so, can you change the php help file?
Delete:
```
  <div id='osk-tablet' data-states='default symbol'></div>
</div>
```
And change the phone section to:
```
  <div id='osk-phone' data-states='default non-amharic punctuation punctuation-2'></div>
</div>
```
I haven't tested the new mobile layout yet as I can't seem to get my phone in sync with the laptop.

dyacob:
Hmm... I delete the tablet layout indirectly.  The new Amharic layout is copied from the Tigrinya-ET one which did not have a tablet layout.  I think it would be best to have the tablet layout, so I'll return it with corresponding updates.

dyacob:
The last update has a revised tablet layout that is ready for testing.

LornaSIL:
I didn't check every keystroke on the mobile layouts, hopefully you got them all right. I do see that on the `punctuation` layer (both phone and tablet) you have a longpress for the digits. The longpress for the `K_1` (and all of those digits) need the "modifier" layer changed from "(current layer)" to "default".

Also, the php file needs to be changed for both the phone and tablet to:
```
  <div id='osk-tablet' data-states='default non-amharic punctuation punctuation-2'></div>
</div>

  <div id='osk-phone' data-states='default non-amharic punctuation punctuation-2'></div>
</div>
```
The php file uses the names of each layer the way you named them.

LornaSIL:
And the "punctuation" layer that needs to change the `K_1` (all digits) to use the default modifier layer?

dyacob:
On the "punctuation" layer (only) for both phone and tablet templates, the modifier has been changed from "(current layer)" to "default" for the K_1 - K_0 longpress keys.  I tested on my Android phone it seemed to work fine, I don't have a tablet to test on (but use the Chrome "Developer Tools").

LornaSIL (approved):
LGTM

PR #1407, opened 2020-12-08
[bj_cree_east_james_bay]  CreeBoard

MakaraSok:
> 1.0 (2020-12-08)
> ----------------
> *  Change Alt to Ralt and Shift Alt to Shift Ralt
> * Revision: 02 May 2001, (BJ) redone for Unicode by Bill Jancewicz

The keyboard version is not mentioned. It's marked as 1.0 for now.
The original version does not have touch layout, should we add one?


LornaSIL:
@makara When I keyboard already existed...but it did not have a version number, I believe we should consider that one version 1.0. That means I would give the new one a version 1.0.1 or 1.1.

At this point, if the keyboard did not have a touch layout, let's NOT add a touch layout. When we try to add behavior it requires getting the community involved in what they want it to look like. I'd like to get the existing keyboards into the repo and then when they are wanting a touch layout we can add it more easily.

LornaSIL (approved):
LGTM.

I removed the .js file from the package. That should only be included when we provide a mobile layout.

PR #1409, opened 2020-12-09
[bj_cree_west_latn] SRO (west)

MakaraSok:
1.1 (2020-12-09)
----------------
*  migrate to public repo
* add touch layout
* add support to all platforms

Touch layouts were added as they are like those of the English and it is presumed that they would be fine to be included.


LornaSIL:
If the "phone" and "tablet" layouts are the same, you can delete the "tablet" layout. That reduces keeping two touch layouts in sync.

The documentation doesn't explain how to get the long macrons. Can you add to both `welcome.htm` AND `bj_cree_west_latn.php` the following:

```
<p>Macrons over long vowels are available by typing two vowels in a row. Thus, typing "aa" would become "ā". The long vowels are aeio and AEIO.</p>
```
Otherwise, it looks fine.

LornaSIL (approved):
LGTM

PR #1410, opened 2020-12-09
[bj_cree_east_latn] Hâts (east) keyboard

MakaraSok:
1.1 (2020-12-09)
----------------
*  migrate to public repo
* add touch layout
* add support to all platforms


LornaSIL:
Similar comments to the "west" keyboard:

If the "phone" and "tablet" layouts are the same, you can delete the "tablet" layout. That reduces keeping two touch layouts in sync.

The documentation doesn't explain how to get the hats or tilde. Can you add to both `welcome.htm` AND `bj_cree_east_latn.php` something like the following:

```
<p>Hats over long vowels (and tilde over "n") are available by typing two letters in a row. Thus, typing "aa" would become "â", and typing "nn" would become "ñ". These are available for aiun and AIUN.</p>
```

Otherwise, it looks fine.

LornaSIL (approved):
LGTM

PR #1411, opened 2020-12-10
[bj_cree_woods] Western Cree (TH-Woods) keyboard

MakaraSok:
* migrate to the public repo
--------------------------

TODO

* fix error

> Two keys (i.e. Î and ÿ) of the keyboards above in the store do not compile. What keys should be used in place of these?
> 
> > c initial consonant keys (The Î is U+1525 which is the sh consonant)
> > c (The ÿ is U+15AE which is the th consonant that looks like y)
> > 
> > store(ConsKey) "ptkcmnsÎyfÿ"
> > store(CapConsKey) "PTKCMNSÎYFÿ"

* update welcome/help/readme and others


LornaSIL:
Old versions of keyboards used to compile even when there were impossible keystrokes. I think that's what you are running into.

If you just remove those characters from the stores so now it would be:

```
store(ConsKey) "ptkcmnsyf"
store(CapConsKey) "PTKCMNSYF"
```
You might also consider removing one of the
```
group(Main) using keys
```
statements. There are two in the keyboard. That might be confusing.

MakaraSok:
The compile does get through when the two keys are removed, but it introduces an issue where pressing "ye" generates ᔐ (U+1510) and "fe" ᔦ (U+1526) which I believe are not correct, and the same for the rest in the stores of "consonant followed by vowel'.

I've asked Bill to test and verify whether the keyboard works as expected after removing the two keys.

LornaSIL:
It looks like a change to aksarabali_panlex.kpj crept in. Could you removed that from the PR?

LornaSIL (approved):
LGTM

PR #1412, opened 2020-12-10
[bj_mista_wasaha_cree] Swampy Cree (Tataskweyak) keyboard

MakaraSok:
* migrate to the public repo
--------------------------

TODO

* fix error

> Two keys (i.e. Î and ÿ) of the keyboards above in the store do not compile. What keys should be used in place of these?
> 
> > c initial consonant keys (The Î is U+1525 which is the sh consonant)
> > c (The ÿ is U+15AE which is the th consonant that looks like y)
> > 
> > store(ConsKey) "ptkcmnsÎyfÿ"
> > store(CapConsKey) "PTKCMNSÎYFÿ"

* update welcome/help/readme and others


LornaSIL (approved):
LGTM

PR #1380, opened 2020-11-05
[Ó_tissi] initial submission

Connman90:
This is named Ó tissi because that is the name in the language and the name the community is referring to it as moving forward. However, anthropologists and linguists may know it as Achumawi (acv).

PR #1421, opened 2020-12-16
[o_tissi] Submit O tissi keyboard on behalf of author

LornaSIL:
The PR is based on PR #1380. I'm resubmitting it as the previous Pull Request had files in the wrong places.

I changed ALT to RALT in the keyboard (with permission of author).

I added a help php file.


DavidLRowe:
readme.htm mentions "alt" key. Should that now be "right-alt"?

Minor: welcome.htm describes U+02BC as "right-slanted" apostrophe, but that description might not be true for all fonts. (Not sure if I have a better suggestion other than just "apostrophe", which, of course, could be confused with U+0027.) So maybe just leave it as is.

welcome.htm says U+02BC is on "v" key, but the OSK (as well as the default layer image used by welcome.htm) has comma (U+002C) on the "v" key, rather than U+02BC, and the keyboard produces U+002C rather than U+02BC.

Minor: The combining acute (U+0301) on the comma key doesn't seem to be mentioned in the documentation, though it is indicated on the OSK. 

Touch layout:
default layer, second row, fifth key: key cap shows doubled U+02D1
shift layer, second row, fifth key: key cap shows doubled U+02D1
numeric layer, third row, ninth key: key cap shows "/" but key produces U+02B0

DavidLRowe:
o_tissi.php and welcome.htm have "The acute accent can be with the comma key...", but I think you need something ("made", "typed", "constructed", ??) after the "can be".

.kmn file has two entries for `+ [K_V]`. I think the second one (under "symbols") should be removed.

BTW, when Zoom is active, it grabs the RALT+SHIFT+T combination for some function (screen capture perhaps?), so I had to quit Zoom to test this keyboard.

DavidLRowe (approved):

PR #1423, opened 2020-12-22
GFF Amharic Classic Keyboard

dyacob:
This is a version of the GFF Amharic Keyboard v2.0 but using the QWERTY-centric touch layout.  The underlying .kmn file is unchanged with the exception of the use of the "classic" identifier string where required.  The keyboard is only intended for use on mobile and tablet platforms, all desktop platforms have been disabled.


keyman-server:

LornaSIL:
@dyacob 
- I also think you should:
   - remove the gff_amharic_classic.kvks file itself
   - delete the reference to the .kvks file in the .kmn file
   - remove the reference to the .kvk file from the .kps file
   - remove the reference to the .kmx file from the .kps list of files.
In the .keyboard_info file I don't think I would deprecate the keyboards that `gff_amharic` deprecates. But, I do think you should link to `gff_amharic`:
```
  "related": {
    "gff_amharic": {
      "deprecates": false
    }
  }
}
```
This will allow people to find the Desktop and Web layout keyboard.

Also, I'd like to see the help file changed to have this removed:
```
<div id='osk-container'>
  <div id='osk' data-states='default shift'></div>
</div>
```

I think that's it. I didn't test the keyboard since this is using the one we had previously.

dyacob:
Hi @LornaSIL , the last commit should address each time raised.  thanks!

LornaSIL (approved):
LGTM

LornaSIL:
The keyboard is online now.

I should have checked your php file better. Can you submit a new PR with these changes?

Phone should be:
```
  <div id='osk-phone' data-states='default symbol'></div>
</div>
```
Tablet should be:
```
  <div id='osk-tablet' data-states='default symbol'></div>
</div>
```

Also, in that PR, can you delete `gff_amharic.keyman-touch-layout` from the source folder? You probably copied that from the other keyboard.

None of this means a version change.

LornaSIL:
Ok, it's got the proper screenshots now!

PR #1427, opened 2020-12-31
[taigi_poj] Add support for e͘

guyemerson:
The e͘ character is used in some Taiwanese dialects/accents, referred to as Lāi-po͘-khiuⁿ or Chiang-chiu-khiuⁿ, for example in Gî-lân (Yilan, 宜蘭).

### DIFF
--- a/experimental/t/taigi_poj/HISTORY.md
+++ b/experimental/t/taigi_poj/HISTORY.md
@@ -1,6 +1,10 @@
 PhahTaigi POJ Change History
 ====================
 
+1.2 (2020-12-31)
+----------------
+* Added support for e͘
+
 1.1 (2020-02-29)
 ----------------
 * Added missing images
--- a/experimental/t/taigi_poj/README.md
+++ b/experimental/t/taigi_poj/README.md
@@ -3,7 +3,7 @@ PhahTaigi POJ
 
 © 2020 Ngô͘ Ka-bêng
 
-Version 1.1
+Version 1.2
 
 Description
 -----------
--- a/experimental/t/taigi_poj/source/taigi_poj.kmn
+++ b/experimental/t/taigi_poj/source/taigi_poj.kmn
@@ -227,6 +227,8 @@ c
 + [CAPS K_W] > U+0057
 + [NCAPS SHIFT K_W] > U+0057
 + [CAPS SHIFT K_W] > U+0077
++ [ALT K_W] > "e͘"
++ [ALT SHIFT K_W] > "E͘"
 
 + [NCAPS K_X] > U+0078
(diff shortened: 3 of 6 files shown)


keyman-server:

guyemerson:
Before submission, I verified the build following the instructions [here](https://help.keyman.com/developer/keyboards/advanced/step-5), and everything compiled successfully.

LornaSIL:
Guy, it won't automatically build for you, but that doesn't mean there is a
problem. We won't be able to check the revision until next week, as the
keyboard checking people are currently on break. We'll take a look next
week.

guyemerson:
Great, thank you!

LornaSIL:
@guyemerson does this update support another language than Min-Nin Chinese [nan]? If so, should we add that language to keyboard_info and to .kps?

guyemerson:
This update adds support for a subvariety of Min Nan.  It would have the same ISO 639-3 code [nan].  (Glottolog has a relevant [more specific code](https://glottolog.org/resource/languoid/id/fuji1236), for anyone interested.)

I suppose additional country codes could be added?  (e.g. for Malaysia, Singapore, Indonesia, the Philippines, Mainland China...)  On the other hand, I'm not sure whether this would be useful, or where to draw the line -- as far as I'm aware, Taiwan is the only country where Min Nan has official status.  The decision to list additional countries would be independent of this update, though, as the use of the <e͘> vowel cross-cuts country borders.

LornaSIL (approved):
LGTM. Ok, we can accept the changes. I'd rather not add country codes.

guyemerson:
How often are keyboards rebuilt?  I'm not in a hurry, but it would be nice to know when this update will be available in the app.  I've searched the documentation but only managed to find a comment [here](https://help.keyman.com/developer/keyboards/) that: "The catalog of Keyman keyboards is automatically rebuilt when updates to the keyboard repo are made, and your keyboard users will receive notifications in their Keyman apps when newer keyboard versions become available."

LornaSIL:
Oh, sorry about this. I didn't realize there was a problem. Unfortunately the most important version number (in the kmn) didn't get bumped when this PR was processed. Let me discuss with someone what we need to do.

DavidLRowe:
The version number in the .kmn file was just bumped by #1434 . The new version 1.2 should show up on https://keyman.com/keyboards/taigi_poj "soon" (though I'm not sure how soon, but hopefully less than 24 hours). Sorry for the delay.

LornaSIL:
@guyemerson v1.2 is now online: https://keyman.com/keyboards/taigi_poj

guyemerson:
Thank you both!

PR #1428, opened 2021-01-03
Add Arno's Engram layout

binarybottle:

PR #1440, opened 2021-01-15
Revise Engram layout

binarybottle:
This PR is in response to https://github.com/keymanapp/keyboards/pull/1428

@LornaSIL -- I have updated the README, .htm pages, and .kmn and .kvks files by hand (because I no longer have access to Windows).  I believe I made all of the fixes you recommended.  

1. I made some changes to the layout and switched the Ctrl layer to Ctrl+Alt.  I assume this means that I would have to press three keys to type any character in that layer?  If so, I would prefer that this layer be accessible by a single key.  I read in Keyman's documentation (https://help.keyman.com/developer/language/guide/virtual-keys) that "Right alt key or AltGr on European keyboards (not found on all keyboards)", so I was concerned about using R-Alt.  Does this mean that if I were to use R-Alt, it won't work on all keyboards?  Is there any reason why I can't/shouldn't use Alt instead of Ctrl+Alt?
2. When I tested out the previous onscreen keyboard on macos, it seems that only the Shift key changes the key identities -- Option, Command, Ctrl, Caps, and any combination of these modifier keys I've tried completely ignores the Engram layout. Do I need to add key mappings for each of these keys in my .kmn and .kvks files?
3. Note: I will need to update the three images in the source directory.


keyman-server:

LornaSIL:
I don't see that you've committed any files. You just brought your own repo into line with the Keyman repo. I *think* before you try to make more modifications you need to get Keyman Developer running again. It would be very hard to modify the On-Screen Keyboard in a text editor and get it accurate.

binarybottle:
Sorry!  I am still getting used to submitting PRs from forks.  I have submitted a new PR with the changes: https://github.com/keymanapp/keyboards/pull/1441

PR #1442, opened 2021-01-15
[shahmukhi_phonetic] New keyboard for Punjabi with rlam

LornaSIL:


DavidLRowe:
Default script for "pa" is Gurmukhi. Does that mean the tag for this keyboard should be "pa-Arab"?

very minor point: Shift+102nd key (left of "z") produces "|", however this key is blank on OSK and image used in welcome.htm. But the default layer for that key also produces "|" so if you don't want to change anything that's fine.

LornaSIL:
Thanks. That was the tag I had from the MSKLC keyboard I used. That was the wrong tag :(
I won't make the change to the OSK at this point.

DavidLRowe (approved):

PR #1444, opened 2021-01-21
[sil_kvl_kayaw] Add touch layout

sdysart:
This adds a touch layout to the keyboard.


keyman-server:

DavidLRowe:
It seems that two unrelated files have slipped into this pull request:
release/sil/sil_myanmar_my3/sil_myanmar_my3.kpj
release/sil/sil_myanmar_mywinext/sil_myanmar_mywinext.kpj
I think they should be removed from the pull request.

Several files have outdated copyright: need "2018" changed to "2018-2021":
- sil_kvl_kayaw.php
- LICENSE.md
- README.md
- sil_kvl_kayaw.kmn [although, optionally, for this file as well as for welcome.htm, you could drop the whole "License" section -- with the copyright statement and MIT license information -- since that appears in the LICENSE.md file]

README.md also needs:
- the version number changed to 1.2
- Under supported platforms, add "Linux" and "Web"

sil_kvl_kayaw.kmn also needs:
- change to: store(&TARGETS) 'any'

"v" on the keyboard now produces "v" rather than  U+030C Combining caron. 
If this is correct, then the OnScreen Keyboard and documentation files should be updated to match. Otherwise the .kmn file needs to be corrected.

The sil_kvl_kayaw.php help file indicate both phone and tablet layouts exist, but the sil_kvl_kayaw.keyman-touch-layout file only contains a tablet layout.

On the touch layout, shift layer, several longpress keys on "." produce the wrong character:
- `,` actually produces `<`
- `'` actually produces `"`
- `\` actually produces `|`
- `;` actually produces `:`

DavidLRowe:
I guess @darcywong00 made some corrections while I was writing my review: that's a fast response!

DavidLRowe:
The PR no longer has the two unrelated .kpj file, but unfortunately, it is now deleting the sil_myanmar_my3.kpj file, which causes that keyboard to fail to compile. I expect the same would be true for the sil_myanmar_mywinext.kpj file. So, either you need to add those files back in, or (perhaps easier) close this PR and open another with just the sil_kvl_kayaw files.

The other changes look good, though I haven't checked them exhaustively.

darcywong00:
> So, either you need to add those files back in

I've put those 2 .kpj files back (whitespace diffs remain)

DavidLRowe:
I guess it's OK to leave the whitespace-only changes to the two sil_myanmar*.kpj files.

I think you've dealt with everything except the change in the .php file. Since the touch layout only has "tablet" (and not "phone"), the .php file needs to reference `id='osk-tablet'` (rather than `id='osk-phone'`).

DavidLRowe:
Hmm. It seems to work anyway, that is, the tablet layout is displayed, even though phone is requested. So, let's go ahead and accept this as is.

DavidLRowe (approved):

PR #1446, opened 2021-01-22
[sil_bru] New keyboard based on an MSKLC keyboard

LornaSIL:


DavidLRowe:
Everything looks fine. The only question I have is the sentence in welcome.htm (and sil_bru.php) that says: "The keyboard produces NFC Unicode." If I type `a` and `5` I get U+0061 U+0300, rather than the NFC U+00E0. 

Is that a problem? Should that statement be removed? Presumably the MSKLC original did the same thing.

DavidLRowe (approved):

PR #1448, opened 2021-01-26
[newa_traditional] fix '7' key in phone numeric layout. remove some repeated characters.

sapradhan:
phone touch layout generated a quote '  instead of 7, this is fixed
Also removed a few characters had multiple bindings


keyman-server:

DavidLRowe:
The touch keyboard changes look fine. Unfortunately, the copyright statement will need to be updated (from "2020" to "2020-2021") in several files:
- README.md
- LICENSE.md
- newa_traditional.kmn
- newa_traditional.kps

Additionally, the copyright statement should either be updated in the same way, or else be removed entirely, for:
- readme.htm
- welcome.htm

Thanks!

sapradhan:
Thanks will do. 
One question regarding copy rights. I noticed I had put (cc) creative commons in those files, but in another submission(nepali_traditional) you specifically asked to put a (c) instead. I would prefer it to stay (cc) if possible, would that be okay ?

DavidLRowe:
I'm not an expert in this area, but my understanding is:
- All the keyboards in the repository (with the exception of some old ones in "legacy") are released under the MIT license.
- The copyright statement is the way that you establish your right to release the material under the MIT license.

I'm not that familiar with CC licenses, but I think the end result it similar. I expect the Keyman team chose to go with the MIT license to avoid having to deal with multiple licenses and differing restrictions.

sapradhan:
dates extended as advised.

DavidLRowe (approved):

PR #1450, opened 2021-01-31
[nepali_traditional] added a frequently used character combination

sapradhan:


keyman-server:

DavidLRowe:
These changes look good. The only thing to ask is that you make the same changes in nepali_traditional.php as you made in welcome.htm. (The welcome.htm is the help that is packaged with the keyboard and available locally. The .php file is the help that is available on the Keyman.com site.) Thanks!

DavidLRowe (approved):

PR #1452, opened 2021-02-05
chore: Update BCP47 codes for release .keyboard_info files

mcdurdin:
Part 1 of a multi-part maintenance job. This updates the .keyboard_info files for release/ but does not touch .kps or other folders.

This brings the canonicalization of BCP 47 codes in line with langtags.json. I have done a lot of verification of the changes.

1. Compare old and new and verify that the canonicalized codes match (should be able to do 1-1 match for 99% of cases):
    * amazigh_lat, amazigh_lat.kmn: these should have special language tags to indicate transcription rather than language? For now, using -t-xxx-Latn but could be better. Note that the amazigh_lat keyboard had one which should have been marked -Latn but wasn't, the older one has many issues.
    * azeri_cyrillic: azb is default -Arab so needed -Cyrl
    * azeri_latin: az is -Latn, but azb has no -Latn variant
    * cayuga_u: lre has no default script (-Zyyy)
    * Various Greek keyboards: grc has a default script of -Cprt (Cypriot Syllabary) so have added -Grek
    * lingfilsemitica: I added -fonipa
    * Portuguese keyboards: pt is for Brazil by default, pt-PT for Portugal
    * sil_euro_latin: removed -Hani tags; why is zyg-Nshu present?
    https://gist.github.com/mcdurdin/ec5cbd1f43dcffaab3a55c2e4b3a0fc0/revisions?diff=split

2. For each .keyboard_info, check that the script is consistent for each lang tag (could also do this for .kps but have not done so yet):
    https://gist.github.com/mcdurdin/1a7c488399a6b4594031e82179a6bd57

Note that this PR will build once #1453, #1454, #1455, #1456, and #1457 are merged into it. The PRs are only split up to make it easier to review. See #1457 to see the passing build.

These files will not build with the 13.0 release of kmcomp.exe; they do depend on 14.0.239-beta or later.


release/b/balochi_latin/balochi_latin.keyboard_info
```diff
@@ -1,7 +1,7 @@
 {
     "license": "mit",
     "languages": [
-        "bcc-Latn"
+        "bal-Latn"
```
darcywong00, line 4: Note for @rmlockwood 

Changing from `bcc` to `bal` will impact the lexical-models:
* sil.bcc-arab.upp_ptwl1
* sil.bcc-latn.upp_ptwl1

Does anyone know if these keyboards distinguish between 

Identifier| Reference Name | 
-- | -- |
bcc | Southern Balochi |
bgn | Western Balochi |
bgp | Eastern Balochi

mcdurdin: Yes, we should update those models to also include `bal` (don't remove `bcc` at this point to maintain compatibility with 13.0 keyboards)

LornaSIL:
> cayuga_u: lre has no default script (-Zyyy)

Cayuga is actually `cay` and not `lre` - so does the package need changing or just the `keyboard_info`

LornaSIL:
I am updating the source for langtags.json to add azb-Latn. I'm not sure when the .json file will be rebuilt.

mcdurdin:
> I'm not sure when the .json file will be rebuilt.

We can update langtags.json in the Keyman repo if we need to; we don't pull that automatically during builds for this kind of reason.

mcdurdin:
After significant [discussion](https://docs.google.com/document/d/1N5CkTX7tlLlmUB9Mbd-LblY1ch1e642XxlTasC_WE_Q/edit#), this PR will be replaced with targeted changes to specific keyboards where BCP 47 codes should be fixed. Given the compat differences between Keyman 13 and 14, canonicalization will be treated as warnings/hints rather than errors, and will not block the keyboard build.

PR #1453, opened 2021-02-05
chore: Update BCP47 codes for release .kps files

mcdurdin:
See #1452 for discussion.

Part 2 of a multi-part maintenance job. This updates the .kps files for release/.


LornaSIL:
I would like to understand better the rationale for changing from an indidual language code like bcc-Arab to bal or `spy` to `'kln`. I do know that the Sabaot [spy] orthography is very different from the standard Kalenjin orthography and I hate to lose that distinction. I suspect that would be the case for other languages.

Besides this big issue in my mind the other changes seem fine.

mcdurdin:
> I would like to understand better the rationale for changing from an indidual language code like bcc-Arab to bal or `spy` to `'kln`. I do know that the Sabaot [spy] orthography is very different from the standard Kalenjin orthography and I hate to lose that distinction. I suspect that would be the case for other languages.

To be honest, this is my concern also. I am just using the canonicalization data from langtags.json -- so the rationale on which language tag to use originates there. In langtags.json, the tags are considered equivalent, with in some cases the macrolanguage tag being considered canonical, and in other cases an individual language tag is considered canonical.

I *could* do another round of changes that splits out the macrolanguage changes -- because I think the rest of it is pretty straightforward. But I would like to get to the bottom of the macrolanguage equivalency in langtags.json and whether or not we need to be cleverer in our canonicalisation (we've still got time before 14.0 stable -- and I will delay stable if I need to in order to get this right).

Per https://github.com/silnrsi/langtags/blob/master/doc/tagging.md#macro-languages:

> Some language tags are best not used and alternative used instead. This is indicated by specifying the basic tag to use instead, in the macro field. This replacement may occur for different reasons, but the default is because this language is the default language for a macro language group and the macro language should be used instead.

But I don't know if that really helps. @mhosken may have some wisdom.

PR #1454, opened 2021-02-05
chore: Update BCP47 codes for legacy .keyboard_info files

mcdurdin:
See #1452 for discussion.

Part 3 of a multi-part maintenance job. This updates the .keyboard_info files for legacy/.


LornaSIL:
I have the same question about using macro languages that I had for a previous PR.

PR #1459, opened 2021-02-05
chore: build script tweaks

mcdurdin:
Allows the build script to be restarted with `-start shortname/prefix` for example: `./build.sh -start sil/sil_eur` will start building all keyboards, alphabetically, beginning with `sil_euro_latin`. This helps when a keyboard fails to build, so you don't have to rebuild all the earlier keyboards again!

Adds some colour to make each keyboard built more visible.


darcywong00 (dismissed):
lgtm

Can we jot a quick note in https://github.com/keymanapp/keyboards#build-instructions ?

darcywong00 (approved):
lgtm

PR #1463, opened 2021-02-13
Fix version number to 1.2 and add support for the CapsLock key in engram.knm

binarybottle:


binarybottle:
@DavidLRowe -- I hope that I properly added support for the CapsLock key, according to instructions in https://help.keyman.com/developer/8.0/docs/reference__struct_vkeys

keyman-server:

binarybottle:
@DavidLRowe -- Do I need to revise all files to say version 1.3 since this is a new PR, or can they all remain as 1.2?

DavidLRowe:
The changes for version number look good, HOWEVER , the changes to the .kmn file for CapsLock are NOT what you want! If I type the "a" key, I get "i" as expected. Pressing "CapsLock" followed by "a" gives me the literal text "BIG A". 

SO, option 1 is to remove those changes from the .kmn file (leaving only the change to the version number) and call that version 1.2. There are plenty of keyboards around where the CapsLock key doesn't function. (You could make the CapsLock changes at a later date and call that version 1.3.)

Option 2:  If you want to make other changes and call them part of version 1.2, I think that will work.

Here's an example from the basic_kbdfr (an emulation of the French keyboard, where the A and Q keys are switched from the QWERTY layout):
```
+ [NCAPS K_A] > U+0071
+ [CAPS K_A] > U+0051
+ [NCAPS SHIFT K_A] > U+0051
+ [CAPS SHIFT K_A] > U+0071
```
These four lines are saying:
- CapsLock off, "a" key produces "q"
- CapsLock on, "a" key produces "Q"
- CapsLock off, Shift+"a" key produces "Q"
- CapsLock on, Shift+"a" key produces "q"

Keyman treats "A" as equivalent of [SHIFT K_A] and "a" as equivalent of [K_A] when on the left side of the rule. So, leaving the CapsLock out of the rules, the first and third rules above could be written:
```
"a" > "q"
"A" > "Q"

If you are wanting this reviewed today, I'd suggest option 1 and hopefully I'll be able to respond today. CapsLock changes would need more review and would need to wait until next week.

DavidLRowe:
More on CapsLock:

help.keyman.com/developer/language/guide/virtual-keys
scroll down to "Caps Lock" heading
"If neither CAPS or NCAPS is specified, then the CAPS LOCK key is ignored. This means that if you do have a rule that uses CAPS, you should make sure that no other rule references that key without NCAPS or CAPS specified."

See also: https://help.keyman.com/developer/language/reference/caps

binarybottle:
@DavidLRowe -- I made the changes you recommended and stuck to v1.2 for now.  Thank you again for your help!

DavidLRowe (approved):
I hope this works. I see now that you had already bumped the package version to 1.2, so maybe that published the keyboard package already. If that's the case, then the changes you made to the welcome.htm with the latest commit won't be in the 1.2 package.

binarybottle:
Thank you, @DavidLRowe!  I'm typing with the new v1.2 layout now!

DavidLRowe:
I'm afraid that I did something wrong. On the https://keyman.com/keyboards/engram page, you can see that version 1.2 is being offered for download combined with the Windows installer, however the keyboard details at the bottom of that page show version 1.1 with a date of January 21. (Hopefully when you submit version 1.3, we'll catch all the needed changes ;) !)

binarybottle:
Yeah, I saw that as well, but all links seem to work and lead to version 1.2.  I will set up for caps lock for v1.3 next!

PR #1466, opened 2021-02-19
[sil_hebrew] Add touch layout

LornaSIL:
Update with a table layout.


DavidLRowe:
The only thing I noticed was on the touch layout, default layer
second row down, second key from left: final longpress display doesn't seem to match what is produced. I think the problem is in the .kmn file:
```
+ [T_05D5_05D5_05B9]  > U+05D5 U+05B5 U+05B9
```
where the second character on the right side is U+05B5 rather than U+05D5

Other than that, it seems fine.

DavidLRowe (approved):

PR #1468, opened 2021-02-23
[o_tissi] keyboard changes submitted on behalf of author

LornaSIL:


DavidLRowe:
Minor observations on touch layout: 

On default layer: `.` and `?` and `!` (longpress on `.` key) switch the keyboard to shift layer, but the `?` (longpress on `.`) does not switch the keyboard to shift layer. Options: leave it as is; add the layer switch to the `?` longpress key; or remove the `?` longpress key.

`s` on default layer has longpress that produces `s̓` (U+0073 U+0313), but `S` on shift layer has no longpress. (This is in contrast to other keys, such as `w` and `W` for example, where the shift layer has a longpress corresponding to the default layer.) Is this by design? or should `S` have a longpress?

These are minor questions. I'll go ahead and approve this PR. If you want to make any changes, dismiss the review. If you don't want to make any changes, then go ahead and merge the PR.

DavidLRowe (dismissed):

DavidLRowe:
Oops. I see that the .php file still mentions "right-alt" in the instructions. It seems that the content should be updated to match welcome.htm.

DavidLRowe (approved):

PR #1472, opened 2021-02-26
[sil_madi] Latin based keyboard for Madi language

MakaraSok:
The keyboard output NFD characters.

TODO:
- [x] pending for feedback from the user community


LornaSIL:
This mostly looks good. 
* There are multiple places where `desgined` is misspelled. It should be `designed`. That's in README.md, readme.htm, keyboard_info, welcome.htm, sil_madi.php, etc.
* The OSK shows `x` but not `X`. I'm not sure what is best since `x` is a deadkey. But, maybe it should be removed from the OSK screenshot? It is documented in welcome.htm that it is a deadkey.
* Some of the copyright dates are in a different order than we like. This `2021 © SIL International` should have the copyright sign at the beginning of the string. We can actually remove the year if you want in all places except the LICENSE.md.
* Since you began work on this keyboard a new version of Charis SIL has been released and because the filenames are different we created a new folder called `charis`. Could you remove from the .kps the Charis SIL files that you have and then add them all back in from a different folder? `release\shared\fonts\sil\charis` instead of `release\shared\fonts\sil\charis_sil`. (I'm not even sure how it's building on the system since we removed `charis_sil`.)

LornaSIL (approved):
LGTM

PR #1474, opened 2021-03-01
[sil_hebrew] add .js file to package so it will work on mobile

LornaSIL:
I forgot to add the .js to package. :(


DavidLRowe (approved):
And I forgot to check that the .js file was added!

PR #1478, opened 2021-03-04
[yiddish_pasekh] Update to keyboard

### LINKED ISSUE #1364: [yiddish_pasekh] User request for an update
Hi, the Keyman Pasekh keyboard is also missing shortcuts for creating the khirek yud with other vowel combinations. For example, it is sometimes necessary to type עיִ or אַיִ, but that doesn’t seem possible with this keyboard. This would be an excellent option to add. Thanks!

https://community.software.sil.org/t/how-to-enter-a-yud-khirek-yud-in-the-pasekh-yiddish-keyboard/2024/5

LornaSIL:
Closes #1364


DavidLRowe:
Minor question on the documentation:
In the welcome.htm file (and presumably the .php help file)
In the first table under "Quickstart"
The table says "c" produces "ץ" (U+05E5), but "c" produces U+05E6 and it is "C" that produces U+05E5.
The table says "x" produces "ך" (U+05DA), but "x" produces U+05DB and it is "X" that produces U+05DA.

And indeed the following table specifies those two keys with SHIFT.

LornaSIL:
So, the problem with this keyboard is you have to understand how it works and it's not documented well. "c" by itself does produce U+05E6, but when you type "c" after something else it produces U+05E5. There are tons of cases like that, so I'm afraid to touch the documentation much. Apparently it's a popular keyboard so people have figured it out.

DavidLRowe (approved):
I wondered if it might be something like that.

PR #1480, opened 2021-03-10
[sil_eastern_congo] Updates per community request

LornaSIL:
Wait for feedback on change in special key.


DavidLRowe:
welcome.htm has two entries for ;'' (U+003B U+0027 U+0027), the first for producing U+02BC and the second for U+A78C. It seems the first of these should be ;? which works to produce U+02BC.

Similar change needed for .php help file.

Everything else looks fine.

DavidLRowe (approved):

PR #1481, opened 2021-03-10
[rawang] Add tablet layout

### LINKED ISSUE #1338: [rawang] User request for touch-optimized layout
A user is using rawang version 1.5 on Desktop. He's also currently able to install the "web" version of the keyboard in Keyman for Android 13.0.

But since a touch-optimized layout doesn't exist, he won't be able to search for the mobile keyboard when Keyman 14 gets released.
e.g. https://keyman-staging.com/keyboards?q=rawang&embed=android

Would it be possible for @sdysart or @MakaraSok to work with the user to add a touch layout?

LornaSIL:
Closes #1338 
This is just a tablet layout. If a phone layout is required, we'll need to open a new issue/PR.

### DIFF
--- a/release/r/rawang/HISTORY.md
+++ b/release/r/rawang/HISTORY.md
@@ -1,6 +1,10 @@
 Rawang Change History
 ====================
 
+1.6 (2021-03-10)
+----------------
+* Added tablet layout
+
 1.5 (2020-04-17)
 ----------------
 * Moved to release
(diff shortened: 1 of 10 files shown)


DavidLRowe:
I think the Shift key on the default layer needs to make the next layer Shift.

DavidLRowe (approved):

PR #1486, opened 2021-03-16
chore: use kmcomp 14.0.263-beta

mcdurdin:
This updates the compiler to a close-to-final beta version which builds all keyboards successfully.

This is ready for merge.

Once we reach release, I will use #984 to enable 64-bit compiler which is necessary for use in WINE on macOS Catalina onward.


DavidLRowe (approved):
Taking your word for it!

PR #1488, opened 2021-03-18
[sil_euro_latin] support n dieresis

### LINKED ISSUE #1484: [sil_euro_latin] KeymanWeb: displayed Digraphs table is off
When using EuroLatin in KeymanWeb, instead of the keyboard layout a page is shown with what to type to get specific characters. The Digraphs table is off: As result of typing `f\l` the result `fi` is shown.

Screenshot from 2021-03-16 11-44-34

### LINKED ISSUE #1485: [sil_euro_latin] Add Letter N with Diaeresis to keyboard
Guinea-Bissau Creole (pov) uses upper- and lowercase letter N with diaeresis. This can't currently be typed with the EuroLatin keyboard. Please add `n` + `:` to produce **n̈** and `N` + `:` for **N̈**.

Also, Guinea-Bissau Creole could be added to the list of supported languages.

DavidLRowe:
Please excuse the wonky branch name!

usage.htm needed `colspan='2'` in the heading for `c\h` to address the issue in #1484 . A similar change was made in sil_euro_latin.html.

fixes #1484 

The remaining changes are for adding n dieresis.
fixes #1485 

I took my best guess at changing sil_euro_latin.html and sil_euro_latin_js.txt (&KMW_HELPFILE and &KMW_EMBEDJS respectively), imitating what was done for p dieresis. I made similar changes in usage.htm.

Also in sil_euro_latin.html the "More help" link referenced 1.8/sil_euro_latin.php, so I changed that to 1.9.4/sil_euro_latin.php, but that won't exist until the keyboard is published.

I inserted the "pov-Latn" code in sil_euro_latin.keyboard_info and sil_euro_latin.kps after "pl" because it looked as though the codes were in alphabetical order. Later I discovered other codes inserted at the end.

Thanks to your tip, I created new lowercase.png and uppercase.png files from the Excel file. The height of some of the rows seems to be off (apparently where the table to the right has a heading?), but it seems that has been the case in previous versions.

Hopefully I found all the places that needed changing, but I could easily have missed some.


LornaSIL:
>Also in sil_euro_latin.html the "More help" link referenced

If you change that link to: `https://help.keyman.com/keyboard/sil_euro_latin`
it will always go to the latest version and won't need to be updated again.

Maybe you could update:
* README.md
* readme.htm
* welcome.htm
* sil_euro_latin.php

where it says something like "The EuroLatin keyboard enables you to type in all European languages which use Latin-script, and most Latin-script languages from the rest of the world. " change to:
"The EuroLatin keyboard enables you to type in all European languages which use Latin-script, and **many** Latin-script languages from the rest of the world. **However, the keyboard should not be expected to be enhanced for languages outside of Europe.**"

Or something like that...what do you think?

Then...I wonder if we should update one of the pan-Africa keyboards to include a mobile layout. That might help this problem...although I'm not sure they are the best keyboards. We would need to review them.

LornaSIL (approved):
LGTM

PR #1489, opened 2021-03-19
chore: add blurb from keyman.com to sil_euro_latin keyboard

mcdurdin:
We want to merge the separate landing page for sil_euro_latin (https://keyman.com/eurolatin) as it is essentially the same content as the keyboard record. We thought we'd just add this blurb to the keyboard description :)


darcywong00 (approved):
lgtm

PR #1490, opened 2021-03-21
GFF Gurage v0.8

dyacob:
These are updates to the GFF Gurage and GFF Gurage & Amharic package for compatibility with the Unicode 14 standard.  The Gurage letters are revised to use the addresses defined in the Ethiopic Extended-B block of Unicode 14.  Documentation and the included fonts are likewise updated.


keyman-server:

LornaSIL:
Do you feel it's "safe" to release a font and keyboard with the new codepoints? I haven't felt like I should release a font with the Unicode 14.0 codepoints until the beta review. The alpha review goes until April 12 and the beta will start in June. It says:
> This early review is provided so that reviewers may consider the character repertoire issues prior to the start of beta review (currently scheduled to start in June, 2021). Once beta review begins, the repertoire, code points, and character names will all be locked down, and no longer be subject to changes.

which made me nervous about releasing something where codepoints could change.

dyacob:
Oh, hrmm, gambling on code points... :-)   I think we should wait, and for the work we're in need of doing in the near term we'll use a private build of Keyman Desktop.  I will make a new branch and PR, of the Gurage Keyboard just to include a note in the README.md and other descriptive pages to warn people that the any materials produced will become incompatible with the Unicode 14 update in about 6 months time.

Issue #1482, opened 2021-03-15
[khmer_angkor] Invalid link on Welcome page


ermshiperete:
`welcome.htm` contains a link to "Khmer Fonts by the Khmer Software Initiative", but the target of the link doesn't exist.

![Screenshot from 2021-03-15 15-53-43](https://user-images.githubusercontent.com/181336/111173124-b143ad80-85a6-11eb-96f1-924e05d20d64.png)

darcywong00:
When khmer_angkor gets an update, can you also include an example in .keyboard_info?
Maybe: <kbd>X</kbd> <kbd>J</kbd> <kbd>Shift</kbd>+<kbd>E</kbd> <kbd>R</kbd> for `ខ្មែរ`

I always forget the "correct" sequence.

PR #1491, opened 2021-03-24
[khmer_angkor] various updates and fixes

MakaraSok:
* fix the broken link on welcome page
* add example to the keyboard_info
* add tablet layout
* update phone layout
* get rid of the Lunar Dates layer on touch layout
* update documentations
* change OSK font to Khmer Mondulkiri


MakaraSok:
fix https://github.com/keymanapp/keyboards/issues/1482

LornaSIL:
* The link for the Mondulkiri font should be changed to: https://software.sil.org/mondulkiri/
* Under "More information" in the welcome.htm file there are two things relating to Unicode 10. Since we are at Unicode 13.0 now, it seems appropriate to remove those references, that is:
  * this text `Khmer character chart (Unicode Standard 10.0)` could just be changed to `Khmer character chart`
  * the link for **Khmer Script by Unicode Standard** should be changed from: `http://www.unicode.org/versions/Unicode10.0.0/ch16.pdf` to `http://www.unicode.org/versions/latest/ch16.pdf`
* All the changes to the welcome.htm file (like these links) must also be made in khmer_angkor.php (help file)

MakaraSok:
Thank you for looking into this update.

> * Under "More information" in the welcome.htm file there are two things relating to Unicode 10. Since we are at Unicode 13.0 now, it seems appropriate to remove those references, that is:
>   
>   * this text `Khmer character chart (Unicode Standard 10.0)` could just be changed to `Khmer character chart`
>   * the link for **Khmer Script by Unicode Standard** should be changed from: `http://www.unicode.org/versions/Unicode10.0.0/ch16.pdf` to `http://www.unicode.org/versions/latest/ch16.pdf`
> * All the changes to the welcome.htm file (like these links) must also be made in khmer_angkor.php (help file)

Unicode Standard evolves over time and it might be better to stick to the version considered when the keyboard was created. What do you think?

LornaSIL:
I would link to the latest. The keyboard may not support newer characters,
but the latest gives the most/best description of Khmer.

release/k/khmer_angkor/README.md
```diff
@@ -15,7 +15,7 @@ Supported Platforms
 -------------------
  * Windows
  * macOS
- *LInux
+ *Linux
```
mcdurdin, line 18:
```suggestion
 * Linux
```

mcdurdin (commented):
I think Lorna captured everything that was important. The changes look good.

MakaraSok (commented):
LGTM

LornaSIL (approved):
LGTM

PR #1494, opened 2021-03-26
Jawa (version 2.0)

bennylin:
Changing default font to Noto Sans Javanese, and other minor improvements.

OK, let's publish these changes first before doing a new project

PR #1495, opened 2021-03-26
[latin_jawa] New keyboard

bennylin:
New layout. Thanks.


keyman-server:

LornaSIL:
This one will take a bit more work. Let me give you some initial feedback.
* The language tags `jv`, `id` and `en-Latn-ID` are all for Latin script, but the keyboard is also supposed to support the Jawa script. I'm not sure how to handle that. It's possible you should also add `jv-Jawa` and `id-Jawa` to both the keyboard_info and the .kps file.
* The welcome folder has a lot of nice graphics, but the welcome.htm file doesn't use any of those graphics. You should probably add those in.
* The `readme.htm` file seems to use the graphics from the `welcome` folder. I believe that it is best for the `readme.htm` to be a simple text file with a short summary and the `welcome.htm` file to be the one that describes the keyboard layout. This is because when a user installs the .kmp the readme does not display the graphics, just boxes. It is the welcome.htm that will display graphics properly. All of your keyboard projects should be changed to follow this practice (`readme.htm` as a short summary of the purpose of the keyboard and `welcome.htm` to describe the keyboard layout).
* Please add a `latin_jawa.kpj` file in the root `latin_jawa` folder. That .kpj file needs to be associated with the `source\latin_jawa.kmn` and `source\latin_jawa.kps` files.
* This project also uses your renamed `noto2.ttf` file. Let's get the filename corrected in PR #1494 first. #1494 will need to be merged before this PR can use that file.
* You have 3 `.ico` files in the PR. You could delete all but `latin_jawa.ico`
* The file `latin_jawa-help.htm` is empty. Unless you are going to update that file, you can delete the file from the PR and also remove this line in the the .kmn: `store(&KMW_HELPFILE) 'latin_jawa-help.htm'`

I didn't test the keyboard yet as I'd like these other issues fixed first.

DavidLRowe:
We need to close this pull request since it is so far out-of-date. Please feel free to resubmit the files in a new PR. Thanks!

PR #1499, opened 2021-03-28
chore(japanese): use unminified version

mcdurdin:
This is part of investigation for #1493. Using the unminified version will make testing easier, and should not have negative consequences for for end users; the file size is slightly larger but not significant.


legacy/j/japanese/japanese.keyboard_info
```diff
@@ -8,7 +8,7 @@
     "ios": "none",
     "android": "none"
   },
-  "version": "1.0",
+  "version": "1.0.1",
```
darcywong00, line 11: For some reason, the PR build fails because it thinks the version should be "1.0" 😕

darcywong00 (approved):
lgtm

PR #1502, opened 2021-03-31
[khmer_angkor] fix the missing image

MakaraSok:
An image in the welcome and help pages does not show up correctly and this is to fix it.

Should the version number be bumped to 1.0.9?

### DIFF
--- a/release/k/khmer_angkor/HISTORY.md
+++ b/release/k/khmer_angkor/HISTORY.md
@@ -1,6 +1,10 @@
 Khmer Angkor Change History
 =======================
 
+1.0.9 (7 Apr 2021)
+----------------------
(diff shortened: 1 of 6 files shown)


DavidLRowe:
The version number will need to be incremented so that the new version of the .kmp will be offered to users. (If the only change had been for the online help, then the version number increment would not have been needed.)

MakaraSok:
> The version number will need to be incremented so that the new version of the .kmp will be offered to users. (If the only change had been for the online help, then the version number increment would not have been needed.)

Not just online help, but the welcome page which is included in the .kmp as well. I think the version number should be bumped now.

DavidLRowe (approved):

PR #1504, opened 2021-04-01
[tangsa_lakhum] initial commit

sven-oly:
First try at submitting to Keyman repository.

Tangsa language keyboard in Lakhum script, newly added to Unicode 14.0.


keyman-server:

sven-oly:
Please advise how I can ask for help with the build.

DavidLRowe:
Thanks for your submission. After a manual review of the pull request, the reviewer will initiate a build.

LornaSIL:
Welcome! There are several things that must be done and several things that would be nice.
Must do:
* The folder name needs to be the same as the file names. I assume that means you would change `tangs_lakhum_unicode` to `tangsa_lakhum_unicode`
   * A good change, but not required, would be to remove the  `_unicode` from all the files and folder names. I think it's superfluous and makes from an extra long url as the folder name is what is used in the url.
* The files you have in the root of `tangs_lakhum_unicode` should go into a folder called `source`
* Then in the root folder you need a kpj file (called `tangsa_lakhum.kpj` or `tangsa_lakhum_unicode.kpj` depending on what you name the folder
   * The .kpj file is then used to reference both the .kmn and the .kps file.
* In the .kps file
  * you will need to add a language tag. Go to the "Keyboards" tab, click "Add" under "Languages" and add "nst". Then under "script" it looks like we aren't supporting "Tangsa" yet. You can still just type "Tnsa" in for the script code.
  * Also, it looks like the .kps is referencing the .ttf from the wrong place. You may need to remove the .ttf from "Files" tab and then re-add it in from the `source` folder.
* Also, in the root folder you need a keyboard_info file.
  * It should be called `tangsa_lkhum.keyboard_info` (or include the `_unicode` if that's what you decide.
  * The contents should be something like this. You'll need to confirm the language code is correct.
```
{
    "license": "mit",
    "languages": [
        "nst-Tnsa"
    ],
	  "description": "Experimental keyboard for the Tangsa language using the Tangsa Lakhum script which will be encoded in Unicode 14.0"
}
```

Nice to do:
* As mentioned above, it would be nice to remove the `_unicode` as it is superfluous. 
* Since this is in the experimental folder, you don't have to have all the metadata files, but you really should add:
   * README.md
   * LICENSE.md
   * HISTORY.md 

I hope this isn't too confusing. I'm willing to modify files and send them back to you zipped for you to resubmit if you would find that helpful.

sven-oly:
Thank you for these very helpful comments. I'll update the files as suggested!

sven-oly:
Updated with renamed directory and files.

LornaSIL:
It is looking better! You still have a `tangsa_lakhum_unicode` folder in your Pull Request (PR). Could you delete that folder?

Then, you've committed the `build` folder in `tangsa_lakhum`. The `build` folder can be deleted from the PR because the Keyman system does the build. We don't want the build as part of the PR.

If you could move README.md, LICENSE.md and HISTORY.md to the root of `tangsa_lakhum` instead of in the `source` folder.

There's a readme.htm in the root and another in `source`. Choose the most up-to-date version and put it in `source`.

I'll give feedback on the keyboard itself when these changes are done.
Thanks!

sven-oly:
Thanks for your comments and suggestions. Please take another look!

sven-oly:
Updated the file paths. Thank you for your review.

LornaSIL:
It's looking a lot better. I do have quite a few comments! Since this is a new script to Unicode and new to me, we may have a few hiccups along the way.

* My personal recommendation is that we start with just a Desktop keyboard and don't try to do the mobile layout yet.
   * If you do that, you'll need to remove the targets in README.md for:
``` * iPhone
 * iPad
 * Android phone
 * Android tablet
 * Mobile devices
 * Desktop devices
 * Tablet devices
```
  * Open the .kpj project and then open .kmn through the **Project / Keyboards** menu.
    *  In the .kmn **Details** tab change **any** to **windows**, **macosx**, **linux**, **web**
    * The `tangsa_lakhum-help.htm` and `tangsa_lakhum.css` files are empty, so unless you plan to add something to those files, I would remove them from the **Features** section.
    * There is no `tangsa_lakhum-code.js` in your PR so I would also remove that from the **Features** section.
    * Please add a copyright statement under **Details** eg "(c) 2021 Craig Cornelius"
    * You have a lot of rules for the CTRL key. While that is acceptable, people might run into problems with that because Windows especially makes heavy use of CTRL+C, CTRL+V, CTRL+A, etc. I like to use the RALT key instead of CTRL because it has less negative ramifications. However, you can leave it CTRL if that is your preference. Are they the same as what you've put on the deadkey ( [ )? If so, maybe you don't need them on the CTRL? Or maybe since it's experimental you want to see what people prefer. I'm okay with that.
    * The touch layout has two platforms: tablet and phone. Both of them have Latin script on the keycaps. It looks to me like you just pulled in the template and if that is the case, you will get unsatisfactory results. That is why I'd like it if you removed that from this version. 
      * To remove it, you'll need to go back to the **Details / Features** section and remove "Touch-Optimised Keyboard"
      * If you want to keep the touch layouts, then they'll need more work.
    * Lastly, go to **Keyboard / Fonts** and make sure your font is selected for the On-Screen Keyboard.

* Assuming you are removing the files I mentioned above, let's go to the .kps file by going back to the .kpj file and then clicking on the "Packaging" tab and clicking on the .kps.
   * In the **Files** tab you should click **Add...** and go to the build folder and add in `tangsa_lakhum.kmx`
   * Then go to the **Keyboards** tab and select your **Keyboard font** and **Display font**. 
   * Under **Languages** click on **Add...** and type in **nst** for **Language tag:** and **Tnsa** for **Script tag:**
   *  Please add a copyright statement under **Details** eg "(c) 2021 Craig Cornelius" (I know, it's a pain to do it in multiple places. We've asked if this could be simplified.)

I *think* that's it. I *am* having a bit of a problem with the font. Some applications don't seem to recognize the font and others do. I guess that's because the script is going into Unicode 14.0. There isn't a Unicode bit for Tangsa yet, but if you have any control over the font, you might want to have the designer set bit 57 (Non-Plane 0) for the Unicode Character Range (https://docs.microsoft.com/en-us/typography/opentype/spec/os2)

sven-oly:
Thank you again. I really appreciate your careful review and suggestions. I hope this continues progress!

LornaSIL (approved):
I went ahead and made some changes to your files so that we can go ahead and approve this (rather than keep sending it back to you for updates). The most significant was that the .kmx file wasn't included in the list of files in the .kps. Without a .kmx, the Desktop keyboard wouldn't work. 

You can now see how it works and update the keyboard as necessary. Remember to pull from master so the changed files will be available to you. 

Remember when you update the keyboard you have to bump the version number.

LornaSIL:
Keyboard is now available here: https://keyman.com/keyboards/tangsa_lakhum
I see the font is not available. I'll try to fix that.

sven-oly:
thank you very much for your patient help!

LornaSIL:
Do take a look at how I revised the keyboard_info file to call the font for the Web keyboard.

In the future, if you update the font (which is quite likely), could you please remind us to update the font on the "other" server? That way we can keep the web font updated too.

PR #1505, opened 2021-04-01
chore: Update kmcomp to 14.0.271 stable

darcywong00:
Updates keyboard compiler to the latest stable version to incorporate fix from keymanapp/keyman#4832


mcdurdin (approved):
LGTM if build passes

PR #1506, opened 2021-04-01
Nias keyboard version 1.0

sslaia:
I am glad to find this open-source keyboard and this is my first contribution. Hope this would help Nias community to preserve their fast disappearing language in the digital age.


keyman-server:

LornaSIL:
Thanks for your keyboard submission! Overall it looks very good, but I do have a few comments.
* The touch layout seems to be almost the same for both tablet and phone. If they are intended to be the same, then you should just delete one of them because maintaining two equivalent keyboards is unnecessary. I would keep the phone one because if you want to keep the tablet one, there is one change.
   * For the table layout, numeric layer, the SHIFT key doesn't do anything. You should either "Next layer" to "shift" or remove the SHIFT key.
   * For BOTH phone and tablet layouts, on the numeric layer, the `[`, `{`, `]`, `}` keys all produce [{]{ and not öÖŵŴ. You should change the `Code` from `K_LBRKT` to `U_005B`, etc. (This is an example of where you need to do this in both the tablet and phone layouts (unless you choose to delete the tablet layout.
* If you do delete the tablet layout, you should change the help file from `<div id='osk-tablet' data-states='default shift'>` to `<div id='osk-phone' data-states='default shift'>` (from tablet to phone). 
*

sslaia:
> * The touch layout seems to be almost the same for both tablet and phone. If they are intended to be the same, then you should just delete one of them because maintaining two equivalent keyboards is unnecessary. I would keep the phone one because if you want to keep the tablet one, there is one change.

Thank you for this valuable feedback. I'll try to implement just that. Even more as I realised, we will use this keyboard a lot for writing entry in Wiktionary, so it'd also good to have certain keys directly available on the touch layout.
How would I proceed? Should I close this PR first and submit a new one later?

LornaSIL:
If you know how to push to the same Pull Request you can make the changes in this one. It depends on how you are working.
I do:
`git checkout nias`
make my changes
`git status`
make sure all the changed files are what I want. If so,
`git add .`
`git commit -m "[nias] Changes per feedback"
`git push origin nias`

That should push your changes to this PR. If you are just uploading via a web interface, you may want to create a new PR. Whichever is easiest for you.

sslaia:
> That should push your changes to this PR. If you are just uploading via a web interface, you may want to create a new PR. Whichever is easiest for you.

Thank you for this. I think I just close this one first. The holy week will be busy, so I'll create a new PR after Easter. Happy Easter to you and your family.

LornaSIL:
Thank you! And to you.

PR #1509, opened 2021-04-05
Engram v2.0

binarybottle:
I have updated the Engram keyboard layout based on feedback from the ergonomics community.

Version 2.0 uses a revised scoring model that better accounts for finger dexterity and strength.


keyman-server:

PR #1510, opened 2021-04-05
Engram v2.0

binarybottle:
I have updated the Engram keyboard layout based on feedback from the ergonomics community.

Version 2.0 uses a revised scoring model that better accounts for finger dexterity and strength.


keyman-server:

DavidLRowe:
@binarybottle Did you intend to remove the rules to make CapsLock work?

If you're using Keyman Developer version 14, you can now instruct the compiler to create those extra rules for specific keys. Furthermore the compiler now supports ranges, so you don't have to list each character.

For information on CasedKeys, see https://help.keyman.com/developer/language/reference/casedkeys
For information on Range expansion, see https://help.keyman.com/developer/language/guide/expansions

store(&CasedKeys) "a".."z"
would indicate that all the a-z keys on a US English keyboard should obey CapsLock. But you may not want that since you have TYGHBN keys as punctuation, and have letters on ;,./['

store(&CasedKeys) "'" ",".."9" ";" "=" "[".."]" "`".."z"
would indicate that all the keys should obey CapsLock, which you likely don't want either!

BTW, you might want to change the comments near the top of the file, since "a" doesn't produce "q" on your keyboard.

binarybottle:
Hello and thank you, @DavidLRowe!  Have I fixed the CAPS with the changes I made?:

https://github.com/keymanapp/keyboards/pull/1510/commits/c54392ee6affe12459cc7a1f9dc4171cc357dd8e

DavidLRowe:
This applies CapsLock to all the keys on the keyboard. That is, with CapsLock on, pressing the "2" key would produce "=". Is this what you want? This is the way, for example, the French keyboard works, but not the way the US English keyboard works, where only keys producing alphabetic characters are affected by CapsLock.

I'm happy to approve this as is, if the behavior is what you want, but wanted to give you the chance to change it if you want to limit CapsLock to alphabetic characters.

binarybottle:
Yes, that would be perfect -- THANK YOU, @DavidLRowe!

DavidLRowe:
Just to clarify, @binarybottle: You're happy with the PR as it currently is and don't want to make changes.

binarybottle:
The caps lock behavior you described is exactly what I want, so yes!

DavidLRowe (approved):

binarybottle:
Thank you once again, @DavidLRowe!

PR #1511, opened 2021-04-07
[indonesia] keyboard layout version 2.0

bennylin:
update touch layout to harmonize with my other keyboards


keyman-server:

LornaSIL:
Thanks for this update.
I have two issues:
* The copyright that shows on the .kmp is still 2020. If you could change that in the .kps file **Details** tab, **Copyright** field, that would be great.
* The other issue would be a bigger change. It looks to me like when you install the package, the preview **Readme** does not handle graphics. So, I would recommend that you change the `readme.htm` to just have a quick summary like the first 2 paragraphs WITHOUT the graphics. THEN, add a welcome.htm file which is the equivalent of what you have in the readme.htm. The end result would be when the user installs the package they can click on **Readme** and it gives the quick summary and THEN after the user clicks on "Install" the welcome.htm file will popup and that one does display graphics properly. This would better follow how we handle other keyboard packages.

Neither of these is required for me to approve the keyboard, but it would be better. If you do want to add the welcome.htm, remember to **Add** it to the list of files in the .kps file. I don't think you have to do anything else there since the graphics are already included.

LornaSIL (approved):
I'm approving this. There were a few changes I would have liked to see, but they are not essential. I added those in a "To do" in the README.md file.

PR #1514, opened 2021-04-12
[sylheti_nagri] New keyboard on behalf of STAR

LornaSIL:
This keyboard is ready to be checked...but I'm double-checking the language name example in `keyboard_info` (so it's not ready to merge if everything else is okay).


LornaSIL:
Keyboard is ready for checking now.

DavidLRowe:
In welcome.htm (and .php file), the second example under "Each of these have a different appearance..."
- The example has r + SHIFT+Y + z but the result looks like l + SHIFT+Y + z. Could you verify?
- Also, shouldn't that introductory sentence start "Each of these has..." (rather than "Each of these have...")?

I couldn't get the font to display correctly in the mobile emulator, so I didn't do much checking of the mobile layout.

DavidLRowe:
Oh, I see now there's a second "Each of these have" sentence in welcome.htm and the .php file. I wouldn't have mentioned the other except that I think the file might need to be changed anyway. Up to you whether to to change the "have" to "has", but if you change one, change both of them!

LornaSIL:
I think I'm going to leave the word "have" because it's referring to the following 3 examples and because that's the way the author wrote it. I did change the 2nd example to use "r" rather than "l".

DavidLRowe (approved):

PR #1519, opened 2021-04-15
Obolo Chwerty (v 1.1.2)

katelem24:
Version 1.1.2, an update to Version 1.1.1


keyman-server:

DavidLRowe:
The structure of this pull request is good. There are a few issues with some files, but you can fix them by adding revised files to the "obolo_chwerty-(v-1.1.2)" branch (which is the basis for this pull request).

The OSK (on screen keyboard) shows Ctrl+Alt and Shift+Ctrl+Alt layers, but they don't appear in the keyboard. Either the layers should be removed from the OSK, or else the relevant commands should be added to the .kmn file to support these layers. At any rate, the OSK should match what the desktop keyboard actually produces.

The chwertyftkey1.png file in the source/welcome folder seems to be damaged, which means the image does not appear when welcome.htm is displayed.

In the .php help file:
- the default and shift layers of the desktop keyboard are displayed twice: once using the .png images, once using `<div id='osk' data-states='default shift'>`. This wouldn't stop me from approving the keyboard, but you might want to remove the reference to the two .png images from the .php file and the images themselves from the help folder. But if you want to leave it as it is, that's fine.
- under the Mobile/Tablet Layout heading, you have `<div id='osk-tablet' data-states='default shift numeric'>`, which displays the default and shift layers, but not the numeric layer, since the third layer is actually named "alt" in your touch layout. You should either change the name of the layer in the touch layout, or else change "numeric" to "alt" in the `<div>` line in the .php file.
- the references to the three Enter key images do not match the filenames in the help folder, which means these images do not display.

In the touch layout, shift layer, the "^" key produces "₦". Because the code for that key is K_6 and you are on the Shift layer, [Shift K_6] is passed to the .kmn file, which produces "₦". If you change the code for that key on the shift layer to U_005E, it should produce "^" (U+005E), and match what is displayed on the key itself. 

In the .kmn file, your CapsLock rules work for ch, sh, ọ, n̄, but, without more work, CapsLock won't apply to the other letters. What I think is needed is:
- Add `store(&CasedKeys) 'a'..'b' 'd'..'g' 'i'..'p' 'r'..'w' 'y'..'z'` to the .kmn file (or you could use the equivalent `[K_A]..[K_B]` etc.)
- Add rules for all those letters:
```
+ 'a' > 'a'
+ 'A' > 'A'
(etc)
```
or equivalently:
```
+ [K_A] > 'a'
+ [SHIFT K_A] > 'A'
(etc)
```
These are the default rules for the US English keyboard that Keyman uses unless you redefine them, however it seems they must be explicitly stated in order to work with the &CasedKeys store.

I'm fine if you want to leave this as it currently is and deal with CapsLock in a later version of the keyboard. There are plenty of keyboards in the repository that do not handle CapsLock at all!

katelem24:
I've committed a new chwertyftkey1.png file to replace the previous one.

The OSK in this keyboard had only two layers - default and shift. I can't see any option in Keyman Developer to remove the other layers.

On what the keys output in OSK: Two keys - the backquote and backslash keys - in default and shift layers are used to place diacritics on the base letters. If those keys are pressed independently they produce what is on the keycap, but when pressed after "aeioọumn" it should those letters with accent. So it's not really clear to me what should be on the OSK, whether it should be the backquote/backslash or the tone marks.

katelem24:
I want [SHIFT K_6 ]  to produce ₦ while ]+[SHIFT K_6] should produce ^

I've tried the store(& CasedKeys). I wanted unshifted state to produce lowercase (ch, sh, ọ, n̄), shifted state to produce title case (Ch, Sh, Ọ, N̄), and capslock to produce uppercase (CH, SH, Ọ, N̄). I couldn't figure out how to do "override" as was said in the webinar so I commented out that line. Maybe I will be take care of it in the next version. My question is here: https://community.software.sil.org/t/store-casedkeys-with-many-characters-but-few-keys/4660

DavidLRowe:
Sorry I wasn't clear. The "^" that produces ₦ is on the touch layout, shift layer. That is the key displays "^" but doesn't output "^".

You are right that there doesn't seem to be an option to remove layers from the OSK. I expect one could use the "Fill from layout" button and that would wipe out the existing data and generate new data from what is actually in the .kmn file. This, however, would mean that you'd have to redo the backslash key, etc. 

Another option would be to use the "Code" tab at the bottom of the OSK pane, then carefully delete from `<layer>` to `</layer>` for `shift="CA"` and `shift="SCA"` and `shift="C"`. (The third one is a space on the Ctrl layer, which isn't necessary, so you might as well delete it as well.)

I understand what you say about the accent keys on the OSK. What you put on the OSK keys doesn't affect what is actually produced when you type. You could, for example, use something like `^ \` for the backslash key that also puts a circumflex on vowels. I'm not sure I'd bother for the backquote (since it more or less looks like a grave accent).

katelem24:
Done. I've made those changes but I've not committed them yet.

I need you to help me out in one thing: the space bar in layer "alt" for touchscreen is not working. When I tap it, it doesn't respond. Could you help me check what may be the reason.

DavidLRowe:
In Keyman Developer, for the obolo_chwerty.kmn file, Touch layout tab, select Layer: "alt". Click on the space bar. Change the Modifer: value from "(current layer)" to "default". Now when you select the space in the alt layer of the touch layout, it will send [K_SPACE] (with no modifier, since it is "default" layer) rather than [ALT K_SPACE] (that is the space with the "alt" modifier from the "alt" layer of the touch layout). Many keyboards use "numeric" as the layer name at this point, and avoid this problem.

katelem24:
Thank you. The space bar is working now.

katelem24:
I'm done. Please cross-check.

DavidLRowe:
I added a commit to your pull request to remove the unwanted layers from the OSK.

Even though the "^" on the shift layer of the touch layout is producing the Naira sign, I think it's OK because you can get the "^" on the alt layer. It would be good to correct that in the next version of this keyboard.

DavidLRowe (approved):

katelem24:
Thank you

PR #1520, opened 2021-04-16
[sil_eastern_congo] Updated one keystroke that was problematical and …

LornaSIL:
…added NFC


DavidLRowe:
Just some questions:

`!!o` gives me `ő` (U+006F U+030B) as expected, but `!!;o` gives me `!ɔ̌` (U+0021 U+0254 U+030C), which was not expected. It turns out I need `'';o` to get `ɔ̋` (U+0254 U+030B). Is that an oversight? or by design?

Should double acute over OoUu use composed forms (as you have done for double grave)?

For the touch layout, I type `o` and double grave accent and I get U+006F U+030F. Should that be outputting U+020D instead? Or is it only the desktop layout that produces NFC?

LornaSIL:
The client is changing his mind about something, so please don't approve!

DavidLRowe (approved):

PR #1522, opened 2021-04-22
[Baybayin] New Keyboard

Ilhamkang:
Please review this request, and let me know if there is something to adjust. Thanks in advance.


keyman-server:

LornaSIL:
Thanks for the submission! I haven't checked your keyboard, but two things I noticed:
* Keyman really likes htm suffixed files (not html), so if you could change the name of the `welcome.html` to `welcome.htm` it will open it when the package is installed.
* I don't see a minimum Keyman version set in the `.kmn` file. That should be set, so add this line at the top of the file:
```store(&VERSION) "10.0" ```

**Comment by Ilhamkang** [comment]
@LornaSIL I have revised the two files.

**Comment by LornaSIL** [comment]
I apologize for delaying in my review.

* There is no `baybayin.kpj` file. You need to add that, and it should be in the root of the `baybayin` folder and it should reference both the .kps and the .kmn file which are correctly in the `source` folder.
* Once you associate those two files, you can go to the .kmn file and then click on **Keyboard / Fonts** and associate the **Noto Sans Tagalog** font with this keyboard. Click **Save** and it will save the font association too.
* Also, in the kmn file, when you have a script like this with a small number of characters, I find it helpful to make a rule that prevents the unused keys from outputting Latin characters. Adding this to your rules will do that:
```
store(nul) "acefjoqrvxzBCDEFGHJKLMOPQRSTVWXYZ"
+ any(nul) > nul
```
This only prevents producing Latin letters. It allows for the input of the digits and punctuation. If you don't want them to produce anything else, just add that to the `store(nul)`
* In the `baybayin.kps` file it might be best to go to the `Details` tab and remove the version and then click on "Package version follows keyboard version". I cannot think of any good reason NOT to do this and it will save problems in the future so you don't have to remember to update that field when you update the keyboard.
* When I open the `welcome.htm` file, it does not display the graphics.
     * Change `` to `` and do the same for the other graphic.
* In the baybayin.php (help) file you can delete this line: ` ` That image is not available and it is unnecessary. This line `<div id='osk' data-states='default shift'>` will automatically create the images for you.

I think that's all the changes needed.

Ilhamkang:
Great! No problem for the delay. Thank you for the review. I will do the needed changes.

Ilhamkang:
@LornaSIL please review the changes. Thank you.

LornaSIL:
Thanks for fixing those. There are a few remaining issues.
* The `keyboard_info` file needs a comma at the end of the description line. It looks like you deleted it in the last round of edits:
`"description": "This keyboard supports the Tagalog language using the Baybayin script.",`
* The build path is wrong in the .kpj. Open the .kpj in Keyman Developer, go to **Project / Project Settings** and set the **output path** to `$PROJECTPATH\build`
* In the .kmn file, under "group(main) using keys" I would add this rule (you added the store, but not the rule):
`+ any(nul) > nul`
* The `welcome.htm` file still has the wrong paths for the graphics. Look at my comments above about getting those correct. 
   * (Also, the filename kbd1.PNG must match what you have in the welcome.htm file including casing. So, you should either rename the files to use lowercase, eg kbd1.png (that would be my preference) or change the name in the welcome.htm from kbd1.png to kbd1.PNG. This is because casing matters in Linux and won't display properly on Linux if the filename is uppercase and the welcome file is calling for a lowercase filename.
* In the .kps file, you should add the files: welcome.htm, kbd1.png, and kbd2.png. 

I think everything else looks okay.

Ilhamkang:
Dear @LornaSIL , thank you for pointing out the missed parts. I think I have made the whole needed changes. But there is a problem. When I put the rule `+ any(nul) > nul`, the .kmn could not be compiled locally on my computer. Is this a bug or a normal thing?  

On the .htm file, I changed the image paths to `` and ``, hopefully the images will appear to you. I see your suggestion on the previous comment to change the path was identical with the original path I've made before. Please confirm if it has been corrected at the moment. 

Thank you.

LornaSIL:
I tried to commit changes to your repo. If you can accept and merge them I think the keyboard will be okay.

- I deleted welcome.htm
- I renamed welcome.html to welcome.htm
- I changed the filenames in kps to welcome.htm
- I renamed kbd1.PNG to kbd1.png (and kbd2)
- I moved your "nul" rule down to below the "group(main) using keys" line as it must be in that group.

These are all important changes for making the keyboard work better.

Ilhamkang:
Thank you for the corrections @LornaSIL. Somehow I could not rename the .PNG to .png in my computer, or even I renamed/changed to the new file, the image format keep appears .PNG. I will find a way to resolve this for my computer and will make sure to use the .png format for the future keyboard proposal.  

Hopefully this Baybayin keyboard will work well.

LornaSIL (approved):
LGTM.
I did have to make some more changes including fixing paths for some of the compiled files. Please make sure you update your repo to match master before you make any new Pull Requests.

LornaSIL:
>Somehow I could not rename the .PNG to .png in my computer, or even I renamed/changed to the new file, the image format keep appears .PNG. I will find a way to resolve this for my computer and will make sure to use the .png format for the future keyboard proposal.

I used `git mv kbd1.PNG kbd1.png`. That seemed to do it.

Ilhamkang:
> > Somehow I could not rename the .PNG to .png in my computer, or even I renamed/changed to the new file, the image format keep appears .PNG. I will find a way to resolve this for my computer and will make sure to use the .png format for the future keyboard proposal.
> 
> I used `git mv kbd1.PNG kbd1.png`. That seemed to do it.

Okay this is helpful. Thanks.

LornaSIL:
This PR has been merged. If there is something you would like changed, could you create a new issue?

Ilhamkang:
Thank you Lorna. Sure, I will create a new issue for the future change.

PR #1525, opened 2021-04-28
[hieroglyphic] Fix platforms in keyboard_info file

darcywong00:
While investigating keymanapp/keyman#4990 I was able to install hieroglyphic keyboard on Keyman for Android.
I wondered why:
1. The keyboard wasn't working (e.g Shift-H)
2. The bottom right key on the default layer had no output
hieroglyphic default layer

Turns out the hieroglyphic keyboard should not be available on mobile apps though
https://keyman.com/keyboards/hieroglyphic lists:

hieroglyphic platforms

hieroglyphic.kmn only targets 
```
store(&TARGETS) 'windows macosx linux web'
```

I suspect kmcomp sees the .js file in the .kmp file and erroneously includes Android and iOS.

This PR fixes the platforms that get listed in the keyboard_info file.

### DIFF
--- a/release/h/hieroglyphic/HISTORY.md
+++ b/release/h/hieroglyphic/HISTORY.md
@@ -1,6 +1,10 @@
 Hieroglyphic Keyboard Change History
 =======================
 
+1.3.2 (29 Apr 2021)
+-------------------
+* Remove JS file from keyboard package
+
 1.3.1 (23 Apr 2019)
 -------------------
 * Rebuild to resolve compatibility issue with Chrome
--- a/release/h/hieroglyphic/HISTORY.md
+++ b/release/h/hieroglyphic/HISTORY.md
@@ -3,7 +3,7 @@ Hieroglyphic Keyboard Change History
 
 1.3.2 (29 Apr 2021)
 -------------------
-* Remove JS file from keyboard package
+* Remove JS file from keyboard package as the keyboard does not currently support mobile devices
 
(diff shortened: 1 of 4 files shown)


mcdurdin:
Technically the .js should not be in the .kps, because it's only for web.

release/h/hieroglyphic/HISTORY.md
```diff
@@ -1,6 +1,10 @@
 Hieroglyphic Keyboard Change History
 =======================
 
+1.3.2 (29 Apr 2021)
+-------------------
+* Remove JS file from keyboard package
```
mcdurdin, line 6: This doesn't really explain why we are doing that -- perhaps append "as the keyboard does not currently support mobile devices"?

release/h/hieroglyphic/source/hieroglyphic.kmn
```diff
@@ -3,10 +3,10 @@ store(&NAME) 'Hieroglyphic'
 store(&BITMAP) 'hieroglyphic.ico'
 store(&INCLUDECODES) 'codes.txt'
 
-store(&COPYRIGHT) '© 2010-2019 Christian Casey'
+store(&COPYRIGHT) '© 2010-2021 Christian Casey'
```
mcdurdin, line 6: This date shouldn't change for minor edits

mcdurdin (approved):

PR #1528, opened 2021-04-28
fix: rewrite .kmp files that are incompatible with Keyman 8

### LINKED ISSUE #1527: bug: Manually-edited .kmp files are not compatible with older versions of Keyman Desktop
Regression from #1405.

It looks like this update may have made the .kmp files incompatible with older versions of Keyman Desktop -- the .zip file format used is not compatible. We need to rewrite these packages to fix that.

* For future reference, do you know which ZIP tool you used to edit the packages?
* Are there likely to be other legacy packages with the same issue? (I can fix those in the same way)

## Proposed Fix

My suggestion: I will unzip each of the files in question, and re-zip with a hacked together tool that uses the same zip builder that the .kmp compiler does. For this process, I won't update the version number -- I'll manually replace the existing .kmp files on the download server when the PR is approved and merged.

## Original report (via email)

I've just noted that the current version of your Keyman package grkpoly2.kmp distributed at

https://keyman.com/keyboards/grkpoly2

is corrupt and therefore refuses to install. Error message:

The file grkpoly2.kmp could not be installed because of the following error: Not a valid zip file!

This should be installable in Keyman Desktop 8.0 for Windows, because the minimum version required is 7.0, but the Keyman installer doesn't accept the file. The .kmp file itself can be opened in WinRAR without errors, and the included .kmx file installs just fine and works correctly when it is extracted from the .kmp and installed, so this is a problem with the .kmp package. 

I still have an older version of the grkpoly2.kmp package which installs and works fine, so apparently something went wrong when the .json file was added to the package in december 2020.

mcdurdin:
Fixes #1527.

This simply rebuilds these kmp files to be compatible with Keyman Desktop 8; there are no changes to the contents of the files. I will manually upload these to the download server, because given the file versions have not changed, they will not otherwise be updated.

I have tested these new files with Keyman Desktop 8 and they do work.


mcdurdin:
This is the tool (including source) that rewrites the .kmp files, in case we need this in the future for any other similar cases:
[RewriteKmp.zip](https://github.com/keymanapp/keyboards/files/6395494/RewriteKmp.zip)

DavidLRowe (approved):

mcdurdin:
Okay, I have manually uploaded the corrected .kmp files to downloads.keyman.com and purged the relevant files from the Cloudflare cache.

PR #1539, opened 2021-05-10
[sil_bunong] Bunong (SIL) Keyboard

MakaraSok:
Description
-----------

This keyboard is designed for typing Bonong language using Khmer characters.

Supported Platforms
-------------------
 * Windows
 * macOS
 * Linux
 * Web
 * Mobile Web
 * iOS
 * Android


tbequette:
On Android the character ៈ is not combining with the preceding vowel.

tbequette:
Looks good! Here are two more requests:
Please add the rotation - > — on the Android keyboard. The symbol — its needed in Paratext. 
Also, a dot in the top right corner of the ៖ key and the <space bar> would be a helpful reminder that these keys have rotations.

MakaraSok:
> Please add the rotation - > — on the Android keyboard. The symbol — its needed in Paratext.

The rota is in place on desktop layout. Press the hyphen key twice to get the em dash and press it again to get back to a regular one.

> Also, a dot in the top right corner of the ៖ key and the would be a helpful reminder that these keys have rotations.

@mcdurdin Is this possible to add a little dot with a different color on the top right corners of keys with rotas? This would be a helpful hint for users. If not possible, probably coloring the letter on the keycap may be a workable alternative.

tbequette:
Hi Marc,

Greetings from Mondulkiri where I'm making work for Makara ;-) But,
don't worry he's getting lots of things done!

Regarding the dots in the top right-hand corners on Android keyboards, I
thought the rotation options and long-press options for a key would be the
same, but I guess they don't have to be. Either way, it would be nice to
have some indication that a key provides a rotation.

Warm regards,
Todd

mcdurdin:
I think @DavidLRowe has a good suggestion -- it also gives discoverability for the rotas.

Otherwise, a different key cap colour may be a helpful alternative.

Rotas can be problematic if there is ever a need to type the same letter twice; in Keyman 15 we plan to introduce 'multitap' for mobile devices which will work the same as rotas but will be timing-based (this is part of the support for LDML keyboards that we need to include as well).

mcdurdin:
We'll also be adding a `hint` property to each key which will allow the keyboard designer to specify a tiny letter to go in the top-right of the key (in place of the dot).

MakaraSok:
Thank you @DavidLRowe and @mcdurdin for the prompt responses to this.

For now, I'll add longpresses to those keys with rotas.

In the future version, we may follow suits with the multitaps and the little character hit.

MakaraSok:
Problem solved for the touch layouts, but now it looks like it may take CSS to take care of the hint on OSK. Letter on the keycap coloring probably?

tbequette:
I installed sil_bunong_may20.kmp on PC and phone and it works well. Thanks!

LornaSIL:
* In the .kmn file, could you assign a minimum Keyman version? `store(&VERSION) '9.0'`
* In the `README.md` file the language name is misspelled (or at least inconsistent): Bonong > Bunong
* I think you could delete the file: `sil_bunong.png` from the repo.
* It's not a big deal, but pressing SHIFT+\ produces `|`. The OSK doesn't indicate it should produce anything.
* On the tablet layout there is the 102nd key. Since the desktop and OSK do not use the 102nd key, I think it could be removed from both layers of the tablet layout. Otherwise, it just adds to confusion.
* In keyboard_info the sample keys you list are incorrect. I think `VUn,g` should be `FUn,g` 

I've done minimal testing of the keyboard - I'm assuming since @tbequette checked the keyboard that it works correctly.

MakaraSok:
Thanks, @LornaSIL, for the feedback. All is well now.

LornaSIL (approved):
LGTM.

PR #1550, opened 2021-05-16
[sil_kmhmu] feat: add special font for the On Screen Keyboard to improve display of diacritics

### LINKED ISSUE #1542: [sil_kmhmu] four vowels on OSK shown as boxes
Four vowels (three on the first row and one on the second of the Default layer) and one mark on the Shift layer are shown as boxes instead of the expected glyphs of 

- U+0EB8 LAO_VOWEL_SIGN_U,
- U+0EB9 LAO_VOWEL_SIGN_UU,
- U+0ECD LAO_NIGGAHITA, and
- U+0EB3 LAO_VOWEL_SIGN_AM
- U+0ECC LAO_CANCELLATION_MARK

Currently shown on 

Default layer:
image

Shift layer:
image

The expected appearances should be like those of what are shown on the [help page](https://help.keyman.com/keyboard/sil_kmhmu/1.6/sil_kmhmu).

---

Even though the keys are shown up in boxes, they still output the expected characters which look as expected on the screen.

- ຸ
- ູ
- ໍ
- ຳ
- ໌

---

Font set for the OSK is [Kmhmu OT 2.002.ttf](https://github.com/keymanapp/keyboards/tree/master/release/shared/fonts/kmhmu).

image

mcdurdin:
Fixes #1542.

With thanks to John Durdin for creating the Lao Osk font and updating the .kvks file.

Question: is this work needed also on the .keyman-touch-layout?

Note: the `Lao OSK.ttf` font would also improve some of our other Lao keyboards in the same way. Something to consider for a future update?

## New OSK

(note the different styling on the diacritics as well as the immediate fixes to the issues shown in issue #1542):

### Base layer

image

### Shift layer

image


LornaSIL:
Then, let's put the font in the `shared` folder so other keyboards can
benefit and we don't end up with multiple copies.

mcdurdin:
> Then, let's put the font in the `shared` folder so other keyboards can benefit and we don't end up with multiple copies.

Yep, already done; it's in: release/shared/fonts/kmhmu/Lao OSK.TTF

Could move it to release/shared/fonts/laoo folder instead if you prefer.

LornaSIL (approved):
LGTM

Issue #1544, opened 2021-05-12
[sil_ethiopic] the help page is not well aligned


MakaraSok:
Boxes are shown inline. It does not look like they are intended.

Link: https://help.keyman.com/keyboard/sil_ethiopic/2.0.3/sil_ethiopic

![image](https://user-images.githubusercontent.com/28331388/117949122-fcc2cf80-b33b-11eb-9877-7e823287cfbd.png)

LornaSIL:
This is still not correct. Sigh.

DavidLRowe:
Looking at https://help.keyman.com/keyboard/sil_ethiopic/2.0.3/sil_ethiopic, I'm not seeing any issues, but you know that script far better than I do!

LornaSIL:
It is improved, not perfect.

PR #1551, opened 2021-05-19
[fix_silEthi_php] Fix help file to look as expected by Keyman

LornaSIL:
#1544 
My previous attempt did not look as the Keyman help files are supposed to look. It just looked like an htm page without Keyman branding.

This one puts the Keyman help header info in there and hopefully will look better.

However, the tables will probably still need some scrolling as we see with: https://help.keyman.com/keyboard/sil_ethiopic_power_g

### DIFF
--- a/release/sil/sil_ethiopic/source/help/sil_ethiopic.php
+++ b/release/sil/sil_ethiopic/source/help/sil_ethiopic.php
@@ -1,17 +1,7 @@
 <?php
-  $pagename = 'SIL Ethiopic Keyboard Help';
+  $pagename = 'SIL Ethiopic Help';
   $pagetitle = $pagename;
   $pagestyle = <<<END
- #section2 td p a {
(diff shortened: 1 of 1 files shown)


DavidLRowe:
Are you happy with the table? I found I had to go all the way to the bottom of the page to find the left-right scroll bar in order to see the rightmost column of the widest table. But there are only 3 characters in that column, so I guess that's okay. I don't have a clue about all the HTML stuff displaying the table, so don't have any suggestions for changes!

I'll approve the PR and you can merge it (or dismiss my review and change it if you want).

DavidLRowe (approved):

LornaSIL:
I'm not entirely happy, but the author is trying to automate his documentation and so I can't really change the structure of the table without it being a problem when he needs to update the documentation again. This is better than it was anyway.

PR #1557, opened 2021-05-24
[dega_keyboard] Dega keyboard

MakaraSok:
Description
-----------
Dega Keyboard is created for indigenous people in the Central Highlands of Vietnam, to preserve their writing, this Dega Keyboard can be used by Êđê, Jơrai, Kơ’Ho, Mơnông, Bahnar and many other indigenous ethnic groups.

Supported Platforms
-------------------
 * Windows
 * macOS
 * Linux
 * Web
 * iPhone
 * iPad
 * Android phone
 * Android tablet
 * Mobile devices
 * Desktop devices
 * Tablet devices


LornaSIL:
@MakaraSok in the future my preference is NOT to have `keyboard` in the filename/foldername/keyboard name. It is redundant since everything in this repo is a keyboard. I guess you don't have to change it this time since it's a lot of work, but I would prefer we get out of the habit of doing this. 

I also prefer NOT to have `Unicode` in the filename/foldername/keyboard name. This is not the case for this keyboard, I just wanted to say it again :)

MakaraSok:
For the goodness of the community, if not yet, we should have this mentioned in a white paper or a keyboard creation guideline somewhere. Probably specifically recommend the do's and don'ts.

LornaSIL:
>I also prefer NOT to have Unicode in the filename/foldername/keyboard name. This is not the case for this keyboard, I just wanted to say it again :)

At one point @DavidLRowe did document this somewhere, but I'm not seeing it.

DavidLRowe:
See https://github.com/keymanapp/help.keyman.com/pull/403

MakaraSok:
User testing has been done and it is affirmative that this keyboard is working as expected.

MakaraSok:
All is done for this version. @LornaSIL, ptal. :)

LornaSIL:
The only change I see that should be made is regarding the language tags. @DavidLRowe and I feel like we should continue to add the script code unless the IANA subtag registry says to suppress it. So, for all of these, the code should be change in keyboard_info and in .kps to:
```
rad-Latn
kpm-Latn
cmo-Latn
jra-Latn
bdq-Latn
```

DavidLRowe:
Unless @mcdurdin has different advice.

LornaSIL:
> Unless @mcdurdin has different advice.

Yes, and if that's the case, there is already a keyboard with `jra-Latn` and so in order to share lexical models it would need the same tag probably?

And `cmo` also uses Khmer and we haven't determined whether Latn or Khmr are the default, so it needs to be `cmo-Latn` even if you don't change the other tags.

DavidLRowe:
Lexical models support multiple BCP 47 tags, so it is possible to specify both `jra` and jra-Latn` so that (IIUC) the lexical model can connect with a keyboard using either tag.

mcdurdin:
I was told that the IANA BCP 47 subtag registry suppress-script data is quite old and very much obsoleted by CLDR data. We now use data sourced from CLDR (via langtags.json) so that should be more authoritative than the BCP 47 suppress-script data.

However, for better compatibility with earlier versions of Keyman, it may be helpful to keep using the script tags for now... For version 14 and later, the lack of a script tag, where the langtags.json data provides a default, should not cause any trouble.

mcdurdin:
> Lexical models support multiple BCP 47 tags, so it is possible to specify both `jra` and jra-Latn` so that (IIUC) the lexical model can connect with a keyboard using either tag.

Yes

MakaraSok:
@LornaSIL Technical issue aside, the user prefers not to see the (Latin) bit in the keyboard name on the space bar of keyboard on phone. This info is not critical to them because they found it unnecessary to be told that the keyboard is in Latin script given the fact it's self evidential when one see the keyboard layout.

I'll put the `Latn` back in. Note that the recommendation notified in the keyboard compiler is:

```
dega.kps: Keyboard with ID 'dega' has a BCP 47 warning: 'rad-Latn' is a valid tag but is not canonical: it should be 'rad'.
dega.kps: Keyboard with ID 'dega' has a BCP 47 warning: 'kpm-Latn' is a valid tag but is not canonical: it should be 'kpm'.
dega.kps: Keyboard with ID 'dega' has a BCP 47 warning: 'jra-Latn' is a valid tag but is not canonical: it should be 'jra'.
dega.kps: Keyboard with ID 'dega' has a BCP 47 warning: 'bdq-Latn' is a valid tag but is not canonical: it should be 'bdq'.
```

mcdurdin:
> the user prefers not to see the (Latin) bit in the keyboard name on the space bar

This is an issue in Keyman, not the keyboard. Let's get it fixed in Keyman: https://github.com/keymanapp/keyman/issues/949 (vote for this issue!)

MakaraSok:
@LornaSIL @DavidLRowe Let me know if anything else need to be updated. Thanks.

LornaSIL (approved):
LGTM

PR #1558, opened 2021-05-24
Add devanagari common keyboard

Raji1205:
We have developed a keyboard, 'Common Devanagari'. Please review and let us know if you have any comments/corrections. 
Thank you.


keyman-server:

LornaSIL:
Thanks for your submission! I haven't done a thorough review yet, but there are some things that need changing first.

* The keyboard foldername and filenames must match. In this PR there is a foldername of `devanagari_common` and all the filenames are `common_devanagari` (the order is reversed).
   * The easiest thing for you to do would be to change the foldername `devanagari_common` to `common_devanagari`. However, if you do that...you must also move the folder to the `c` folder (from `d`).
   * If you want it to stay in the `d` folder, then you should change all the filenames to  `devanagari_common`. This would mean you would need to change all the internal references in the .kpj, .kps, and .kmn files.
* Both the readme.htm and the welcome.htm files have a line that should be deleted. It starts with `<script>...</script>`
* The `Help` folder must be renamed to be lowercase: `help`. This folder is used for the online help system.
* The file in the `help` folder must be renamed from `win_hindi_winscript.php` to `common_devanagari.php` (or whatever name you end up choosing to use for the folder/filenames.

I'll let you fix these issues before I do any more review.

LornaSIL:
I've been testing a few things and I believe the main changes are with the documentation:
* For the welcome.htm
  * Please add a `<style>` right after the title.
  * For `jx/w` you should remove the `/`
  * For `MM` I'm getting two of that character, not just one. I think the documentation should just list `M`?
  * The two rules to get fullstop and ellipses ("|\`" and "|\`\`" will never fire because `|` turns into a closing quote. Either the documentation needs changing or the rules need modifying.
  * `Ralt + o` does not produce anything. Either the documentation needs changing or the rules need modifying. (The uppercase one works, it is the lowercase that does not)
  * All the documentation says "Ralt" but the keyboard rules (in the .kmn) actually just say `ALT`. I would personally change your rules to use `RALT`. This is because some Windows programs use "Alt" for certain sequences. If you make it RALT that allows the user to keep using the Windows keys. Also, I believe the RALT will then work as the AltGr key on macos.
* The BCP-47 tags in `keyboard_info` and in the `.kps` file must match. Personally, I think it should just be `hi` (not `hi-IN` nor `hi-Deva-IN`)
* In the help file, I don't think you need any of the style information. I think the standard styling will work for what you have. 
   * Delete from ` <style>` down through the `</h1>` (the header for the php gives the title automatically). 
   * Delete the final ` </body>` tag
   * You only include one of the images. Is that intentional? You could choose to not include any of the images and let our system automatically generate the layout. If you want to do that, this would be the syntax to use:
```
</div>
```
(This will only work if you changed the rules in the .kmn to use the RALT keys.)
 
I think that's about it. It looks pretty good.

LornaSIL:
It looks like you deleted all the ALT keys. I assume that was intentional. If so, 
- you'll need to regenerate the graphics for both the welcome file and the help file
- and delete the graphics no longer needed
- delete the reference to the graphic in the help and welcome file

LornaSIL (approved):
LGTM.
I changed a few things like copyright dates to beginning of string and langtag.

PR #1561, opened 2021-06-02
[cook islands maori] updated touch layout

victoriaq22:
updated touch layout and made a few small fixes to layout.  I'm keeping this in experimental for now as I still need to hear back from my on-site contacts.  I hope that's ok!  Thanks so much


keyman-server:

DavidLRowe:
cim/source/Thumbs.db -- needs to be removed from the project. (This is a hidden file that Windows uses to cache thumbnails of images in the folder.)

In the cim.kmn file, the second occurrence of
```
begin Unicode > use(main)

group(main) using keys
```
(down with the added logic for full stop on mobile) should be removed. This seems to be causing the rules below that point to be ignored.

Currently the keyboard only produces dk(007e) and dk(0060), so the rules in group(deadkeys) that deal with dk(00b4), dk(005e) and dk(00a8) will never be used (and could be removed if you're sure they won't be needed).

victoriaq22:
took out the thumbs.db file, and the extraneous rules/repetitive code

DavidLRowe:
The touch layout generates vowel + macron, but the desktop layout generates vowel + grave and vowel + tilde. Is this intentional? We really prefer that the two layouts have a similar repertoire of characters.

victoriaq22:
Ok, I think I fixed it.  Thanks for catching that!

DavidLRowe:
Some compiled files (cim.js, cim.kmx, sim.kvk) have slipped into the source folder. They need to be removed from the source folder. 

How do you type macron over aeiouAEIOU on the desktop keyboard? We'd still like the desktop and touch layouts to be able to type the same set of characters.

victoriaq22:
Deleted the extra files.  You type the macron using alt-vowel or shift-alt-vowel, so there aren't any deadkeys.  I did run into an issue actually with some operating systems using shift-alt to cycle between keyboards, so I'm waiting to hear back on if those are actually needed in this language, or what the best alternative would be.  I'll bump that message and try to fix it asap

DavidLRowe (approved):
Sorry for the confusion. I was still looking for dead keys.

DavidLRowe:
I'll approve this as is. If you find you need to change something, you can make a new version.

One option would be to pick a key (hyphen, for example) and type that after a vowel to get the macron. (If hyphen is used in the language, that would not be a good choice, so equals might be better.) This would mean typing two characters (a-) but that might be nearly as easy as Alt+a.

victoriaq22:
sounds good, thank you!

PR #1567, opened 2021-06-11
[basic_kbdsn1] Add tablet layout

### LINKED ISSUE #1565: add touch layout for Sinhala Basic keyboard
Original post: https://community.software.sil.org/t/bluetooth-keyboard-on-a-samsung-tab-10-obo-user/4838/8

LornaSIL:
Closes #1565


DavidLRowe:
On the shift layer there are two keys labeled "ZWJ" that produce U+00A0 (no-break space). (On default layer there are two keys in corresponding positions labeled "ZWJ" that do indeed produce U+200D.) These keys on the shift layer of the desktop layout are blank. Perhaps they could be blank here also? Or "NBSP"? But it seems misleading to have them labeled "ZWJ" when they produce a different character.

DavidLRowe (approved):

PR #1573, opened 2021-06-21
across all 4 Arabic keyboards: several additions and changes

rmlockwood:
Some keyboards didn't have the zal or hamza letters. These were added.
Left and right single/double quotations added for the desktop these are consistently mapped to ctrl/shift--/+
En-dash, RTL, and LTR characters added. Some keyboards had the marks, but now all have them consistently on ctrl/shift-,/.
Instead of 'abc' on the touch keyboards use اب‌ج for returning to the default layer
Use the new ZWNJ default key to get the new image. For desktop onscreen, I added the bitmap of the iOS image for ZWNJ.
Increased the font size of touch keyboards to 1.4em.


DavidLRowe:
(I think I managed to keep the right comments with the right keyboards!)

balochi_inpage
- .kps file has version 1.1 (rather than 1.4)
- image for Left Control has "ZWJ" on space bar, but space produces space, not ZWJ (either image or the .kmn file should be changed).
- touch layout: nothing to indicate that Shift+space produces U+200C.
- minor point: would you want to change the heading in welcome.htm from "Left Shift Control" to "Left Control Shifted" to both match the previous headings and also reduce ambiguity?

balochi_urdu
- .kps file has version 1.1 (rather than 1.4)

FYI: You are not required to put copyright, license or version history information in the welcome.htm or readme.htm files. Many people prefer not to include that information in these files in order to reduce the number of files that need to be changed when a keyboard is updated.

rmlockwood:
Thanks David. I’ve fixed these things and committed them.

keyman-server:

DavidLRowe (approved):

PR #1513, opened 2021-04-11
GFF Amharic v2.1

dyacob:
This update applies CSS based key coloring to the touch layouts.  The full effect is only available on iOS and phone and tablet platforms. Minor fixes were made to some touch layout mappings.

PR #1575, opened 2021-06-22
GFF Gurage v0.8.1

dyacob:
The only change to the keyboard file is the VERSION string is set to '10.0' from '14.0', which is erroneous and caused an issue on macOs (discussed in: https://github.com/keymanapp/keyboards/pull/1513#issuecomment-818269381).  The fix corresponds to the reported issue: https://github.com/keymanapp/keyman/issues/5316

I had made this error earlier with gff_amharic , I had actually updated gff_gurage evern earlier but did not submit it, then forgot about the VERSION string.

### DIFF
--- a/release/gff/gff_gurage/HISTORY.md
+++ b/release/gff/gff_gurage/HISTORY.md
@@ -1,5 +1,8 @@
 # ጉራጊና (Gurage) Change History
 
+## 2021-06-22 0.8.1
+* Change of language version from '14.0' (non-existent) to '10.0'.
+
 ## 2021-06-09 0.8
 * Update for Ethiopic Extended-B block from Unicode 14.
 * The Gurage Wookianos font added from the "Gurage Font Collection".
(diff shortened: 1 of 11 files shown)


keyman-server:

LornaSIL:
@dyacob you'll need to remove this statement from both `keyboard_info` files:
`  "minKeymanVersion": "14.0",`
I think just remove the line. It's not really helpful to include it.

LornaSIL:
I went ahead and edited your files. Let's see if it works.

dyacob:
Sorry about this, I missed an update and got out of synch.  I've committed some last changes that remove reference to the "Zebidar" name which was the experiment, pre-Unicode 14, experimental keyboard and font.

LornaSIL (approved):
LGTM. Is it ready to merge or do you still have changes?

dyacob:
@LornaSIL , I have no more changes to commit, all done from here :-)

PR #1578, opened 2021-06-25
[indigenous_nt] Update to 4.0.1

ind-nt:
I had to revert the vowels with diacritics to their NFC on the touch layout, as otherwise the standard iOS wasn't displaying them correctly.


keyman-server:

LornaSIL (approved):
LGTM

PR #1586, opened 2021-07-01
[charis_kbds] Update multiple keyboards to use new version of Charis SIL (new filenames)

LornaSIL:
basic_kbdgn
embera_north
haroi
sil_bari
sil_busa
sil_eastern_congo
sil_indic_roman
sil_ipa
sil_moore
sil_sahu


DavidLRowe:
sil_indic_roman.kmn seems to be UTF-16, but that doesn't seem to be a problem except when viewing diffs of the file.

I'll approve this as is. You can decide whether to merge or to update that file first.

DavidLRowe (approved):

DavidLRowe (approved):

PR #1589, opened 2021-07-05
obolo_chwerty_1.2.0

katelem24:


keyman-server:

LornaSIL:
Before I start checking these keyboards, I wanted to let you know that we cannot distribute the Tahoma font in our keyboard packages. This would be a violation of the Tahoma license.

* You can have the keyboard request to use the Tahoma font which is probably already on most computers, but we cannot include it in the packages.
* OR, you can use another font that supports the Obolo requirements, such as Charis SIL which is in the `release\shared\sil\charis` folder

LornaSIL:
Somehow the obolo_qwerty keyboard now has 3 `welcome.htm` files!
The only one you should have is the one in the `obolo_qwerty\source\welcome` folder.

Since you have made so many changes, it would be best to close this PR and submit each keyboard in a separate PR. At the same time, fix the issue of removing the Tahoma font from the packages and this welcome.htm issue.

Thanks.

katelem24:
Ok

katelem24:
@LornaSIL 
Do you accept Roboto font? It uses Apache License 2.0.
More details and license are here: https://fonts.google.com/specimen/Roboto?category=Sans+Serif&subset=latin-ext

PR #1592, opened 2021-07-06
[cdga6001] Updated CDGA fonts to 6.001

LornaSIL:
Updating fonts to use newest version.
Deleted old folders for `abyssinica_sil` and `charis_sil`.


DavidLRowe:
looks like we missed libtralo

DavidLRowe (approved):

PR #1597, opened 2021-07-15
[wakhi_standard] New keyboard submitted on behalf of author

LornaSIL:


DavidLRowe:
I'll approve this as is, though I'll add some comments that may be helpful for a later version of the keyboard.

For desktop keyboard, there are a number of uppercase combinations that are not present: capital schwa, acute accent over a,e,i,o,u,schwa. Including these in the keyboard doesn't mean that people have to use them, but it does allow that option if (for example) someone wants to type a heading which contains a schwa in CAPITAL LETTERS.

Similarly for the touch layout, perhaps you would want to add the uppercase for:
Schwa, U-bar
acute accent over A,E,I,O,U,Schwa,U-bar
T-dotunder

I'm curious about the inclusion of the Cyrillic character U+044B. The documentation lists ";/" for ы́, but doesn't mention ";u" for ы. There are no uppercase versions: uppercase-Ы (U+042B) and uppercase-Ы acute, either on the desktop or touch versions.

DavidLRowe (approved):
Approved as is. If you want to make any changes, dismiss the review and I can look at it again.

LornaSIL:
* I updated this although I don't know if these combinations are required. 
* Because of the rule for ;/ to produce yeru and not being an obvious way to type the capital in the same way I added a rule for accessing it with deadkey ;u followed by grave.
* I'm not sure what `T-dotunder` means as the uppercase is already on the touch layout.

DavidLRowe (approved):

PR #1599, opened 2021-07-16
[naijatype] Initial Commit of Naijatype

rowbory:
Previous attempt had just created a file instead of the folder for reasons I don't understand. I'm trying again.


keyman-server:

LornaSIL:
Thanks for your keyboard submission! There are some structural issues that need fixing as well as a few other things. I can't really test the keyboard until these are fixed.

* The `naijatype.kpj` file should be in the root of `naijatype` (not in source). Then it will need to point to the .kmn and .kps that are in `source`.
   * When you move and modify the .kpj file, you'll need to go to **Project / Project Settings** and add this for the **Output path**: `$PROJECTPATH\build`  
   * You have two different .kps files associated in the .kpj. Remove the `david_rowbory` one.
* Also, in the root of `naijatype` should be 4 more files: README.md, LICENSE.md, HISTORY.md and `naijatype.keyboard_info`. You can see the sample files in `keyboards\release\template`. The contents of `naijatype.keyboard_info` should be:
```
{
    "license": "mit",
    "languages": [
        "ha",
        "png",
        "ahs",
        "en",
        "fuv",
        "yes",
        "ldb"
    ],
    "description": "This keyboard attempts to help you to type any letters used in Nigerian orthographies."
```
* In order to submit something to the Keyman repo the license must be under the MIT license (hence the need for LICENSE.md). So, you will need to modify your .kmn file to remove the statement `Free to distribute under licence CC-BY-SA`
* In older versions of Keyman we used this filename `naijatype-layout.js` but now we are using `naijatype.keyman-touch-layout`.
   * You should rename the file and then update the name in your .kmn file. There shouldn't be any negative ramifications to just changing the name.
   * As far as I can tell, you can delete this file: `NgUnified.kvk`
* In the .kps file list you can remove the `kmp.inf` file. That is automatically added by the Keyman build system.
   * Also delete that file from the Pull Request.
   * Also, the .kmx and .js files should be removed from the list of files and then added back in from the `naijatype\build` folder. (But don't add the actual .kmx and .js files to this Pull Request. Those compiled files will be automatically generated on the Keyman build system.)
* I would like to see the `welcome.xhtml` file saved to be a regular html file and named `welcome.htm`. Then add it to the file list in the .kps file. If you change it to be .htm then the Keyman package will automatically display the welcome.htm file. If you leave it as html or xhtml it will not automatically display the file. 

I know this seems like a lot, especially for a keyboard going into the experimental section, but it won't build without most of these basic changes. If you feel it's too complicated, I can take your files and resubmit them for you.

rowbory:
Many thanks for the feedback. I'm trying again.

LornaSIL:
It's looking much better. Here are a few more things that need changing. 

* In HISTORY.md, please put the keyboard name where the underlines are.
* `naijatype.keyboard_info` requires a closing `}` on the last line.
* `naijatype.kpj` needs the buildpath to change still. It must be `$PROJECTPATH\build` (remove the LexicalModels stuff. The build all happens within this folder. The lexical stuff happens separately.
* The .kps calls for this file `..\promotional\naijatype_400.jpg`. That is not in the Pull Request. Either remove that name from the .kps or add the .jpg into the repo.
* This isn't required, but the welcome.htm file doesn't discuss the way to type with CTRL and ALT. I couldn't figure out how to type a semi-colon (since it's a special key) without looking at the code.
* In general, we prefer NOT to explain to people how to install the keyboard (How to install it) whether Desktop, Mac, Android, IPhone, etc. It's much better to have people to go to the Keyman site so they get the most up-to-date instructions. Since this is in experimental, I'll leave it for now, but it needs to be removed in the future from the welcome.htm and the welcome.odt and pdf.

By the way, when you make these changes, there is no reason to bump the version. We only need to bump it when the keyboard is changed in the future.

Otherwise, it looks like I can approve it with these changes.

We will need a online help file, but I can add that once the keyboard is online. I would base it on what's in `naijatype-help.htm`

rowbory:
Thanks for the direction. I'll come back to finish that later today or tomorrow, I hope.

rowbory:
OK, I've had another go, I think dealing with all the issues you mentioned. 
* History changed
* keyboard_info done
* .kps graphic is correctly included in source/icons_pictures now
* welcome.htm mentions how to use ALT to get semicolon.
* Some unnecessary files used during development removed.

LornaSIL (approved):
LGTM.

PR #1598, opened 2021-07-16
[masaram_gondi] Add files via upload

hinditutorpoint:
added itrans masaram gondi keyboard layout

PR #1600, opened 2021-07-16
[masaram_gondi] added masaram gondi keyboard layout

hinditutorpoint:
added itrans masaram gondi keyboard layout


keyman-server:

LornaSIL:
I haven't reviewed the keyboard yet, but 
* it looks like you will need to add .jpg files into the welcome folder and commit those.
* The font is being pulled from your own downloads folder. I would recommend you create an `assets` folder in your root of the keyboard: `masaram_gondi\assets` and put the font in there. Then you will need to update the pointer to the font in the .kps. (Do make sure the license for the font allows redistribution).
* If this is considered an ITRANS layout, would it be worth changing the keyboard name to "Masaram Gondi (ITRANS)"? That would allow for another "Masaram Gondi" layout in the future. Just something to think about.

LornaSIL:
I'm confused. Is this Pull Request the correct one to review or is #1598? Please close the invalid one. Thanks!

LornaSIL:
If it's easier to close this and then recommit, that's fine. Or you could just submit changes.

PR #1602, opened 2021-07-16
CIM touch layout updates

victoriaq22:
fixed a nextlayer issue; goes with an updated lexical model


keyman-server:

LornaSIL:
Remember the compiled files (kmx, kvk, js) should not be committed. You just want the source files (kmn, kvks, keyman-touch-layout) should be committed.

victoriaq22:
Ah sorry about that - thank you!

DavidLRowe:
It seems that you removed the `"nextlayer": "shift"` from the saltillo (U+A78C) for the default and shift layers, but not the numeric and symbol layers? Was that an oversight? or intentional?

At any rate, I'll approve this as is and you can decide whether you need to make any changes.

DavidLRowe (approved):

PR #1605, opened 2021-07-18
feat: xmllint .kps files

mcdurdin:
This adds xmllint checks for .kps file to the build. If xmllint is not on the path, then the XMLLINT variable can be set to point to it; if no xmllint can be found then .kps file structure validation will be skipped (not a failure).

Found minor errors in one .kps file (apart from balochi_phonetic, which is already resolved in a separate branch).

Relates to keymanapp/keyman#5464 and uses .xsd from keymanapp/api.keyman.com#158. Do not merge until api.keyman.com PR is approved (as we may need to propagate changes for the .xsd to this branch).

Note: I will update the build agents to ensure xmllint is available on them too.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -3,8 +3,8 @@
 #
 # This script is built with commands available to Git Bash on Windows. (mingw32)
 #
-# This script and scripts in resources/ should be identical in the 
-# https://github.com/keymanapp/keyboards and 
+# This script and scripts in resources/ should be identical in the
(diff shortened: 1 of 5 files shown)


darcywong00 (approved):
lgtm

> Note: I will update the build agents to ensure xmllint is available on them too.

Will that just be on the Windows ba's?

mcdurdin:
> Will that just be on the Windows ba's?

Yes. Done.

PR #1606, opened 2021-07-20
GFF Amharic v2.2

dyacob:
This pull request contains updates to reference version 2.1 of the Abyssinica SIL font, it adds the TypeTuned "Abyssinica SIL (Connected)" font, and adds Unicode 14 Gurage letters to the non-amharic layer of the tablet layout.


keyman-server:

DavidLRowe (approved):

PR #1608, opened 2021-07-22
GFF Ge'ez v1.6

dyacob:
This update adds a mobile and tablet keyboard.


keyman-server:

DavidLRowe:
I'm not that familiar with this script, so perhaps I've misunderstood how things should work, but here are some (non-exhaustive) comments: 

default layer: `?` key (longpress on `!` key)  produces U+1367. The code for that key is K_SLASH, so it inherits the output of the desktop keyboard K_SLASH. (Selecting "?" again turns the U+1367 into "?".) Using U_003F as the code would avoid that.

extra-geez layer: `1` key (longpress on `፩` key) doesn't produce "1". That key on the touch layout has the code "K_1" so it inherits the output of the desktop keyboard K_1. You may want to change the label to "U_0031" to output "1" directly. Similar for other keys on this layer (and on the punctuation layer).

extra-geez layer: `?` key (longpress on `።` key) produces U+1367. See note above.

punctuation layer: the rightmost longpress key on `'` displays `’’` (looks like U+201D, but is actually U+2019 doubled) but produces U+2019 since the code is U_2019. It seems that the displayed character should be changed.

punctuation-2 layer: "?" produces U+1367. See note above.

dyacob:
Thanks @DavidLRowe , thanks for the thoroughness that caught these issues.  I believe the fixes are now in place and the keyboard is ready for another round of testing.

DavidLRowe:
I think the only one left is on the extra-geez layer: "?" longpress key on "።" produces U+1367 rather than "?".

dyacob:
Thanks @DavidLRowe , the new commit fixes "?" on the "።" longpress.

DavidLRowe (approved):

PR #1611, opened 2021-07-23
obolo_qwerty_1.1

katelem24:


keyman-server:

LornaSIL:
I should have commented on this for your previous chwerty keyboard. It concerns me that you are adding rules for tilde n:
`Ññ (to be used in special cases such as in presentations where the software doesn't support N̄n̄, though Ññ is not part of the Obolo orthography)`

This feels like a recipe for really confusing users and they may end up having a really messy orthography with some people using ñ and some n̄. 

Anyway, that's just a concern I have for your model, but it doesn't keep us from approving the keyboard.

I am glad you are adding a help file. I had not realized we were missing that. It does need to be renamed from `qwerty.php` to `obolo_qwerty.php` or it will not show up.

Please add your copyright years to LICENSE.md. We've decided this is the most important place for the years to be located. The other places can be gradually deleted. So, for LICENSE.md it should be:
`© 2020 - 2021 Rogers Katelem Edeh`
All other places we can gradually (as we think of it) remove the years and just have `© Rogers Katelem Edeh`
That will reduce having to update files every time.

katelem24:
I've updated the LICENSE and HELP files.

For the issue of Ññ:
I do presentations for my church during Sunday services and other services within the week. Some presentation softwares do not render combining diacritics well. In some softwares, the combining macron mark usually cut through the N as if it ₦ with one stroke, while n-macron will appear as "n–" with the combining macron displaced to the top right of the n. I never knew that will happen until when the service had started. A similar thing happens to M̀m̀ and Ọ̀ọ̀. That can be really confusing, so Ññ was brought in to help the situation, but for the other two letters, there are no alternatives. It is not an attempt to confuse users.

Again, N̄n̄ is really rare to come by in mobile devices and desktop. It seems only Obolo and Yoruba are using n-macron in Nigeria while every other language is using n-tilde. So people have already been using Ññ, but they are always reminded that Obolo uses N̄n̄.

I may remove that statement in the next version, but the problem will still persist since those characters are not available in UNICODE.

LornaSIL (approved):
LGTM

katelem24:
Thank you.

PR #1612, opened 2021-07-23
[basic_kbdrum] Revert desktop rules to what MS has

### LINKED ISSUE #1610: [basic_kbdrum] Fix "c" "C" logic to match source
Hello,

("c" = U+0421. Mapped to s S).

Regarding the Russian Mnemonic keyboard uploaded via #858, the logic with the "c" is not exactly as in Windows 10. I've looked at the code but not entirely sure how to fix it, hope you can help me.

Context:
- When I click on the key for the "c", both keyboards use it as a deadkey which awaits further input. This is a feature of many of the keyboards which have more characters than the latin alphabet.

Logic present on both systems:
- If the next pressed key is one of the pre-configured ones, the first touch is not registered and the second provides the actual character (e.g. s S + c C = щ Щ)

Logic present on Windows 10:
- If the next pressed key is NOT one of the pre-configured ones, the first touch is not registered and the second provides "c" and the other character (e.g. s S + d D = сд СД)

Logic present on the basic_kbdrum
- If the next pressed key is NOT one of the pre-configured ones, the first touch is not registered and the second provides "c" and the other character (e.g. s S + d D = д Д)

Best,

Hugo

LornaSIL:
Fixes #1610

### DIFF
--- a/release/basic/basic_kbdrum/HISTORY.md
+++ b/release/basic/basic_kbdrum/HISTORY.md
@@ -1,6 +1,10 @@
 Russian - Mnemonic Basic Change History
 ====================
 
+1.1 (2021-07-23)
+----------------
+* Revert desktop deadkey behavior to Microsoft's
+
 1.0 (2019-08-13)
 ----------------
 * Imported from Win10 by Makara SOK using the import tool
--- a/release/basic/basic_kbdrum/LICENSE.md
+++ b/release/basic/basic_kbdrum/LICENSE.md
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-© 2019 SIL International
+© 2019-2021 SIL International
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
--- a/release/basic/basic_kbdrum/README.md
+++ b/release/basic/basic_kbdrum/README.md
@@ -1,9 +1,9 @@
 Russian - Mnemonic Basic keyboard
 ==============
 
-© 2019 SIL International
+© SIL International
 
-Version 1.0
+Version 1.1
 
 Description
 -----------
@@ -28,8 +28,3 @@ Supported Platforms
  * Desktop devices
  * Tablet devices
 
-To do
------
-* Fix typing mistakes in welcome.htm and basic_kbdrum.php
-  * "This keyboard uses five deadkeys, such as:" > "This keyboard uses nine deadkeys, on the following keycaps:"
-  * "Mnemonic Basic Basic" > "Mnemonic Basic" (3 places in welcome.htm)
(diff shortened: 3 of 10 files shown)


DavidLRowe:
The keyboard seems to do what is documented:
- typing s<space> (с<space>) gives с
- typing sh (cх) gives ш

(along with upper case combinations)

I'm not sure if it addresses issue #1610 which seems to be talking about undocumented combinations. For the example given in the issue, on this keyboard:
- typing sd (сд) gives д

whereas (IIUC) the author of the issue would expect сд

So then the question is do we follow Microsoft in this behavior or not. But maybe I'm totally misunderstanding the issue.

At any rate, I'll approve this as is, since it seems to be working as designed. If you want to make further changes, dismiss the review.

DavidLRowe (dismissed):

hugoncosta:
@DavidLRowe that is correct, currently the logic is close to the Windows 10 one. The issue is that in order to have "c", we have to click s + spacebar, while on windows, if you click any key that isn't part of the pre-configured pairs, you get both "c" and the other key. Let me give you an example:
- Windows 10 - спасибо - keys pressed: s p a s i b o 
- kbdrum - спасибо - keys pressed: s spacebar p a s spacebar i b o
- kbdrum using W10 keys - паибо (if spacebar is not pressed after s, the "c" gets ignored)

I've looked through the code and not sure this is reflected, I'll try to compile it and test it today.

DavidLRowe:
@mcdurdin If you have the time, can you comment? The change Lorna made was (IIUC) from a Keyman Developer import of the MS keyboard in question. The issue raised in #1610 is that with the Windows keyboard, the deadkey on the "s" key produces "с" when followed by any other than the few combinations listed, whereas with the Keyman version, one needs to type the "s" key and space to get the "с" character. So, we need some direction on what to do about the proposed change. Thanks!

mcdurdin:
I don't think the proposed changes resolve the core issue. It seems to me that instead of these changes, you'd want to have a default match scenario for the deadkeys. One way to do this may be to use `notany` to generate a default result for U+0441:

```
dk(0441) notany(dkf0441) > U+0441 context(2)
```

And similarly for the other combinations.

But I think the changes currently made in this PR may break the double-deadkey approach (typing <kbd>s</kbd><kbd>c</kbd> would generate `dk(0441) dk(0446)` which should match the rule `dk(0441) any(dkf0441) > index(dkt0441, 2)` in `group(deadkeys)` to `U+0449` ("щ").

So I would revert those changes.

LornaSIL:
@mcdurdin we could close this PR easily 

However, there is obviously something wrong with the Keyman import of Windows keyboards. Is that something that can be improved upon?

mcdurdin:
> However, there is obviously something wrong with the Keyman import of Windows keyboards. Is that something that can be improved upon?

Clearly something we could improve -- I'm not sure exactly what the best approach is, whether what I have suggested is sufficient, or whether we'll need something more sophisticated. I can open an issue but don't have a timeframe for resolution.

DavidLRowe (approved):

PR #1616, opened 2021-07-27
Devanagari common

Raji1205:
We have done some changes in existing keyboard. Please give reviews/comments. Thank you


keyman-server:

LornaSIL (approved):
LGTM. You can see I modified a few files, so make sure to do a pull to get the latest versions.

PR #1619, opened 2021-07-28
lao pali and lao_pali_us keyboards submitted

jmdurdin:
Two keyboards allowing full unicode 13 range for Lao to be entered using (extended) Lao typewriter layout and adapted phonetic layout (following english key sounds).


keyman-server:

mcdurdin:
Congratulations on the PR and thank you for your contribution!

jmdurdin:
Dear Marc,

Thank you for expediting the acceptance of the two keyboards!

But I have another question that I would like to ask your help with – it won’t take long, I promise!

I have been testing my variable fonts on different platforms and devices with regard to the behaviour of the mark  and  mkmk  features and they mostly behave exactly as intended – though not when installed on Windows, but that isn’t surprising and I believe is generally known.  Firefox or any chromium-based browser correctly renders the text on a web-page with the embedded font using ***@***.***).

But when I test these fonts on my iPhone (with Safari or any other browser, which appears to use the same rendering), the mark feature is processed correctly but mkmk  seems to be ignored.  Exactly the same behaviour is observed when I install a variable font on macOS and look at any of the font instances created by macOS with TextEdit, Pages, etc.

The problem I have is that neither my iPhone 6 nor my macMini can be upgraded to the current versions, so I do not know whether or not this behaviour remains the same with current versions.  If it does, I want to report it to FontLab to be sure there isn’t something quirky about their compilation of those tables, as Ben Mitchell has written that they have not had any problems with those features when compiled with Glyphs.

So what I would like to ask you to do is to check a link showing a sample text on my website with Safari on an iPhone that uses the current version of iOS and/or on a mac running macOS 11, and send me a couple of screenshots so that I can check the rendering.

The link is https://laoscript.net/sengbuhan/ which needs to have the Variable font option selected:
***@***.***
and the text that is most helpful to look at is towards the bottom of the page:
***@***.***
especially the syllable at the right hand side of the Mai kan, mai kong adjustment examples:
***@***.***
as that shows how the two marks should be centred over a wide base consonant.

If you do have a few minutes more (unlikely!) and can install the attached Sengbuhan font on a mac running macOS 11, it would be great to know also whether the text on the above web page when copied into pages or Text Edit and set to the installed (Regular or any other weight) SengBuhan font is now rendered correctly or not.

When the mark feature works but not mkmk, that last syllable will appear more like:
***@***.***
with the tone mark not positioned correctly with respect to the MAI KAN beneath it.

If neither feature is being processed correctly, it will probably look like:
***@***.***
with both marks in their default positions (towards the rhs of the syllable).

Thank you!

Love from Papa

DavidLRowe:
The protocol for copyright statements is evolving. Our current thinking is to include it in the .kmn file and the LICENSE.md file and omit it from all the other files (or include a copyright statement with a name but no date, so that it doesn't need to be updated when the keyboard is revised). I see both 2019 and 2021 as dates, so perhaps you could include in the LICENSE.md and the .kmn file: 
(c) 2019-2021 John Durdin 

It would be helpful to have a description in the .keyboard_info file. Perhaps you could take something from readme.htm, such as:
"The keyboard uses the official Lao standard layout, adapted to support entry of the Lao Pali and Lao Sanskrit characters included in Unicode 13."

The welcome.htm file needs to explicitly reference the lao_pali_*.png files (as you have done in the lao_pali.php file).

In the package source (.kps):
- lao_pali.ico does not need to be included (since this information will have been compiled into the .kmx file), so the .ico file should be removed from the list of files
- The BCP 47 tag should be "lo" (rather than "lo-Laoo"). It should match the tag in the .keyboard_info file.

DavidLRowe:
I may have more comments once I get a chance to test the keyboard.

DavidLRowe:
The above comments refer to lao_pali. I forgot that there were two keyboards in this PR.

DavidLRowe:
Comments for lao_pali_us:
- There is a "setup.log" file that should be removed.
- It would be helpful to change the description in the README.md file and the .keyboard_info file. You could use text from readme.htm: `The Lao Pali (US) keyboard follows the Lao Original (Lao Pali) key layout for easier typing for users familiar with the standard US English keyboard.`  (If you retain the `"` characters from the original, you'll need to escape them `\"`.)
- The welcome.htm file needs to reference the two .png files in order to display them (as is done in the lao_pali_us.php file).

DavidLRowe:
The lao_pali_us keyboard has two touch layouts (tablet and phone) that differ only in the "fontsize" specification. The lao_pali keyboard also has two nearly identical touch layouts, though the leftmost key in the top row of the shift layer has `"nextlayer": "default"`. You may have better information from an inside source :), but, since Keyman will use a phone layout on a tablet if there is no tablet layout and vice versa, we generally recommend that keyboards not have both tablet and phone layouts unless they are intended to be different (for example, if the tablet layout has 5 rows and the phone layout 4 rows).

jmdurdin:
Thanks for all the feedback.  I will resubmit the keyboards after dealing with each point.

However, the different font-sizes for tablet/phone layouts was by design.  The issue is that with stacked diacritics, it is difficult to display characters on the keycap without truncating the top or bottom of the glyphs used.  The alternative is to use a much smaller font-size, but that makes the difference between some characters too small to tell them apart.  This is partly why I use a specially created font, as with a font intended for general use it is not possible to achieve a satisfactory compromise between character size and distinctness.  But I will revisit the question, and if it is possible to eliminate one or the other, I will do so.

DavidLRowe:
For lao_pali keyboard, on the OSK, for the "e" key, the display has U+0EBF, which is an unassigned code point. The key actually produces U+0EB3, so it seems as though the OSK should change to use U+0EB3. 

Other than that, all looks fine (though I am not familiar with this script).

For the lao_pali_us.kps has the same issues as the lao_pali.kps: 
- the .ico file should be removed from the list of files
- the tag should be "lo" rather than "lo-Laoo"

jmdurdin:
That is by design, and applies to both keyboards (on different keys).

U+0EB3 does not render correctly by itself - it is treated as a mark by the system, but is more like a ligature in behaviour, as it is split apart by shaping engines and reordered with respect to any preceding tone marks.  It is one of the biggest difficulties when working with Lao (or Thai) script fonts.  This is another reason why a special OSK font is used.  If U+0EB3 is used for the keycap, the missing base glyph will be inserted, making the keycap text much too wide.  The alternative method would be to use an image, but images on keycaps scale badly and do not match the remaining keycap (text) labels.

jmdurdin:
David, I do not understand your comment:

The welcome.htm file needs to explicitly reference the lao_pali_*.png files (as you have done in the lao_pali.php file).
As far as I can see, the referencing to the images is identical in welcome.htm and lao_pali.php, which is hardly surprising, as the content for lao_pali.php was taken from welcome.htm and minimally changed for use as a php (help) file.

Can you please clarify what it is you think needs to be changed, and how?  For both files, the images are in the same folder as the file referencing them, so a relative reference seemed appropriate, but I do not know how the php help files are actually used.

At present two copies of each set of images is included, as that was the model keyboard package that I based the current package design on.  It seems to me that only one set of images should be needed, with the references updated accordingly, but I would need to know exactly where the images should be stored and referenced if so.

Thanks,
John Durdin

jmdurdin:
In the php files for the two keyboards, the descriptive text is appended after the closing ?> php tag, since I understood that to simply copy the html unmodified.  (It does in the php files on my sites.)  Are you advising that this text needs to be included *within* the <?php and ?> tags, e.g. within hereto or some similar way of embedding static html within php?

Thanks.

From: John Durdin
Sent: Tuesday, 3 August 2021 1:39 PM
To: keymanapp/keyboards ***@***.***>
Subject: RE: [keymanapp/keyboards] lao pali and lao_pali_us keyboards submitted (#1619)

David, I do not understand your comment:

The welcome.htm file needs to explicitly reference the lao_pali_*.png files (as you have done in the lao_pali.php file).
As far as I can see, the referencing to the images is identical in welcome.htm and lao_pali.php, which is hardly surprising, as the content for lao_pali.php was taken from welcome.htm and minimally changed for use as a php (help) file.

Can you please clarify what it is you think needs to be changed, and how?  For both files, the images are in the same folder as the file referencing them, so a relative reference seemed appropriate, but I do not know how the php help files are actually used.

At present two copies of each set of images is included, as that was the model keyboard package that I based the current package design on.  It seems to me that only one set of images should be needed, with the references updated accordingly, but I would need to know exactly where the images should be stored and referenced if so.

Thanks,
John Durdin

jmdurdin:
The README.md file appears to have been generated automatically, but I can find nothing about what it is for, what it should contain, or how it can be updated if any of the settings in it are changed.  Why is another description needed when the description is already included in readme.htm, welcome.htm and lao_pali.php?

jmdurdin:
The lao_pali keyboard also has two nearly identical touch layouts, though the leftmost key in the top row of the shift layer has "nextlayer": "default".

That was an oversight and has been corrected.  I have now removed the tablet layer from both keyboards and am testing to see whether the compromise scaling works well enough on both phone and tablet.

Thanks

jmdurdin:
Still not clear about references to png in welcome/lao_pali.php. All other changes now made as suggested, deleting tablet touch layer in both kbds.  Will upload corrected keyboard folders to github as soon as that file reference issue is clarified and fixed.

DavidLRowe:
I tried on a Windows machine with three different browsers, and when I launch welcome.htm, a new page opens in the browser, but without any images displayed. I guess I had assumed that the images were not being referenced, but I see now that they are referenced in the .htm file. I'm not sure what to say, since it must work for you.

LornaSIL:
I just downloaded lao_pali and compiled it. When I installed the .kmp the welcome file displayed with the images for me.

I admit I was a little surprised it works because of this part of the path `./` in the image src. The graphics and the welcome file will be in the same folder when the package is deployed, so that part of the source path seems unnecessary. That is, I would have thought `src="lao_pali_S.png"` rather than this `src="./lao_pali_S.png"` would be best, but as I said, it is working for me on Win10.

DavidLRowe:
Great! Sounds as though you've made all the needed corrections. When you're happy with the touch layout, let us know when the keyboard is ready to review. 

The description in the README.md is for the benefit of people looking at the keyboard repository.

mcdurdin:
The `./` prefix for referencing images within .htm and .php files should make no difference -- they should work identically whether or not the prefix is present.

LornaSIL (approved):
@DavidLRowe is traveling, but he said to go ahead and approve this once you made your changes to the touch layout.

PR #1621, opened 2021-07-29
[uma_phonetic] Initial submission of uma language keyboard

jackrittenberry:


keyman-server:

DavidLRowe:
Thanks for your submission to the keyboards repository. We will need some additional information as described on this page:
https://help.keyman.com/developer/keyboards/advanced/step-0

If this isn't enough information or you have additional questions, please let us know.

LornaSIL:
I'm sorry neither of us have had a chance to review your keyboard yet. It looks like you've made good progress. I have two immediate comments that should be fixed and then we can look at it.
* The folder and filenames need to match. Since most of your filenames are `uma_phonetic` you should change the folder from `uma` to `uma_phonetic` and any files that are `uma_ph` should also be changed to `uma_phonetic`. 
* Secondly, you have a .kpj file in the root of the `uma` folder and another in the `source` folder. The .kpj file must be in the root of `uma_phonetic`. So, delete the other. Then make sure the .kpj file points to the .kps and the .kmn in the `source` folder.

This isn't just me being picky, our on-line build system needs it this way. Once these are in place one of us can do a more thorough review.

PR #1627, opened 2021-08-04
Version 4.0.2 Small update to touch layout functionality

ind-nt:


keyman-server:

DavidLRowe:
I think you need to add `"nextlayer": "default"` to all the longpress keys on the shift layer, so that the longpress keys on the shift layer will also return you to the default layer. For example, currently selecting `E` will return you to the default layer, but selecting `Ē` (longpress key on `E`) will not.

Optional change to HISTORY.md: It might be clearer if you change the text for this version from "All layers" to "All alphabetic layers", but it's not a problem if you want to leave it as is.

ind-nt:
Thanks! I've added the nextlayers to the popup caps, and added the detail to the History.md as suggested.

DavidLRowe:
Almost there! It seems that you added the `nextlayer` line to rows 2 and 3 of the keyboard, but not row 1 (ERTYUIO).

ind-nt:
Thanks for checking. May have not saved when I did them all originally.

I also realized the numeric and symbols pages needs the same default functionality. Hopefully, it is good to go now.

DavidLRowe (approved):

PR #1644, opened 2021-08-20
Update sil_euro_latin.keyman-touch-layout to add Anglo-Saxon support.

mx-ash:
Adding support for Anglo-Saxon letters. Added Æ, æ with macron. Added Ȝ, ȝ (yogh). Other letters used in Old English already supported by the layout, so not many updates needed.


keyman-server:

PR #1645, opened 2021-08-20
Add support for Anglo-Saxon

mx-ash:
Added Old English letters Æ, æ with macron (Ǣǣ) and yogh (Ȝ, ȝ) (g\j to type)


keyman-server:

PR #1646, opened 2021-08-21
[sil_euro_latin] add support for Æ, æ with macron.

mx-ash:
Added support for AE ligature with macron. Type Æ as usual and follow with '-' to type.


keyman-server:

LornaSIL:
Thanks for your submission. Your changes mostly look reasonable.
* We do require a version bump for any change in keyboards. That means you'll need to update the HISTORY.md file as well as the .kmn file. I believe 1.9.5 would the new version.
* For the touch layout U+021C is on the longpress for "Z", but U+021D is not on "z". If you feel "G" is more appropriate you can change it to that.
*

mx-ash:
> 
> 
> Thanks for your submission. Your changes mostly look reasonable.
> 
>     * We do require a version bump for any change in keyboards. That means you'll need to update the HISTORY.md file as well as the .kmn file. I believe 1.9.5 would the new version.
> 
>     * For the touch layout U+021C is on the longpress for "Z", but U+021D is not on "z". If you feel "G" is more appropriate you can change it to that.
> 
>     * Also, you should add the language tag "ang-Latn" to the .kps and the keyboard_info file.

Added the tags as requested and checked the touch layout; I believe the letter you're referring to is an ezh ʒ, which looks similar to yogh in some fonts but is a distinct character. Yogh is rounded at the top. Have bumped the version to 1.9.5 accordingly.

LornaSIL (approved):
LGTM. Thanks for your update!

PR #1656, opened 2021-08-30
[sahaptin_umatilla] Initial submission of Umatilla Sahaptin keyboard

jonathangearyCTUIR:
I tried this once before and failed, I think because I had omitted two files by mistake.


keyman-server:

LornaSIL:
Be aware that the build *will* "fail", so don't close it because of that. One of us will check it.

jonathangearyCTUIR:
Ah, my apologies! Thanks for your help, I’ll leave it alone now while you check it.
Best,
Jonathan

LornaSIL:
Thanks for your submission! Here is some initial feedback.
* When I open the .kpj file there are two .kmn files associated. Please remove the second one. The first one is in the `source` folder which is correct. The second seems to be looking in the root of `sahaptin_umatilla`. Click on **Options** / **Remove from Project**
* HOWEVER, the `sahaptin_umatilla.kmn` file that is in the source folder has no rules! So, I suspect the good .kmn file is in the root. If that is true, it should be copied to `source` and then commit that version to this Pull Request.

I'll need that done before we check the keyboard. 

However, some general feedback that will need doing also:
* LICENSE.md needs the copyright year added after the copyright symbol. It's really the only place we are requiring it now. `© 2021 Confederated Tribes of the Umatilla Indian Reservation`
* A description of the keyboard could be added in several places: sahaptin_umatillia.keyboard_info, README.md and readme.htm in particular. Right now it just says "sahaptin_umatilla generated from template". You might want to say "This keyboard is designed for the Umatilla language of Oregon." (or whatever makes this keyboard distinctive).
* You have an "extras" folder that has the keyboard layout in it. 
    * I would move `Sahaptin keyboard layout.png` to `source` and delete the `extras` folder. 
    * Then, rename the graphic file to `sahaptin_layout.png` so that it follows file naming conventions (in particular the spaces are a problem). 
    * Then I would edit the welcome.htm file, add your description, then where it says `<!-- Insert Keyboard Layout Images or HTML here -->` I would add this line:

```
	<p><a href="sahaptin_layout.png"><img class="keyboard" src="sahaptin_layout.png" /></a></p>
```


    * THEN, I would go to the .kps file and in the "Files" tab I would add  `sahaptin_layout.png`. That way it will be added to the keyboard package and people can see the layout.
* Ideally, I'd like to see the name of the keyboard be "Sahaptin Umatilla" (or something like that) rather than just the file naming. If you want to do that, you would need to change the keyboard name in the .kmn file (under **Details / Name**) and in the .kps file (under **Details / Package Name**).

jonathangearyCTUIR:
K̓ʷałá! Thanks, Lorna! I have edited the files based on your feedback and resubmitted them. Please let me know whether there is anything else that I need to do at this time.

LornaSIL:
Thanks for doing all that. It is looking quite good. I have a few more comments:
* The desktop keyboard does not produce upper case to łɨƛ and yet you have the uppercase on the touch layout. Is that something that is needed in the language? If so, you may want to figure out how to add them to the desktop keyboard.
   * If they are not commonly used, you might add them to `[RALT K_BKSLASH]`, etc.
* On the **touch layout**, **numeric** layer, **Shift** key, the `Next Layer` is set to `(none)`. You should change that to `shift`.
* For any of our keyboards in the `release` part of the repo, we require an online help file.
   * In the `source` folder you should create a `help` folder.
   * In the `help` folder you should add the attached file called `sahaptin_umatilla.php`. You'll need to unzip this file as it wouldn't let me attach the .php file.
   * This file is primarily from your `welcome.htm` file, but with different headings. Also, the layout for the desktop and the touch layout will be autogenerated. 
   * Then, commit the new file to your Pull Request.
  
[sahaptin_umatilla_php.zip](https://github.com/keymanapp/keyboards/files/7092447/sahaptin_umatilla_php.zip)

jonathangearyCTUIR:
Thanks, Lorna! Your feedback was extremely helpful!
 
I added uppercase versions of “ł”, “ɨ”, “ƛ” (we don’t actually use a different symbol for “uppercase ƛ”, so these symbols are the same) as [ALT… ] keys to the desktop keyboard. My coworkers and I hopefully will be revisiting this layout soon to make those letters more accessible (i.e. mapped to [SHIFT…] keys) and to make some other changes to the desktop and touchpad layouts.
 
I’ve made both of the other changes that you mentioned and have committed the new files. Please let me know if there is anything else that I should do, and thank you again for your help!

PR #1658, opened 2021-08-31
[uma_phonetic] update on behalf of author

DavidLRowe:


LornaSIL:
Looks good...except there is something wrong with the help file. See the online version: https://help.keyman.com/keyboard/uma_phonetic/3/uma_phonetic that has the issue as well.

I suspect the graphic is probably too big.

I think it would be best to use the osk syntax so it creates the layout automatically rather than relying on the graphic.

That would mean deleting the graphic in the `help` folder and editing the php file to add something like:
```
</div>

<h2>Mobile/Tablet Keyboard Layout</h2>
</div>
```

LornaSIL (approved):
LGTM

@jackrittenberry You should make sure to pull `master` so you get all these changes in your repo before you modify the keyboard again. Hopefully you won't have to modify it after these changes.

PR #1660, opened 2021-09-02
[masaram_gondi] initial release

hinditutorpoint:
please add this keyboard


DavidLRowe:
Thanks for this submission. There are a few changes needed.

LICENSE.md 
This file needs a date after the (c) and before your name. (Recently we changed from requiring the date in all copyright statement, to just needing it in the LICENSE.md file, so any other copyright statements, such as in the .kmn file can omit the date.)

It would be nice to have a proper description in the .keyboard_info file (which is how people can find the keyboard), and the README.md file. The description in the readme.htm is good, or even something as simple as: "This keyboard is for typing the Gondi language in the Masarm Gondi script."

The .kpj.user file needs to be removed. (I was able to do this since you permitted edits from the maintainers.)

There are two copies of welcome.htm. You need to remove the one in the source folder, and then change the package list (.kps file) to use the source/welcome/welcome.htm (rather than the source/welcome.htm file that you just deleted).

In the .kmn file, I tried to access the link for &KMW_HELPTEXT (https://gondi.bwiit.in/keyboard), but evidently it can't be reached without a password.

I was somewhat surprised to find the keyboard producing Latin characters (qzQWEOZ). Would you rather change the .kmn file to have those keys produce nothing (and also show a blank key in the On-Sreen Keyboard)?

The shift key on the touch layout, default layer doesn't take you to the shift layer.

Let me know if you need more information about fixing any of these. Thank!

hinditutorpoint (commented):
remove duplicate welcome.htm and fixed help keyboard links

hinditutorpoint (commented):
added some querty key QqEzZ and other fixed

LornaSIL:
@hinditutorpoint do you still want this keyboard added? If so, can you make the final changes or would you like me to make the final changes requested on your behalf? Otherwise, we could close this PR.

LornaSIL:
@hinditutorpoint your keyboard is now available online: https://keyman.com/keyboards/masaram_gondi

PR #1665, opened 2021-09-06
Taigi POJ v1.4: ⁿ only after vowels

guyemerson:
With the previous versions, it is impossible to type the syllable "nng", because "nn" is converted to "ⁿ".  This update only allows "ⁿ" to be typed after a vowel.  For example:

"nng5" -> "nn̂g"
"inn7" -> "īⁿ"

### DIFF
--- a/experimental/t/taigi_poj/HISTORY.md
+++ b/experimental/t/taigi_poj/HISTORY.md
@@ -1,6 +1,11 @@
 PhahTaigi POJ Change History
 ====================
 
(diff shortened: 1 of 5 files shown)


keyman-server:

DavidLRowe:
We now only require the date in the copyright statement in the LICENSE.md file, so rather than change 2020 to 2020-2021 in the README.md file, I just removed the date.

DavidLRowe (approved):

Issue #1653, opened 2021-08-30
Shan (Sil) Keyboard Can't install on IOS


SZTSIIT:
https://keyman.com/keyboards/sil_shan?bcp47=shn-mymr

After Shan Sil updated can't install on iOS. It doesn't show up when searching for it in the Keyman app. I believe it doesn't work on android as well. 

It said it doesn't support this device. Please see the screenshot here. 

![A86D0134-6B1A-47E1-916B-A6E13AD6B41E](https://user-images.githubusercontent.com/69423986/131280395-28e1d5e5-ee11-4703-a68f-7363f368ebe8.jpeg)

LornaSIL:
`sil_shan` has never had support for mobile devices. I think when you submitted the first issue you were talking about both `shan` and `sil_shan` and we ended up focusing on `sil_shan`. See if the `shan` mobile layout would work for you. 

If you do need `sil_shan` to have a mobile layout, is the desktop layout sufficient or would you like to give us an idea of what the layout should look like?
One strategy I've used is to take the main Burmese characters (circled in red) and put them on the default and shift layers, and then take the digits and punctuation and put them on a 3rd `numeric` layer. That way the keys on a smart phone don't end up too small. 

However, if you want something right away it would maybe need to be a tablet layout that looks exactly like the desktop layout.

![sil_shanD](https://user-images.githubusercontent.com/7071907/131352107-d2930c50-798a-4629-a4ce-97cfcf5c6863.jpg)

LornaSIL:
There's now a tablet layout available here: https://keyman.com/keyboards/sil_shan

SZTSIIT:
@LornaSIL  Thank You, It is the same as what I was about to give you.

SZTSIIT:
@LornaSIL If it is easy to restack I think this is more user-friendly. If not no worry.
![Uploading 27B02578-0C76-4791-886E-68734A4350B2.png…]()

SZTSIIT:
![E990AD64-2018-453E-885E-F870C4B48154](https://user-images.githubusercontent.com/69423986/134079878-2ef4d145-3cfa-4592-b7da-9d05bb3c2f0d.png)
![D3E96237-9F33-411A-8E93-07659C597EEC](https://user-images.githubusercontent.com/69423986/134079879-fd010fae-4bf2-4e11-b830-edd18072507b.png)

LornaSIL:
[...]

PR #1669, opened 2021-09-20
[add_tablet] sil_shan and mandaic_phonetic

LornaSIL:


LornaSIL:
Closes #1653

DavidLRowe:
sil_shan
- HISTORY.md - add tablet (rather than "table") layout
- touch layout: K_OE2 key on default and shift layers doesn't produce output. Perhaps remove the key?

mandiac_phonetic
- seems OK as is, though K_OE2 key on default and shift layers doesn't add anything new.

DavidLRowe (approved):

PR #1675, opened 2021-09-25
Ge’ez-Berhan Keyboard v1.0 Submission on Behalf of the Author

dyacob:
This is the first release of an Ethiopic keyboard that has the goal of being general-purpose and covering all defined Ethiopic symbols under the Unicode standard (up to Unicode 6.0). This PR is made on behalf of the keyboard author Father Berhanemeskel M. Embaye.


keyman-server:

LornaSIL:
I will start doing some review, but be aware that there is a file conflict with gff_geez.kpj. Maybe you accidentally committed a change?

LornaSIL:
There are a few issues that I see immediately.
* We need a `geezbrhan.keyboard_info` file in the `geezbrhan` folder. Make sure to add all the language tags that are in the .kps file.
* The `geezbrhan.kps` file should be moved to the source folder and then the references to files will need changing.
* My preference for the Word file would be to make a `geezbrhan\assets` folder and put the Word file in there rather than source.
* The Word file (at the end) also have a wrong copyright statement. I believe it needs changing from `© Ge’ez Frontier Foundation, 2009-2020.` 
* At this point we've changed how we require dates for copyrights. We require the year(s) to be listed in the LICENSE.md file (as you've done). We don't require the year anywhere else so that all these individual files don't have to be updated every time. Personally, I still like to have the date in the .kmn and .kps so that when the .kmp is installed the user can see the full copyright including year.
* Is there a reason the keyboard will only work on Windows? If there is no technical reason it wouldn't also work on macOS and Linux, it might be nice to tick those settings in the .kmn.
* In the .kmn file, line 75 is a 2nd version statement. Please delete that one. Also, line 76 has another header statement (message) that should be moved to the top section. I personally prefer all those header statements to be kept together so they aren't lost down in the stores and rules.
* In the .kps file:
   * the welcome.htm file isn't included. I think you should include the welcome.htm or else delete it from the .kps.
   * We would prefer that you also include the other files in the `abyssinica` font folder: AbyssinicaSIL_README.txt, OFL.txt, FONTLOG.txt, OFL-FAQ.txt

I will test the keyboard once these changes are made.

dyacob:
Thanks @LornaSIL I'll be able to address these in a couple of days.  The gff_geez.kpj inclusion is definitely unexpected.  I'll ever to the master branch version on my next update.

LornaSIL:
In testing the keyboard, I'm not sure if it's the keyboard or the documentation that needs enhancing. I'll just mention the inconsistencies and you figure it out:
* page 2, section B and C. Several of the keystrokes don't produce what is documented, for instance `Rje`, `Wje`, `mA`
* page 3, section E. punctuation. Many/most of these seem different. Double-check this whole section.
* page 3, section F. `0o` does not produce the character in the chart.
* In the Ge'ez Typing Table the `mI` produces a different character.
*

PR #1677, opened 2021-09-28
[gff_amharic] Revert pdf help files

darcywong00:
Addresses missing links to pdf files in keymanapp/help.keyman.com#450

#1606 removed pdf help files for gff_amharic. The keyboard help page still refers to them
https://help.keyman.com/keyboard/gff_amharic/2.2/gff_amharic

So this PR reverts the pdf files back to the help/ folder.

No change to version 2.2. of the keyboard.


DavidLRowe (approved):

dyacob:
Sorry about that.  The same files were also residing in the "source" directory and I thought the "help" ones were duplicates. I'll be careful to store them in the "help" folder only going forward.

LornaSIL:
We tended to duplicate files before (in both help and welcome) which took up an unnecessary amount of space. But, if it's needed in the help file that's really the only place it needs to be because you can reference the help folder in the .kps (you just can't test the link very well when it's not in the package).

PR #1678, opened 2021-09-28
chore: fix broken links

mcdurdin:
This only fixes broken links as reported in https://github.com/keymanapp/help.keyman.com/issues/450

I note that the sil_madi help appears to have diverged from the welcome but this PR does not attempt to resolve that.


LornaSIL (approved):
LGTM

PR #1679, opened 2021-10-06
[old_hungarian] Update to change labels for RLM and LRM and add ZWJ

### LINKED ISSUE #1676: Required ZWJ key for Old Hungarian lajout
The ZWJ key is required for the Old Hungarian texts, for example Noto-fonts project have ~100 ligatures.
I will to do it.

LornaSIL:
@istszab

Fixes #1676


istszab:
@LornaSIL What "Review required" means, What I need to do. My English is bad.

PR #1686, opened 2021-11-05
[mon_anonta] New keyboard on behalf of author

LornaSIL:
There's also a lexical model that goes with this. I'm not sure if I should attempt to submit it and what the next steps should be.


DavidLRowe:
AFAICT the touch layout is fine. (One thought would be to trim the predictive text suggestions from the top of the images used in documenting the touch layout, but I can think of arguments to keep them, so I'm fine with leaving them.)

For the desktop layout, Shift+] shows "}" on the OSK, but actually produces U+1021 U+1034.
Also, Shift+. shows nothing on the OSK, but produces ">".

DavidLRowe (approved):

PR #1687, opened 2021-11-09
[mon_phonetic] New keyboard on behalf of author

LornaSIL:


DavidLRowe:
For desktop layout, shift layer
"~" key shows "@", but produces U+100E
Shift+4 shows diacritic(U+1034 ?), but produces "$"
"}" key shows "}", but produces U+1034
Shift+. shows blank, but produces ">"

touch layout seems okay

When I test the help file locally, the keyboard images are huge and only show the left half of the keyboard. But maybe that's a problem with my local setup.

DavidLRowe:
Good catch on the .keyboard_info file!

DavidLRowe (approved):

PR #1688, opened 2021-11-15
Update mon_phonetic.kps

Anontamon:
Change Name


keyman-server:

LornaSIL:
I'm not sure what the purpose of this change is. We cannot change the url. It looks like you are trying to put the keyboard name in the url field on the .kps.

PR #1689, opened 2021-11-16
[khmer_angkor] update rules and add regression test

MakaraSok:
* update rules to account for clusters with HA and exception with BA
* add robat rule
* add regression tests


LornaSIL (approved):
LGTM

PR #1705, opened 2021-12-09
took away underlying letters for the Control versions of the keyboard

rmlockwood:
This has to do with the png images only.


keyman-server:

DavidLRowe (approved):

PR #1706, opened 2021-12-11
feat: externally hosted keyboards

mcdurdin:
See documentation in docs/ext-hosted-keyboards.md

Related:
* [Design document](https://docs.google.com/document/d/1g2oR3xHMvJPIOKOI_gir-AjGd1LA2SiluPlS8EuFDNE/edit#)
* keymanapp/lexical-models#150


experimental/e/external_experiment_3/.gitignore
```diff
@@ -0,0 +1,8 @@
+
+# Only specific files may be added to repository for external references
+*
+.source_is_binary
```
rc-swag, line 4: quoting the readme.MD 
"If the file `external_source` is present, only three files are permitted in the
folder:"

resources/compile.sh
```diff
@@ -131,6 +131,29 @@ function build_keyboard {
...
+
+  if [ -f external_source ]; then
+    if [ -d "build/" ]; then
+      # We want to make sure we rebuild or get correct binaries
+      rm -rf "build/"
+    fi
+
+    retrieve_external_keyboard || die "unable to retrieve external keyboard $base_keyboard"
+
+    if [ -f .source_is_binary ]; then
```
rc-swag, line 147: Oh ok, I think that `externally-hosted-keyboards.md` should be updated to explain the optional use of `.source_is_binary`

resources/compile.sh
```diff
@@ -198,7 +221,9 @@ function build_keyboard {
     ./build.sh $FLAG_SILENT $FLAG_CLEAN $FLAG_DEBUG "$kpj" $FLAG_TARGET "$PROJECT_TARGET" || die "Custom build script failed with an error"
   else
     # We will use the standard build based on the group
-    if [[ $group == release ]] || [[ $group == experimental ]] || [[ $KEYBOARDS_STARTER == 1 ]]; then
+    # Externally sourced keyboards (see above) may have the .source_is_binary flag,
+    # in which case they are trated identically to legacy keyboards
```
rc-swag, line 225:
```suggestion
    # in which case they are treated identically to legacy keyboards
```

resources/external.sh
```diff
@@ -0,0 +1,135 @@
+#!/usr/bin/env bash
+
+#----------------------------------------------------------------------------------------
+# Source keyboard from external repository or binary download location
+#----------------------------------------------------------------------------------------
+
+is_external_source_keyboard() {
+  # https://github.com/(owner)/(repo)/tree/(commit)/(path?)
+  [[ $1 =~ ^https:\/\/github.com\/([^\/]+)\/([^\/]+)\/tree\/([a-z0-9]{40})(\/.+)?$ ]]
```
darcywong00, line 9: Should we check owner/repo aren't `keymanapp/keyboards`?

resources/external.sh
```diff
@@ -0,0 +1,135 @@
...
+
+  rewrite_external_git_ignore
+}
+
+verify_external_target_folder_is_clean() {
+  # Remove all files except:
+  # .gitignore
+  # external_source
+  # README_EXTERNAL.md
+  local files=`find * \! -name '.gitignore' -a \! -name 'external_source' -a \! -name 'README_EXTERNAL.md' -print`
```
darcywong00, line 53: Do we allow the optional file `.source_is_binary`?

experimental/e/external_experiment_3/.gitignore
mcdurdin: Yes, am updating the docs accordingly. Note that this is an ignored temporary build artifact.

resources/compile.sh
mcdurdin: OK

resources/external.sh
mcdurdin: While it would be dumb to do this, technically it would work... The source url will be checked during PR review so I think it's not necessary to add this restriction. I'm curious though, what was your concern here?

resources/external.sh
mcdurdin: Good question. We certainly could, because it is a build artifact and will be created and/or deleted as required. I will add that file to the list.

resources/external.sh
darcywong00: No concern. Just asking since the function is checking if a keyboard is "external".

rc-swag (approved):
lgtm

mcdurdin:
Note: I want to add a sha-256 hash for externally hosted files for reliability and safety, so moving this back to draft.

docs/externally-hosted-keyboards.md
```diff
@@ -66,13 +66,15 @@ github.com.
 
 
 The file `external_source` must contain one line for each file to be downloaded,
-in a `file=url` format. For example:
+in a `file url sha256` format. For example:
```
darcywong00, line 69: How does a keyboard developer get/generate the sha256 value?

resources/external.sh
```diff
@@ -1,5 +1,8 @@
 #!/usr/bin/env bash
 
+#set -u
+#set -e
```
darcywong00, line 4: Are you planning to eventually have these flags set?

docs/externally-hosted-keyboards.md
mcdurdin: I've intentionally left this out of the document, because we'll only accept these from orgs with a contract with us. I used sha256sum but there are plenty of ways to get the sha256 of a file.

resources/external.sh
mcdurdin: Not here -- I should probably remove these I guess... they were just for testing purposes. This is an source included file (perhaps should rename to .inc.sh?) and it's best not to set global flags in an included file, rather leave that to the script that includes it.

rc-swag (approved):
lgtm

darcywong00 (approved):
lgtm

PR #1707, opened 2021-12-16
Desktop keyboard for fv_dane_zaa_zaage

caforbes:
Added a desktop keyboard to the Dane-Z̲aa Z̲áágéʔ project -- approved by @madebybridget (I'm a new FirstVoices team member).

Merged together some old/new project files in producing this so hopefully everything works together. I'm specifically not sure about the role of the `source/help/(...).php` file versus `source/welcome.htm` and `source/readme.htm`, and what should be in the project file.


keyman-server:

LornaSIL:
Thanks for your submission! I'm glad you are joining the team!

The readme.htm file shows up when the package is installed:
image
Then, the welcome.htm file is displayed after you click `install`.

The .php file is used for the online help: https://help.keyman.com/keyboard/fv_dane_zaa_zaage and does not need to be included in the package.
* Since you've added the desktop layout, I'd recommend you add these lines to the .php file (either before or after the tablet layout lines). This will give the desktop layout for people who are looking at the help file.
```
</div>
```

Regarding the touch layout:
* We currently recommend that if both the tablet and phone layout are the same (which these appear to be), you should delete one of the layouts. Then it will work on either tablet or phone. This allows you to only have to make changes in one place and is easier to not make mistakes.
* On the `shift` layer for both tablet and phone it might be good to change the shift key from `Special` to `Special (active)`.
* On the `numeric` layer for both tablet and phone the `shift` key should have `next layer` changed from `none` to `shift`.
* If you keep the tablet layout, you can leave the .php file "as-is". If you choose to keep the phone layout and not the tablet layout, you'll need to change the .php file to say `osk-phone` instead of `osk-tablet`.

caforbes:
Thanks for the tips! I've made the requested changes.

LornaSIL (approved):
LGTM

PR #1710, opened 2021-12-22
feat(khmer_angkor): adds custom OSK font for better render

mcdurdin:
Relates to keymanapp/keyman#5986.

This swaps the on screen keyboard / touch layout keyboard font for a custom font called Khmer Busra Kbd which resolves issues with shaping across platforms by including custom PUA glyphs for any characters which require shaping.

This is not yet complete:
- [x] Three characters are missing from the font: U+2248, U+17C7, U+17C8.
- [x] The font is not yet optimised for size.
- [x] I have only updated the OSK and tablet, not yet the phone layout.

Still, this should be good for some testing, and will pick this up again post-holidays.

Sample images:

image

image

image


MakaraSok:
1. check in the Windows Character Map to verify that nothing is left out.
> Single combinable characters (non-private use) such as U+17B6-U+17C6, U+17C9-U+17D2 and U+17DD are shown with an additional dotted circle before them in the character map app, i.e.

image

the multiple combinable vowels and coengs are not shown so though, i.e.
imageimage

For coeng ឡ, we might want to use the one that's better rendered as:
image

Instead of this unexpected one:
image

We might want to add these vowel combinations for future use, if not now.
image

**Is there any way to mark special characters, such as: ZWSP, ZWNJ, ZWJ visually on the OSK?**

2. What's shown on the OSK matches the output
* on Windows 10 21H1:
  - Default - OK
  - Shift - OK
  - AltGr - A key is not presented correctly. It's shown as a box with question mark rather than the “≈” (U+2248) <br/>
  image <br/>
  - AltGr+Shift - OK

* on Linux - Ubuntu 20.04.3 LTS (x64)
  - two issues are experiences in general (on all layers): (1) some keys are not show correctly and (2) others output a different character (see the screenshot below):
    - Default 
    image <br/>
    - Shift 
   image<br/>
    - AltGr 
    image<br/>
    - AltGr+Shift  
    image <br/>

_Note that layout on Ubuntu is kind of European layout based rather than US based._

-----
* on macOS - **pending...**
* on Android - **pending...**
* on iOS - **pending...**
* on Web - **pending...**

mcdurdin:
> on Linux - Ubuntu 20.04.3 LTS (x64)
> 
> * two issues are experiences in general (on all layers): (1) some keys are not show correctly and (2) others output a different character (see the screenshot below):

This suggests that we are not able to set the font for Onboard. @ermshiperete is that true?

I will follow up with Didi on the remaining characters for the OSK font.

mcdurdin:
> **Is there any way to mark special characters, such as: ZWSP, ZWNJ, ZWJ visually on the OSK?**

See https://github.com/silnrsi/font-keymanweb-osk/issues/3

ermshiperete:
Can I download that keyboard from somewhere, or how do I test this?

mcdurdin:
> Can I download that keyboard from somewhere, or how do I test this?

I've built and uploaded the latest version here: https://drive.google.com/file/d/1gTvQe1wbJyzl4Wfo1M_eVoFap0xXQgfq/view?usp=sharing

This is not the final version!

ermshiperete:
> This suggests that we are not able to set the font for Onboard. @ermshiperete is that true?

yes, that's true. It's not possible to set a different font for a specific keyboard.

mcdurdin:
> yes, that's true. It's not possible to set a different font for a specific keyboard.

Hmm. this is a significant blocker for this keyboard and several others. How hard would it be to patch Onboard to support specifying a font?

ermshiperete:
> Hmm. this is a significant blocker for this keyboard and several others. How hard would it be to patch Onboard to support specifying a font?

I guess that would be several days of work. Onboard currently supports only one font, defined in the theme (and a fallback font). There is no keyboard-specific font option. Also, in onboard-keyman we currently use ldml for the keyboard layout which doesn't provide a way to specify a font, so we'd have to store that information in a different place.

mcdurdin:
> on Linux - Ubuntu 20.04.3 LTS (x64)
> 
> * two issues are experiences in general (on all layers): (1) some keys are not show correctly and (2) others output a different character (see the screenshot below):

Given that this is a limitation in Keyman for Linux (specifically, Onboard), we don't need to continue testing this there until that is resolved. See keymanapp/keyman#6186 for details.

@MakaraSok are you able to test other platforms?

MakaraSok:
@mcdurdin What build should be used to test this on other platforms? Since this PR is now in draft state, should I wait till it's done?

mcdurdin:
> What build should be used to test this on other platforms?

Any stable version of Keyman should be fine to test this keyboard.

> Since this PR is now in draft state, should I wait till it's done?

I think this is ready to go, just waiting for your testing and approval before we mark it as ready for code review.

MakaraSok:
Tested with the keyboard package attached [above](https://github.com/keymanapp/keyboards/pull/1710#issuecomment-1023591052) against the stable version 14.0.294.

- Windows - the OSK looks good and works as expected when either using the physical or the OSK to type.
  image
  image

- macOS - the OSK looks good and works as expected when using the physical keyboard to type.
  image
  image

- Android - The OSK looks good and works as expected.
  image
- iOS - The OSK looks good, but the keys with two characters do not function as expected. 

  https://user-images.githubusercontent.com/28331388/173313599-f3223087-85b7-48fb-ab3b-085cbda45174.mov

- Web - the OSK doesn't look good, but the keyboard works as expected. 
  image
  image

This was tested by compiling the Khmer Angkor keyboard with the "Khmer Busra Kbd" font selected for OSK and touch layouts.
  image

mcdurdin:
Reference for Khmer Busra Kbd font PUA characters:

image

MakaraSok (commented):
The name is a little confusing as it's `khmer_busra_osk` here and `Khmer Busra Kbd` there. To be specific, the name after the font is installed is `Khmer Busra Kbd`. The font file name itself is `khmer_busra_os`. However, the name doesn't stop the font from working correctly.

mcdurdin:
> * Web - the OSK doesn't look good, but the keyboard works as expected.

I think this was only because the Khmer Busra OSK font was not selected in Keyman Developer's "Character Font" setting?

I realise now that this PR did not have a version number bump so this change will not have deployed...

PR #1714, opened 2022-01-06
[chinuk_wawa] initial submission of chinuk_wawa keyboard

cachang:


keyman-server:

DavidLRowe:
Thanks for this contribution. Generally, this looks pretty good (much better than my first keyboard submission!), but we do need some changes to fit with our automated system and to provide support for keyboard users.

LICENSE.md needs a date (2022) and name in the copyright statement. The name needs to be the name of a person or a legally-recognized organization. (We used to require the date and name in all copyright statements, but have changed to requiring the name everywhere, but the date only in the LICENSE.md file.)

There are several places where a description would be useful. The description can be what you have in the chinuk_wawa.kmn file ("this is a chinuk wawa keyboard") or "a keyboard for typing chinuk wawa" or something similar. 

README.md needs a name in the copyright statement, as well as the description mentioned above.

The chinuk_wawa.keyboard_info file needs the language code ("chn"), as well as the description mentioned above.

In the chinuk_wawa.kmn file, the Targets should be just 'any' (rather than 'any windows macosx linux mobile'), although see comment below. Also, the copyright statement needs the name as mentioned above.

The readme.htm file needs the description (as mentioned above) rather than 'generated from template'

The welcome.htm file needs the description plus some information on using the keyboard. This could be with images of the keyboard (default and shift layers), or could be a list of how to type characters not on the keyboard: for example, "type shift+e to obtain 'ə'.

The chinuk_wawa.kps file seems to be duplicated. The copy in the chinuk_wawa folder should be removed, leaving the copy in the chinuk_wawa/source folder. That file needs to be connected to the project. Copyright information should be updated.

Above I advised changing the Targets to 'any', that is covering both desktop and mobile options, because I saw that you had a chinuk_wawa.keyman-touch-layout file, used to define the mobile layout(s). But in that file I don't see that you've added any means to type the extra characters added to the desktop keyboard (and I see capital letters which don't appear on the desktop keyboard). So, perhaps you are not intending to include a mobile layout, in which case the Targets could be changed to 'desktop web' which would cover Windows, Linux, Mac and use on the web.

If you do intend to include a mobile layout: 
(a) You don't need to have both tablet and phone layouts if they are identical. Keyman will use the phone layout for a tablet if there is no layout for tablet (and vice versa). 
(b) You should include the extra characters (ə ʰ ɬ Ɂ) present on the desktop layout and consider eliminating the shift layer if no capital letters are needed.

Please let me know if anything is not clear, or you need more information.

cachang:
Thank you so much for the help, I'll get working on these updates right away! I couldn't get the Touch layout to work, it shows up as a white blank screen. I tried removing and readding mobile as a target, and manually removing and adding the touch layout feature, but no luck. Do you have any recommendation or place I should go look for help troubleshooting it?

cachang:
And...removing the conflicting targets so that all it has is "any" fixed it. Nevermind!

DavidLRowe:
The copyright statement in the .kmn file needs your name.
In Keyman Developer, click the Keyboards tab at the bottom of the window, then select chinuk_wawa.kmn, Details tab on the left, change Copyright field to include your name.

There are two chinuk_wawa.kps files. You need to delete the one in the chinuk_wawa folder and keep the one in the chinuk_wawa/source folder.

However, the packaging information (.kps file) is not connected to the project. 
In Keyman Developer, click on the chinuk_wawa.kpj file at the top (just under the File menu), then the "Add existing package..." button, then navigate to the chinuk_wawa.kps file. But note (as mentioned above) that there is an extra .kps file that needs to be removed. The file you need to you want is in the chinuk_wawa/source folder (not directly in the chinuk_wawa folder itself).

Once you have the packaging file included in the project, you need to add  the "readme.htm" and "welcome.htm" files to the list of files in the package.
In Keyman Developer, click on the Files tab on the left, the "Add..." button at the bottom of the list and add the "readme.htm" and "welcome.htm" files.

Then click the Details tab on the left, and in the "Readme file:" field, select the "readme.htm" file you just added. You should also add your name to the Copyright field.

For the desktop layout, I see that Shift + b generates "n". That's not a problem, but unexpected. Was that intentional? It's fine to leave it, but if you change it, you'll need to update the OSK (on-screen keyboard) as well.

For the touch layout, none of your longpress keys are working because you are using T_xxxx identifiers, which need a corresponding match in the .kmn file. I'd suggest changing those to use an identifier of U_0259, which will directly generate the Unicode character (without needing a match for T_0259 in the .kmn file). 

Also note that you don't need the "nextlayer:" "default" for keys (including longpress keys) that are on the default layer. It's not wrong to have it, but it isn't necessary.

For the schwa longpress key, do you mean to use U+1D4A "ᵊ" as the "text" value, rather than U+0259 "ə"? I guess it doesn't matter much, since Keyman scales the character to fit the key.

As I mentioned last week, you don't need to maintain both a phone and a tablet layout, if they are identical. It's not wrong to have both layouts identical, but it's unnecessary and is extra work to maintain them. You might consider making a separate tablet layout that uses 11 keys across (instead of the 10 that the phone layout has) then move the longpress keys to their own keys. (So, for example, the top row might be qweərtyuiop, and so on.) Then people using a tablet wouldn't need to use longpress keys at all.

Let me know if you have questions.

cachang:
Thank you for all of the detailed help - this is my first time using Keyman
Developer so I appreciate all the time you are spending giving me tips and
steps to follow. I think I've implemented all of your suggestions, I'm just
stuck on one longpress key- the x key when you longpress it should be a
combination of U+0078 and U+0323, so I'm not sure what to use for the
identifier, would I need to add this to code directly for it to match the
kmn file?

DavidLRowe:
You've addressed most of the issues I raised. I'll repeat a couple items you missed:

---

There are two chinuk_wawa.kps files. You need to delete the one in the chinuk_wawa folder and keep the one in the chinuk_wawa/source folder.

However, the packaging information (.kps file) is not connected to the project.
In Keyman Developer, click on the chinuk_wawa.kpj file at the top (just under the File menu), then the "Add existing package..." button, then navigate to the chinuk_wawa.kps file. But note (as mentioned above) that there is an extra .kps file that needs to be removed. The file you need to you want is in the chinuk_wawa/source folder (not directly in the chinuk_wawa folder itself).

---

In addition, there is one place where you do indeed need a T_xxxx identifier with a corresponding match in the .kmn file. The x with dot below is actually two Unicode characters and currently the U_xxxx notation can only handle one Unicode character. So, in the touch layout for the 'x̣', use T_0078_0323 for the Code value. Then in the .kmn file, replace your current rule (+ [U_0078+0323] > 'x̣') with:
```
+ [T_0078_0323] > 'x̣'
```

(The T_xxxx string can be anything, T_x-dotunder, for example. I chose T_0078_0323, because in a future verison of Keyman Developer, it will be possible to use U_0078_0323 to get the two characters without the need for the T_xxxx rule in the .kmn file.)

I think with those changes, the keyboard should be ready. Thanks for your perseverance!

cachang:
Thank you! I've done the bit about removing the second .kps file and
mapping it to the project, but the second file keeps coming back - is there
a common thing I might be doing to cause that? I'll try deleting it and
connecting the project again, and fix the kmn file and the identifier.

DavidLRowe:
Hmm. You did delete the spurious .kps file, but unfortunately you also deleted the project file (.kpj) that is essential to have. I also see that the build results (.kmn, .js, .kvk, .kmp) are now in the source folder rather than the build folder.

I think you have all the files needed for the keyboard submission, we just need to get them in the proper places.

You may find it easier to close this pull request and open a new one with the files in the right places (rather than trying to rectify the existing PR). Make sure you have all your files in a safe place, create a new branch and put all the files in proper locations. For reference, there are some guides on the Keyman help site for submitting keyboards using the GitHub interface (https://help.keyman.com/developer/keyboards/github/) and using the traditional git process (https://help.keyman.com/developer/keyboards/advanced). But let me know if you need anything else.

cachang:
Thank you - I'll try again with a clean branch, after making the changes
you listed. I've been following the guides, but I'm new to github so it's
been confusing. I'll follow them slower this time.

DavidLRowe:
replaced with #1725

PR #1716, opened 2022-01-14
[ukwuani] Ukwuani keyboard

MakaraSok:
Ukwuani language keyboard is developed to type Ukwuani Language using the right orthography of Ukwuani language. It is Latin based and additional characters added to the US standard Keyboard layout. They can be found in the AltGr (Right Alt) and Shift AltGr (Shift + Right Alt) on computer (i.e. Windows, macOS, Linux and web) and long presses of related keys on touch layers (i.e. Android and iOS devices).


MakaraSok:
Note that the output of longpress keys on Shift layer don't output anything for the time being until this PR https://github.com/keymanapp/keyman/pull/6100 is merged to stable. :)

DavidLRowe:
@MakaraSok Can you ping me (on this PR) when it is ready for review? Thanks!

MakaraSok:
It's ready to be reviewed.

DavidLRowe:
@MakaraSok Everything looks fine, although I wonder why the use of "ukw-Latn-011" in .keyboard_info and .kps files. It seems that "ukw" would be sufficient. (Also, I'm not sure why "011" would be used instead of "NG".)

So, I'm suggesting changing the BCP 47 code to "ukw" (or providing justification for the existing tag). Thanks!

MakaraSok:
The language tag has been changed to "ukw" with the confirmation from the keyboard requester.

DavidLRowe (approved):

PR #1717, opened 2022-01-17
[old_turkic_udw21_qwerty] New keyboard on behalf of both authors

mehmetoguzderin:
First of all, thanks to everyone for maintaining Keyman; it is a great project and provides the helping hand when one needs it with its excellent technical design. Also, we apologize for any further burden caused by our PR.

Our recently published research has developed a Keyman keyboard for a script that does not have representation here, which consistently enables input of the target corpora despite some digitalization challenges, fitting the metrics defined in the paper. Furthermore, we believe it would be helpful to others if this keyboard is part of Keyman's repositories due to ease of finding and the fact that it is a bit scary to maintain our package and promise trust (keyboards can contain JavaScript). Also, this could help reduce potential duplicate work due to centralized visibility.

While developing this keyboard, we have tested quite a few different implementation approaches. Our final design is consistent with paper, allows for input even without full access, and results in the same output as our application/library implementation when considering lowercasing and more. It is simply a QWERTY keyboard based on US English that closely follows the paper.

We use `otk-Orkh` as a tag in this keyboard per the paper, and it also stays consistent with Noto Font's Old Turkish Old Turkic classification, a font which we do not bundle in this keyboard since the keyboard itself does not utilize. We name this keyboard with paper's venue to underline the close relation to the perspective outlined there.

Paper: https://aclanthology.org/2021.udw-1.11
Transliteration Tables: https://github.com/mehmetoguzderin/UD_Old_Turkish/tree/dev/not-to-release#transliteration
Transliteration Implementation App (iOS): https://apps.apple.com/app/ascii-turk/id1601998507
Base Keyboard: https://keyman.com/keyboards/basic_kbdus
Font: https://fonts.google.com/noto/specimen/Noto+Sans+Old+Turkic


keyman-server:

mehmetoguzderin:
@DavidLRowe is there a way we could help with the manual build?

DavidLRowe:
Thanks for your contribution! Currently automatic builds are disabled for outside contributors purely as a security precaution. After the PR is reviewed (sometime in the upcoming week), the build will be manually launched. Once the PR is approved and the build successful, the PR can be merged, which will make the keyboard available for download.

mehmetoguzderin:
@DavidLRowe thank you very much for your clear explanation! I will try to correspond with any deficiencies we might have unfortunately baked into our initial attempt.

LornaSIL:
Thanks for your submission! I do have a few comments.
* There is not a clear description of how to use the keyboard, do you feel the paper is sufficient to describe the keystrokes? If so, perhaps you could link to the paper in the welcome.htm file and the help file. (I realize you have links in the README.md file, but that file is only part of the repository, not part of the keyboard installer).
* You say you are not distributing the font because it's already available in the applications that need it. I would recommend:
   * Provide a link in the readme.htm or welcome.htm file for downloading the font (I realize you have links in the README.md file, but that file is only part of the repository, not part of the keyboard installer).
   * Because the font is **not** available in most phones/tablets, I would include the font in an assets folder `old_turkic_udw21_qwerty\assets` and then include it in the .kps files (and also in the **Keyboards** tab **Keyboard font / Display font**
   * FYI: If someone already has the font installed, this package will not overwrite the font they have installed, so if someone installs a newer version there is no need to worry that this package will overwrite a newer version.
* The touch layout has a CTRL layer that could be deleted. Also, if you do not need the SHIFT layer, that could be deleted.
   * Then, delete the SHIFT and CTRL keys on the default layer. If you do need the CTRL and SHIFT layers, then you should change the "Next layer" on the SHIFT key to go to the `shift` layer.
* In the help file you should delete these lines since there is no phone layer:
```
<div id='osk-phone' data-states='default'>
</div>
```   
If you want to make it clear that your mobile layout works for both Tablet and Phone you can change the line "Tablet Keyboard Layout" to "Mobile Keyboard Layout"

   * Delete the `<h1>Old Turkic UDW21 (QWERTY)</h1>` line in the help file since our automatic system already creates a header line.

* Lastly, in the `keyboard_info` file I would add a reference to the font like this:
```
{
  "license": "mit",
  "description": "Old Turkic UDW21 (QWERTY) is an Old Turkic script keyboard for Old Turkish language per Universal Dependencies for Old Turkish paper at UDW21, based on US English keyboard and aimed at case insensitive ASCII input without special characters using QWERTY to output a subset of Old Turkic Unicode with minimal variables and restricted keyboard access.",
  "languages": { 
       "otk-Orkh": { 
            "font": { 
            "family": "Noto Sans Old Turkic", 
            "source": "NotoSansOldTurkic-Regular.ttf" 
         } 
      }
   }
}
```
I think this is essential for making it display properly in mobile devices.

I think that's it. Since I'm not familiar with the script I didn't spend much time actually testing the keyboard.

mehmetoguzderin:
@LornaSIL Thank you very much for your detailed review and feedback. I tried to address all problematic aspects you mentioned, but I would like to preserve the SHIFT key as it inserts the actual space character when combined with the SPACE key (is it possible to keep the modifier without layer if used in such way?). You are right about the input approach's non-trivialness. Still, given the outlined constraints, we feel like this could intuitively serve people familiar with the script and crucially the language by the paper's description (I did link thanks to your suggestion).

LornaSIL:
>I would like to preserve the SHIFT key as it inserts the actual space character when combined with the SPACE key (is it possible to keep the modifier without layer if used in such way?

That's interesting! I'll try it tomorrow. I don't mind keeping it if it works.
Otherwise, I think everything is in good shape.

mehmetoguzderin:
@LornaSIL Thank you so much, and apologies again for the sudden burden of this PR!

mehmetoguzderin:
@LornaSIL You are right, I was only testing on-screen, which had working Shift + Space, but mobile didn't 🙀! So I replaced the shift key there with an explicit space character. Thank you very much; I should have been more careful...

LornaSIL (approved):
This looks good! Thanks for your submission.

mehmetoguzderin:
@LornaSIL thank you very much!

PR #1719, opened 2022-01-18
[srr_ajami_qwerty] Keyboard for upcoming workshop in February

LornaSIL:


DavidLRowe:
Looks good. The only thing I noticed was that Shift+O and Shift+F produce O and F, even though on the OSK those keys are blank. Ditto for touch layout.

DavidLRowe (approved):

PR #1724, opened 2022-01-25
Naijatype experimental

rowbory:
I've updated some of the key sizes as they weren't consistent between layers and the space bar was a little too small so that people were touching the punctuation key instead of it. Αlso the ? and ! hadn't been generating the right characters on the default layer so tried to correct that.

### DIFF
--- a/experimental/n/naijatype/HISTORY.md
+++ b/experimental/n/naijatype/HISTORY.md
@@ -1,13 +1,19 @@
-Naija Type Keyboard Change History
+Naijatype Keyboard Change History
 =======================
+
+31-33 (23 Nov 2021)
+-----------------
+
+* Fixed ? and ! keys that were not outputting correctly.
+* Updated keyboard version to match Google App version.
 
 30 (21 July 2021)
 -----------------
+
 * Initial commit (more pre-history to dig up later)
 * Updated documentation.
 * Prepared for upload to Keyman site.
 * Bug fixes around capitalisation at end of sentences.
 
 29 (16 July 2021)
------------------
 * Instructional video made.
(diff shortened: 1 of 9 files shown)


keyman-server:

DavidLRowe:
Because this keyboard is in the experimental portion of the repository, we have lower requirements for documentation and support files, however, the one thing that needs to change is the LICENSE.md file.

Typically the copyright statement has a date (year) when the keyboard is first released. If it is updated, the copyright statement has a range from the year of the first release through the year of the latest update. (We have changed from requiring the date in every copyright statement, to requiring it only in the LICENSE.md file; other copyright statements can just mention the copyright holder.) 

So the LICENSE.md file needs "(c) 2021-2022 David Rowbory", but other places can just have "(c) David Rowbory" (to avoid having to change so many locations when the keyboard is updated).

Other things to consider changing:
- HISTORY.md mentions version 33; README.md and .kmn file have 34.
- Normally the "readme.htm" file is in the source folder. You have it in the source/doc folder. This provides brief information about the keyboard on the point of installation.
- The "help" folder (under the "source" folder) and the naijatype.php file has disappeared. This provides the online help for the keyboard (paralleling the welcome.htm offline help bundled with the desktop keyboard).
- The .ico file is not needed in the packaging (.kps) file. It is already compiled into the .kmx file.
- Any files referenced by the welcome.htm file (7 .png files and 1 .jpg file) need to be included in the list of files for the package so that they are included in the package and available for the welcome.htm file to reference.

DavidLRowe (approved):
I updated the date in the LICENSE.md file, so this should be fine for "experimental".

PR #1725, opened 2022-01-25
submission of chinuk_wawa keyboard

cachang:


keyman-server:

DavidLRowe:
This is looking much better. Just a few small tweaks needed:
- chinuk_wawa.keyboard_info - add a short description to replace the default one (this information is used to help people find your keyboard when searching online)
- Currently for the touch layout you have both a "tablet" layout with your changes and a "phone" layout that just has the default keys. (So a user with a tablet would see the new layout, but a user with a phone would not.) The best solution is to delete the "phone" layout, then Keyman will use your "tablet" layout for both phone and tablet. (In Keyman Developer, click on the "Touch Layout" tab on the left, then switch the "Platform:" value to "phone" then click the "Del" button, and save the changes.)

Since you're changing those things, you could also change: 
- HISTORY.md - add your name
- README.md - add short description

We will need an online help file, but I can add that afterwards based on your welcome.htm text.

DavidLRowe (approved):
Great! I'll add the online help in another pull request.

PR #1729, opened 2022-01-29
[koalib_rere__] Initial submission of koalib_rere keyoard.

W-omer:
For any questions, additions and concerns email weni.omer@gmail.com


keyman-server:

DavidLRowe:
Thanks for your contribution. Don't be discouraged by the length of this feedback. A lot of the changes requested are to keep our automated build system happy!

The name of the project / folder ends with double underline. Consider changing the name (although this will mean a lot of renaming of filenames plus references in certain files). The name of the project folder cannot include letters other than a-z, but the name of the keyboard itself (the &NAME line in the .kmn file) can include anything you like.

The file named "sample" in project folder should be removed

LICENSE.md
This file needs the date (2022) and the name of a copyright holder (the name of a person or the name of a legally recognized entity). Other places (such as the .kmn file and the package source (.kps) file) can have a copyright statement without the date, but with the same name as included here.

README.md
I suggest removing the empty copyright line and also the version number line. (Then you won't need to update them in this file should the keyboard be revised someday.)
In the Description, you have "Koalib_Rere_Ŋenuvǝ". Are the underlines necessary? Are "Koalib", "Rere" and "Ŋenuvǝ" all names for the same language? Or is "Koalib Rere Ŋenuvǝ" together the name of the language?
The reference to "Readme.html" should be changed to "readme.htm" (see below).

.keyboard_info file
The description line should be something short, but meaningful, similar to the first line of the Description in the README.md (or maybe the &MESSAGE line in the .kmn file). Include information that will help people searching online to find this keyboard.

.kmn file
The &NAME line should be the name of the keyboard. Are the underlines needed?
The &COPYRIGHT line needs the same name as in the LICENSE.md file. (You can omit the year from this statement and then won't have to change it should the keyboard be updated someday.)
The &TARGETS line should just have 'any' (since 'any' implies all the others you have listed).

The file "deletethistoo" in the "source" folder should be deleted.

The file "Readme.html" in the "source" folder should be renamed to "readme.htm" to match our conventions. The "readme.htm" file is displayed when the user begins to install the keyboard on a computer. It is generally a short file, sometimes just a one-line description. (Once the keyboard is installed, the "welcome.htm" file, which contains information about using the keyboard, is displayed.)

Eventually, we will need an online help file (keyboard name + ".php") in a "help" subfolder in the "source" folder. This will contain the equivalent information contained in the welcome.htm file. (It is possible for us to approve the keyboard without the online .php help file and then add that later from the information in the welcome.htm file.)

It's not clear to me what purpose the "guide_ENG_html_*.png" files serve. They seem to be duplicates of the "Readme_html_*.png" files. If they are not used, they should be deleted from this keyboard submission.

The "Readme.pdf" appears to be a printable copy of the help. 

The .kpj (project file) has a reference to a non-existant Kib_KB.kmn file. This reference should be removed from the keyboard.

Desktop keyboard layout:
The hyphen used as a dead key, but can't type it twice to get an actual hyphen (as is possible with other dead keys).
The On-Screen Keyboard (OSK) layout displays a blank for the "`" key and for the "'" key (both dead keys), but not for "^" (also a dead key). It would be preferable if these were consistent.
I had some trouble with the Ctrl+d and Ctrl+t combinations. You may prefer using the RightAlt key to produce your extra characters. But, in the end, if these work for you and your users, that's fine.

Touch layout:
The touch layout seems to have both a "phone" and a "tablet" layout, which are almost, but not quite identical. It's not wrong to have both, but it is a lot of extra work to keep both synchronized. Unless you plan to have a different layout for the two, it's better to delete one of them, because Keyman will use the "phone" layout if no "tablet" layout is available and vice versa.

Phone: default layer, longpress key on "c", displays "ɔ" but generates "ç".
Phone: shift layer, "." key produces ">". (Change the "Code" value from K_PERIOD to U_002E to avoid linking to the .kmn file.)

On the Tablet layout, shift layer, the longpress lists for "U", "I", "O" and "A" include extra characters that are not on the corresponding lowercase letters (nor on the phone layout's default layer or shift layer). If you don't need these characters to type the language, you should probably omit them. If you do include them, it would be better if they were consistently available.

If you have any questions about anything I've written, please let me know!

Thanks again for taking the time to submit your keyboard.

W-omer:
Hi David,

I've resubmitted the keyboard with the suggested fixes. It is just waiting
approval and needs a manual build.

Weni

DavidLRowe:
replaced by #1755

PR #1730, opened 2022-01-31
[fv_all] Associate sil_euro_latin with English

darcywong00:
We were checking out the associated languages for the fv_all keyboard package, and noticed the build scripts are pulling the last associated language *Zapotec (Latin)* for the sil_euro_latin keyboard. This tweaks the fv_all build script to use *English*.

This is for integrating dictionary support in the First Voices apps in the upcoming month.

Note, the build script is also generating a lot of warnings about the current language tags being valid, but not canonical. 
Per @mcdurdin, we'll address those at another time.

Tagging @madebybridget

### DIFF
--- a/release/packages/fv_all/HISTORY.md
+++ b/release/packages/fv_all/HISTORY.md
@@ -1,5 +1,8 @@
 # fv_all Keyboard Package
 
+## 12.4 (02 Feb 2022)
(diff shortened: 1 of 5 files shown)


mcdurdin (approved):
LGTM

Issue #1731, opened 2022-01-31
[basic_kbdhebl3] Request to update OSK for AltGr layer


darcywong00:
From https://community.software.sil.org/t/hebrew-standard-basic-keyboard/5829

> The keyboard help page correctly describes the symbols available on Desktop for this keyboard:
https://help.keyman.com/keyboard/basic_kbdhebl3/1.0/basic_kbdhebl3
>
>Unfortunately, the onscreen Keyboard for Desktop does not include all the symbols on the AltGr (Right Alt) layer. (The touch layout seems to be missing some symbols, but I haven’t looked into it closely.)
>
>Can the onscreen Keyboard for Desktop be updated?

darcywong00:
Hmm, some of the key caps on the OSK are really tiny!

![image](https://user-images.githubusercontent.com/7358010/151738248-54a294f1-77a9-4e20-8399-f572befb5aa6.png)

mcdurdin:
This is again a font issue. The ideal would be to have a custom font where the marks are enlarged so they are easier to tell apart -- and where they are made into non-spacing, non-shaping marks for rendering on the OSK. I wonder if this is something that WSTech would be willing to look at preparing? It would work in a similar way to the custom KhmerBusraKbd font for khmer_angkor (see #1710) -- putting these non-spacing marks into the PUA to avoid shaping behaviours. Possibly add a dotted circle? @LornaSIL what do you think?

Ref: https://marc.durdin.net/2015/01/how-to-rendering-combining-marks-consistently-across-platforms-a-long-story/

LornaSIL:
Maybe it would just need a better font associated with the keyboard? Currently no font is specified. I'm quite sure we don't have the capacity to create a new Hebrew font. We already do *not* plan to update Ezra SIL, but we could possible use the **Ezra SIL** or **SBL Hebrew** font with this package.

DavidLRowe:
For the hausa_ajami_qwerty keyboard, @LornaSIL used bitmap files for ZWJ, etc. 

Would it be possible to use create images of the keycaps you want for the various Hebrew diacritic marks and include them in the OSK as bitmaps?

mcdurdin:
It's possible specifying an existing font would help, but the differences in rendering across platforms tends to cause grief.

[...]

PR #1732, opened 2022-02-02
[basic_kbdhebl3] Update OSK and touch

LornaSIL:
OSK was missing some characters
touch layout was missing most ralt characters
Added SBL Hebrew font to hopefully improve display.


LornaSIL:
Helps #1731 but may not fix the font issue that @mcdurdin discussed.

PR #1739, opened 2022-02-11
Added the Postmodern English US Natural Keyboards

PostmodernEnglish:
The Postmodern English project is an open source effort at re-working the English writing system to be more phonetically accurate, compact, and consistent. For more info on this project, please visit http://www.postmodernenglish.org .

We have not included an on-screen keyboard because Keyman Developer appears to lack support for key-specific deadkeys in that area. Specifically, our keyboards require the use of ; as a deadkey, instead of something typical like shift or alt. This is mainly to prevent collisions with common shortcut combinations present in many popular desktop applications. We also feel that the ; key is much easier to access on a regular basis over something like shift or alt. Please correct us on this if this is wrong.


keyman-server:

DavidLRowe:
I didn't realize that you were still making changes. Here are comments from my review earlier today.

Is there a reason that you don't want to enable "web"? There's nothing extra that you have to do to enable a website to use the keyboard. My suggestion would be that you tick the "any" box and untick all the others, then your keyboard will cover all platforms.

Touch layout: You currently have both phone and tablet layouts that seem to be identical. If these are indeed identical (or intended to be so), it's better to delete one, rather than try to keep the two synchronised. (Keyman will use a phone layout on a tablet, if a tablet layout is not available, and vice versa.) I'd recommend keeping the one labeled "tablet" (that is, deleting the "phone" one). That would leave the option, for example, to someday create a "phone" layout that had fewer keys and used more longpress options ('á', 'ā', 'å' as longpress options on 'a', if that would make sense).

We do ask for an onscreen keyboard (OSK) as help information for desktop keyboard users. In Keyman Developer, with the .kmn file selected, click on the Details tab on the left, then "Add..." button at the bottom of the window. Select "Desktop On Screen Keyboard" from the list, then OK. If you then click on the newly added On-Screen tab on the left, you can use the "Fill from layout" button to put the basic information on the keys. You'll then need to indicate the ";"/":" dead key in some way. You could put "Alt" on that key on the OSK, but that might be confused with the physical Alt key. Not sure I have a good alternate suggestion. You'd want to coordinate your documentation (welcome.htm and the .php help file).

By the way, you currently have:
```
+ [SHIFT K_COLON] > dk(colon)
+ [K_COLON] > dk(semi)
```
and a set of changes that is repeated for each deadkey. I believe you could have:
```
+ [SHIFT K_COLON] > dk(semi)
+ [K_COLON] > dk(semi)
```
and omit the set of changes using the dk(colon) deadkey. This would simplify the keyboard and perhaps make it easier to maintain.

Touch layout: Default layer: Longpress for "q" (on the "ћ" key) produces "ћ". Because you are specified the code for that longpress key to be "K_Q", you've connected to the logic in the .kmn file which produces "ћ". If you change that code to "U_0071" it will produce the "q" without linking to the .kmn file. This also applies to "x", "c", "Q", "X", "C" longpress keys (on default and shift layers).
On the numeric layer, you have a similar situation with almost all the keys. Because you've specified the code to be "K_1" you have linked to the change in the .kmn file and get "á" instead of "1". If you change the code to U_0031 you'll get the "1" you want.

I have a further question about your choice of "en" for the BCP 47 tag. Have you considered adding a variant tag to the base "en"? That would allow, for example, use of a spell checker specific to this orthography.

PostmodernEnglish:
Thank you for your feedback- we have addressed most of the issues you pointed out. We still have a few questions regarding some of your other comments however:

The reason we included deadkeys that seem redundant, such as
+ [SHIFT K_COLON] > dk(colon)
+ [K_COLON] > dk(semi)
is because we want the users to be able to type lower or upper case symbols purely based on if the shift key is held down in the moment, and regardless of which case of deadkey they pressed. For example, if the user wants to type in a lower case þ, this allows them to press the lower case (;) deadkey and then ћ to get þ, or, if they accidentally press the upper case deadkey first (:), they can still choose not to hold shift after the fact, and still get a lower case þ. So effectively, ; + ћ = þ and : + ћ = þ. This also works for the inverse case, where a lower case deadkey is pressed first on accident, and then shift + the other key is pressed, resulting in an upper case form of a letter. Would omitting these seemingly redundant key instructions still let the users do this?

Regarding the BCP 47 tag- Postmodern English is designed in such a way that spelling will vary depending on major defined regions of accents. Thus, a single keyboard layout style will not be strictly associated with any particular way of spelling. We tried defining a custom tag: en-PME, but Keyman labels this as invalid. Two questions about this: 
-Will assigning a custom tag for Postmodern English just associate the keyboard with spell checking of any PME variant in general?
-Is there any ways you would suggest formatting the custom tag? Our original en-PME tag was rejected.

Thank you.

PostmodernEnglish:
Adding a note-
After removing the "phone" type touch keyboard, since you mentioned that Keyman will use the tablet layout for phones if a phone layout is missing, we attempted to install this keyboard on a phone. However, Keyman indicates "The keyboard package doesn't have any touch keyboards to install", so we are unable to install the keyboard. Is there something more we need to do to tell Keyman to use the tablet touch keyboard in place of the missing phone touch keyboard?

PostmodernEnglish:
> Adding a note- After removing the "phone" type touch keyboard, since you mentioned that Keyman will use the tablet layout for phones if a phone layout is missing, we attempted to install this keyboard on a phone. However, Keyman indicates "The keyboard package doesn't have any touch keyboards to install", so we are unable to install the keyboard. Is there something more we need to do to tell Keyman to use the tablet touch keyboard in place of the missing phone touch keyboard?

Scratch that, we've figured out the issue. For some reason, the package file was not properly updating when certain files, like the touch keyboard, were added. Being omitted from the package, it makes sense why Keyman said no touch keyboard was available. We've added all needed entries, including images, to the package file manually.

DavidLRowe:
Deadkey comments: Rather than have the shifted (:) and unshifted (;) keys linked to separate deadkeys (colon and semi) each with its own set of rules, you could have the : and ; keys link to a single deadkey (semi, for example) and just have one set of rules. So you would have:
```
+ [SHIFT K_COLON] > dk(semi)
+ [K_COLON] > dk(semi)
```
plus the current rules for the "semi" deadkey, but that deadkey would be triggered by either : or ;.

Unfortunately, Keyman v14 (and the upcoming v15) don't support variants for the language tag, so you'll need to stick with "en" for now. 

But going off on a tangent, here are some options:
(a) You can use a private tag: "-x-" followed by a string of up to 8 characters, for example: en-x-postmod. You can do this now, but only those who know what "-x-postmod" means will understand.
(b) You can, at some point in the future, propose a variant tag: "-" followed by a string of 5 to 8 characters, for example en-postmod, which you could then use if the variant is accepted. For more information see https://www.rfc-editor.org/rfc/rfc5646.html, specifically sections 2.2.5, 3.5 and Appendix B. You would need to provide some indication of how many people are using this variant.

Using such a tag could make it possible to have (for example) a spelling checker that was specific to this writing system, that is the text would be tagged "en-x-postmod" and the spelling checker also identified as "en-x-postmod". If your text is tagged "en", you would not be pleased with the results from an existing "en" spelling checker.

DavidLRowe:
Just a few changes and this should be ready:

Sorry to be picky, but we prefer version numbers to be 1.0 (or 1.0.1), that is, two or three numbers (with no extra leading zeros) separated by ".". Our automated build system sometimes has trouble with leading zeros (as in your current 1.000). So, could you change your version number to "1.0" in the .kmn file (and in the HISTORY.md and README.md files).

For the touch layout, shift layer, 
`!` longpress on `.` key produces `,`. 
`"` longpress on `.` key produces `'`. 

This is due to an incorrect U_00xx code on those longpress keys.

Thanks!

DavidLRowe (approved):

PostmodernEnglish:
Thank you for taking the time to review our keyboard. Now that this is merged, when can we expect to see this keyboard available on the Keyman website?

DavidLRowe:
It should show up at https://keyman.com/keyboards/postmodern_english_us_natural "soon", normally within a few hours, I think, but it was there when I just checked. Certainly with 24 hours, unless there's an issue with the automated system that puts the keyboards on the site.

DavidLRowe:
You should be able to see it now.

DavidLRowe:
Oops. I goofed. I didn't catch the mismatched name of the .php file (postmodern_english_us_natural_desktop.php when it should be postmodern_english_us_natural.php, so that the base name matches the project name). This causes a "Documentation: Help not available" message to show up on that keyman.com page, indicating that the online help is not available. I'll see what I can do to fix that.

DavidLRowe:
Once #1742 is approved and merged, the online help should be available.

PR #1740, opened 2022-02-15
Devanagari Romanized (SIL) Release

sil-internationalNP:


keyman-server:

LornaSIL:
The keyboards are currently in a folder called `sil\Devanagari_Romanized (SIL) v1.6`. The keyboard folder will need to be changed to match the keyboard filenames: `sil\sil_devanagari_romanized`.

It might be better to close this PR and recommit it with the proper folder name.
You've associated the keyboard with `ne-NP`. I'm wondering if you want to add the other languages of Nepal that the keyboard would work with. That way if someone searches for a keyboard that supports another language, they will find it. It will also allow you to associate lexical-models for those other languages when you are ready to do that.

LornaSIL:
A few more comments about things you can do before you recommit. Usually the welcome.htm file has instructions on how to type the keystrokes. You can just put images in there if you want, OR you can link to the .pdf. If you want the .pdf to be distributed, you should add that file in the .kps files.

The help file must have the same filename as the keyboard filenames. Currently it is `sil_devnagari_romanized` and you should change that to `sil_devanagari_romanized` (add an "a"). This is part of our automatic build system and it is required to be spelled the same in order to display.

Since you don't have a tablet layout, you'll also need to remove the following lines from the help file:
```
Similar to desktop layout for most part, except for Alt-Ctrl. Symbols that required dead key combinations (especially independent nepali vowel letters) can be typed with their respective dependent vowels symbol.
</div>
```

We do not recommend that a keyboard try to support two scripts, in this case both Devanagari and Latin. It is better for people to just switch to another keyboard if they want to type in Latin. I would recommend removing the rightctrl and rightctrl-shift layers.

If you do that, you are left with default, shift, rightalt, rightaltshift, CtrlAlt
That means in the help file you can update the last two lines to:
```
<div id='osk-phone' data-states='default shift rightalt rightaltshift CtrlAlt'>
</div>
```

PR #1742, opened 2022-02-16
Rename postmodern_english_us_natural .php file

DavidLRowe:
Rename postmodern_english_us_natural_desktop.php to postmodern_english_us_natural.php


LornaSIL (approved):
LGTM
FWIW, when you include the h1 it will have an extra title which doesn't look great.

PR #1745, opened 2022-02-16
Changed the keyboard info description for PME US Natural to be more informative

PostmodernEnglish:
The PME US Natural keyboard's keyboard info was previously just "Postmodern English US Natural" on the Keyman download page. We've updated it to be more informative.

### DIFF
--- a/release/p/postmodern_english_us_natural/postmodern_english_us_natural.keyboard_info
+++ b/release/p/postmodern_english_us_natural/postmodern_english_us_natural.keyboard_info
@@ -3,5 +3,5 @@
     "languages": [
         "en"
     ],
-	  "description": "Postmodern English US Natural"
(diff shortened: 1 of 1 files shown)


keyman-server:

PostmodernEnglish:
We would like to embed the link mentioned in the description, if possible and permittable. Would using a typical html <a href> tag do the trick, or does the http:// prefix alone cause the Keyman webpage to embed it?

LornaSIL:
There is a "link" field you can add. Take a look at this page: https://keyman.com/keyboards/blackfoot_syllabics_u

And here is the source for that: https://github.com/keymanapp/keyboards/blob/master/legacy/b/blackfoot_syllabics_u/blackfoot_syllabics_u.keyboard_info

This has a lot of extraneous fields that we don't recommend because so much of this information is autogenerated now, but you can at least look at the link field and figure out what you need.

There is a way to do it in the description field, but I'm not finding a sample for that. If that's really where you want it, let me know and I'll search some more.

PostmodernEnglish:
Thanks for the info. We've added a link on the Keyman download page for both this, and the PME UK natural PR.

PR #1747, opened 2022-02-17
Initial submission of Devanagari Romanized (SIL)

sil-internationalNP:


keyman-server:

DavidLRowe:
Thanks for your submission. There are some things that need correction.

The sil_devanagari_romanized.kpj.user file should not be part of the keyboard submission. Please delete this file. (BUT make sure you don't delete the sil_devanagari_romanized.kpj file which is needed for the keyboard build.)

The HISTORY.md file has a non-UTF-8 encoding due to characters in the entry under v1.6.

The welcome.htm should contain information on how to use the keyboard. Currently there is not much help there. You could include information from the .php help file in the welcome.htm. (Alternatively, you could include the PDF file and link to that from welcome.htm.)

There are some problems with the .php help file. 
- the name doesn't match -- it needs to be sil_devanagari_romanized.php to match the keyboard project name (currently it is misspelled as sil_devnagari_romanized.php, that is, "vn" rather than "van")
- the names for "data-states" need to match exactly what is used for the names of the layers in Keyman Developer (and the .keyman-touch-layout file). You have "rightalt-shift Ctrl-Alt" but they are actually "rightaltshift CtrlAlt".

It would be better to omit the layers for the QWERTY keyboard (rightctrl and shiftrightctrl), and let the user switch keyboards to type in Latn script. 

Some minor suggestions/questions: 

The touch layout seems a bit crowded. You could, for example, move some punctuation keys off the default layer (or make them longpress keys on a single punctuation key). But perhaps keeping the same layout as the desktop keyboard is more important?

For the .kmn file, you just need "any" if you are supporting all platforms. You can untick all the other boxes. (This makes it easier for someone examining the keyboard.)

Let us know if you have any questions.

DavidLRowe:
I think you've dealt with all the problems, so I could approve this as it stands, but I'll raise two things you might want to change:
- You've included SILkeybRom.png in the package. Is this a file to replace the default Keyman image when the desktop keyboard is being loaded? If so, you'll need to go to the Details tab, and select the file for the "Image file:" field. Ideally the image should be 140x250 pixels. If the image is a different size, Keyman will try to make it fit which may distort the image.
- Currently you have specified the language tag (BCP 47 code) as "ne-NP" in the sil_devanagari_romanized.keyboard_info and in the sil_devanagari_romanized.kps files. The canonical form for this tag is just "ne", which would be preferable. 
- If the keyboard can be used to type other languages written in the Devanagari script, you can include the corresponding BCP 47 codes  in the sil_devanagari_romanized.keyboard_info and in the sil_devanagari_romanized.kps files. (Or possibly you would want to wait until the next time this keyboard is updated.)

DavidLRowe:
The only issue I saw was that you didn't actually reference the newly resized image file, so I made that change in the .kps file.

DavidLRowe (approved):

sil-internationalNP:
Thanks for the upload. But, we've got some problems with the pull.

The source in the pull doesn't match our branch. We have lots of errors: web keyboard not working, web help file has wrong text and is missing keyboard images, etc.

Any idea why we are having these probems?

LornaSIL:
The help file has the wrong syntax, so I'll try to fix that for you. I've uploaded your font to another repo to see if that helps. I'm not quite sure about the web keyboard, but we'll look into it. Have you tried the keyboard on Desktop? Is it actually correct? If so, then it's possible it's just we didn't get some of the structural issues right.

LornaSIL:
The help file doesn't match the text in the welcome.htm file. Do you want the online help to match what is in the welcome.htm file?
I can fix that without bumping the version number.

LornaSIL:
Adding your font to our site has helped the web keyboard situation. Please double-check that. Then let me know if you wish for me to update the online helpfile to match welcome.htm.

sil-internationalNP:
> The help file doesn't match the text in the welcome.htm file. Do you want the online help to match what is in the welcome.htm file? I can fix that without bumping the version number.

Yes, we want text in the welcome.htm to match the help file. Also, we want keyboard description in both help and welcome file to match with the description in read.md or keyboard_info file. Thank you for the fixes. Now, the web keyboard is working fine.

LornaSIL:
See #1761. I updated only the online help file with the next from welcome.htm *except* I used the description from README.md as you asked. You'll need to be the one to update welcome.htm next time you update the keyboard. The only time we can update the welcome.htm file is when you bump the keyboard version and I didn't want to do all of that for the help file.

LornaSIL:
I updated the help file, so I hope it looks okay to you now.

DavidLRowe:
https://help.keyman.com/keyboard/sil_devanagari_romanized

sil-internationalNP:
In the code we checked in, the welcome/readme/help first paragraph all matched. So we are confused why after the pull they were different?

sil-internationalNP:
> I updated the help file, so I hope it looks okay to you now.

Yeah it looks fine. Thank You! for the update.

LornaSIL:
>In the code we checked in, the welcome/readme/help first paragraph all matched. So we are confused why after the pull they were different?

I suspect some of the files you changed didn't make it in to the Pull Request. I know this is a confusing process.

PR #1752, opened 2022-02-18
Added the Postmodern English US DualStroke Keyboard

PostmodernEnglish:
The Postmodern English project is an open source effort at re-working the English writing system to be more phonetically accurate, compact, and consistent. For more info on this project, please visit http://www.postmodernenglish.org/ .

This is a United States Postmodern English keyboard, in the DualStroke layout style. The DualStroke style makes use of 3 dead keys to type the 3 different accent marks of Postmodern English, plus the alternate characters on the keyboard.


keyman-server:

DavidLRowe:
@PostmodernEnglish Everything looks fine, except some things that I missed on your previous submissions. (Oops!)

If you go to https://help.keyman.com/keyboard/postmodern_english_uk_natural you'll see that the images are too wide. I should have caught that. Sorry! (The welcome.htm file has an entry in the style element: `img { width: 75%; height: 75%; }` that scales all img elements.) So, one way to address the problem is to include scaling information either in the img element (width="95%" or whatever) or via a style. The other way would be to manually reduce the resolution of the images themselves; in this case we'd suggest 920 pixels wide or less.

The same is true for the postmodern_english_us_natural keyboard. 

I'd suggest that you make a new PR to correct the widths of the images for postmodern_english_us_natural and postmodern_english_uk_natural. (Again, sorry I didn't catch that before.) Normally when modifications are made to a keyboard, the version number needs to be incremented (1.0 to 1.1 or to 1.0.1, for example) so that the build system knows that this is a new keyboard that should be offered to users. The exception is in cases like this where the only change is to the `help` folder and the on-line help. So, for this case, you don't need to modify anything except the files in the help folder.

Then in this PR, you'll need to make similar adjustments for the widths of the images in the .php help file.

Thanks!

DavidLRowe:
I spoke too soon! See the comment I made in PR #1756 about extra files in the package. You should remove:
- the three *.md files
- the *.keyboard_info
- the source files: *.ico, *.kps, *.keyman-touch-layout, *.kmn, *.kvks, *.kps

DavidLRowe (approved):

PR #1759, opened 2022-02-24
Postmodern English US Natural Misc Fixes

PostmodernEnglish:
Misc. fixes suggested for US Natural.


keyman-server:

DavidLRowe:
Rather than changing the version number in README.md to 1.1, I just removed the line. We've found that having the version number in multiple places makes extra work when updating the keyboard.

DavidLRowe (approved):

PR #1763, opened 2022-03-02
update touch layout

hinditutorpoint:
hi team update change according this commit and fixed touch layout shift not working yet

PR #1768, opened 2022-03-04
[masaram_gondi] Update keyboard with hopefully correct changes

LornaSIL:
This was rather complex since the keyboard was not updated from the previous package structure.
This should hopefully replace the changes that were made in PR #1763.


DavidLRowe:
For desktop layout, default and shift layers, the "[" "]" "{" "}" keys generate "[" "]" "{" and "}", but the characters displayed on the OSK (and with the welcome.htm) are different. Not sure which is correct.

I'm having trouble testing the touch layout, but it seems that the shift button doesn't work.

DavidLRowe (approved):

hinditutorpoint:
hello team there was a font issue in welcome.html can you please fixed it

LornaSIL:
Ok, I'll take a look at it. The revised help file also is in bad shape.

PR #1769, opened 2022-03-05
Vietnamese Telex Keyboard

miketvo:
After trying out Keyman, I didn't find any implementation for the [Telex input method](https://en.wikipedia.org/wiki/Telex_(input_method)) for Vietnamese, so I decided to write my own implementation for Keyman.

I generated all possible Vietnamese Telex key sequences with Python (see `extra/` folder), then copied those rules to the keyboard's .kmn file.


keyman-server:

LornaSIL:
Thanks for your keyboard submission! Of course, I have a few comments! 

This is quite the amazing keyboard with 57,000+ lines of code!

* Would you mind moving this to `v` folder rather than creating a new `vi` folder. We basically just organize the repo alphabetically and not by language tag.
* The `keyboard_info` file could use a longer description. As long as it's all on one line, you could include everything you have in the `README.md` description. The `keyboard_info` file is used for the website for the keyboard so it's helpful for people to see why they would want to use this keyboard.
* I would ideally like to see all these fonts in a `release\shared\fonts\noto\Latn` folder. That way other keyboards could benefit from the fonts being available and we don't have to add the fonts to every keyboard source folder in the repository. Would you be able to add them to that folder, or would you want me to? That would mean you have to change the .kps file to reference them in the `shared` folder. I think longterm that is the best option so we don't keep adding fonts to individual keyboard folders. Your package would still distribute the fonts in the same way, they just wouldn't be in the source keyboard folder.
* Regarding the touch layout. It looks like your phone and tablet keyboards are the same. When that is the case, we recommend deleting one or the other. If there is no phone layout, a tablet layout will work on the phone and vice versa. This allows us to have less work in making sure both layouts are in sync.
* We do require a `help` folder and file. Sometimes I'll do it for people, but based on this PR I assume you can do it.
    * Add a `help` folder in the `source` folder. Copy the welcome.htm file to that folder, rename it to `vietnamese_telex.php` and then edit the file in a text editor.
    * Delete all the top lines through the `<h1>`. Then, add the following as a header:
```
<?php 
  $pagename = 'Vietnamese Telex Keyboard Help';
  $pagetitle = $pagename;
  require_once('header.php');
?>
``` 
Then, delete the last two lines in the file:
```
</body>
</html>
```
This file will be used for the online help.

I think that's it. Let me know how you want to handle the font situation.

miketvo:
Hello Lorna!

Thank you for your reply. I have finished everything you suggested and test build the keyboard successfully on my local machine using Git Bash and the `build.sh` script.

I have also moved the fonts to `release\shared\fonts\noto\Latn` as you suggested.

Please advise on what I should do next.

LornaSIL (approved):
LGTM. Thanks for the fixes. This should be online within a few hours.

LornaSIL:
Keyboard is now available: https://keyman.com/keyboards/vietnamese_telex

miketvo:
Thank you for your support Lorna!

PR #1772, opened 2022-03-07
[srr_ajami] Changed two characters per user request

LornaSIL:


DavidLRowe:
The version in the .kmn file is 1.1, but in HISTORY.md it is 1.2.

Can you check the shift layer on the touch layout? It seems that, for the character being displayed, the U+0684
didn't get changed to U+08A2.

DavidLRowe (approved):

PR #1776, opened 2022-03-10
[burushaski_girminas] Version Change and Development 

quwatsunny:
Based on your recommendation, I have open a new pull.


keyman-server:

LornaSIL:
Thanks for resubmitting this. I do have some things that need changing, but you should be able to update the files in this Pull Request.
* In LICENSE.md, please change the copyright date from `© 2020 Quwat K. Sunny` to `© 2020-2022 Quwat K. Sunny` (adding the ending year. In all other locations you can completely remove the year if you want so that you don't have to keep updating that. We just required LICENSE.md to be kept updated.
* Please delete the .kmn, and .kps files that are in the `burushaski_girminas` folder (and keep the ones in the `source` folder).
* The `keyboard_info` file needs some adjustment. When you were removing the extra languages I think you accidently removed the "description". Also, the final double quote needs to be on the previous line at the end of the description. The whole file needs to look like this:
```
{
    "license": "mit",
    "languages": [
        "bsk-Latn"
    ],
	  "description": "Burushaski Girminas is latin/Roman based keyboard developed for Buruśo Mahraka, a representative body of all three varieties of Buruśaski from Hunza Nagar and Yasin."
}
```
* In the .kmn file:
   * Either remove this line `store(&KMW_HELPFILE) 'burushaski_girminas-help.htm'` or add that file into the Pull Request (and in the .kps).
   * On line 63 you have this `U+1E8Ec` please delete the `c` as that makes it the wrong character.
   * When the tablet and the phone touch layouts are the same, we recommend deleting one of them. This is because if there is no phone layout, the tablet layout will work on a phone (and vice versa). The reason we recommend this is that it is easier to just keep one layout up-to-date rather than having to update both layouts and test both layouts. You can delete one of them by selecting "Platform: phone" and then clicking on the "Del" button to delete that layout. Then save.
   * The way to access the "Symbol" layer is in a very odd location. I had to look very hard to figure out where to find it. Would you consider moving it back to where it was before (lower left, just above the abc key)?
   * The key labeled `U_1F821` has the glyph for U+1F621 on the keycap. You should either change the glyph on the keycap or change the label to `U_1F621`. U+1F821 appears to be an up arrow so I think you should change the label.
* In the .kps file:
   * Please add in your welcome.htm
   * Please add in your readme.htm
   * On the Keyboards tab, the BCP 47 tag should match what is in the `keyboard_info` file. In this case, `bsk-Latn`. You can take `-PK` off the tag name.
   * Under the Details tab
      * add in the readme.htm
      * Delete the version number and check "Package version follows keyboard version". This is a good example of why this field is needed since your keyboard version is different than what you have here.
      * You can delete the copyright year here if you want, but leave the copyright holder name.
* In the help file (burushaski_girminas):
   * This is a php file that should have the regular html header info removed. Please delete *everything* through this line `<h1 class="style11">Overview</h1>` and then paste the following to the top of the file:
```
<?php 
  $pagename = 'Burush&atilde;ski Girminas - Roman/latin Keyboard Help';
  $pagetitle = $pagename;
  require_once('header.php');
?>

    <style type="text/css">
<!--
.style3 {
	color: #FF3300;
	font-weight: bold;
}
.style9 {color: #FFFFFF}
.style10 {color: #000000}
.style11 {font-family: Calibri}
.style13 {color: #000000; font-family: Calibri; }
.style16 {
	font-family: Verdana, Arial, Helvetica, sans-serif;
	font-size: 12px;
}
.style18 {font-size: 14}
body,td,th {
	font-family: Times New Roman, Times, serif;
}
-->
    </style>
```
Then, delete the last two lines in that file:
```
</BODY>
</HTML>
```

I think that's everything.

quwatsunny:
@LornaSIL  Thanks for the detailed review. I have made all the recommended changes. I have tried to follow your instruction to the best of my ability. Please let me know if anything else I need to do. Otherwise, you can merge it.

quwatsunny:
Thank you, Lorna, for your detailed review.  I have followed almost all your i

LornaSIL:
Looks much better.  A few things:
* The `keyboard_info` file needs a double quote at the end of the description.
* You now have two php files. Please delete the one in the `welcome` folder. They look like they are exactly the same.
* I still find the access key to get to the symbol layer (on the touch layout) in a strange location, but if that's where you want it, I'll let that one go.

That's it. With that, I should be able to approve it.

quwatsunny:
Hi Lorana, 

I understand your concern about the symbol.  Let's just leave it the way, I have updated it. 
I have made the other two changes.

LornaSIL (approved):
LGTM. Please note that I had to change your keyboard_info file. So, make sure next time you update the keyboard that you get the latest version of the files.

quwatsunny:
> LGTM. Please note that I had to change your keyboard_info file. So, make sure next time you update the keyboard that you get the latest version of the files.

Ok, thanks.  Thanks for approving. 
Also Lorna,  the web version of the the keyboard is picking up some different keyboard. 
Is there any issue.  The use Keyboard online option does not pick up the burushaski_girminas. 

https://keymanweb.com/#bsk-latn,Keyboard_burushaski_girminas

LornaSIL:
I think you just have to be patient for a few hours and eventually that
part will build properly too.

PR #1781, opened 2022-03-11
Latest update for Devanagari Romanized (SIL)

sil-internationalNP:


keyman-server:

DavidLRowe:
It seems that the tablet layout you added is identical to the phone layout. In such a case, we strongly recommend that you keep only one of the layouts (that is, delete either phone or tablet layout). Keyman will use a phone layout on a tablet if no tablet layout is specified and vice versa. So if the two are identical, you greatly increase the maintenance burden of keeping the two synchronized whenever there's a change. (The only reason to have both layouts is if you intend to make the tablet and phone layouts different. For example, you might make a phone layout that has one less row by moving the top row of digits to a numeric layer.) If you really want to keep both layouts (phone and tablet), I guess that's okay, but in our experience, it's just extra work.

When you delete one of the layouts, you'll need to adjust the .php file accordingly.

I don't have much experience with the Devanagari script, so the following comments may just reflect my ignorance!

The on-screen keyboard (OSK) shows a character on the RAlt+x key, but pressing that combination doesn't produce anything. The images included with the welcome.htm are similar. Does this only work in specific contexts? Or is the OSK in error (as well as the welcome.htm image)?

The same seems to be the case for the LCtrl+LAlt+x

The Shift+A produces "A", although the OSK and welcome.htm indicate it should produce अ.

Just one more comment about the documentation in the welcome.htm. Sometimes the QWERTY keyboard keys are listed (as in the "Conjuncts" subsection of the "halanta" section) and sometimes they aren't (as in the other two subsections of the "halanta" section). It may be that the QWERTY key info isn't needed for someone who knows the script, but I find it really helpful! This is not a request for changes to the documentation, but just a comment from someone new to the script.

DavidLRowe:
Everything looks good, except for one small item with the help file. The .php file has 
```
<div id='osk-phone' data-states='default shift rightalt rightaltshift'>
```
but the name of the layer is `rightalt-shift` (rather than `rightaltshift`) so currently that layer is not displayed.

With that change, the keyboard should be ready.

Optionally, you might consider changing the touch layout to remove the `leftctrl-leftalt` and `leftctrl-leftalt-shift` layers (along with the "LAltLCtrl" key to the left of the space bar) since these two layers are duplicates of the `rightalt` and `rightalt-shift` layers.

DavidLRowe:
Thanks for submitting this update. The 1.7 version of the keyboard should be available at https://keyman.com/keyboards/sil_devanagari_romanized once the keyboard build system runs. (That link takes you to a page that currently shows version 1.6, but will show version 1.7 when it is available, usually within 24 hours.)

Issue #1778, opened 2022-03-10
Update masaram_gondi welcome, readme and help file


LornaSIL:
The welcome and readme files don't display the font properly when you install the keyboard. The help file doesn't display the tables properly...probably because it is using an h2 tag with div tags. It probably needs to be changed to use proper tables.

PR #1782, opened 2022-03-15
[masaram_gondi] Update welcome.htm to use tables so it can be used in help file

LornaSIL:
… help file

* Updated readme.htm to change how font is called so it would display the font when the .kmp is being installed.
* Updated welcome.htm to use tables and changed how font is called so it would display the font when the .kmp is being installed.
* Updated php file to use new welcome.htm code so that tables will be displayed properly.
* Updated `keyboard_info` to use correct font filename.
* Bumped the version number

### DIFF
--- a/release/m/masaram_gondi/HISTORY.md
+++ b/release/m/masaram_gondi/HISTORY.md
@@ -1,6 +1,11 @@
 Masaram Gondi (ITRANS) Change History
 ====================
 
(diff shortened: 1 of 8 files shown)


LornaSIL:
Fixes #1778. Note that readme.htm still does not display font in .kmp UNLESS the font is already installed.

DavidLRowe (approved):

PR #1785, opened 2022-03-16
Vietnamese Telex (Legacy Mode) - new keyboard

miketvo:
Due to request from an user, I decided to write another implementation of the [Vietnamese Telex keyboard](https://keyman.com/keyboards/vietnamese_telex) that uses the older convention for tonal diacritic placement instead.

This is intended as an alternative version to  [Vietnamese Telex keyboard](https://keyman.com/keyboards/vietnamese_telex) for users who prefer the old (and currently more popular, although not officially endorsed by the Vietnamese government anymore) way.

## Request from user:

image_2022-03-16_200356

## Extra information:

[Quy tắc đặt dấu thanh kiểu cũ (Translation: Older convention for tonal diacritic placement)](https://vi.wikipedia.org/wiki/Quy_tắc_đặt_dấu_thanh_trong_chữ_quốc_ngữ#Kiểu_cũ)


keyman-server:

LornaSIL:
Here are a few changes.
* In `HISTORY.md` please change the keyboard name to include `(Legacy Mode)`
* I would personally change the name everywhere to just have the word `(Legacy)` so it would be **Vietnamese Telex (Legacy)**. This is purely a personal preference to get the name a bit shorter.
* I would delete the version everywhere except in `HISTORY.md` and the `.kmn`, including and especially the readme.htm title. This just makes it easier to update the files the next time you do an update. (We used to include versions everywhere in the templates, but it has proved to be a bit of a pain to maintain consistently.)
* It's best to delete the `<h1>` in the help file. This is because the help system already adds a header and it just ends up looking superfluous. See how it looks on this page: https://help.keyman.com/keyboard/vietnamese_telex
* The readme.htm points to the `vietnamese_telex` `HISTORY.md` file. You probably want that changed to `vietnamese_telex_legacy`. I know it's not live yet, but it will be when we approve the keyboard.
* Otherwise, I think it looks good.

miketvo:
Hi Lorna,

Thank you for your help! I've made the changes you mentioned. Please let me know if I missed a version tag somewhere beside `HISTORY.md` and the `.kmn`.

I've also took a look at the help page: https://help.keyman.com/keyboard/vietnamese_telex. Seems like my `$style` variable on line 4 of the help `.php` doesn't work. I'll update the help `.php` in both `vietnamese_telex` and this keyboard to have inline style instead.

image_2022-03-17_081055

Best regards,
Mike Vo.

LornaSIL:
If you are just updating a help file and nothing that actually goes in a keyboard package (like vietnamese_telex.php) you do not need to bump the version number in that keyboard. We can just update the .php file. In fact, you are welcome to update that php file in this Pull Request.

miketvo:
All done. Thank you!

LornaSIL (approved):
LGTM

PR #1786, opened 2022-03-23
[sil_akha_act] created based on an existing .kmx with modification

MakaraSok:


LornaSIL:
The keyboard seems to be fine. The font kind of worries me. It claims to have Doulos SIL origins, but the filename uses the word `times`. If the real origin is Times New Roman, we should **not** distribute it. Other properties in the font do indicate that it derived from our old Encore font *SILDoulosL* which would be acceptable. But, I'm really unhappy with the filename having the word "times" in it. "DoulosSILALU-Regular.ttf", "DoulosSILALU-Bold.ttf", "DoulosSILALU-Italic.ttf", "DoulosSILALU-BoldItalic.ttf" would be more appropriate.

MakaraSok:
I've put the comment through to Scott. Let's wait and hear what he has to say.

sdysart:
The font is an SIL font, I know it was derived from Doulos SIL.  The WS Tech team will have to speak to the name of the font.  I have no opinion on the name.

LornaSIL:
Since it's derived from Doulos I think these should be the filenames:
"DoulosALU-Regular.ttf", "DoulosALU-Bold.ttf", "DoulosALU-Italic.ttf", "DoulosALU-BoldItalic.ttf"

MakaraSok:
@LornaSIL @sdysart Should the files simply be renamed?

LornaSIL:
Yes, just the filenames. But then the .kps needs to reference the new filenames.
au_timesr.ttf -> DoulosALU-Regular.ttf
au_timesb.ttf ->DoulosALU-Bold.ttf
au_timesi.ttf -> DoulosALU-Italic.ttf
au_timesbi.ttf ->DoulosALU-BoldItalic.ttf

LornaSIL (approved):
LGTM.

PR #1789, opened 2022-03-25
Changed the Postmodern English UK Natural Keyboard BCP 47 tag to en-QP

PostmodernEnglish:
This change is to align they keyboard language code to the new lexical model.


keyman-server:

LornaSIL (approved):
LGTM

PR #1790, opened 2022-03-25
Changed the BCP 47 code to en-QP for Postmodern English US DualStroke Keyboard

PostmodernEnglish:
This change is to align the keyboards with the new lexical model.


keyman-server:

LornaSIL (approved):
LGTM

PR #1793, opened 2022-03-25
Change BCP 47 code to en-QP for Postmodern UK Dualstroke Keyboard

PostmodernEnglish:
This is to match up the BCP 47 code with the recent lexical model that was added.


keyman-server:

LornaSIL (approved):
LGTM

PR #1800, opened 2022-04-01
[fv_uwikala] layout updates

caforbes:
Updates to key layouts, character normalization, and documentation.


keyman-server:

DavidLRowe:
I could approve this as is, but I have a minor question. 

On the shift level of the touch layout, pressing most keys will insert the character and return you to the default layer. This is not the case with the Ə key (as well as with some of the accents). Perhaps that is by design, but perhaps it is an oversight.

Otherwise this looks very good. 

Sorry for the delay in reviewing this submission.

caforbes:
Thank you! I have fixed the layer switch after schwa but will leave the behavior after accents.

DavidLRowe (approved):

DavidLRowe:
The updated v9.2 keyboard should be available at https://keyman.com/keyboards/fv_uwikala before too long.

Issue #1798, opened 2022-03-29
Remove Mingrelian from listed language in sil_euro_latin


LornaSIL:
https://community.software.sil.org/t/mingrelian-keyboard/4128/40
lzz and xmf

DavidLRowe:
Is this a matter of removing xmf (Mingrelian) from the .keyboard_info and .kps files?

Not sure what the reference to lzz is. I'm not seeing that in either file.

Should the "-Latn" be removed from the tags (at least for languages where Latn script is the default)?

Any other changes needed for sil_euro_latin?

DavidLRowe:
I don't think any lexical models will be affected. It seems the only overlap is "en"!

I think these tags need to keep -Latn: 
bgx-Latn, crh-Latn, fub-Latn, fuh-Latn, rmy-Latn, zqe-Latn
I propose removing -Latn from all the rest that have it.

.kps and .keyboard_info have:
"zyg-Nshu" Yang Zhuang
"zyn-Hani" Yongnan Zhuang

langtags has:
zyg = zyg-Hani Yang Zhuang
zyn = zyn-Hani Yongnan Zhuang

Is zyg-Nshu correct? 
Is zyn-Hani correct? Is it typed with Latin characters even though it's not a Latin script?

Is it worth putting all the codes in alphabetical order?

LornaSIL:
There's something wrong with those. I thought we removed them.

DavidLRowe:
Are you suggesting we remove zyg-Nshu and zyn-Hani?

What about putting the codes in alphabetical order? That would generate a lot of changes, but it might make long term maintenance easier.

PR #1803, opened 2022-04-12
Sil euro latin

DavidLRowe:


DavidLRowe:
It may be easier to review this in stages.

At this point, 10 BCP 47 codes have been removed from both .kps file and .keyboard_info file:
zch-Hani
zeh-Hani
zgb-Hani
zgm-Hani
zgn-Hani
zln-Hani
zlq-Hani
zyg-Nshu
zyn-Hani

LornaSIL:
Fixes #1798

DavidLRowe:
I'm in the process of alphabetizing the tags, which I'll put in the next stage. After that I'll remove the -Latn from tags that don't need it. Hopefully this will make it easier to review. (And, yes, I'll need to update version and HISTORY.md.)

DavidLRowe:
At this point the two lists of BCP 47 codes (in .keyboard_info file and in .kps file) should be in alphabetical order. Note that .keyboard_info file has "pov" whereas .kps file has "pov-Latn", but that will be resolved in the next stage.

DavidLRowe:
At this point -Latn removed from tags, except: bgx-Latn, crh-Latn, fub-Latn, fuh-Latn, rmy-Latn, zqe-Latn
Metadata files updated, however, I'm not sure what to do about readme.htm file. It currently has about 120 languages listed, but .kps has 372 BCP 47 tags.

PR #1512, opened 2021-04-07
Batak

bennylin:
New layout. Thanks.

PR #1807, opened 2022-04-14
[batak] New keyboard on behalf of author

LornaSIL:
Replaces #1512. Issues Makara requested have been addressed.


DavidLRowe:
All looks fine with desktop layout, except Shift+K, which seems to produce the wrong character (at least it doesn't match documentation / OSK).

I'm not sure how to test the touch layout. I'm not seeing the font.

DavidLRowe:
Actually, on the desktop layout, is there a way to generate U+1BFC through U+1BFF?

DavidLRowe:
Touch layout: 
- keys to right of space bar don't work on shift layer
- also the Shift+K issue mentioned above shows up here as well

LornaSIL:
* Once the `SHIFT K_K` got fixed, it should then work on touch layout. Did you repull?
* I fixed the keys on shift to the right of the spacebar
* These 4 keys for U+1BFC..U+1BFF were apparently intended to only be on the touch layout. See the code for .kmn where it says `c Hanya touch layout / touch layout only`. I don't feel comfortable adding them to the .kmn when it seemed intentional.

DavidLRowe (approved):

bennylin:
Will take a look at it once I got things sorted. (PC motherboard broken again for the 2nd time, and other IRL issues)

PR #1808, opened 2022-04-17
[adiga_danef] First release of the Adiga Danef keyboard.

showgan:


keyman-server:

showgan:
Hi,
I'm not sure what's the next step. Will someone on your end review my pull
request and let me know if I need to fix something?

Thanks,
Haroon

DavidLRowe:
Thanks for your submission. Yes, one of us will review your pull request within a few days.

DavidLRowe:
Looks very good (much better than my first submission!).

Touch layout, shift layer, the key labeled "I" produces U+0131 (rather than U+0049).
Touch layout, numeric layer, the key labeled "&" produces "^".

Since the desktop layout has no control layer, the `<div id='osk' data-states='default shift ctrl'>` statement in the adiga_danef.php file shows an empty keyboard for the control layer. It would be better to remove `ctrl` from that statement. 

It would be good to add to the adiga_danef.php file the information that you currently have in the welcome.htm file (`;s => ş`, etc.). (The .php file provides the online help and the welcome.htm provides the help packaged with the keyboard. Unfortunately, this means some repetition of information.)

Optional: If you want, you could add
```
<h2>Touch Keyboard Layout</h2>
</div>
```
to the end of that file in order to add the touch layout information to the online help. (Not required, but potentially useful for someone investigating the keyboard.)

showgan:
Thanks a lot David for the detailed feedback!
- the U+0131 is intentional. The character looks the same as the English
"I", but it has a different uni-code.
- I fixed the printed character for "&"
- I was't sure about the format of the php file. I made the changes you've
recommended.

I made another pull request with these changes.

Thanks again!
Haroon

p.s. how long does it take from approving the merge till the new keyboard
appears on the website for download?

DavidLRowe:
Once the keyboard is approved and merged, it normally takes less than 24 hours until the keyboard is available for download, often much less than that.

About the two "I" keys on the touch layout:
(1) Above the backspace key: 
- default layer: U+0069 (LATIN SMALL LETTER I) is displayed on the key, and using K_I for the ID links to the .kmn file which produces U+0069 (since no rule redefines "i").
- shift layer: U+‎0130 (LATIN CAPITAL LETTER I WITH DOT ABOVE) is displayed on the key, and using U_0130 for the ID produces that character.

(2) To the right of the "u" key:
- default layer: U+0131 (LATIN SMALL LETTER DOTLESS I) is displayed on key and using U_0131 for the ID produces that character.
- shift layer: U+‎0049 (LATIN CAPITAL LETTER I) is displayed on the key, but using U_0131 for the ID produces something different that what is displayed on the key.

My concern is that last combination listed, where the character being produced by pressing the key (LATIN SMALL LETTER DOTLESS I) is different from what is displayed on the key (LATIN CAPITAL LETTER I). This is generally not done with touch layouts. If you want that key to produce a U+0131, even on the shift layer, you should display the U+0131 character on the key. On the other hand, if the key should produce U+0049 (as displayed), it may be best to change the ID to U_0049 (rather than using K_I).

Similarly, you may want to to change the ID of the first combination listed above from K_I to U_0069.

showgan:
Thanks David. I changed the ID to K_I for Latin capital letter I.

DavidLRowe:
I added two `</div>` lines to the .php file so that the help file will display correctly.

DavidLRowe (approved):

DavidLRowe:
Should show up at keyman.com/keyboards/adiga_danef before too long.

showgan:
David, thanks a lot for all your help! and sorry for the mistakes I made in the php file.

DavidLRowe:
Not a problem at all. We don't have much documentation on the .php file format and you can't immediately see the result (without setting up a local server).

PR #1809, opened 2022-04-20
[batak_php] Remove spurious space for OSK

LornaSIL:
The extra space in the php file made it think there was another layer.


LornaSIL:
I submitted a fix for the sil_akha_act php file also.

DavidLRowe (approved):

PR #1811, opened 2022-05-06
[sil_bolivia] Keyboard for Bolivian languages

LornaSIL:
I need to review it once more.


DavidLRowe:
Oops I see there's still a WIP label. Maybe this review is too early.

For the touch layout, rightalt layer, the barred-i and barred-I keys don't seem to produce any output.

LornaSIL:
@DavidLRowe I changed the rules so it wouldn't be dependent on Keyman 15 and now it builds. I think it's ready.

DavidLRowe:
On the touch layout, dieresis and tilde dead keys plus barred-i don't seem to work to create an accented barred-i, whereas dead keys plus other vowels do work. (And the dead key combinations with barred-i work on the desktop layout.) If this is working as intended (since those characters are available on long press keys), I can approve it as is.

LornaSIL:
Sorry for all the changes. Since the deadkeys weren't working for barred i, I saw more things that could be improved.

DavidLRowe (approved):

PR #1813, opened 2022-05-09
Mozhi malayalam Update

beniza:
This PR includes a completely re-written code for the `mozhi_malayalam` keyboard. The rules in this PR are auto-generated from the Spec using a tool developed by @cibucj.

A new layout image is added.


keyman-server:

PR #1821, opened 2022-05-10
[sil_cameroon_azerty] Add NCAPS to some rules

darcywong00:

The upcoming kmcmp.exe release will generate warnings about inconsistent matches needing NCAPS (https://github.com/keymanapp/keyman/pull/6347)

```
```


LornaSIL (approved):
LGTM

PR #1824, opened 2022-05-10
[sil_moore] Add NCAPS to some rules

darcywong00:

The upcoming kmcmp.exe release will generate warnings about inconsistent matches needing NCAPS (https://github.com/keymanapp/keyman/pull/6347)

```
```

This keyboard involved a lot more changes, so initially making the PR in draft


darcywong00:
No longer needed due to #1883 where NCAPS warnings no longer fail the build

PR #1844, opened 2022-05-12
[indonesia] Add NCAPS modifier to some rules

darcywong00:

The upcoming kmcmp.exe release will generate warnings about inconsistent matches needing NCAPS (keymanapp/keyman#6347)

```
```


LornaSIL:
@darcywong00 is this still in progress?

darcywong00:
No longer needed due to #1883 where NCAPS warnings no longer fail the build

PR #1845, opened 2022-05-12
[itrans_bengali] Remove duplicate ©RIGHT store

darcywong00:

* Remove the duplicate ©RIGHT store


LornaSIL (approved):
LGTM

PR #1849, opened 2022-05-12
[sil_tai_dam_lao] Remove duplicate MESSAGE store

darcywong00:

* Remove the duplicate MESSAGE store


LornaSIL (approved):
LGTM

PR #1853, opened 2022-05-13
[ukwuani] Remove extraneous group(main)

darcywong00:

* Remove extraneous group(main) since it's already on l.14


LornaSIL (approved):
LGTM

PR #1855, opened 2022-05-16
[cim] Remove extraneous rules for default keys

darcywong00:
The upcoming 15.0 kmcomp will warn about needing to include NCAPS modifier to avoid inconsistent matches.

This removes extraneous  rules for default keys, as they're handled in the rest of the kmn.


LornaSIL (approved):
LGTM

PR #1856, opened 2022-05-16
[nailangs] Add NCAPS modifier to some rules

darcywong00:
In prep for 15.0 release, some of the keyboards need to be updated to add NCAPS modifier to some rules to avoid inconsistent matches.

Settings this to draft as there's several rules to investigate


LornaSIL:
@darcywong00 is this still in progress?

PR #1857, opened 2022-05-16
[taigi_poj] Add NCAPS modifier to some rules

darcywong00:
In prep for upcoming Keyman 15.0 release, this adds NCAPS modifier to some rules to avoid inconsistent matches


LornaSIL (approved):
LGTM

PR #1858, opened 2022-05-17
[santali_latin] Remove extraneous default rules

darcywong00:
The upcoming 15.0 Keyman compiler warns about adding NCAPS modifier to avoid inconsistent matches.

For this keyboard, can remove the default keys defined from basic_kbdus


LornaSIL (approved):
LGTM (@DavidLRowe agreed I could do the review)

PR #1860, opened 2022-05-23
Updating Language ID

rdai:
Small change, only updating language id in kps file to allow applications to detect language correctly. This fixes kern issue in Microsoft Word

### DIFF
--- a/release/s/shan/HISTORY.md
+++ b/release/s/shan/HISTORY.md
@@ -1,7 +1,7 @@
 Shan Change History
 =======================
 
-1.1 (27 Oct 2020)
+1.1 (27 May 2022)
 -------------------
 * Change language id to "my" to allow Microsoft Word to pick up the correct locale.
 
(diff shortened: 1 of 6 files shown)


keyman-server:

LornaSIL:
There are two places where the language tag is listed in a keyboard package. One is in `keyboard_info` and one is in the `.kps` file. You need to make them consistent. My recommendation is to leave `shn` in the keyboard and add `my` if you wish. That way when applications better support `shn` they can begin doing so without having to update the keyboard again. 
So, `keyboard` info may need the line changed to:
```
    "languages": [
        "my",
        "shn"
    ],
```

LornaSIL:
I'm sorry I wasn't clear. Could you add `shn` back into the .kps file so it lists both `my` and `shn`? Then I think it's okay.

LornaSIL:
@rdai I can make this change if you don't have time to do so.

rdai:
Sorry, life got busy, changed per your comment. Please verify? Thanks.

LornaSIL (approved):
LGTM

PR #1869, opened 2022-06-02
fv_tlingit: desktop keyboard and mobile updates

caforbes:
- added desktop keyboard (based on languagegeek Tlingit-Yukon in use by YNLC)
- added keys for macron to desktop/touch
- improvements to diacritics on desktop and touch
- documentation etc.


keyman-server:

DavidLRowe:
Looks good! A few comments:

In the .kmn file, line 37 is:
```
+ [SHIFT K_BKSLASH] > '̨'

```
Version 15 of the Keyman compiler warns that this rule will never be matched (since the rule on line 27 will take precedence). Not sure what the intent of the rule is, but it will never do anything as currently written, so should be revised or removed.

Very minor: The comment on line 36 mentions 'right-alt' but the rules actually use ALT (not RALT). So you might want to change the comment.

I see the line for the icon file is commented out in the .kmn file. Do you want the icon for the keyboard to be something other than the default Keyman key? If so, you can uncomment this line and include the .ico file. Not a problem either way.

Currently you have the welcome.htm file in the source folder and the .png file it references in the help folder. When these two files are included in the package, the linkage won't work because they will end up in the same folder. One solution would be to make a welcome folder under the source folder and put the welcome.htm and a copy of the .png file in that folder. Then (after adjusting the reference inside the welcome.htm file), they will work locally as well as when included in the package.

Optional: BTW, we have relaxed requirements for the keyboard version number and dates in the copyright statement. The only copyright statement that requires the date (or date range) is the LICENSE.md. Everywhere else you can just use (c) First Voices etc. We need the version number in the .kmn file and in HISTORY.md, but it can be, for example, omitted from README.md. This may help reduce the number of files that need to be updated when the keyboard is revised.

DavidLRowe:
I'm not sure that the .php file in the help folder is working properly. I'll have to check tomorrow whether there's a problem with my setup or with that file. 

It also seems that that file is not referencing the .png file at all, so you could just move the .png file to the newly-created welcome folder for use with welcome.htm. (No need to keep a copy in the help folder if it isn't being used there.)

It also occurred to me that since you have so much extra space on the touch layout, would you want to add five keys for AEIOU+ ogonek.

DavidLRowe:
The .php file is fine. If you want, you could include the numeric layer in the display of touch keyboard layers by using:
```
id='osk-tablet' data-states='default shift numeric'
```
(that is, adding 'numeric').

caforbes:
Thanks for the comments! Great catches! Fixed:
- line 37 was a duplicate, removed
- fixed comment on line 36 regarding (right) alt
- moved the .png to the source folder with welcome.htm

stuff I didn't change:
- I didn't edit line 6 re: the icon, or at least didn't intend to change any icon behavior, it must be from the original template used to build this so I'll just leave it.
- it's my understanding that ogonek is a niche character for this language and there were no user complaints about it, so I will leave the touch layout as-is.

DavidLRowe:
Oops. I missed the fact that the package now needs to have the .kmx and .kvk files included from the build folder. Also the .png file reference is still pointing to the help folder, so that needs to change since the file is now in the source folder.

Hopefully that should be all!

DavidLRowe (approved):
I took the liberty of correcting the references in the package file, so that this keyboard update could be finalised.

PR #1873, opened 2022-06-08
[colchis_latin] Revised mobile layouts per user request

LornaSIL:
Added tablet layout and revised phone layout.


DavidLRowe:
tablet layout, rightalt layer, the key to the right of "æ" is blank, but produces "Æ". 

phone layout, default layer, typing ç twice produced çç rather than ç̆ (Also for upper case)

PR #1876, opened 2022-06-10
[ajami_hausa] Update to use Alkalami and Ruwudu fonts

LornaSIL:
No change to the keyboard.

I also updated the Padauk font while I was at it.


DavidLRowe:
Is it intentional that the change to hausa_ajami_qwerty.php removes information about rotas? Is this bringing the documentation in line with the actual function?

The v15 compiler notes for line 93:
This rule will never be matched as another rule takes precedence Overridden by rule on line 92

Not sure if you want to do anything about these or not. I'll approve this PR as is, and you can decide whether to make changes or not.

LornaSIL:
Since the keyboard is working, I will merge it. The warning doesn't make sense because the two rules do not interact at all. And when I comment out line 93 the whole base layer stops functioning.

>Is it intentional that the change to hausa_ajami_qwerty.php removes information about rotas? Is this bringing the documentation in line with the actual function?

This was a duplicate section. It is still in there elsewhere.

PR #1884, opened 2022-06-17
[sil_jarai] Jarai ចារាយ (SIL) keyboard

MakaraSok:


MakaraSok:
known issue: https://github.com/keymanapp/keyman/issues/6792

DavidLRowe:
Other than the failure to build on TeamCity, everything looks fine except one comment:

In welcome.htm, should the "To stay on the Shift layer, double tap on the Shift key." sentence be moved down to the "Touch layout" section (as is the case in sil_jarai.php)?

DavidLRowe (approved):

PR #1887, opened 2022-06-21
[vietnamese_vni] New Vietnamese keyboard

miketvo:
My implementation of the Vietnamese VNI input scheme: https://en.wikipedia.org/wiki/VNI#VNI_Input_Method

Heavily based on my other keyboard [Keyman Vietnamese Telex](https://keyman.com/keyboards/vietnamese_telex).


keyman-server:

LornaSIL:
Have you upgraded to Keyman 15? It might be helpful to do so as it gives helpful warnings. There are a LOT of warnings with messages such as:
`vietnamese_vni.kmn (365): Hint: 10AE This rule will never be matched as another rule takes precedence Overridden by rule on line 360`

The rule on line 365 is indeed a duplicate of the rule on line 360. Since this is such a large keyboard file, I believe it would be important to eradicate these warnings. There are 1184 of these type of warnings. It's possible it wouldn't take so long to compile.

I didn't do any other testing, but I did look at all the peripheral documentation and it seemed good.

miketvo:
Thanks Lorna!

I was indeed using Keyman Developer 14 and its CLI tools.

After upgrading to version 15, I was able to reproduce 591 of those duplication warnings you mentioned during the keyboard build process for desktop devices; and 591 of the same warnings during the keyboard build process for mobile devices. This makes to a total of 1182 warnings, 2 less than what you found. Perhaps you overcounted because of these 2 lines in the log that were inserted by the compiler between the two build processes?

```
...
vietnamese_vni.kmn: Success: '%path_to_project%\keyboards\release\v\vietnamese_vni\source\vietnamese_vni.kmn' was compiled successfully for Windows, macOS, Linux, Desktop devices to '%path_to_project%\keyboards\release\v\vietnamese_vni\build\vietnamese_vni.kmx'.
vietnamese_vni.kmn: Compiling '%path_to_project%\keyboards\release\v\vietnamese_vni\source\vietnamese_vni.kmn' for Web, iPhone, iPad, Android phone, Android tablet, Mobile devices, Tablet devices...
...
```

I'm working on addressing these warnings. I'll let you know when I'm done.

On the same subject, I also tested my other keyboard [vietnamese-telex](https://github.com/keymanapp/keyboards/tree/master/release/v/vietnamese_telex) against Keyman Developer 15 and found that it also has 607 unique warnings of the same type. Should I open a new separate PR to address that?

miketvo:
Hi Lorna, it's fixed. There should be no warnings now. Functionally no features were affected.

LornaSIL (approved):
LGTM

PR #1890, opened 2022-06-24
[basic_kbdrum] Russian - Mnemonic Basic Improve dead-key pair processing

dotland:
This change makes it possible to input some words without a space between dead keys. For example, the Russian word *сюжет* can be typed without a space between the dead keys `S` and `Y`.

In the following repo, I have rewritten the Russian Mnemonic keyboard layout with improved dead-keys processing, complementary special and math symbols accessed by the keys `AltGr`, `/` and `\`.

[https://github.com/dotland/mnemonic-kb-ru](https://github.com/dotland/mnemonic-kb-ru)

### DIFF
--- a/release/basic/basic_kbdrum/source/basic_kbdrum.kmn
+++ b/release/basic/basic_kbdrum/source/basic_kbdrum.kmn
@@ -69,28 +69,28 @@ group(main) using keys
 + [CAPS SHIFT K_C] > dk(0446)   c ц (small)
 
 c цц (cc)
-dk(0446) + [NCAPS K_C] > U+0446 U+0446
-dk(0426) + [CAPS K_C] > U+0426 U+0426
-dk(0446) + [CAPS SHIFT K_C] > U+0446 U+0446
-dk(0426) + [NCAPS SHIFT K_C] > U+0426 U+0426
+dk(0446) + [NCAPS K_C] > U+0446 dk(0446)
+dk(0426) + [CAPS K_C] > U+0426 dk(0426)
+dk(0446) + [CAPS SHIFT K_C] > U+0446 dk(0446)
+dk(0426) + [NCAPS SHIFT K_C] > U+0426 dk(0426)
 
 c цс (cs)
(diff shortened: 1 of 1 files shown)


LornaSIL:
Can you clarify if your change is intended to make the keyboard better follow the Windows keyboard? Or is it intended to provide different behavior? If this is to make the keyboard better follow the intended Windows keyboard layout, then it's fine to update this keyboard and we will check it. I do know we got some of the behaviors wrong (with dk) when we imported them. 
If it's intended to give enhanced behavior it would be better to rename the keyboard.

dotland:
I think that every keyboard layout should process adjacent dead keys as described in this change because it is the natural way to input dead keys consequently. You can ensure that the dead keys are implemented this way in macOS. 
Let's consider for example the Russian word **сюжет** (transliterated **syuxet**). When inputting that word by the keyboard Russian - Mnemonic (Windows), we have to type a space after **s**, because otherwise, we would get the word **сыужет**. And this change eliminates the need for such redundant spaces during inputting a word and therefore speeds up typing. That is, when typing **sy**, the input engine should be in awaiting state of the dead key **y**.
I think Microsoft should review the implementation of dead keys and it can be improved as is done in macOS.

LornaSIL:
From my testing the output of the Windows **Russian - Mnemonic** keyboard with the changes you have made, I believe the _behavior_ of the keyboard is different from the Windows keyboard. Even though it is only a slight difference, I'm going to have to ask you to resubmit your keyboard under new filenames and new keyboard name. This is because the **basic** keyboards are supposed to follow the Windows keyboard layouts.

You will also need to copyright it someone else (such as yourself).

All of this is very much as you did for the Armenian Mnemonic keyboard.

dotland:
Indeed, I followed the same design as for the Armenian Mnemonic keyboard.
I agree to have two variants of the Russian Mnemonic keyboard. I will submit the new keyboard with the filenames russian_mnemonic and with the keyboard name Russian Mnemonic R  (R for Redesign, Rewrite, Restyle) to differentiate it from the Windows variant. It will also include supplementary special and math symbols.

PR #1891, opened 2022-06-26
[basic_kbdhym] Initial submission of basic_kbdhym keyboard

dotland:
This is the initial submission of the Armenian Mnemonic keyman keyboard. It supports any desktop, mobile and web platform. The full documentation can be found at [https://github.com/dotland/mnemonic-kb-hy/blob/main/README.md](https://github.com/dotland/mnemonic-kb-hy/blob/main/README.md).


keyman-server:

LornaSIL:
Thanks for this keyboard submission! I haven't checked the keyboard, but I have two initial questions. We use the `basic_kbdxxx` for any keyboards that we've imported from Windows keyboards. We have permission to do that, but unless this is a Windows keyboard, you should rename the files and put them in another folder. It would be better to rename it something like `armenian_mnemonic` or something like that. 
Also, the only fonts we are allowed to distribute are ones whose license allows for distribution. Sylfaen is a Microsoft font which we are not allowed to distribute in Keyman packages. You'll need to just include in the documentation the information that Sylfaen is a good font to use, or you could include a Noto font whose license is OFL which we can include in Keyman packages.

dotland:
I got it. Thanks for the explanation of the project structure and what fonts to choose. I have chosen the **Sylfaen** font because it is nice and it has Armenian ligatures within. There are fonts with Armenian ligatures (e.g. **Arial Unicode MS**, **Mshtakan**, **TITUS Cyberbit Basic**, **GHEA Mariam**), but they are not in the list of [OFL Fonts](https://scripts.sil.org/cms/scripts/page.php?item_id=OFL_fonts). So I will choose the default Tahoma font, I will rename the Keyman project, and create a new PR placing the source files under the directory `keyboards/release/a/`.

dotland:
Created a new PR [1892](https://github.com/keymanapp/keyboards/pull/1892/).

PR #1892, opened 2022-06-28
[armenian_mnemonic] Initial submission of armenian_mnemonic keyboard

dotland:
This is the initial submission of the Armenian Mnemonic keyman keyboard. It supports any desktop, mobile and web platform. The full documentation can be found at [https://github.com/dotland/mnemonic-kb-hy/blob/main/README.md](https://github.com/dotland/mnemonic-kb-hy/blob/main/README.md).


keyman-server:

LornaSIL:
I haven't tested the keyboard yet, but structurally it all looks good. Would you be able to create a `armenian_mnemonic.php` file that would go in a `source\help` folder? In general we would expect the contents of that file to match what is in the welcome.htm file. This is what would show on the online help page. It would need the html header stuff stripped and php header added in. If you don't feel comfortable doing it we could add it as soon as the PR is merged.

dotland:
Strongly agree. Indeed, frequently updated version and date should be defined in one place.
Please, change as you described. 

Thanks

LornaSIL (approved):
LGTM

PR #1913, opened 2022-07-11
chore: cleanup build scripts 🚦

mcdurdin:
This is part of a larger set of PRs coming which will automate the build of keyboards during the Keyman Developer build.

Refactors and rearranges various parts of build scripts:

* Moves functions to search for kmcomp, etc to environment.sh -- this way we can use the search functions from any script

* Moves parse_args from util.sh back to build.sh -- parse_args is specific to build.sh so should not be in a shared util script

* Moves kmcomp downloaded files to tools/kmcomp/ and cleans up older compiler files in tools/ -- this makes it much easier to cut and shut new versions of kmcomp

* Adds color flags to build.sh and related scripts for controlling color output in logs

* Cleans up some mess from .gitignore and ensures that kmcomp and future regression outputs will be ignored.

Note: additional cleanup of build scripts is in follow-up PR #1914.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,17 +8,21 @@ packages/*/build/
 .cache
 servervars.sh
 /.project
-release/j/jawa/jawa.kpj.user
-release/j/jawa/source/jawa-help.htm
-release/j/jawa/source/jawa.keyman-touch-layout
 .vscode/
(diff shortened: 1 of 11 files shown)


rc-swag (approved):
lgtm

rc-swag (approved):
lgtm

PR #1914, opened 2022-07-11
feat: regression builds 🚦

mcdurdin:
This is part of a larger set of PRs coming which will automate the build of keyboards during the Keyman Developer build.

Adds scripts to do a full build of .kmx and .js files with a specific version of the compiler. Puts results into `output/<version>/` so that we can easily run diff scripts against different versions.

Will first be used in the Keyman Developer CI build, which will download latest version of compiler in the same tier, as well as using the build within the Keyman Developer `developer/release/<version>` folder, which should make for reasonably straightforward comparisons.

If the kmcomp.zip archive includes kmdecomp.exe, the script can also decompile the .kmx keyboards back to .kmn, again in the `output/<version>/` folder, so that we can do textual comparisons of the results.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,5 +1,8 @@
 #!/usr/bin/env bash
 
+set -e
+set -u
+
 #
 # This script is built with commands available to Git Bash on Windows. (mingw32)
 #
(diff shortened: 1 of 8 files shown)


mcdurdin:
See https://build.palaso.org/buildConfiguration/Keyman_Developer_Test/327110 for sample build with regression test results.

rc-swag (approved):
lgtm

PR #1915, opened 2022-07-11
Initial submission of Nisenan keyboard

JGearySSBand:


keyman-server:

DavidLRowe:
Thank you for submitting this keyboard. The files are in very good shape. I just have a few observations:

(1) You use Ctrl as a modifier. This could interfere with other computer programs, for example, Ctrl+C is often used as a shortcut for Copy, but with this keyboard it will generate ć instead. Often people use the RightAlt layer for added characters (and Shift-RightAlt for upper case). I'm willing to approve the keyboard as it is, but you may want to consider such a change. (Note, however, that Alt-Shift-K and Alt-Shift-O are, by default, hot keys used by Keyman itself, so if you do opt for using the RightAlt key, you may need to include instructions to help users change or disable those Keyman hot keys.)

(2) It is possible with this keyboard to type Ctrl+o and get ŏ (U+014F), but also to type "o" then Ctrl+- to get ŏ (U+006F U+0306). Computer programs are supposed to treat such equivalent encodings identically, but sometimes that is not the case. It would be possible (either now or sometime in the future should the problem arise) to add logic to the keyboard to ensure that only one form is produced no matter which way it is typed.

(3) For the touch layout, the shift key on the numeric layer currently does nothing. If that is intentional, that's fine. Alternatively, you could have that key take you to the shift layer. Either way is fine.

(4) The welcome.htm file provides the help that is packaged with the keyboard. We will need an equivalent PHP file to provide the online help. To create the PHP file:

Start with a copy of the text of your welcome.htm file.
Delete everything from the top of the file down through and including the `<body>` element and the first `<h1>` element.
Delete the 
```
</body>
</html>
```
at the end of the file.
Then add the following heading at the top of the file:
```
<?php 
  $pagename = 'Nisenan Keyboard Help';
  $pagetitle = $pagename;
  require_once('header.php');
?>
```
Save as nisenan.php, in a "help" folder in the "source" folder.

Copy the two image files ("keyboard_ctrl_layout.png" and "keyboard_ctrlshift_layout.png") referenced by the text from the source folder into the "help" folder you just created. 

It is possible for us to accept the keyboard without the online help and then add it later on by using the welcome.htm information.

mcdurdin:
> (Note, however, that Alt-Shift-K and Alt-Shift-O are, by default, hot keys used by Keyman itself, so if you do opt for using the RightAlt key, you may need to include instructions to help users change or disable those Keyman hot keys.)

This is no longer an issue as Keyman hotkeys only operate with the Left Alt key, so any combination using Right Alt will be fine.

JGearySSBand:
Se˙kas’ĭn David, Marc,
 
Thank you both for your help! I made the changes you both suggested.
 
1)	I changed all uses of Ctrl to RightAlt.
2)	I added some code to change combinations of characters (e,i,o,u +  ̆, c,k,p +  ́ ) to the appropriate single character (ĕ,ĭ,ŏ,ŭ, ć,ḱ,ṕ), and it looks like that is working on my end.
3)	Huh. I assigned the shift key on the numeric layer to take the user to the shift layer.
4)	I made the “help” folder as requested.
 
I’ll go ahead and reupload these to github. Thanks again for your help!
Jonathan

DavidLRowe:
Replaced by #1917 so closing this one.

PR #1917, opened 2022-07-12
Initial submission of Nisenan keyboard - Attempt 2

JGearySSBand:
Resubmitting files for the initial submission of Nisenan keyboard, based on feedback from DavidLRowe and mcdurdin.
 
(My apologies in advance if I shouldn't have uploaded the files this way! Github always trips me up...)


keyman-server:

DavidLRowe (approved):
The only problem was the file list in the package source (.kps) file that still had "ctrl" rather than "ralt". I was able to change that (though it took me two tries to get it right!).

Thanks again for your submission.

PR #1920, opened 2022-07-14
fix: add false parameter for rsync

mcdurdin:
After enabling `set -u` a missing parameter for rsync upload was encountered. This fixes that.


rc-swag (approved):
lgtm

PR #1922, opened 2022-07-18
Sil yupik cyrillic ru

JBrase:
Note both Russian and English versions of the welcome and help files.


keyman-server:

DavidLRowe:
(1) This pull requests contains two files that belong with another keyboard:
```
release/sil/sil_yupik_cyrillic/sil_yupik_cyrillic.kpj
release/sil/sil_yupik_cyrillic/source/help/sil_yupik_cyrillic.php
```
Is that intentional? We generally limit information in a pull request to just one keyboard (unless it's the same small change that is needed for a number of keyboards, then we might lump them together).

I think you need to remove those two files from your sil_yupik_cyrillic_ru keyboard (which will then remove them from this pull request).

(2) Likewise a spurious "setup.log" file slipped into the files in this pull request. It should be deleted.

(3) We no longer require the date to be included in the copyright statements, except for the LICENSE.md file. The LICENSE.md file is fine, but the .kmn file still has 2020. Please either update that to 2022 or remove the date range from the .kmn file. (Actually I see it has 2009-2020. Is 2009 a typo for 2019? Or should the dates for the LICENSE.md change to 2009-2022?)

At any rate, they should all be consistent. Other files (README.md, readme.htm, .kps, all three .php files, all three welcome*.htm files) also need a consistent copyright statement. I personally think the best thing is to remove the date from everything but the LICENSE.md file to reduce the number of files (where you're just changing the date) when you update the keyboard.

(4) On the touch layout, numeric layer, the "/" and "\" produce the opposite of what's displayed. You may want to use U_005C for "\" and U_002F for "/" to avoid linking these keys to the .kmn file.

(5) Additional comment: I'm curious why you included uppercase longpress keys on the default layer. For example, the "а" key has longpress keys of "а̄" and "А̄", whereas I would have expected only "а̄". Up to you whether or not you change that.

PR #1926, opened 2022-07-21
Sil tai dam typewriter

JBrase:


keyman-server:

DavidLRowe:
@JBrase This PR contained two .kpj files that belonged to other keyboards. (Hence the Conflicting files message.) I thought I could just remove the files from the PR by deleting them (see the two commits I made). This seems to have worked for one file, but not the other, and the conflict message still lists both.

It may be easiest to close this PR and start another one. Or maybe you can see how to remove the spurious files from the PR (without the PR actually deleting the files from other keyboards!).

DavidLRowe:
replaced with #1952

PR #1929, opened 2022-07-27
correct font name

MakaraSok:
>    "source": ["khmer_busra_kbd.ttf"]
> This should have been khmer_busra_osk.ttf! We will need to fix and re-deploy... 😢


LornaSIL:
Do you want to fix the family name so it's not confusing?

mcdurdin:
The font is in use already so I'm not sure it's worth the effort of renaming it? We'd either need to update the font to change its family name internally, or rename the file (I think the second option would be better in some ways), but both changes would mean updates to other keyboards as well.

LornaSIL:
I meant in keyboard_info file.

mcdurdin:
I wouldn't want to use one font family here and another say for khmer_angkor -- I think that would be worse than the current situation :grin:

I also would like to continue to use the actual font family name from the font itself, so if a user has the font installed, it doesn't download it.

mcdurdin:
Now that I look more closely, khmer_busra_osk.ttf is only referenced by this keyboard and khmer_angkor. So it's not that big a deal to fix the references -- we don't even need to update khmer_angkor version as this doesn't need to deploy.

So, after all I just said above, let's go and rename the font to khmer_busra_kbd.ttf, instead, and update s.keyman.com. This is a one-time fix, after all, and then we won't have this confusion in the future!

DavidLRowe:
@MakaraSok Will you be able to make the changes @mcdurdin has suggested?

DavidLRowe:
Is everyone (@mcdurdin @MakaraSok @LornaSIL ) happy with these changes?

mcdurdin (approved):
Changes LGTM

PR #1930, opened 2022-07-28
[armenian_mnemonic_r] Initial submission of Armenian Mnemonic R keyboard.

dotland:
In Armenian, there are soft and hard letters for `T` and `P`.
The original Armenian Mnemonic keyboard uses soft letters for keys  `T` and `P`, that is **Թ** and **Փ** respectively.
This keyboard uses hard letters for keys  `T` and `P`, that is **Տ** and **Պ** respectively.


keyman-server:

LornaSIL (approved):
LGTM

PR #1939, opened 2022-08-02
[khmer_angkor] update OSK and documentation

MakaraSok:
Updates:

- OSKs are now having combining vowels rendered nicely together, i.e. ុំ, ាំ, ុះ, េះ, ោះ
- Documentations in both EN and KM have also been updated accordingly


DavidLRowe:
The .kps file references `..\..\..\..\..\..\Users\Keyman\Documents\splash.gif` which can't be resolved. If you want to use this file, you need to place a copy in the source folder.

You may want to look at the `This rule will never be matched` warnings. Maybe they are fine.

MakaraSok:
Thanks for catching these. 💯 

The `.gif` will be updated. I've missed that in the process.

The warnings are ignored as the keyboard's rules will be revisited.

MakaraSok:
Will this be conflicting with the changes made in https://github.com/keymanapp/keyboards/pull/1929?

DavidLRowe (approved):

PR #1951, opened 2022-08-09
Changes to miluk_hanis_siuslaw keyboard

jagdouglastavani:
Hello there. I believe I have made the necessary changes to my keyboard. Thank you for you support and troubleshooting.


keyman-server:

DavidLRowe:
Hi Jordan,

It's looking better!

The HISTORY.md file is good. 
The HISTORY.md.txt file should be deleted.
The LICENSE.md file is good.
The LICENSE.md.txt file should be deleted.
(Windows by default hides some file extensions, so it may be hard to see the difference.)
It would be helpful to have a more informative description in the README.md file. You could use what you wrote before in the README.md.txt file.
The README.md.txt file should be deleted.
The .keyboard_info file looks good.
We need the .kpj (keyboard project) file, but...
the .kpj.user file (where Keyman Developer stores some configuration info) should be deleted.

The .kmp, .kmx and .kvks files need to end up in the "build" folder, not the "source" folder. They should be deleted from the source folder. (And, in fact, these compiled files should not be part of the keyboard submission at all. They will be built from source by Keyman's automated build system.)

The readme.htm file is displayed as a user is on the point of installing the keyboard. The description you have is good, but it would be better to remove the "miluk_hanis_siuslaw 1.0 generated from template." line right before it, since that won't mean anything to users. You wrote "these keyboards" (perhaps referring to the desktop keyboard and the touch keyboard?), but we generally say "this keyboard", since the desktop and touch layouts are generally related.

The welcome.htm file provides the help that is packaged with the keyboard. It is displayed after the keyboard is first installed or when a users asks for help. It needs to have enough information to enable a user to figure out how to type on the keyboard (at least the desktop layout). This can be done by listing the keys that are used to make a character not found on the keyboard, or by including an image file that shows where keys are located on a keyboard, or a combination.

Additionally we need a help file in PHP format. This is often a copy of the body of the welcome.htm file with an added PHP header. I can point you to some examples, if you want to make this file, or we can add it later based on the information in the welcome.htm file.

I haven't tested the keyboard, but I see in the .kmn file that you have some inconsistent references. The .kmn file has 

'CLUSI 1.ico'
'CLUSI 1.kvks'
'clusikeyboard1.keyman-touch-layout'

but the files in the source folder are actually (and correctly):

miluk_hanis_siuslaw.ico
miluk_hanis_siuslaw.kvks
miluk_hanis_siuslaw.keyman-touch-layout

The .kmn file also references two files that don't exist ('CLUSI 1-help.htm' and 'CLUSI 1-codes.txt') which aren't usually needed by keyboards.

Let me know if anything is unclear.

jagdouglastavani:
Hi David, 
I believe I've addressed the points that you've made. If you are willing to make the PHP file, I would greatly appreciate it. As you can tell, this isn't exactly something I'm incredibly proficient in.

DavidLRowe:
If it's any consolation, my first keyboard submission had lots of problems!

There are still a few issues with the file structure, but I can now look at the actual keyboard.

For the desktop layout, it seems that this language is using only lower case. Is that correct? You should be aware that any key that you don't map ([SHIFT K_M} for example) will output the character on the US English QWERTY layout ("M" in the example). If you don't want that, you'll need to specifically output `nul` for such combinations.

You don't need touch layouts for both "tablet" and "phone", unless they are different. Keyman will use the phone layout on a tablet if not table layout is available and vice versa. Trying to keep two identical layouts in sync is extra work. I'd suggest you delete the tablet layout and just work with the phone one.

The touch layouts (both "tablet" and "phone") reference a "shift" layer which doesn't exist. If you don't have uppercase characters, you can eliminate the shift key (and therefore the reference to the non-existent shift layer).

jagdouglastavani:
Thank you David. 

I've put nul values for the shift characters. It is indeed a writing system without capitalization -- so removing the option for default QWERTY behavior is definitely the goal.

I've removed the tablet layout as per your advice. I also removed the shift layer from the mobile.

Thank you!

DavidLRowe:
@jagdouglastavani There were still some structural issues with your pull request. I took your files and submitted them as a new pull request (#1961). You can see that pull request for details. 

I'm also going to attach a zip file of what I submitted. It would be most helpful if you could archive all your existing copies of the files associated with this keyboard, then unzip this file and use the resulting folder for future work.

I put your keyboard in the "experimental" section of the keyboards repository. The keyboard will show up the same as if it were in the release section, and you can have people try it out. If you want to change the keyboard, make sure that the version number is changed (from 0.1 to 02, for example) and submit a new pull request.

Thanks for tacking this project. I hope that all goes well with your work.

PR #1952, opened 2022-08-09
[sil_tai_dam_typewriter] fixing repo

JBrase:


keyman-server:

DavidLRowe:
The only thing I wonder about is that the .keyboard_info file has "blt-tavt" and the .kps file has "blt-Tavt-LA". Normally these should match. It doesn't seem that the -LA is needed. (That is, the language wouldn't be written differently in the Tavt script between VN and LA, would it?)

DavidLRowe:
There are 40 or so extraneous files that need to be removed from this pull request. (Or else this PR can be closed and another one submitted.)

JBrase:
Files copies to wrong place. Need to clean up repo.

PR #1954, opened 2022-08-10
[sil_yupik_cyrillic_ru] new keyboard

JBrase:
Removed old branch and set up new one to correct repo errors.
Corrected errors in touch screen layout: slash & backslash swapped output; removed upper case chars from long press output.


keyman-server:

DavidLRowe:
This is looking very good. There are some changes needed in a few files.

The .keyboard_info file has some "related" entries. (I don't understand this very well, so I consulted Lorna.)
- There is no "cyrillic" keyboard, so it won't work to deprecate a non-existent keyboard.
- You should not have a reference to this keyboard itself.
- If you want to indicate that the sil_yupik_cyrillic keyboard is related, you can use something like:
```
"related": {
    "sil_yupik_cyrillic": {
      "deprecates": false
    }
``` 
but you need `"deprecates": false` so that you don't deprecate the sil_yupik_cyrillic keyboard.

It would also be fine to omit the "related" section entirely.

The language tag in the .keyboard_info file should be "ess-Cyrl" (since "ess" implies "ess-Latn").

The language tag in the .kps file should be "ess-Cyrl" (rather than "ess-Cyrl-RU").

The language tags in the two files should match.

The sil_yupik_cyrillic_ru.php file references "sil_yupik_cyrillic_ru_english.php" but the filename actually has a capital "E" (sil_yupik_cyrillic_ru_English.php), so this is likely to fail on any non-Windows machine. The best thing would be to rename the file, since the reference already has lower case. But it may be tough to do that on Windows (perhaps rename it to x.php, then rename x.php to the lower case version?). The alternative would be to change the reference in the file to match the case of the filename.

DavidLRowe (approved):

DavidLRowe:
@JBrase This keyboard should be available at https://keyman.com/keyboards/sil_yupik_cyrillic_ru shortly (within a few hours usually).

PR #1964, opened 2022-08-19
fix: Lowercase png and jpg filenames for help

darcywong00:
While working on migrating the help.keyman.com site from Windows to Linux, I noticed the case-sensitivities of PNG, JPG, etc filenames were triggering some broken links.

https://github.com/keymanapp/help.keyman.com/pull/588#issuecomment-1220335292

This standardizes the filename to lowercase (e.g. from `.PNG` to `.png`. )
I'm not looking to re-deploy any keyboard packages. Just hoping this will cause the appropriate help.keyman.com files will pick up the updated filenames...

Scope:
I **won't** be changing filenames inside  legacy .kmp files.


release/u/uma_phonetic/source/welcome/welcome.htm
```diff
@@ -20,7 +20,7 @@ <h1>Start Using Uma Phonetic</h1>
 
 <h1>Keyboard Layout</h1>
 
-<img src="keyboardlayout.PNG" alt="Uma Keyboard Layout">
+<img src="keyboardlayout.png" alt="Uma Keyboard Layout">
```
darcywong00, line 23: The .php help file doesn't use the png file

darcywong00:
Hmm. I think a similar change will be needed for .JPG files too. Putting this back to draft

mcdurdin:
I'm not sure if you've done this already or not, but can you make sure that .php and welcome.htm references to these images are all updated to use the same casing -- otherwise they'll break on case-sensitive file systems?

LornaSIL (approved):
LGTM

PR #1967, opened 2022-08-25
[dogra_inscript]   Initial submission

WseriesY:


keyman-server:

LornaSIL:
Thanks for your keyboard submission! We don't have any keyboards with the Dogra script yet, so this will be helpful to people.

There are several issues that need to be addressed.
- The first is that the keyboard folder has been put in the wrong place in the repo. It is in `keymanapp\keyboards\d\dogra_inscript` and it should be under `keymanapp\keyboards\experimental\d\dogra_inscript` _or_ `keymanapp\keyboards\release\d\dogra_inscript`. We are more rigorous in our checking of keyboards if they are in "release".
- Next, the copyright date in LICENSE.md is `2022-2031`. I think that needs to change to just `2022` for now. You can extend the date when/if you update the keyboard in subsequent years. You also have those dates in other locations. Those should either be changed to `2022` _or_ you could entirely remove the date from the other locations. We only require it in LICENSE.md now.
- The keyboard_info file has information that is included on the main distribution page for this keyboard. I'd recommend you include more description. I've suggested some wording below. Also, I don't think mobile devices will include support for the Dogra script without the font. I'd suggest you replace everything in the keyboard_info file with what I've got below. This will mean the font will be used in the Keyman app on Android or hopefully ios:
```
{
    "license": "mit",
  "languages": { 
       "doi-Dogr": { 
            "font": { 
            "family": "Noto Serif Dogra", 
            "source": "NotoSerifDogra-Regular.ttf" 
         } 
      }
   },
	  "description": "This keyboard is designed for the Dogri language, using the Dogra script with an Inscript layout"
}
```
- For the .kmn file:
  - If you could add this line at the top `store(&VERSION) '10.0'` it will let Keyman know the minimum version of Keyman that is recommended.
- For the touch layout:
  - I would recommend that you delete either the tablet or the phone layout. They are the same layout and it will be easier to maintain just one layout in that case. If you don't have a tablet layout, the phone layout will be used on a tablet...and vice versa, so there is no need for both.
- In the .kps file:
  - the font has a wrong path. You'll need to remove the font and then re add it.
  - You could include the readme.htm and the welcome.htm in the list of files. I would recommend you use the same description in those files as you use in the keyboard_info file. If you want to make the welcome.htm file even more useful you could include a graphic of the layout. In that case you would want to make sure to include the graphic in the .kps.
  - On the "details" tab, you could remove the version number and tick "Package version follows keyboard version". That will make it easier to update the files the next time.

That's it for now. I know it seems like a lot. Much of this is required for our build system to have things in the proper place. Once you resubmit, I'll recheck everything. If you want to resubmit in a clean Pull Request that might be easier for you.

PR #1969, opened 2022-08-27
Release of Chechen Latin based keyboard

khashashin:


keyman-server:

LornaSIL:
Thanks for your submission! In general it looks very good. I have a few suggestions and questions:
I see these two lines in the .kmn file:
```
store(&KMW_HELPFILE) 'chechen_latin-help.htm'
store(&INCLUDECODES) 'chechen_latin-codes.txt'
```
Are those needed? The files are blank. It seems like those lines (and files) could be removed.

The description in the README.md file is very good. It would be helpful for you to add that to the readme.htm or the welcome.htm. The README.md file is primarily just there in the repo and so having that information in either the readme.htm or welcome.htm file would be a great addition. 

I see several places where there is a reference to version 1 (README.md, readme.htm, welcome.htm and chechen_latin.php). You've properly done a version bump in the .kmn file and you have the .kps file following the version number from the .kmn so my recommendation is to remove the version number in these other files. That makes it easier to maintain the keyboard in the future so you don't have to update the version in so many places. 

Otherwise, I think it looks good.

khashashin:
@LornaSIL thank you for your feedback. I have now made the changes you mentioned. You are right, it would be much easier to maintain the version if we kept it simple. Otherwise there could be confusions, this PR was a good example of that.

I also added description in `readme.htm`

LornaSIL (approved):
LGTM

LornaSIL:
@khashashin Your new version is now online. We had some issues with our build system, but I believe everything is fine now.

khashashin:
@LornaSIL thank you!

Issue #1942, opened 2022-08-03
chore(infrastructure): Add woffs to shared folder


LornaSIL:
...in preparation for automatic deployment
If a woff and woff2 exists, just include woff2

Could also update if Noto Sans is in multiple keyboards...put in shared folder and include woff2 if available.

LornaSIL:
@mcdurdin Is it okay if I put the fonts in a folder within the font folder called `web`? This would better follow how we like to distribute font packages. You would then need to automaticly deploy from the `web` folder, eg `release\shared\fonts\sil\charis\web`.
Is that okay?

mcdurdin:
Yes, fine to organise fonts accordingly. I need to make sure I don't lose track of this task for automated deployment of fonts -- I think it's in our discussion notes. Note that we have a challenge around versioning of fonts -- what do we do when different keyboards have different font versions?

LornaSIL:
My assumption is that usually fonts are enhanced and only rarely do they have behavior removed, so I would expect a newer version is better.

DavidLRowe:
@LornaSIL Any more work to be done before this issue is closed?

LornaSIL:
I guess I should re-review the fonts. The goal was for @mcdurdin to add functionality so the fonts would automatically be copied to s.keyman.com so I wouldn't need to do that extra step. I'm not sure where we are in that process either.

mcdurdin:
> I'm not sure where we are in that process either.

Me either. No progress on that. I can't find an open issue on it so I think I've just forgotten about this altogether. Realistically, nothing will happen in the next couple of months on that.

DavidLRowe:
Replaced by #2323

PR #1974, opened 2022-08-30
[sil-woff] Add SIL woff/woff2 fonts to shared folder

LornaSIL:
These will hopefully be used for automatically updating the s.keyman.com repo.


LornaSIL:
Partially implements #1942

DavidLRowe (approved):

PR #1979, opened 2022-09-01
[shaw_2layer] Add help file and font to package

LornaSIL:
I was planning to just add a help file and then I realized we didn't have font support so I added that with permission of author.
@JapanYoshi


DavidLRowe (approved):
LGTM

PR #1989, opened 2022-09-07
[piqad] Initial submission

berrymot:
Notes:

- it should be capitalized as `pIqaD` *exactly* - no capital p if you can :)
- `qoqQuch` (the name in the copyright section) is a Klingon translation of "merrybot", which my github username is a variant of

Includes an onscreen keyboard as well as touch layouts.


keyman-server:

berrymot:
also yes this took 18 commits because I'm bad at github

release/p/piqad/source/piqad.kps
```diff
@@ -0,0 +1,73 @@
...
+      <FileType>.htm</FileType>
+    </File>
+    <File>
+      <Name>readme.htm</Name>
+      <Description></Description>
+      <CopyLocation>0</CopyLocation>
+      <FileType>.htm</FileType>
+    </File>
+    <File>
+      <Name>..\..\..\..\..\Desktop\nerding\code2000.ttf</Name>
```
darcywong00, line 56: Should the code2000.ttf font be attached to this pull request  (within the piqad folder)?

release/p/piqad/source/piqad.kps
```diff
@@ -0,0 +1,73 @@
...
+      <FileType>.ttf</FileType>
+    </File>
+  </Files>
+  <Keyboards>
+    <Keyboard>
+      <Name>pIqaD</Name>
+      <ID>piqad</ID>
+      <Version>1.0</Version>
+      <Languages>
+        <Language ID="tlh-Piqd">Klingon</Language>
```
darcywong00, line 68: If the touch keyboard uses the code2000.ttf font, I'd expect something like
```
      <OSKFont>..\code2000.ttf</OSKFont>
      <DisplayFont>..\code2000.ttf</DisplayFont>
```

berrymot:
ah yeah the font should go in there. i thought i put it in the source folder but i guess that doesn't work?

LornaSIL:
Most of the files are not actually in the Pull Request. They all need to be in the same folder. I'm concerned about several things. I don't know that we have permission to distribute code2000. It is Shareware... @mcdurdin ?

Secondly, I know pIqaD is not in Unicode so I'm thinking it would be best to distribute your keyboard on your own hosted site. We don't want to have custom-encoded keyboards in the keyman repository.

Issue #1633, opened 2021-08-08
[lao_phonetic] build with 15.0 if we want to target web [low-pri]


mcdurdin:
This is low priority, just noting a potential issue for the future.

Due to keymanapp/keyman#5440, the lao_phonetic keyboard would not work correctly if it was built for web.

This is a bug in the 14.0 and earlier version compilers, which has been fixed in 15.0 (see #5525). Once we have a 15.0 compiler in the keyboards repository, we can rebuild this keyboard and make it available for web. (I note the keyboard is also not currently available for linux, which is a shame.)

I have audited all the keyboards in the repository, and this is the only one that is impacted by the described bug.

LornaSIL:
This keyboard is a bit too complex for me to fix. I would be worried I would break it. I also note that Linux is not a target and it could be added.

mcdurdin:
Note: this requires no functional change to the keyboard, only an update to build with the 15.0 compiler (which is now in use in the repo), in order to be able to use it on web.

mcdurdin:
I spoke too soon -- there's a functional issue with `store(C)` which is generating a lot of warnings; looking deeper.

`store(C)` has an 'ñ' to generate 'ຍ' -- but there is no 'ñ' key on the keyboard, so the web compiler generates a warning for it:

```
lao_phonetic.kmn (237): Warning: 209A The rule will never be matched because its key code is never fired.
... many warnings later ...
```

LornaSIL:
>lao_phonetic.kmn (237): Warning: 209A The rule will never be matched because its key code is never fired.
... many warnings later ...

Yeah, maybe I should have told you that was part of why I got scared off. I don't know how to handle it.

mcdurdin:
@LornaSIL see keymanapp/keyman#7216 -- partly it's a bug in Keyman, partly an issue with the `C` and `CLao` stores -- `C` includes the letter `'ñ'` which is never matched. Working on a fix for the Keyman bug and checking with my father on the keyboard issue now :grin:

PR #1994, opened 2022-09-12
feat: support all platforms for Lao phonetic

mcdurdin:
* Add support for web, Linux, and touch devices
* Change ຍ usage in `CLao` to prevent issues with unreachable keys, specifically removing the key code ñ as it can never be generated on US English.
* Update version to 1.1.2

Note that the issue with repeated warnings has been resolved by removing the key code ñ. The corresponding ຍ in the `C` store was moved to the end of the store to ensure that the rest of the keyboard continues to work without changes.


LornaSIL (approved):
Fixes #1633

PR #1997, opened 2022-09-13
[armenian_mnemonic] Fixed Options rendering in Linux

dotland:
In Linux Mint 21 the **Options** screen of the *Armenian Mnemonic* keyboard failed to render because of an incorrect tag pair and escaping ampersand symbol in the `<script>` tag.


keyman-server:

LornaSIL (approved):
LGTM

PR #2000, opened 2022-09-19
[inscript_php] Update php files to use alt instead of rightalt in osk

LornaSIL:
Hopefully a minor issue with php files using rightalt when the keyboard is alt.


LornaSIL:
Fixes [#596](https://github.com/keymanapp/help.keyman.com/issues/596)

DavidLRowe (approved):

PR #2006, opened 2022-09-22
[slc_saliba] New keyboard for Sáliba (ISO 639-3: slc)

eddieantonio:
This is a keyboard for the Sáliba langauge, spoken in Colombia and Venezuela. The keyboard was made on behalf of Jorge Emilio Rosés Labrada, who works with the Sáliba community.

Apart from Basic Latin (i.e., ASCII) characters, the Sáliba orthography uses ã/ẽ/ĩ/õ/ũ for nasal vowels and ñ.  In talks with the community, they felt that the <kbd>Ñ</kbd> key represented "nasality" and wanted a feature similar to the deadkey for typing the stress mark in Spanish. The physical keyboards they use are the Latin American layout have both <kbd>Ñ</kbd> and a deadkey for the stress mark (´). However, for keyboard layouts lacking the <kbd>Ñ</kbd> key, we agreed to use the combination <kbd>Alt</kbd>+<kbd>n</kbd>.

I would appreciate any feedback on this keyboard!


release/s/slc_saliba/source/slc_saliba.kmn
```diff
@@ -0,0 +1,33 @@
...
+store(&ETHNOLOGUECODE) 'slc'
+store(&MESSAGE) 'Sáliba keyboard/teclado sáliba'
+
+begin Unicode > use(main)
+
+group(main) using keys
++ 'ñ' > deadkey(virgulilla)
++ 'Ñ' > deadkey(virgulilla)
+c For keyboard layouts that lack a dedicated [Ñ] key:
++ [ALT K_N] > deadkey(virgulilla)
```
mcdurdin, line 15: Thanks for the keyboard submission!

For improved compatibility across platforms, I suggest you use `RALT K_N` instead of `ALT K_N`, as the Left Alt key is reserved for shortcuts on Windows. Right Alt is generally reserved for extended key combinations such as this. We are trying to standardize on Right Alt through all new keyboard submissions.

eddieantonio:
Hey @mcdurdin! Unfortunately, I'm running into this issue on macOS: keymanapp/keyman#875

Is the workaround mentioned here https://community.software.sil.org/t/cant-install-cipher-music-on-ios-13-and-some-input-does-not-work/2757/6 still acceptable?

---

My userbase are:
 - community members in Colombia and Venezuela using predominantly Windows PCs
 - linguists in Canada predominantly using Macs or running Windows within a VM in macOS

It needs to work on both platforms!

release/s/slc_saliba/source/slc_saliba.kmn
```diff
@@ -0,0 +1,35 @@
...
+
+begin Unicode > use(main)
+
+group(main) using keys
++ 'ñ' > deadkey(virgulilla)
++ 'Ñ' > deadkey(virgulilla)
+c For keyboard layouts that lack a dedicated [Ñ] key:
++ [RALT K_N] > deadkey(virgulilla)
+c Workaround for macOS -- see https://github.com/keymanapp/keyman/issues/875
+platform('mac') + [ALT K_N] > deadkey(virgulilla)
```
eddieantonio, line 17: @mcdurdin I added this workaround. I can't get right <kbd>Alt</kbd> working on my Mac — not on this layout, nor on other layouts I've downloaded that map characters to right <kbd>Alt</kbd>.

release/s/slc_saliba/source/help/slc_saliba.php
```diff
@@ -0,0 +1,198 @@
+<?php
+  $pagename = 'Sáliba Keyboard Help';
+  $pagetitle = 'Sáliba Keyboard Help';
+  require_once('header.php');
+?>
+
+<nav>
+  <p lang="es"> <a href="#slc_saliba-es">Lee este documento en español.</a> </p>
+</nav>
```
eddieantonio, line 9: Would it be appropriate to swap the English and Spanish versions of the documentation? The primary user group are more likely to know Spanish than English.

LornaSIL:
I thought I had given you similar feedback regarding `RALT` yesterday, but I must not have submitted it :( `RALT` should be equivalent to `AltGr` on the mac.

My other suggestion was for keyboards that don't have a "ñ" you could make the "tilde" be the deadkey instead of RALT, because most Spanish keyboards would have "ñ" and most other keyboards would be US English which has a tilde. You already have a rule turning "~ " into a regular "tilde".  I'm not sure if that rule would still work or if you'd need the rule to be `dk(1) + "~" > "~"`. Experiment if you like that option.

Yes, if you want the Spanish to be first that would be fine. We have a few keyboards with just Spanish so it's fine to make that primary since you want your users to find it easy to use.

release/s/slc_saliba/source/slc_saliba.kmn
mcdurdin: Yes, this is good. And as you have linked to the relevant issue on the Keyman repo, if we do ever manage to find a good general fix for the base cross-platform question, we'll see this keyboard and be able to issue an update for it. (Yeah, it's difficult to make it back-compat and cross-platform-compat and not break anything!)

release/s/slc_saliba/source/help/slc_saliba.php
mcdurdin: It's perfectly fine to put the Spanish documentation first. A link near the top of the documentation to other languages is helpful, too.

eddieantonio:
> I thought I had given you similar feedback regarding `RALT` yesterday, but I must not have submitted it :( `RALT` should be equivalent to `AltGr` on the mac.

Blocked by https://github.com/keymanapp/keyman/issues/875 unfortunately :/


> My other suggestion was for keyboards that don't have a "ñ" you could make the "tilde" be the deadkey instead of RALT, because most Spanish keyboards would have "ñ" and most other keyboards would be US English which has a tilde. You already have a rule turning "~ " into a regular "tilde". I'm not sure if that rule would still work or if you'd need the rule to be `dk(1) + "~" > "~"`. Experiment if you like that option.

I made a version with this change to my collaborators and I am awaiting feedback. Once this is done, this pull request will be ready for final review!

> Yes, if you want the Spanish to be first that would be fine. We have a few keyboards with just Spanish so it's fine to make that primary since you want your users to find it easy to use.

Done!

LornaSIL (approved):
LGTM

LornaSIL:
Keyboard is now online! https://keyman.com/keyboards/slc_saliba

PR #2008, opened 2022-09-23
Changes to miluk_hanis_siuslaw

jagdouglastavani:
Hello! After talking with the community and doing some testing, we have decided to change the name of the keyboard to hanis_shayuushtla_quuiich_alsea (while leaving the language as Miluk so that each language shows up only once in the app). This was a highly political point that was insisted upon by the community. 

I have fixed the issues with the ALT, insofar as I can tell, by removing the alt characters altogether and instead having those on the shift layer. There was also an update to two of the characters, moving diacritics around. 

Please let me know if there are any issues. 

Best, 

Jordan


keyman-server:

jagdouglastavani:
@DavidLRowe I believe it wanted me to tag you.

LornaSIL:
David is not available now. I'm not sure when he will be. 

Now we have the problem that there is a keyboard that is already online with the folder/filename of `miluk_hanis_siuslaw` (https://keyman.com/keyboards/miluk_hanis_siuslaw)

and now a new keyboard that you would like added with the folder/filename: `hanis_shayuushtla_quuiich_alsea`

It would be possible (and preferable from a technical standpoint) for you to keep the `miluk_hanis_siuslaw` files, but just change the keyboard name to something like "Hanis" or "Hanis Shayuushtla". The shorter the name the better because once people use it on a mobile device they won't see the whole name anyway. The problem is we cannot delete the `miluk_hanis_siuslaw` keyboard once we've accepted it (@mcdurdin is that true if it's in experimental?). We would need to go through a process of deprecating it.

If the community is still discussing the acceptance of the folder/filename issues, then it would be better if you do more testing distributing the keyboard yourself rather than in the Keyman repository. Our repo needs the folder/filenames to remain stable. The keyboard name in the .kmn/.kps/.kmp can change. For example, the keyboard NAME of **IPA (SIL)** has a folder/filename of `sil_ipa`. Another example where the name is **Bulgarian (Typewriter) Basic** has the folder/filename of `basic_kbdbu`.

jagdouglastavani:
Hi Lorna,

If we have to go back to the miluk_hanis_siuslaw I think that will be okay
-- but I know that they wanted the spelling of siuslaw to change
to shayuushtla and the addition of quiich and alsea.  The thing that was
bothering some folks was how this was displayed on the spacebar in the
keyman app.

Now, in the supporting documents I can put all of this there. The big
question that I have is it possible to have it be that we keep that name
for backend purposes but have it so if 'quuiich', 'alsea', or 'shayuushtla'
are searched, that it brings this keyboard up? That is, I think, the most
important thing. This is the difficult part of the process, as there are 5
languages using the same keyboard: miluk, hanis, sha'yuushtl'a
(shayuushtla), quuiich, and alsea. If it were to just be 'Hanis
Shayuushtla' as the name, that would certainly cause some political issues.
Is there a technical issue or reason to have a shorter name? I can also
float the idea of having the keyboard name be the tribe's initialism:
CTCLUSI (again, keeping the miluk_hanis_siuslaw folder structure).

Best,

Jordan AG Douglas-Tavani

LornaSIL:
I have to say, I really like the idea of the *name* being "CTCLUSI" because that removes the issue of which name is politically correct or incorrect.

Right now you only have Miluk [iml] associated with the keyboard.

It is possible to associate many languages with a keyboard, so I think that would be `iml` for Miluk, `sis` for Siuslaw, `csz` for Hanis. You would need to add that to both the `.kps` file and the `keyboard_info` file.

The keyboard_info would have these lines:
```
    "languages": [
        "iml"
    ],
```
changed to:
```
    "languages": [
        "csz", "iml", "sis"
    ],
```

Unfortunately, right now those language codes will use the default language name that is used in the standards (ISO 639-3, etc), not what it seems people are wanting, BUT including the new description in the `keyboard_info` file is good because I believe those names in the description will also be used by the search engine. The `keyboard_info` file will be the information displayed on this page: https://keyman.com/keyboards/miluk_hanis_siuslaw If there are different spellings for language names, include them all in the description!

mcdurdin:
## Deprecation of experimental keyboards

> (@mcdurdin is that true if it's in experimental?)

I think we can be relaxed about not requiring deprecation of experimental keyboards. It will be fine to rename it. The main issue is that existing users will not get updates if the keyboard is renamed without using a deprecation file. They'd have to uninstall and reinstall the new version themselves. (We do require deprecation of keyboards in the release folder, for stability reasons.)

## Language Names

In terms of the space bar text, it shows the `&name` of the keyboard from the .kmn file, and the "language" that the keyboard is installed with (usually from the .kps), by default. This is configurable as a user option.

You can update the language name shown in the keyboard package source file, if that is necessary, if the ISO639-3 name does not match the preferred name. This language name will be displayed in many locations, but not all -- there are a few locations where the system language name must be displayed, such as the Windows language bar.

image

It's better to leave the language information out of the .keyboard_info file if you are able to specify it all in the .kps file -- the .keyboard_info language metadata will be automatically generated in that case.

## Keyboard App Builder

Finally, if the majority of your users are Android users, you may also find it worthwhile to make a [Keyboard App Builder standalone app of your keyboard](https://software.sil.org/keyboardappbuilder/), where you can control the default text displayed on the keyboard space bar, and list all the languages for your communities in the installation user interface.

jagdouglastavani:
Thank you Marc!
Jordan AG Douglas-Tavani

jagdouglastavani:
Thank you for this information! I'll certainly add the other languages into
the code. For now, the language coordinator for the tribes has told me to
revert back to the miluk_hanis_siuslaw name and push the other changes,
with the name needing to go before the language and culture committee. I'll
work on making the necessary changes and then submit a new pull request.
Jordan AG Douglas-Tavani

PR #2009, opened 2022-09-23
Changes to miluk_hanis_siuslaw (back to original name)

jagdouglastavani:
Hello. As per our early conversation, I've gone back to the old name and have only put in the non-name related changes. The tribe hasn't decided on whether or not to use CTCLUSI -- but it looks like that will be a good compromise (everyone who has heard it thusfar is on board). 

Thank you!


keyman-server:

LornaSIL:
You added these language to `keyboard_info`:  aes, cxz, iml, sis Can you add them to the .kps file?

Then, if you are happy with the name `miluk_hanis_siuslaw` I can approve it as is.

jagdouglastavani:
I have added them to the .kps file. Do you need me to submit a new pull
request?

The tribe is okay with the name miluk_hanis_siuslaw for now. They are
discussing the CTCLUSI name (which does seem favored) -- but it has to work
through the tribal political system first.

Best,

Jordan AG Douglas-Tavani

LornaSIL:
Apparently you have the keyboard in both `experimental` and in `release` which we cannot allow. It should stay in `experimental` at this point I believe. Can you delete the one in `release`? I think the `experimental` one is the most up-to-date anyway, is that correct?

jagdouglastavani:
That is indeed correct that it was at one point in release and
experimental. I had thought David had deleted the release version. I can do
so though if you need me to.

Definitely should keep it as experimental right now.

Best,

Jordan AG Douglas-Tavani

jagdouglastavani:
it's been removed
Jordan AG Douglas-Tavani

LornaSIL (approved):
Thanks! LGTM.

PR #2011, opened 2022-09-25
[karakalpak] Initial submission of karakalpak keyboards

atabekm:
Initial attempt of adding support for Karakalpak language keyboards. Consists of two layouts:
- _cyrillic_, not official script at the moment, but due to historical reasons, still commonly used
- _latin_, current official script


keyman-server:

atabekm:
Oh, seems the build has failed, I couldn't access the TeamCity to see the problem. I could try fixing the problem if anyone can provide some details.

LornaSIL:
Thank you for your keyboard submission. Sorry to not respond sooner. In order to reduce churn in our build system, the build doesn't kick off without one of us hitting the right buttons.

This is just a review on karakalpak_cyrillic. It looks very good. 
- The description in `keyboard_info` and the `welcome.htm` could be improved if you use the description from the `readme.htm`: "This keyboard is designed for Cyrillic script of Karakalpak Language."
- You have the correct bcp-47 tag in the `keyboard_info` file, but the .kps file should be changed from `kaa-Cyrl` to just `kaa`. They should be consistent and I believe that is the correct tag according to CLDR.
- In the .kps file you should also include all the graphics (.png files) that are required for displaying the `welcome.htm` file properly.
- The mobile layout is fine, but I wanted to let you know (for future keyboards) that IF you have a tablet and a phone layout that are exactly the same, you do not need to include both. We recommend deleting one or the other in order to reduce the maintenance of the keyboard. (If there's only a "phone" keyboard it will work on both tablet and phone, etc.)

I will review the Latin keyboard next.

atabekm:
Thank you @LornaSIL for your thorough and quick review, I think I've fixed all the problems you have mentioned. I have one questions though regarding `bcp-47 tag`. In cyrillic version you suggested to use `kaa`, but nothing was said about the latin version, which uses `kaa-Latn`. Is this fine, or both of them should utilize `kaa`? If former, shouldn't the default tag (`kaa`) be used for latin one (since it is the current official script), and cyrillic use `kaa-Cyrl`?

LornaSIL:
> In cyrillic version you suggested to use kaa, but nothing was said about the latin version, which uses kaa-Latn. Is this fine, or both of them should utilize kaa? If former, shouldn't the default tag (kaa) be used for latin one (since it is the current official script), and cyrillic use kaa-Cyrl?

The BCP-47 tag indicates which language AND which script, so they need to be different.

At this point anyway `kaa` by itself indicates Cyrillic in the CLDR. So, include `-Latn` for the Latin keyboard.
CLDR has this in the `likelySubtags.xml` file
```
		<likelySubtag from="kaa" to="kaa_Cyrl_UZ"/>
		<!--{ Kara-Kalpak; ?; ? } => { Kara-Kalpak; Cyrillic; Uzbekistan }-->
```

It would not be wrong to make the Cyrillic keyboard have `kaa-Cyrl` instead of just `kaa`. So, if that feels better to you, I would be okay with you changing it to `kaa-Cyrl` (and the Latin one would still be `kaa-Latn`). That way if CLDR ever changes `kaa` to represent `Latn` then the keyboard wouldn't need the BCP-47 tag changed.

LornaSIL:
I can approve this if you are happy with the language tag.

atabekm:
> I can approve this if you are happy with the language tag.

Thanks Lorna, I am quite happy with the current version, so let's go ahead with it. Thanks.

LornaSIL (approved):
LGTM

LornaSIL:
Your keyboards are now online!

https://keyman.com/keyboards/karakalpak_cyrillic
https://keyman.com/keyboards/karakalpak_latin

atabekm:
Thanks Lorna, quite happy for this.

PR #2013, opened 2022-09-29
Himyarit Musnad Keyboard v1.0

dyacob:
This submission is made on behalf of the author, Riyadh Alfareh.  The release provides a desktop keyboard and touch layout for the Musnad (aka Old South Arabian) script.  The Musnad mnemonic associations follow the correspondences to Arabic phonology for easy input for users familiar with an Arabic script keyboard.


keyman-server:

LornaSIL:
Could you add a `keyboard_info` file? That is essential for the web page to display.

For the touch layout (which says "phone"), we would prefer not to have a mixed script keyboard. It is better to create a separate Arabic and separate Latin script keyboard that meets the needs of the community. I would prefer those layers get removed, but I guess I won't require it. The Latin one is especially problematical because this is a RTL keyboard! Also, the `Musnad_Arabic` layer doesn't produce Arabic, it is producing the Old South Arabian script. Is that what is desired? If that is what is desired I guess it's because there might be users who want to see Arabic script to know what to type?

The php file has this `<div id='osk-tablet' data-states='default'>` and the word "tablet" should be "phone". Also you could include "Punctuation" and "Punctuation-2" if you wanted.

The PR includes two fonts: sultanmusnad.ttf (name is sultan-musnad) and SultanMusnad-Normal.ttf (name is SultanMusnad). Both are v1.100. I suspect you only need to include one of those. Currently, the .kps file calls for `sultanmusnad.ttf`. If you decide to include the other one instead you should switch out the font in the .kps. Also, you should go to the "Keyboards" tab in the .kps and select the correct font that you want to be the default...either the SultanMusnad or Qataban.

For the `keyboard_info` file you will definitely want to include a font because this is an unusual script which will not have many available fonts.

Something like this would be good:

```
{
    "license": "mit",
    "languages": {
        "xsa-Sarb": {
            "font": {
            "family": "SultanMusnad",
            "source": "SultanMusnad-Normal.ttf"
            }
        }
    },
	  "description": "This keyboard is designed for the Dogri language, using the Dogra script with an Inscript layout."
}
```

Lastly, you used `xsa-Sarb` for the BCP-47 code. It seems that just `xsa` would also be acceptable and mean the South Arabian script. Whichever you choose to use, be consistent between the .kps and the keyboard_info. Then, if you ever create a lexical model you would use the same BCP-47 tag.

dyacob:
Hi Lorna,

Thanks for the great review and feedback.  Some quick follow-up:

* himyarit_musnad.keyboard_info is added.
* "xsa-Sarb" is now just "xsa"
* "osk-tablet" is now "osk-phone" in the PHP file.
* The SultanMusnad-Normal.ttf font is set as the default for the keyboard and display.

Notes:

The font file names are unfortunately similar (though styles are different), but we want to avoid changing the names from what the font company uses to avoid potential confusion.  We can emphasize the name issue to the vendor who is a friend of the author.

Regarding the layers, the intent with the `Musnad_Arabic` layer is to make it easy for the Arabic speakers to enter Musnad where they can type phonetically as per the Arabic mappings.  This was also a solution to the limitation whereby an Arabic letter could not be shown on the same key as a hint (like in a corner and smaller).  I've read that this will be possible in Keyman 16, so in an update, the `Arabic_Musnad` layer could be eliminated and the Arabic hint would appear on the default layer.

The `Arabic` layer is different and is a representative Arabic layout.  Arabic layouts can change a little, and there was a desire to consistency with the layout shown in the `Arabic_Musnad` , but like the English layer, it is also there for convenience. I believe the author wants to keep the layers as-is for now until some user feedback is received, then make an update when version 16 is available.

Thanks, I hope I didn't miss anything :-)

LornaSIL (approved):
LGTM. I still think Latin will be a problem with the RTL settings, but I'll let you try. Also, if you ever set up a lexical model it won't know what to do with the Arabic and Latin text. :(

mcdurdin:
> I've read that this will be possible in Keyman 16, so in an update, the `Arabic_Musnad` layer could be eliminated and the Arabic hint would appear on the default layer.

Just FYI, this probably won't land in 16.0 -- the design tools will be there but the development of the touch layout and gesture changes is taking longer than we hoped. Have not yet managed to update the roadmap...

RiyadhAlfareh:
thanks all for your support and your cooperation

RiyadhAlfareh:
I build lexicon to add to keyboard

PR #2015, opened 2022-09-30
[dogra_inscript] First update

WseriesY:


keyman-server:

LornaSIL (commented):
For some reason the keyboard doesn't have a keyboardversion statement. That's not good! I should have caught that in the first place. Can you make it version 1.1 or 1.0.1?

Also, if you could update the HISTORY.md file that would be helpful.

LornaSIL (approved):
LGTM

PR #2018, opened 2022-10-03
[remington_gail] Updates to kbd

LornaSIL:


LornaSIL:
The desktop changes have been well tested and they are the most important to get in.

The mobile changes have not been well tested, but they are not as important at this point. Since the keyboard is in experimental, I think that's okay.

DavidLRowe (approved):
Minor point: The version number in HISTORY.md doesn't match that in the .kmn file, but that can be fixed later.

PR #2024, opened 2022-10-06
[nisenan] Initial submission of Nisenan 2.0

JGearySSBand:
We have made extensive changes to the SSBMI Nisenan orthography, and so need to release a new version of the Nisenan keyboard.


keyman-server:

DavidLRowe:
Thanks for your work on this keyboard.

It would be helpful to add the date and a change summary line to the HISTORY.md file.

Currently the on-screen keyboard (OSK) only shows a few characters on the RAlt and Shift-RAlt layers. Using the On-Screen tab in Keyman Developer, you can tick the "Auto-fill underlying layout" box at the bottom of that screen, then select the "Fill from layout" button at the top of the screen. This will fill in the characters on the default and shift layers, which will make the OSK complete.

Currently in the touch layout you have both a "tablet" layout and a "phone" layout, however the phone layout is just the default layout with no added characters. If this keyboard is used on a phone, none of the extra characters needed for this language will be available. The best thing to do is to remove the phone layout, leaving only the tablet one, which Keyman will use for both tablets and phones. (Click the "Touch Layout" tab on the left, select "Platform:" as "phone" , then click the "Del" button to delete the phone layout.)

These should be easy changes to make, then the keyboard will ready. Thanks!

JGearySSBand:
Thanks for your help, David! I think I have made all of the necessary changes and will go ahead and reupload the files.
 
Best,
Jonathan

DavidLRowe:
I see you added an icon file. It needs to be referenced by the .kmn file in order to be used. I just tried adding that from with Developer and crashed the program! I'm not sure if that's because the icon file is larger than expected or some other reason.

Normally the icon file is 16x16 pixels and displayed in place of the keyboard key icon ( image ) which is used by default. Sadly, you are probably better off replacing that nice image with something simpler (that can fit in 16x16 pixels). 

If you select the Packaging tab at the bottom, then select the .kps file, then the Details tab on the left, one of the options is Image file. This allows you to include a 140x250 pixel image that is displayed when the user begins the installation process. Your 256x256 image is still too big for that, but perhaps you can adapt it.

JGearySSBand:
Thanks again for your help, David! I also just tried adding an icon to the .kmn file in Developer, and Developer keeps crashing for me before I select any icon file (error identifier: 6ACBC9A5BD3145080A70EDDC0F56B67C). Do you know why it might keep doing that when I select "Icon" as the feature I want to add but before I select the actual icon file?
 
In any case, we don't really need the icon right now. Would it be possible to just omit that file?

DavidLRowe:
I wrote up the issue with Developer as https://github.com/keymanapp/keyman/issues/7419

I'd suggest deleting the .ico file for now. You can experiment with the 16x16 size icon and see what you think. For now it'll be fine with the default Keyman icon.

DavidLRowe:
I went ahead and deleted the .ico file (and a stray .kpj.user file that crept in at some point).

DavidLRowe (approved):

PR #2025, opened 2022-10-07
Refactor boonkit to allow \ as same as ` key

mhosken:
The Boonkit currently uses ` as the key to change a consonant into a subjoined consonant by inserting a U+1A60 sakot before it. Unfortunately, most Thai users have the ` and ~ wired to switch keyboards. Changing this behaviour involves a lot of detailed magical clicking and could be considered a regression for the user.

This change adds the same ` subjoining behaviour to the \ key, which previously output (. Notice that there is no corresponding key that outputs ) on the keyboard, thus making the \ somewhat useless. Seems like a perfect compromise. The existing behaviour of ` is left unchanged for backward compatibility.

Warning. This change has not been tested (even syntactically).


keyman-server:

LornaSIL (approved):
LGTM

PR #2026, opened 2022-10-08
[younger_futhark_short_twig] Initial submission

WseriesY:


keyman-server:

LornaSIL (commented):
You could add a rule with all the unused Latin keys, like:
```
store(nul) "ABCDEFGHIJKLMNOPQRSTUVWXYZcegjpqvwxz"
+ any(nul) > nul
```
That would help the keyboard not to produce or show Latin letters.

PR #2028, opened 2022-10-10
[runeboard] Initial submission of runic keyboard

Runeboard:
This is my first ever attempt at doing this so my apologies if I've made any mistakes. I welcome advice on how to improve. Thank you for considering this keyboard for inclusion in Keyman.


keyman-server:

LornaSIL:
Thanks for your submission. You've got a bit of a problem with your file naming.
The only files in the root of the `runeboard` folder should be: README.md, LICENSE.md, HISTORY.md, runeboard.kpj and runeboard.keyboard_info. All the other files should be deleted.

However, runeboard.kpj references `rune.kmn` that is in the root. You need the correct `.kmn` file in `source` and it **must** be named `runeboard.kmn` **not** `runic.kmn`. So, you may need to copy `runic.kmn` to `source` and rename it `runeboard.kmn`. Then, you'll need to edit the .kmn and rename the header statements that call for 
```
store(&VISUALKEYBOARD) 'runic.kvks'
store(&LAYOUTFILE) 'runic.keyman-touch-layout'
```
change to:
```
store(&VISUALKEYBOARD) 'runeboard.kvks'
store(&LAYOUTFILE) 'runeboard.keyman-touch-layout'
```
Basically, you need to remember that the foldername, `keyboard_info` file, `.kpj`, `.kmn`, `.kvks` and `keyman-touch-layout` files must all have the same name. So, either choose `runic` for all of that or `runeboard`.

Ideally you might change the `ico` to use the same name as well. Currently you have several `ico` files. you could delete the unnecessary ones.

You have a bunch of language codes in the .kmn file. That is now not recommended for usage. I recommend you include all the language codes that need associating with the keyboard in the .kps file. The same language codes must be included in the `keyboard_info` file.

The package is over 2Mb. You might consider removing some of the fonts and just include the one that you feel is best use. 

You could also consider removing the `junicode-1.002` `doc` and `util` folders and their contents since those are unnecessary to the keyboard function.

I haven't done any testing. I just was checking the structure of the package. We'll do better testing once these file names get straightened out.

Runeboard:
Thanks for your help Lorna :)

I modelled this on the keyboard I built for my own private testing in Keyman 9.0 some years ago. As it's been a while, I thought I'd probably make a few mistakes so having someone to point out where is much appreciated.

In Keyman Developer, is there a way to change the names of files after they have been created?

Regarding the fonts, Quivira (serif) and Junicode (sans serif) are both large files because they're designed for a large array of scripts. This is helpful to the user as it means they'll have fonts that can display just about any language. On the other hand, although the BabelStone fonts are really only designed for runic and Latin, they do have the specialised ability to display bindrunes whereas the other fonts do not. Still, you're right that much of the documentation that comes along with all those fonts can be removed.

Other than that, I'm not sure I've perfectly fixed everything but I believe I've made significant improvements. Yet as I can't find any way of updating files on this site, I'm guessing I have to start this process again by submitting a new pull request.

PR #2031, opened 2022-10-11
[runeboard] 2nd attempt

Runeboard:
I just tried to do this a moment ago but it appeared not to work so I'm trying again.


Runeboard:
You're welcome for the submission David. Thank you for your patience with me as I learn how things work around here and for helping to guide me through the processes. As I see that I was meant to base the pull request on a fork rather than the master branch, I've created a new fork to make the new pull request.

Having noted that Keyman already hosts an Ogham keyboard, I've removed that layer entirely. I've also removed the Latin characters but left the symbols and numbers as they don't belong to any particular script and I observe that various other keyboards hosted by Keyman already include such non-script bound characters.

Regarding the layout presented on the Welcome page, I've kept the complete layout for display (including Ogham, etc.) but added a note explaining Keyman's limitation. As the full keyboard is functional, I could potentially distribute it on my own to anyone who may want it. Displaying the full layout also explains to users why certain keys won't work on the version available for download from Keyman. If necessary, I could name &/or number the version I distribute privately different from that available through Keyman in order to avoid confusion.

Regarding languages, I've now tried to limit the list to those languages which have some historical connection to runes for which there may be a purpose for writing them in runes.

As you suggested, the "phone" layout has been removed. Will this create any problems when trying to create a version for Android? Or is that a question I would need to ask somewhere else?

The documentation has been updated. I hope I've understood what's required correctly.

DavidLRowe:
Replaced by #2036

DavidLRowe:
You asked "As you suggested, the "phone" layout has been removed. Will this create any problems when trying to create a version for Android?"

No. The Keyman engine that runs on Android will look in the keyboard package for a phone layout first, but, after not finding one, will then load the tablet layout.

Runeboard:
Okay. Thank you for letting me know. Sorry I only just found this.

PR #2036, opened 2022-10-12
[runeboard] 3rd attempt at submission

Runeboard:
Still not sure I'm doing this right but I think I'm getting closer.


keyman-server:

DavidLRowe:
Yes, this is looking much better. Now that you have your pull request connected to the "runeboard" branch (in your fork of the repository), you will make any further changes directly in that branch and they will automatically become part of this pull request (so no need to create a new pull request to add changes to this keyboard, just make sure you switch to the runeboard branch before uploading changed files). Once the pull request is merged, you can delete the "runeboard" branch. 

In the Packaging section (.kps file), there are some changes needed. 
Select the "Files" tab on left in order to change the files included in the keyboard package:
- runeboard.ico is not needed (since it has already been compiled into the .kmx file), so Remove it.
- you need to add runeboard.kvk and runeboard.js files from the build folder (so that the on-screen keyboard and the touch keyboard layout will be included in the package)
- you need to add all the image files referenced by welcome.htm

Because this keyboard is in the "experimental" section, we are less worried about documentation details, but I'll point out some in case you want to deal with them sooner rather than later (when you move the keyboard to "release").
- one of your images has an uppercase "JPG" extension. This works on Windows, but not on other platforms with case-sensitive filenames. Using all lower case filenames can avoid this problem.
- you may want to avoid mentioning a specific version number in descriptions and other documentation so that you don't need to change the version number each time you revise the keyboard.
- your package currently includes four fonts; are all of them needed?
- currently the On-Screen Keyboard (OSK) shows blank keys for ~ ^ | and upper case ASD, however the keyboard produces these characters when the keys are typed. Any key on the default or shift layer will output the equivalent en-US (QWERTY) key unless you specifically include a rule to output nothing.

Eventually, we'll need a runeboard.php (online help) file  that provides the equivalent information as welcome.htm (the help packaged with the keyboard). But that can wait.

I was amused to find the Bitcoin symbol U+20BF on the runic keyboard, but maybe it feels at home there!

Runeboard:
As best as I can tell, the only way to proceed is to start the whole process afresh every time. I can't find any way to make changes once I've uploaded a folder sorry. Thus, I'm about to try starting again.

I've tried to follow the rest of your instructions as best as I can.

There are four fonts because 2 are serif fonts while 2 are non-serif fonts, and 2 are rune specific (and thus capable of creating bindrunes) while 2 are capable of being used with a wide array of other scripts but not capable of creating bindrunes.

The blank keys were not originally meant to be blank. Rather, they were meant to have Latin letters on them which are historically related to runes (i.e. æ, œ, þ, ð, ȝ, & ƿ). However, Lorna instructed me to remove them because apparently Keyman prohibits the use of multiple scripts on the same keyboard. Consequently, I've left them blank so that users who only need a basic but complete runic keyboard may still be able to access this one through Keyman, while those who require access to these additional characters may be able to obtain it from me privately.

As Keyman Developer does not accept the Unicode Character Value "U+0000", it would appear as though a nul character can't be encoded. Is there a way around this?

As you may be able to see on the tablet layout, the key with the Bitcoin symbol is repositioned to be beside the spacebar near ᛒ. Originally I thought to include the Bluetooth symbol there yet one doesn't seem to exist in Unicode, which makes sense as it would be logical to create it as a bind of 2 Younger Futhark runes instead. Consequently, ₿ seemed like the next best option for a somewhat related symbol.

DavidLRowe:
Once a pull request is submitted, any changes that are made to the branch which was used to create the pull request, become part of the pull request.

In the instructions on the page https://help.keyman.com/developer/keyboards/github/step-3 , we have a paragraph:
"This pull request is linked to the branch (of your fork of the keyboards repository) which you used to create the pull request. Any changes made to the branch become part of the pull request. If you are asked to change a file, you'll open the branch and drag the new version of the file to the proper folder and commit the changes."

If you have suggestions for how we could make that clearer, please let me know (or suggest them yourself, by clicking on the "Edit page" tab at the bottom of the page).

Adding
```
store(nooutput) 'ASD'
any(nooutput) > nul
```
should suppress output for those keys.

Continuing work in #2044

Runeboard:
You suggested adding the following:

store(nooutput) 'ASD'
any(nooutput) > nul

Doing that creates an error which causes the keyboard compilation to fail.

DavidLRowe:
You are correct. My apologies. The second line needs a "+":
```
+ any(nooutput) > nul
```

PR #2038, opened 2022-10-13
Multani inscript

WseriesY:


keyman-server:

DavidLRowe (approved):
The .ico file doesn't need to be included in the package (since it is already compiled into the .kmx file), but everything else looks fine, so we'll leave it.

PR #2042, opened 2022-10-13
GFF Gurage v0.9.1

dyacob:
This is a simple fix for the phone touch layout that had an incorrect እ key assignment.


keyman-server:

dyacob:
Something went wrong my tool configuration and it appears every file was touched during my last commit.  I'm going to delete this branch and start over.  There should be only 7 files that update.

PR #2044, opened 2022-10-14
[runeboard] 4th submission attempt

Runeboard:


DavidLRowe:
(see additional comments in previous PR #2036)

DavidLRowe (approved):
The BCP 47 codes in the .keyboard_info files need a comma between them. (That is, a comma after each one except the last.) Because you permitted developer edits, I was able to make that change. (For future reference, you can use https://jsonformatter.curiousconcept.com/ to check the syntax of the .keyboard_info file.)

The .kps file had invalid paths for two items. I removed "..\Deleted4Upload\" from the path for .kvk and .js files.

You'll need to incorporate these changes into your files before submitting any revisions.

DavidLRowe:
Eventually, when you next revise the keyboard or move it to "release", it would be helpful if the order of the codes in the .keyboard_info file and the order in the .kps file were the same. This is not a technical requirement, but it does help the reviewers and anyone contributing to the keyboard. Thanks!

Runeboard:
I've rearranged the language codes in the .keyboard_info file and the .kps file so that they're in the same order for you.

The instructions on your website say to delete the build folder. As it seemed likely its contents would subsequently be needed, I moved it and renamed it. May I please ask why the Keyman Developer tool puts the .kvk and .js files in a folder which we're told to delete if those files are actually needed? This seems a little odd. Anyway, I've now moved them into the source folder instead.

DavidLRowe:
Your keyboard is now available at https://keyman.com/keyboards/runeboard

In the instructions on page https://help.keyman.com/developer/keyboards/github/step-3 we try to explain the reason for deleting the build folder: "Before making your submission on GitHub, you'll need to use Keyman Developer to compile and test the keyboard and the package. Once you're satisfied that everything works, delete the "build" folder and its contents. (This may seem odd, but only the source files are submitted to the Keyman keyboards repository. All the keyboards on the keyman.com site will be built from the source files.)"

In this case your build was successful because the Deleted4Upload folder was still on your computer, but it failed on our build system because that folder was included in the path. The build folder has to be in the right place (under the project folder at the same level as source) for the system to work.

DavidLRowe:
1. Currently the "Shift" key on the default layer does nothing to change layers. Select Touch layout on the left side of the screen, Platform: tablet, Layer: default, and select the Shift key. You can then change the Next layer value to "shift". You will similarly want to select Layer: shift (above the keyboard image) and select the Shift key and change the Next layer value to "default". 

2. I installed the keyboard (via keyman.com) on my Android phone, but am not seeing the behaviour you describe. You might try uninstalling the keyboard, and reinstalling. If the problem disappears, then fine, but otherwise you can report a bug by raising a GitHub issue (https://github.com/keymanapp/keyman/issues). If you have the means to record a video of your Android device, that can be helpful in showing the developers exactly what you are seeing.

Runeboard:
Thanks for your help. I think I've fixed the first problem with the shift key now. Lest I confusedly do something wrong again, which file should I upload where to distribute the correction please? If it matters, I have updated the history file and the version numbers where needed.

In regards to the 2nd problem, it seems to only be an issue in the Keyman app itself but not when the keyboard is used in other apps. As such, it's probably not worth worrying about.

DavidLRowe:
I'm not an expert with GitHub, but I think this should work:
- Go to https://github.com/runeboard/keyboards. This is your "fork" of the keyboards repository. By default the "master" branch is  displayed. You can see that your fork is "29 commits behind keymanapp:master". 
- Follow the instructions in the "Updating your fork" section at the end of the https://help.keyman.com/developer/keyboards/github/step-3 page to bring your fork up-to-date. In particular this will bring in the updates I made to several files.
- If you need to get a copy of the runeboard keyboard (so that you'll be sure to have the same files as a starting point), one way to do that is to use https://www.gitkraken.com/learn/git/github-download#how-to-download-a-folder-from-github to download a zipped file with the runeboard folder and its contents (using the https://download-directory.github.io/ page and giving it https://github.com/keymanapp/keyboards/tree/master/release/r/runeboard as the path to this keyboard).
- To submit a pull request with your changes (based on an up-to-date copy of the keyboard), you can follow the procedure on the https://help.keyman.com/developer/keyboards/github/step-3 page. (It may be that you can "reuse" the existing runeboard branch, but I'm not expert enough to know if that will work. I'd suggest creating a new branch, "runeboard2" for example, just to be safe.) Even though the runeboard keyboard already exists, you can still upload the entire runeboard folder. GitHub will figure out which files haven't changed and exclude them from the pull request.

Hope all works for you. If not, ask again! (And if you have any suggestions for improvements to the documentation, let me know.)

mcdurdin:
> 2\. The keys either side of the spacebar do not return to normal after being released. Instead, they remain displayed large as when one holds down a key.

Could you share a screenshot and steps with us at https://github.com/keymanapp/keyman/issues/new/choose? It would help us to improve the quality of Keyman :grin:

Runeboard:
> > 2. The keys either side of the spacebar do not return to normal after being released. Instead, they remain displayed large as when one holds down a key.
> 
> Could you share a screenshot and steps with us at https://github.com/keymanapp/keyman/issues/new/choose? It would help us to improve the quality of Keyman 😁

Okay. I'll get onto it soon.

Runeboard:
> I'm not an expert with GitHub, but I think this should work:
> 
>     * Go to https://github.com/runeboard/keyboards. This is your "fork" of the keyboards repository. By default the "master" branch is  displayed. You can see that your fork is "29 commits behind keymanapp:master".
> 
>     * Follow the instructions in the "Updating your fork" section at the end of the https://help.keyman.com/developer/keyboards/github/step-3 page to bring your fork up-to-date. In particular this will bring in the updates I made to several files.
> 
>     * If you need to get a copy of the runeboard keyboard (so that you'll be sure to have the same files as a starting point), one way to do that is to use https://www.gitkraken.com/learn/git/github-download#how-to-download-a-folder-from-github to download a zipped file with the runeboard folder and its contents (using the https://download-directory.github.io/ page and giving it https://github.com/keymanapp/keyboards/tree/master/release/r/runeboard as the path to this keyboard).
> 
>     * To submit a pull request with your changes (based on an up-to-date copy of the keyboard), you can follow the procedure on the https://help.keyman.com/developer/keyboards/github/step-3 page. (It may be that you can "reuse" the existing runeboard branch, but I'm not expert enough to know if that will work. I'd suggest creating a new branch, "runeboard2" for example, just to be safe.) Even though the runeboard keyboard already exists, you can still upload the entire runeboard folder. GitHub will figure out which files haven't changed and exclude them from the pull request.
> 
> 
> Hope all works for you. If not, ask again! (And if you have any suggestions for improvements to the documentation, let me know.)

Thanks for all your help David :)

I think I managed to follow all of this correctly except for the bit at the end where I errantly named the pull request "Bug fixed".

All in all, the documentation is quite good. The inclusion of pictures helps. It's just that the whole process is so complex that it's easy to become lost while trying to grasp hold upon the learner's ropes. There's a lot of jargon to learn so following directions which may be easy for those familiar with the site takes some effort on the part of a new user to understand.

Thanks again for your patience with me.

PR #2045, opened 2022-10-15
[runeboard] 5th submission attempt

Runeboard:


keyman-server:

DavidLRowe:
Your keyboard is already submitted (see https://keyman.com/keyboards/runeboard). 

The compiled files should (.kmx, .kvk, .js) should not be in the source folder. They need to be in the build folder (so that the paths to them are correct) but the build folder should be removed before submission so that all the compiled files can be built directly from the source files you submit.

Now that the keyboard is published, any subsequent changes to the keyboard will need a version number increment (that is, a change to &KEYBOARDVERSION in the .kmn file, plus other places where the version number is mentioned). For a minor change, that could be from 1.0 to 1.0.1, or from 1.0 to 1.1. For a major change, you might change from 1.0 to 2.0. The change in version number allows Keyman to know that a new version of the keyboard is available.

Save the changes to the order of BCP 47 tags in the .kps and .keyboard_info files until you next revise the keyboard. Perhaps you'll want to get feedback from users and possibly make some changes to the keyboard.

Runeboard:
Thanks. This message is just to let you know I've read this. Although I'm still confused about the process, thank you for helping to make it work.

PR #2046, opened 2022-10-18
GFF Ge'ez v1.7

dyacob:
This is an update to the GFF Ge'ez language keyboard that fixes an issue with Zaima notation input that started appearing last year (I *think* with Keyman 14), and adds support for input of extra symbols supported in the Abba Garima font (now included with the package).

**Note:**   This keyboard contained a workaround for an "outs" menu issue, I believe devised by @mcdurdin , and this note has appeared in the keyboard `remove 'using keys' and +any() when KMW bug fixed`  I believe the bug referred to has been fixed since a new related issue appeared around Keyman 14 that may have lead the workaround to exhibit an undesired behavior.  I've fixed the issue with the unwanted behavior by removing the workaround.  _However_... I'm not sure what `KMW` is, "Keyman Web" maybe?  In this case, I have not tested the keyboard on Keyman Web, I don't believe I can test in this environment.


keyman-server:

mcdurdin:
> **Note:** This keyboard contained a workaround for an "outs" menu issue, I believe devised by @mcdurdin , and this note has appeared in the keyboard `remove 'using keys' and +any() when KMW bug fixed` I believe the bug referred to has been fixed since a new related issue appeared around Keyman 14 that may have lead the workaround to exhibit an undesired behavior. I've fixed the issue with the unwanted behavior by removing the workaround.

I don't recall what the issue was that required this workaround. I'm not seeing any changes I made to the keyboard in the [history](https://github.com/keymanapp/keyboards/commits/50df804364d1d6e0f2cb64ae15561d7563307998/release/gff/gff_geez/source/gff_geez.kmn), so it's something that happened more than 5 years ago. 

What was the undesired behavior that you were seeing with Keyman 14?

> _However_... I'm not sure what `KMW` is, "Keyman Web" maybe?

Yes, KeymanWeb.

> In this case, I have not tested the keyboard on Keyman Web, I don't believe I can test in this environment.

You can test with KeymanWeb through the Keyman Developer test environment. On the Build tab in Keyman Developer keyboard editor, click **Test Keyboard on web**, and then **Open in browser**. 

image

[More information](https://help.keyman.com/developer/15.0/context/server)

dyacob:
Ohhhh... I use that interface all the time to test the mobile keyboards, I forgot or didn't realize that it also worked for the keyman web interface.  I've tried it now, and see no issues.

The issue that showed up in Keyman 14, or possibly 13 (its all a blur now), was one where you type a zaima mark like:

`᎖`

then go to type a 2nd zaima mark _immediately_ after the first (they often go together), then the Ethiopic gemination mark is output unexpectedly like so:  

`᎖፟ `

As an image:

Screen Shot 2022-10-18 at 3 21 30 PM


Underscore is used to invoke the zaima menu where the gemination mark is the last item.  In the non-Ge'ez GFF keyboards, 2 underscores would produce the gemination mark, some of that logic was still present in the Ge'ez keyboard also.  I forget what the earlier issue was that led to an extra "first" group being used in the keyboard (the work-around), but all was well until KM 14 or 13.  I don't see the issue with this update to the GFF Ge'ez keyboard using KM 15, I haven't tried it on any earlier versions.

DavidLRowe (approved):
(after minimal review) LGTM

PR #2047, opened 2022-10-21
[sil_limbu_typewriter] correct OSK states

DavidLRowe:
Doesn't need version increment since only the online help is changing.


darcywong00 (approved):
lgtm

PR #2048, opened 2022-10-21
Initial Release of Devanagari Typewriter (SIL) 1.1

sil-internationalNP:


keyman-server:

DavidLRowe:
A few minor changes are needed:

The "sil_devanagar_typewriter.kpj.user" file should be removed. (But be sure to leave the "sil_devanagar_typewriter.kpj" file!)

There are a number of empty layers in touch layout:
leftctrl-leftalt-shift
rightctrl-shift
leftctrl-leftalt
rightctrl
It seems that these could be removed (along with the longpress keys on the RAlt/Default key for each of the remaining layers).

The BCP 47 codes in .keyboard_info file and the .kps (package source) file need to agree. Currently the .kps file has the "-NP" region qualifier for all the codes and the .keyboard_info file does not. Developer is recommending that you drop the "-NP" from all except "kgy-NP" and "mai-NP" (which are spoken in countries other than NP). However, you would only use the region qualifier if there are orthographic differences in the way that the language is written between the two countries (as in the case, for example, of UK and US English).  

I expect that you can drop the -NP for kgy-NP and mai-NP as well. If so, then that would match the .keyboard_info file. If you decide to keep the -NP for either or both of those, then you need to make the corresponding change in the .keyboard_info file.

DavidLRowe:
(Optional) BTW the image file is expected to be 140x250 pixels (PNG or JPEG). If it is larger, as is the case here, the file is scaled to fit. You will get better results if you adjust the image to the expected size yourself.

DavidLRowe:
Sorry I wasn't clear. The "-NP" should be omitted from the BCP 47 codes for all that languages listed in both the .kps file and the .keyboard_info file. (The only exception would be for kgy and mai, and then ONLY if the language has a different orthography in Nepal than in other countries where the language is used.)

Other than that, the changes you've made look good.

DavidLRowe (approved):

PR #2051, opened 2022-10-22
GFF Amharic v2.3

dyacob:
This update addresses a small issue with a vowel composition, optimizes the layout of popups, and repositions a frequently used punctuation layer for quicker access.

**_Note_**: I've tested on an Android Phone, iPhone, iPad Mini, Mac and Windows 11 computers.  There seems to be an issue with Keyman 15 on an iPad whereby it does not use the "tablet" layout, but instead uses the "phone" layout.  I've observed this issue with another keyboard as well.  I suspect it is an issue with the KM and not the keyboard, but do let me know if there is something broken in the `.keyman-touch-layout` file.  The tablet layout otherwise tests fine in the web browser based testing that is launched from KM Developer.

iPadMini:  iPadOS 15.7 (19H12),  KM: 15.0.270


keyman-server:

dyacob:
Closing this PR and will create another later today.  It appears that a merge with remote updates went bad.

PR #2057, opened 2022-10-27
Updates to Cameroon Qwerty and Azerty

MattGyverLee:
Fixing my part of https://community.software.sil.org/t/bug-introduced-in-cameroon-keyboard-by-km-15-web-android/6834/5
Reverting "Latn" addition for languages where Canonical is Latn.
Switching to &CasedKeys instead of NCAPS/CAPS
Switching both keyboards to fonts from the Repo


keyman-server:

DavidLRowe (approved):
LGTM (after a cursory check of changes and successful build on the CI).

PR #2064, opened 2022-11-03
[rejang] new keyboard

Ilhamkang:
Dear @LornaSIL  , I've created a new keyboard to support Rejang script. Could you please to check whether this keyboard needs any adjustment? Thanks in advance.


keyman-server:

DavidLRowe:
@Ilhamkang Thanks for your submission. A more complete review is coming, but I notice that your pull request includes a change to an unrelated keyboard: a version number change in release/s/sundanese/sundanese.kpj . That change should not be included in this pull request. Thanks!

Ilhamkang:
@DavidLRowe thank you for noticing it! 
Please exclude the version change on Sundanese keyboard. That should be included to another future PR.

DavidLRowe:
Although there is a rejang.keyman-touch-layout file included in the source/ folder, it is not referenced by the rejang.kmn file (since there is no `store(&LAYOUTFILE) 'rejang.keyman-touch-layout'` statement in the rejang.kmn file). I think in this case the desktop keyboard layout is shown on the touch device. 

It seems, however, that the existing rejang.keyman-touch-layout file (other than specifying the font) is just the generic US English one, so it wouldn't add any value to your keyboard. 

If you want to make a keyboard for only web and desktop (Windows, Mac, Linux) devices (and not touch devices such as phone and tablet), you could change the &TARGETS from 'any' to 'desktop web' and delete the existing rejang.keyman-touch-laytout file.

If you want to include a touch layout, I'd suggest going to the Details tab on the left, then using the Add button on the right to add the "Touch-Optimised Keyboard". This should give you a default keyboard for both tablet and phone. You'll then likely want to delete one of these (unless you're planning to develop two different touch layouts, one for phones and one for tablets), since Keyman will use a tablet layout on a phone if a phone layout isn't available and vice versa. You can use the "Import from On Screen" button to fill in some of the keys. If you need more information, check the resources at https://help.keyman.com/developer/current-version/ or ask a question here.

Ilhamkang:
Great! Thank you for your inputs @DavidLRowe! I will revise the keyboard soon.

Ilhamkang:
I've updated the touch keyboard. Please check if this works or not.

DavidLRowe:
@Ilhamkang I'm not familiar with this script, so my testing has been very minimal. 

For the touch layout, there is a problem with some of the longpress keys on the "." key, which are giving the wrong output. For example, the ":" produces ";". Currently you have the ID as "K_COLON" and the Modifier as "(current layer)". Since the current layer is "default", this links to the .kmn file as [K_COLON] and produces ";". If you change the Modifier to "shift", then the link to the .kmn file would be to [SHIFT K_COLON] and would produce ":". 

But for these punctuation characters, you may prefer to change the ID (for ":" for example) to U_003A, which will bypass the .kmn file and produce ":" (U+003A) directly.

On the numeric layer, there is a similar problem with some incorrect values produced for some of the keys. I think you'll be better off using U_xxxx to produce the U+xxxx value directly, rather than trying to link to the .kmn file.

Since you don't have a "shift" layer, it may be better to remove the shift (up-arrow) key.

Ilhamkang:
@DavidLRowe got it! will be corrected soon. Thanks!

Ilhamkang:
The touch layout has been adjusted.

DavidLRowe (approved):
Looks good!

Ilhamkang:
Thank you!

DavidLRowe:
@Ilhamkang Oops, I didn't notice that the keyboard doesn't have an online help file. This would be rejang.php in folder called "help" under the "source" folder (so, source/help/rejang.php). Often you can take the body of the welcome.htm file and add a PHP header. You can either copy the image files from the welcome folder to the help folder, or change the references to use keyboard images created from the keyboard information.

Have you done that before for another keyboard? If you need more information, please let me know. 

Because there's no change to the keyboard itself (only the online help), you do NOT need to increment the version number or change any other files. Just add a help folder with rejang.php (and possibly image files) to the source folder in a new pull request.

Or if you'd rather that I do that, just let me know. Thanks!

Ilhamkang:
Ah, my bad. I missed the help part. I've done it before and will add the help file soon. Thanks for noticing it!

Ilhamkang:
@DavidLRowe I've made PR for this: #2083

PR #2066, opened 2022-11-06
Mon UniBoard

ntun1203:
Mon UniBoard is Mon unicode keyboards with Mon A1 loyout. This keys mapping is compline with all the Mon Unicode fonts and Myanmar Unicode fonts.


keyman-server:

ntun1203:
> This pull request is from an external repo and will not automatically be built. The build must still be passed before it can be merged. Ask one of the team members to make a manual build of this PR.

PR #2068, opened 2022-11-12
GFF Amharic v3.1

dyacob:
This is an update to the GFF Amharic keyboard where the CSS definitions are simplified significantly, where the keyboard `.css` is reduced from 14,662 lines to 517.  The last layer for numerals now uses a pseudo-sliding numeral row.


keyman-server:

LornaSIL (approved):
LGTM

PR #2069, opened 2022-11-12
GFF Mesobe Fidelat v1.0

dyacob:
This is a language-neutral touch (only) keyboard for mobile devices. The keyboard provides all Ethiopic letters in active use on a single layer in a circular, semi-clockwise orientation. It is meant to be intuitive for new typists and fits more comfortably on a tablet device than on a mobile phone.


keyman-server:

LornaSIL:
I updated the `keyboard_info` file thinking that I could approve the keyboard. But, I found that some of the longpress keys, like on the punctuation, are producing latin punctuation rather than Ethiopic. You probably want to fix that.

Also, you've listed a lot of languages in the .kps file. Those need to be included in `keyboard_info` (preferably in the same order so I can more easily check them). The complicating issue with that is that you will need to do the font syntax for each language.

Both of these issues are the same in PR #2070

dyacob:
@LornaSIL thanks for the fix-ups along the way.  I'll give the punctuation popups a review today.  I had gotten into cycles of fixing one thing, and breaking another, apparently I missed these.  Thanks for catching!

dyacob:
Thanks, I used the JSON validator with my last commit.  I think all issues are addressed now.  On a related note, next week I'll offer an update for the Wookianos font that includes Unicode 14 additions.

LornaSIL:
I'm going to do the Abyssinica release tomorrow. Do you want me to wait to approve these until we put the new Abyssinica and new Wookianos in the repo?

dyacob:
Oh, I would prefer not to wait.  These last two keyboards have mobile targets _only_ so the web font wouldn't be a factor, also they don't provide the Ethiopic Tonal Marks where the update happened in Abyssinica.

LornaSIL:
I think that entry is used in both the KeymanWeb (which this keyboard doesn't need) and also in the help file maybe? Are you calling the font in the help file? If not, it is unneeded.

If that's the case, something like this is sufficient:
```
    "languages": {
        "anc-Latn",
        "ann-Latn",
        "atg-Latn",
        "bom-Latn",
        "bzw-Latn",
        "enn-Latn",
        "gkn-Latn",
        "ig-Latn",
        "igb-Latn",
        "ige-Latn",
        "ikw-Latn",
        "jbu-Latn",
        "yaz-Latn"
    },
```
It doesn't have to be on one line per language altho I do find it easier to parse.

dyacob:
Thanks @LornaSIL , I've committed and pushed in updates to remove the font references.  The JSON validator didn't like the structure with `"languages": { "am-Ethi", "agj-Ethi", ... }`  I think because `languages` is not an array type that would have used `[]` in place of `{}` .

So I got around it by adding the empty `"{ }"` bits after a language name.

dyacob:
I've removed the script code in the commits just made, save for byn-Ethi which matches the setting in the gff_blin keyboard.

LornaSIL:
Is `mvz` definitely written with Ethiopic or is that an assumption that it would be written if someone develops an orthography?

mcdurdin:
FYI, you can spec the array format for languages if extra metadata is not required:

```
    "languages": [
        "anc-Latn",
        "ann-Latn",
        "atg-Latn",
        "bom-Latn",
        "bzw-Latn",
        "enn-Latn",
        "gkn-Latn",
        "ig-Latn",
        "igb-Latn",
        "ige-Latn",
        "ikw-Latn",
        "jbu-Latn",
        "yaz-Latn"
    ],
```

Note also that if those languages are specified in the .kps file, then you don't need to add them to the .keyboard_info as well.

dyacob:
I see, the same list is in the `.kps` file, so I'll remove them from the `.keyboard_info` file.

mcdurdin:
> Do you mean that the list of BCP 47 tags in the .kps file doesn't need to match that in the .keyboard_info file?

Yes, there are cases when it can be helpful to have divergent codes in the .keyboard_info, for example, if you want to use a subtag such as `-fonipa` which we don't yet fully support in Keyman .kps.

> Can it be omitted entirely from the .keyboard_info file (unless other info such as font or example is added)?

Yes, the language data in the .keyboard_info will be populated from the .kps, if there is not already a language element in the .keyboard_info

dyacob:
As there isn't a need for divergent codes, I'd like to keep it simple and maintain the language codes in a single place only (the `.kps` file).  Thanks.

dyacob:
@LornaSIL would you like me to add the Abyssinica v2.200 fonts to this PR, or do you plan to do that separately?

LornaSIL:
I just added them in a separate PR. I guess @DavidLRowe needs to approve it and then you'll need to merge master into your PR.

DavidLRowe:
@dyacob PR #2074 to add the v2.200 Abyssinica fonts has been merged.

dyacob:
@LornaSIL I ran the merge with master for this PR and #2070 . Thanks.

LornaSIL:
Is this ready for me to do a final review?

dyacob:
Yes.

LornaSIL:
There's still a problem with the `keyboard_info` file and I don't know what it is. It may be we still need at least one language since that used to be a required field or it wouldn't build. Awaiting feedback from @mcdurdin

dyacob:
Thanks for checking, I've now added the last version of the languages list back in the array format.

DavidLRowe (approved):

DavidLRowe:
@dyacob Keyboard is merged. It'll take a while (few hours?) until it's available on the Keyman site.

PR #2070, opened 2022-11-12
GFF Harege Fidelat v1.0

dyacob:
This is a language-neutral touch (only) keyboard for mobile devices. The keyboard provides all Ethiopic letters in active use on a single layer in a circular, semi-clockwise orientation. 

The Harege keyboard is similar to the related Mesobe keyboard but is meant for more experienced users.  The layout is more compact and so fits better on mobile phones, it avoids the use of popups for vowel selections, and starting in the 6th order syllables should allow for faster typing overall.


keyman-server:

LornaSIL:
Besides the issues I noted in PR #2069 I would recommend you add to `keyboard_info` something like this:
```
  "related": {
    "gff_mesobe_fidelat": {
      "deprecates": false
    }
}
```
This will link these two keyboards together since you also try to reference them in the welcome file. It doesn't deprecate the keyboard since it is "false" it just says they are related.

This would mean `gff_mesobe_fidelat` needs approving first, then you may need to merge master into this PR after mesobe is approved before I can do a proper build of this one.

dyacob:
The last commit fixes the issue with punctuation and adds the recommended relationship to the `.keyboard_info` file, which also has more languages added (json validated as recommended)

DavidLRowe (approved):

PR #2073, opened 2022-11-17
sil_senegal_bsc_azerty

madskinner2:
Initial production release of sil_senegal_bsc_azerty keyboard supporting Windows/Linux/Mac/Web/|Android/iOS


keyman-server:

DavidLRowe:
Thanks for your submission @madskinner2 

- We'll need to have you remove the build/ folder and everything in it from this pull request. (The Keyman site will rebuild everything from source.)
- Also the sil_senegal_bsc_azerty.kpj.user file should be removed from this PR.
- The Copyright date in LICENSE.md should be 2021-2022 (to match HISTORY.md).
- I suggest removing the copyright statement and the version number from README.md, so that this file doesn't need to be changed when the keyboard is updated (but if you keep the copyright statement, you should remove the date or change it to match LICENSE.md).

In the source folder:
- As with README.md, the copyright statement in readme.htm should (a) be removed, (b) omit the date, or (c) have the date that matches LICENSE.md.
- sil_senegal_bsc_azerty.htm is an empty file and should be removed.
- Also, sil_senegal_bsc_azerty-help.htm, which is listed as the Embedded Help file (&KMW_HELPFILE in .kmn file), is empty. You may want to remove this feature and this file. 

Touch layout:
default layer:
- longpress for NBSP (on space) doesn't seem to produce any output. You may wish to use U_00A0 as the ID to generate U+00A0 directly without relying on the .kmn file to change your T_nbsp to the right thing depending on layer modifier.
numeric layer:
- key labeled "*" produces "8". You may wish to use the U_002A form of the ID to generate * (U+002A) directly without going via the .kmn file (as happens when K_xxxx is used as the ID).
other layer:
- key labeled "µ" produces nothing.

Package:
- The icon.ico file is 16x16 pixels. The expected size for the Image file used in the splash screen is 140x250. Keyman scales your file to fit. You'll probably like the results better if you create a 140x250 image yourself.
- Do the .odt and .pdf files need to be included in the keyboard? I'm not seeing where they are referenced (but maybe I'm missing something?).
- Perhaps it would be better to 
  - remove them from the list of files included in the package
  - move the files from source/welcome/ to extras/

madskinner2:
suggested mods/fixs made, pdf file are added to windows Startup Menu.

madskinner2:
fixes/mods made. PDF files are added to Windows Start Menu.

DavidLRowe:
Two minor nits to pick:

I think the copyright statement in the .kps file should include the 2021-2022 date range (or omit the date range altogether).

Similarly in readme.htm, the `<p>©2021 SIL International</p>` line should be removed or changed to 
`<p>© SIL International</p>` or
`<p>©2021-2022 SIL International</p>`

Thanks!

PR #2081, opened 2022-11-18
sil_senegal_bsc_azerty

madskinner2:
Hopefully the final initial production release of sil_senegal_bsc_azerty


keyman-server:

DavidLRowe (commented):
Looks good. Thanks!

DavidLRowe (approved):

LornaSIL:
@madskinner2 
The online help file doesn't look great: https://help.keyman.com/keyboard/sil_senegal_bsc_azerty

At a minimum you should add the same graphics files to the `help` folder that you used in `welcome.htm` (the `help` folder gets automagically migrated to another repo and doesn't have access to the graphics in any other folder).

I think you also need to include the `kb.css` and `latin.css` since the table isn't very well formatted.

This would require a new Pull Request, but since you are just modifying the online help file, there is no version bump needed on the keyboard itself.

PR #2082, opened 2022-11-19
GFF Harege Fidelat v1.01

dyacob:
This update addresses a number of the `nextlayer` attributes that were unset or not set correctly.  The issue is difficult to test thoroughly given the high number of both layers and keys, so I wrote a simple keyboard layout validation script which has also been added to the repository.


keyman-server:

LornaSIL (approved):
RSLGTM

PR #2083, opened 2022-11-20
[rejang] added rejang.php help file

Ilhamkang:
Dear @DavidLRowe, I've created the help file for the Rejang keyboard but when I compared the help file on the other keyboards that I've created before: Babyayin, Buhid, Hanunoo, Sunda, they seem don't have the image files in the help folder. 

So, I use the same way with the other keyboards help file. Hopefully it will work. Your suggestion for the keyboard image files on the help folder is rather new to me. Is there any change for displaying the keyboard there? I'm freely to adjust the method if it necessary.


keyman-server:

LornaSIL (approved):
I think it looks correct. Very often understanding the `osk` syntax is difficult for people which is likely why @DavidLRowe suggested graphics.

DavidLRowe:
The help file looks good. My comment was a general one (either use the graphics used in the welcome.htm or the keyboard image syntax supported by the PHP file). Since you had experience with creating the PHP file, you already knew how to use that method. 

But this made me realize that we need some documentation on the options for creating the PHP file, which is now available at: https://help.keyman.com/developer/keyboards/phphelpfile

Ilhamkang:
Thanks @DavidLRowe dan @LornaSIL !

PR #2085, opened 2022-11-21
NaijaType readme update

rowbory:
I just improved the first paragraph of the readme file as I realised it wasn't the most helpful for the Keyman installation page.

### DIFF
--- a/experimental/n/naijatype/HISTORY.md
+++ b/experimental/n/naijatype/HISTORY.md
@@ -3,6 +3,7 @@ Naija Type Keyboard Change History
 
 October 2022
 ------------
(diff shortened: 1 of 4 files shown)


keyman-server:

LornaSIL:
It looks good...except that it is the `readme.htm` file that is being displayed in the .kmp (as referenced in the .kps file).
I think the one you changed is used for Keyman web, so it's good that it got changed too. If you make the same changes to `readme.htm` (in the `doc` folder) it should be fine.

LornaSIL (approved):
I'm sorry. I saw the wrong file. You did fix the correct one.

PR #2087, opened 2022-11-22
GFF Mesobe Fidelat v1.01

dyacob:
This is a very minor update that changes the "ፀ" key on the "ፀ-layer" from the "Deadkey" type to the "Default" type.


keyman-server:

LornaSIL (approved):
LGTM

PR #2088, opened 2022-11-22
[gf_ethi] Add GF Ethiopic

LornaSIL:
These fonts support the Unicode 14.0 Gurage characters needed for the GFF keyboards.


DavidLRowe (approved):
RSLGTM

PR #2092, opened 2022-11-24
Update to 4.0.3 to include sentence capitalization feature

ind-nt:


keyman-server:

PR #2093, opened 2022-11-25
sil_senegal_cou_azerty

madskinner2:
Initial production release of keyboard supporting Windows/Linux/Mac/Web/Android/iOS for the Konyagi/Wamey language.


keyman-server:

PR #2094, opened 2022-11-25
sil_senegal_cou_azerty

madskinner2:
Initial production release with revised help files.


keyman-server:

DavidLRowe (approved):

DavidLRowe:
Looks good! I could approve this as it is (and indeed that's what I did), then I noticed that the splash screen image has "bsc" and is being pulled in from sil_senegal_bsc_azerty. There's a "cou" splash screen image included in the source folder, which I think is what you intended to use. 

If you want to use the "bsc" image, we can go ahead and merge this. Otherwise, if you update the .kps file to point to the splash screen image in the source folder, I'll re-review it.

PR #2097, opened 2022-11-29
chore: Remove extraneous begin statements

darcywong00:
From keymanapp/keyman#7583, the upcoming kmcomp 16.0 will flag keyboards that have duplicated `begin` statements.

No functional changes so I'm not bumping any keyboard versions.


DavidLRowe:
These changes do not increment the version number, but IIUC that's fine because this is just to have a version in the keyboards repository that will compile without warnings when the new compiler is used. There won't be any difference for the keyboard that is generated.

DavidLRowe (approved):

mcdurdin:
> These changes do not increment the version number, but IIUC that's fine because this is just to have a version in the keyboards repository that will compile without warnings when the new compiler is used. There won't be any difference for the keyboard that is generated.

Correct.

PR #2100, opened 2022-12-02
GFF Harege Fidelat & Mesobe Fidelat Adjustments for Keyman Web

dyacob:
These are updates to the `.keyboard_info` files to Harege Fidelat and Mesobe Fidelat to avoid appearing as Keyman Web options.

Paragraph justification is added to the respective `.php` help pages where it was missing.

These updates are as per the recommendations in the Keyman Community form discussions here:  https://community.software.sil.org/t/question-on-keyman-web-the-kmd-web-target/6972/1


keyman-server:

Issue #2072, opened 2022-11-16
Update of Wookianos Web Font


dyacob:
Keyman Web uses a web font (.woff) version of the Wookianos font that is referenced by many or most of the Ethiopic keyboards in the `.keyboard_info` files as per:

```
      "font": {
        "family": "GeezWeb",
        "source": [
          "wookianos.woff",
          "wookianos.ttf",
          "wookianos.mobileconfig",
          "ETHIOPI0.eot"
        ]
```

Wookianos, in turn, targets the Unicode 3.0 standard so is not suitable for a number of languages like Blin, Gurage and others. It also has ቬ misencoded in the position of ቭ which is only an occasional problem.   The Wookianos font was updated to support Gurage as per (Unicode 14) and Blin (as per Unicode 4.1) languages and can be found in the [gff_gurage](https://github.com/keymanapp/keyboards/tree/master/release/gff/gff_gurage) project in the `source` folder with the name `GurageEthiopicWookianos-Regular.ttf`.  Note that the Gurage fonts in the `gff_gurage` project in turn originate from the [gurage-font-collection](https://github.com/raeytype/gurage-font-collection) project.

Attached to this ticket is the same font, but renamed **"Keyman Wookianos"** , the archive includes `.woff` and .`woff2` web formats.  The refinements made are detailed in the  [FONTLOG-GurageEthiopicWookianos.txt](https://github.com/keymanapp/keyboards/blob/master/release/gff/gff_gurage/source/FONTLOG-GurageEthiopicWookianos.txt) in the `gff_gurage` project which can be repurposed as a FONTLOG file for "Keyman Wookianos".

I could not locate any of the files in the "GeezWeb" collection in keyman repositories.  If the "keyboards" project is the wrong place for this ticket, I can recreate the ticket in the appropriate repository if directed to it.

Food for thought, while the attached update to Wookianos is more complete than the original, it still does not provide every Ethiopic symbol defined in the Unicode standard.  The same I expect is true of the `ETHIOPI0.eot` font (which I could not locate).  It might be better to migrate to more modern and actively maintained fonts such as [Noto Ethiopic](https://github.com/notofonts/ethiopic/releases/tag/NotoSerifEthiopic-v2.102) or Abyssinica SIL which have full Unicode coverage for Ethiopic.

dyacob:
[...]

PR #2108, opened 2022-12-09
Migration to the Noto Ethiopic Web Fonts.

dyacob:
This PR changes all references to the Wookianos (and other) web fonts referenced in the `.keyboard_info` files of the GFF keyboards.  The `fonts` field entry was also added to a number of keyboards that did not have it set.  An exception is the gff_ethiopic_7 (language-neutral) keyboard where a correct update would need more time and so will be revisited later.

The formatting of the `.keyboard_info` files was also synchronized.

This PR should complete #2072


keyman-server:

LornaSIL (approved):
LGTM
I think I didn't need to delete the gff_geez name, but at first I thought the issue was with the apostrophe. Then I figured out it was a spurious comma.

PR #2109, opened 2022-12-12
sil_senegal azerty for dyo, gsl, krx, ndv, sav, snf, srr and wo

madskinner2:
initial production release of azerty keyboards for  dyo, gsl, krx, ndv, sav, snf, srr and wo,


keyman-server:

LornaSIL:
Wow, lots of new keyboards! This is great! I have a few structural things that would be helpful to get right first:

* These keyboards should get the minimum version set in the header: dyo, gsl, krx, ndv, sav, srr, wo. (`snf` has it set correctly):
`store(&VERSION) '10.0'`
* `srr` has the `.kps` file in the root folder, not in source. That will need to be moved to `source`. Then you'll need to re-associate the `.kps` file correctly in the `.kpj` file. You can either do that through Keyman Developer (go to "Packaging" tab and remove the existing one, then re-add it in from the correct `source` folder (or, you can use a text editor to edit the .kpj file, to correct the path, but you have to do that very carefully).
* All your keyboards are using the same font. I'd really rather we not put the same font in every keyboard folder. That makes it really difficult if we ever need to update the font again, we have to touch every keyboard. Instead of distributing the font in `extras` you could reference (in the .kps) the font that is in `\release\shared\fonts\sil\andika_subsets`. Then you could remove the fonts in each keyboard folder.
   * When you do this (in the .kps), the main issue I've run into is that I need to re-select the "Keyboard font" in the "Keyboards" tab of the .kps file.
* README.md
   * Some of the keyboards have a link to your work email address, one of them (wo) has a link to your personal email address, and some don't have a link. You should probably standardize. I personally wouldn't want my personal address there and not even my work address. Perhaps a link to the community site, or to the keyboard site. Many of our README.md files just link to the online help page for that keyboard.
```
Links
-----
https://keyman.com/keyboards/basic_kbdadlm
```

madskinner2:
sil senegal azerty keyboards for bsc, cou, dyo, gsl, krx, ndv, sav, snf, srr and wo:
font set to shared, README.md links removed

DavidLRowe:
srr has the .kps file in the root folder, not in source. That will need to be moved to source. Then you'll need to re-associate the .kps file correctly in the .kpj file. You can either do that through Keyman Developer (go to "Packaging" tab and remove the existing one, then re-add it in from the correct source folder (or, you can use a text editor to edit the .kpj file, to correct the path, but you have to do that very carefully).

All your keyboards are using the same font. I'd really rather we not put the same font in every keyboard folder. That makes it really difficult if we ever need to update the font again, we have to touch every keyboard. Instead of distributing the font in extras you could reference (in the .kps) the font that is in \release\shared\fonts\sil\andika_subsets. Then you could remove the fonts in each keyboard folder.
When you do this (in the .kps), the main issue I've run into is that I need to re-select the "Keyboard font" in the "Keyboards" tab of the .kps file.

bsc: 
After the font change, OSKFont, DisplayFont are no longer set in .kps file. Is this intentional? I see the same for cou and all the new keyboards. It's up to you whether you make this change or not.

bsc: 
- The source folder has empty files: sil_senegal_bsc_azerty.htm, sil_senegal_bsc_azerty-help.htm which should be removed.
- Is dad.jpg file needed? I'm not seeing a reference to it.

srr: 
This keyboard project still has the old .kps in the wrong location (root instead of source) in addition to having the new .kps file in the source folder. The new .kps file in the source folder (the correct location), unfortunately has incorrect references to file paths. (Need to remove "source\" at the beginning of most files; change "build\" to "..\build\" at the beginning of others, etc.) May be simplest to delete the package file and build from scratch. (See comment above from @LornaSIL )

Note: Optional: If using Keyman 15 or later, you can use, for example, U_0068_0303 (instead of T_h_tilde) as the name of the key in the touch layout, and then the `+ [T_h_tilde] > U+0068 U+0303` rule won't be needed in the .kmn file. (But, if you do opt for this format, the .kmn file should include `store(&VERSION) '15.0'` to indicate that Keyman 15 compiler is needed.)

(Keyboard approval work will resume in January.)

PR #2115, opened 2023-01-01
Added Right Alt modifiers for numbers and symbols on the top row of keys

PostmodernEnglish:
Postmodern English keyboards are populated with extra letters on the top row of keys, which are usually used for numbers and symbols on standard keyboards. Before this update, a user would need to press the dead key before each number or symbol was typed. Since this was tedious, a Right Alt modifier key was added, which allows a user to hold down RALT, and then type numbers or symbols without any further extra keystrokes. This modifies the UK Natural keyboard.


keyman-server:

LornaSIL (approved):
Cursory review....LGTM

PR #2119, opened 2023-01-06
[sil_khamti] Initial creation based on khamti_uni

laineyhm:
This is checking in the Khamti Shan keyboard for desktop layout (previously khamti_uni)
Created by @mhosken and @lornaSIL

* Also added OSK and icon

Next month we plan to add a touch keyboard layout


keyman-server:

LornaSIL:
Thanks for your submission! 

There are 3 warnings when I compile:
```
sil_khamti.kmn (182): Hint: 10AE This rule will never be matched as another rule takes precedence Overridden by rule on line 84
sil_khamti.kmn (86): Hint: 10AE This rule will never be matched as another rule takes precedence Overridden by rule on line 85
sil_khamti.kmn (81): Hint: 10AE This rule will never be matched as another rule takes precedence Overridden by rule on line 80
```
* Basically, the rule on line 182 is the same as line 84 although the syntax is different. You can deleted line 182
* I don't see line 86 and line 85 being exactly the same, but I do see that there is overlap in some of the stores: `baseK` and `takesVSK`. Right now line 85 takes precedence so you'll want to figure out which rule is correct.
* Similar situation for line 80 and line 81. 

You don't *have* to deal with these warnings, but it would be good to get it correct.

You have the keyboard name in the .kmn and .kps as "SIL Khamti". My own preference would be to make it "Khamti (SIL)" so the language name is more important than the organization name. Do not change the filenames/foldername, just the name of the keyboard unless you really want to keep it as "SIL Khamti". (This recommendation is based on the past few years of working with naming keyboards...originally I was putting "SIL" first.)

In the **Files** section of .kps you could remove `LICENSE.md`, `README.md`, `HISTORY.md`. They are really there just for the repository and don't need distributing. You don't have to do it, it just makes it cleaner.

In the **Details** section of the .kps file, you add `readme.htm` to **Readme file**. That is the file that will be displayed when someone first opens the .kmp.

Did I really help with this keyboard? I don't remember, and I'm not seeing it in our repository. If it's in the repository already we could do some code to deprecate the original keyboard.

Ok, none of my suggestions are show stoppers, so we could approve this as is, or you could fix the suggestions above. Let me know.

darcywong00:
> There are 3 warnings when I compile:

Yeah, we saw the compiler hints and originally tried commenting out the lines. But then the keyboard didn't work the same as khamti_uni. Some of the keys would output Latin characters. So when we added the lines back, it worked.

We didn't understand why or investigate further.

------

> Did I really help with this keyboard? I don't remember, and I'm not seeing it in our repository. If it's in the repository already we could do some code to deprecate the original keyboard.

My bad. I thought I heard or read that somewhere. Guess we can list @mhosken as the sole author on readme.htm.

laineyhm:
Thank you @LornaSIL! I have pushed commits with these changes:

- Remove redundant line 182 rule. I left the other two in for now because we don't have a good idea about which rule to keep, and while not as succinct as the compiler wants, the two rules together *seem* to produce a good result. We will evaluate all of this soon with colleagues in the language community. 
- Renamed the keyboard and package to Khamti (SIL)
- Removed LICENSE.md, HISTORY.md, and README.md from the package file
- Added readme.htm to the package details
- Removed your name

LornaSIL:
I just realized Khamti is based on Shan, and I *did* help move Shan to the repository, adding all the necessary extras that were not a part of the origin package...but I don't consider that I am an author, just doing the extra necessary bits.

LornaSIL:
Keyboard is now available: https://keyman.com/keyboards/sil_khamti

PR #2120, opened 2023-01-10
Fixed a few mobile keyboard layers not working on all PME keyboards

PostmodernEnglish:
This bug was caused by a lack of a key ID for certain keys. Adding the needed IDs resolved the problem in our tests.


keyman-server:

LornaSIL (approved):
LGTM

PR #2126, opened 2023-01-16
[ausephon]

benaustralia:
Initial submission of ausephon keyboard


keyman-server:

DavidLRowe:
Thanks for your submission. There are a few changes needed:

- It would be nice to change the description in the readme.htm and ausephon.keyboard_info files (to be something other than "generated from template"). Perhaps "a keyboard for phonemic (broad) transcription of Australian accented English speech" (lifted from the welcome.htm). Whatever description you decide to use could also go into README.md.

- In the ausephon.keyboard_info file and the ausephon.kps file, the BCP 47 language tag is specified as en-AU. That is probably the best choice. If this keyboard is generating IPA data, then the preferred tag might be en-AU-fonipa, but support for the -fonipa subtag is not universal, so sticking with en-AU is best for now.

- You've included the aus_phonemic_3.ico file in the source folder, but the .kmn file doesn't reference it. With Keyman Developer closed, edit the .kmn file with a text editor and add:
```
store(&BITMAP) 'aus_phonemic_3.ico'
```
When you restart Keyman Developer, it should recognise the icon. Note that the icon is compiled along with the .kmn file into the .kmx file, so no need to include the .ico file in the list of files included in the package.

- The on-screen keyboard (OSK) currently only shows the keys that have changed from the underlying keyboard. It would be preferable to show all keys that produce output. Using the "On-screen" tab on the left, you can tick the "Auto-fill underlying layout" box at the bottom of the screen, then use the "Fill from layout" button at the top to fill in all the characters. But see the next two comments before you do that.

- The OSK has "." on key to left of right shift, but this key really produces "/". The procedure in the previous paragraph should fix that (by making the OSK reflect the keyboard layout).

- Currently the keyboard produces upper case letters (QWRUPFGHJKLXCVBM) on the shift layer for keys you haven't reassigned. If you want those keys to produce no output, you need to include a rule to do that explicitly. For example, you could add
```
store(nooutput) "QWRUPFGHJKLXCVBM"
```
at the end of the list of initial `store` commands, then add (after the `group(main) using keys`):
```
+ any(nooutput) > nul
```

- The package source file (.kps) has some problems with the paths to the four font files and to the layout150.png and keyboardblack.jpg files. For example, the path for the layout150.png file is:
```
..\..\..\aus_phonemic\aus_phonemic_3\source\layout150.png
```
It will compile on your machine because that path is present there, but it won't work on the Keyman build system. The easiest way to fix the problem is to use Keyman Developer, select the .kps file in the Packaging section, then select the Files tab on the left. You can then remove the four .ttf files, the aus_phonemic_3.ico file, the layout150.png file and the keyboardblack.jpg file. Then add those files (except the .ico file) back into the list of files in the package, but navigate to the correct location before selecting the file. As mentioned above, there's no need to include the .ico file in this list since the contents of the file have already been included in the .kmx file.

If there's anything unclear, please let me know. (If it's any consolation, this is in better shape than my first keyboard submission!)

benaustralia:
Thanks David for your generous and detailed feedback.

I've made all the changes you suggest (or at least I think I have...it's late here) except one.

> * Currently the keyboard produces upper case letters (QWRUPFGHJKLXCVBM) on the shift layer for keys you haven't reassigned. If you want those keys to produce no output, you need to include a rule to do that explicitly. For example, you could add
> 
> ```
> store(nooutput) "QWRUPFGHJKLXCVBM"
> ```
> 
> at the end of the list of initial `store` commands, then add (after the `group(main) using keys`):
> 
> ```
> + any(nooutput) > nul
> ```
We don't mind those keys producing output. The Australian IPA we output is all lowercase. That said, we are new to keyboards and will gladly listen to your advice if you can explain it more to us.

Again, thanks for reviewing us and we hope you don't mind having a second look.

DavidLRowe:
The changes to the description in various files looks good, and you've successfully added the icon.

The paths in the keyboard package still need correcting.

Suppressing the output of the capital letters is connected to the OSK issue. The bottom line is that we would like the OSK to reflect what the keyboard actually does. If there's a blank key, it should produce no output. If there's something printed on the OSK key, that should be what the key produces when typed. Right now (for example):
- the key to left of right shift has "." on the OSK, but produces "/"
- the capital "Q" key is blank on the OSK, but produces "Q"
- the lower case "s" key is blank on the OSK, but produces "s"

The key that produces "/" should have "/" on the OSK.

Keys like "s" that produce corresponding characters should show up on the OSK.

It's up to you whether you permit the unused keys on the shift layer to produce capital letters or not, but the OSK should match what the keyboard produces. Thus, you should put the capital letters on the OSK, or change the keyboard to suppress output for those keys. (My personal opinion would be to suppress the capital letters, thereby reserving those keys in case you decide to add an additional IPA character sometime in the future.) 

First decide what you are going to do about capital letters (leave them as is, or suppress output for them), then use the tool to fill the OSK from the layout. This will deal with all the OSK issues.

Our reason for wanting the OSK to match the keyboard is a selfish one. We want users of your keyboard to have as much information as possible so they don't write to us complaining that the keyboard doesn't work!

benaustralia:
Many thanks again John for guiding us through this.

I think I have now made all the changes you suggest.

I ended up editing the .KPS XML in my text editor (the Keyman Dev app seemed OK with that). I edited the those long pathways to look more like this:

```
    <File>
      <Name>CharisSIL-Italic.ttf</Name>
      <Description>Font Charis SIL Italic</Description>
      <CopyLocation>0</CopyLocation>
      <FileType>.ttf</FileType>
    </File>
```
If by chance I haven't managed to make all the changes you suggest, please let me know and I'll get my co-developer to work on them: Kym is the one with the actual IT degree, I just have a library/IS degree and a lot of persistence.

DavidLRowe (approved):
This will need a help file for the Keyman site. (The ausephon.php will provide the online help, modeled on the welcome.htm which provides the help packaged with the keyboard.) But that can be added later.

DavidLRowe:
I edited the .kvks (source for the OSK) to remove the capital letters in order to match the change you made to the .kmn file.

DavidLRowe:
@benaustralia The keyboard should be available at https://keyman.com/keyboards/ausephon once the Keyman site builds it and makes it available.

DavidLRowe:
@benaustralia A help file was added as part of PR #2133 . The keyboard was also changed to pick up the fonts from the shared folder. (This makes for less maintenance of keyboards when fonts are updated.)

If you end up submitting another keyboard or revising this one, you can duplicate the folder structure of the repository so that your keyboard references the shared font files correctly. In some folder "xxx", create these paths:
xxx/release/shared/fonts/sil/charis - contains your local copies of the Charis font files
xxx/release/a/ausephon - contains your keyboard files

PR #2129, opened 2023-01-19
[temiar] initial submission of Temiar keyboard

lublinkhof:


keyman-server:

LornaSIL:
Thank you for your keyboard submission! I don't think we have any keyboards for Malaysia as yet so that will be very useful.

Unfortunately, the structure of the files is not quite what our build system is expecting.

Basically, we need you to create a `source` folder within the `temiar` folder (so you would have `temiar\source` and then copy all your existing files into `source`. There are additional files that are needed, such as a `temiar.kpj` file which would be in the `temiar` folder. You can start by reviewing this page:
https://help.keyman.com/developer/keyboards/advanced/step-0

The other files that belong in the root of `temiar` are:
- LICENSE.md (must be MIT license)
- README.md
- HISTORY.md
- temiar.keyboard_info

The contents of that last file would be something like this:
```
{
    "license": "mit",
    "languages": [
        "tea"
    ],
	  "description": "This keyboard uses a simple set of key sequences for typing Temiar, based on standard English or Malay keyboard layouts."
}
```

The contents of this last file is what is used on the home page for where people download the keyboard.

LornaSIL:
@lublinkhof will you be able to fix this?

DavidLRowe:
This PR replaced by #2202.

PR #2132, opened 2023-01-20
[sil_boonkit] Update README.md

ermshiperete:
This change updates the README.md file for sil_boonkit so that it lists Linux as supported platform.

The keyboard already supports Linux, but the readme didn't reflect that until now.

### DIFF
--- a/release/sil/sil_boonkit/README.md
+++ b/release/sil/sil_boonkit/README.md
@@ -2,7 +2,7 @@ Keyboard Data
 -------------
 
 * Name:           Boonkit
-* Copyright:      2008-2018(C) SIL International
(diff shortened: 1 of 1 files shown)


LornaSIL:
Is this change sufficient for what is needed? The .kmn file doesn't even have a targets header statement:

`store(&TARGETS) 'desktop'`

I'm a bit surprised that we aren't getting an error.

If we do that change, that would mean a version bump.

mcdurdin:
If not present, `&targets` is assumed to be 'desktop'.

LornaSIL (approved):
Thanks. LGTM

PR #2133, opened 2023-01-20
[ausephon_runeboard] Add help files and change to using shared fonts

LornaSIL:


DavidLRowe (approved):

DavidLRowe:
@benaustralia We made some updates to your keyboard in order to add the online help that's available for keyboards on the keyman.com site. We took the information from the welcome.htm file, which provides the help packaged with the keyboard. (Someday we hope to have a unified source for the two help files.) 

We also took the opportunity to change the location of the fonts used by the keyboard package. Instead of storing separate copies in the keyboard source folder, the keyboard package now references copies from a shared folder. This means that when these fonts are updated in the shared folder, this keyboard will have that new version as well. 

Normally a new keyboard starts at version 1.0, but this one seems to have started at 10.0 (perhaps because of a confusion of keyboard version number and Keyman version number). We incremented the version number to 10.1 with this change. (Whenever the contents of the keyboard package change, the version number needs to increase so that Keyman will know there's a new version available.

Thanks again for your contribution. If you have any questions, please feel free to ask!

DavidLRowe:
The online help is now available at https://help.keyman.com/keyboard/ausephon/10.1/ausephon

PR #2136, opened 2023-01-23
[fv] documentation/readme cleanup

caforbes:
- fixed broken links to website or wiki
- fixed broken filename references
- updated READMEs in all FV files to remove duplicated copyright/version info


keyman-server:

LornaSIL:
I like the changes you are making. There is one unfortunate thing. If something that goes in a .kmp is changed, then we *must* bump the version number of the keyboard. This means that if you are just changing `README.md` then you don't need to bump the version number of the keyboard. But, if you are changing documentation that goes in the package (eg `welcome.htm`) then you need to bump the keyboard version. The reason for this is that unless the version is bumped, existing users of a keyboard will not be served the new content (or behavior).

So, since you are making changes to `welcome.htm` you should change the version number in the .kmn. It can be a tiny bump (9.1.1 > 9.1.2). Also, if you have `Package version follows keyboard version` set in the .kps file, then the only thing you *must* change is the version in the `.kmn` file. Of course, it would be nice if the change were documented in the `HISTORY.md` file, but I realize that would be a lot of extra work.

LornaSIL:
@caforbes will you be able to bump the version number on these keyboards?

caforbes:
@LornaSIL Yes sorry, I will try to finish it today or at least this week.

caforbes:
@LornaSIL done! please let me know if I have missed anything (else)! Thank you!

LornaSIL (approved):
LGTM.

PR #2139, opened 2023-01-25
chore: make resources/ scripts executable

darcywong00:
This just sets ci.sh and all the other BASH scripts in resources/ executable.

I note several .sh scripts use `#!/bin/bash` instead of `#!/usr/bin/env bash` but I hesitated from changing them


rc-swag (approved):
lgtm

PR #2141, opened 2023-01-30
fix(sil_euro_latin): repeated ik, ff, gn, qu language codes

mcdurdin:
Relates to keymanapp/keyman#8118

### DIFF
--- a/release/sil/sil_euro_latin/source/sil_euro_latin.kmn
+++ b/release/sil/sil_euro_latin/source/sil_euro_latin.kmn
@@ -14,7 +14,7 @@ c KeymanWeb-specific header statements
 store(&KMW_HELPFILE) 'sil_euro_latin.html'
 store(&KMW_EMBEDJS) 'sil_euro_latin_js.txt'
 store(&LAYOUTFILE) 'sil_euro_latin.keyman-touch-layout'
-store(&KEYBOARDVERSION) '2.0.2'
+store(&KEYBOARDVERSION) '2.0.3'
 store(&TARGETS) 'any'
 
 begin Unicode > use(Main)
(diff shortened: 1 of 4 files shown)


darcywong00:
I'd have warmer fuzzies if someone could load this .kmp on an iOS device/emulator :)

release/sil/sil_euro_latin/source/sil_euro_latin.kmn
```diff
@@ -14,7 +14,7 @@ c KeymanWeb-specific header statements
 store(&KMW_HELPFILE) 'sil_euro_latin.html'
 store(&KMW_EMBEDJS) 'sil_euro_latin_js.txt'
 store(&LAYOUTFILE) 'sil_euro_latin.keyman-touch-layout'
-store(&KEYBOARDVERSION) '2.0.2'
+store(&KEYBOARDVERSION) '2.0.3'
```
darcywong00, line 17: Should we also update l.149 in sil_euro_latin.kps?
```
     <Version>2.0.1</Version>
```

I think the fv_all build script uses the keyboard version defined in the .kps files

darcywong00 (approved):
lgtm

release/sil/sil_euro_latin/source/sil_euro_latin.kmn
mcdurdin: Problem here is that we use `<FollowKeyboardVersion/>` so Keyman Developer will remove the version number on save from the .kps. So a manual patch here is not going to work long-term. We may need to fixup fv_all/build.sh to read from the .kmp files.

mcdurdin:
> I'd have warmer fuzzies if someone could load this .kmp on an iOS device/emulator :)

Done. 2.0.3 installs without issue.

Note: I've now also enabled build artifacts on pull request test builds.

PR #2148, opened 2023-02-12
[alephwithbeth] 1.0.3

tycebrown:
- Fixed a faulty autocorrect.
- Added the title in a few places. 
- I believe I've fixed the issue of an inconsistent version number

One thing I noticed: if I search for the keyboard on keyman, there are zero search results if I include more than one word.

Let me show you what I mean.
The search results for just the first word: [https://keyman.com/keyboards?q=aleph](https://keyman.com/keyboards?q=aleph)
The search results for the first two words: [https://keyman.com/keyboards?q=aleph%20with](https://keyman.com/keyboards?q=aleph%20with)

Can this be fixed?


keyman-server:

DavidLRowe:
> One thing I noticed: if I search for the keyboard on keyman, there are zero search results if I include more than one word.
> Let me show you what I mean.
> The search results for just the first word: https://keyman.com/keyboards?q=aleph
> The search results for the first two words: https://keyman.com/keyboards?q=aleph%20with
> Can this be fixed?

Currently @mcdurdin has limited availability, but would be the one to address that question.

LornaSIL:
I think it would be less confusing if you could delete the version information in:
- README.md
- welcome.htm
- readme.htm
- alephwithbeth.php

And lastly, with the .kps file open, go to the details tab and DELETE the version number and then select **Package version follows keyboard version**. Right now it seems to still be 1.0.2. The other thing you should change on this **Details** tab is the keyboard **Package Name**. You've already changed it many other places, so this should be adjusted as well.

All of these changes will make it easier for you to maintain and update the keyboard without adjusting so many files.

LornaSIL (approved):
Thanks for your changes. I'll approve this. For the future, there is no need to bump the version number each time we request you to make changes. If a Pull Request isn't approved and merged, you can keep it the same version number. Once a Pull Request is merged and the keyboard version is online, then you do need to bump the version number.

mcdurdin:
> > One thing I noticed: if I search for the keyboard on keyman, there are zero search results if I include more than one word.
> > Let me show you what I mean.
> > The search results for just the first word: https://keyman.com/keyboards?q=aleph
> > The search results for the first two words: https://keyman.com/keyboards?q=aleph%20with
> > Can this be fixed?

This turns out to be an oddity in SQL Server Full Text Search which we haven't struck with other keyboards. 'With' is a 'stop word', and so it gets ignored in a phrase search. This does feel a little broken. It's possible to disable stop words; we'll think about doing that.

PR #2149, opened 2023-02-13
[osage_nation] make old keyboard classic and add reformed keyboard

LornaSIL:
- The `osage_nation` keyboard is now the "Classic" one.
- The `osage_nation_new` keyboard is now the official orthography.
- The `osage_nation_new` does not **deprecate** the Classic one, but I did want them to be related so I added the related info in `keyboard_info`.


DavidLRowe:
On the Files changed tab, using the "gear", there's a setting for hiding whitespace changes:
image

DavidLRowe (approved):

PR #2153, opened 2023-02-15
Bug Fix

Runeboard:
Fixed 3-dot punctuation mark bug (adding needless empty space after character) & updated relevant files.


keyman-server:

LornaSIL:
For the future, it would be easier to only have the version in the HISTORY.md file and in the .kmn file. Having a version number that must be updated every time you update the keyboard in readme.htm, README.md, etc just makes more work.

The same goes for tracking all your changes. You should only need to include them in HISTORY.md, not everywhere else.

LornaSIL (approved):
You can wait until you want to update the keyboard again.
This looks good and I'll approve it.

Runeboard:
> You can wait until you want to update the keyboard again. This looks good and I'll approve it.

Okay. Thank you.

PR #2154, opened 2023-02-15
[txo_toto] Add new keyboard on behalf of author

LornaSIL:
Submitting on behalf of author.


DavidLRowe:
Minor nits: 

Are these needed?
- toto_icon_blue.bmp
- toto_icon_blue.png
- toto_icon_blue_2.png
They show up in the .kpj and .kps files, but I don't see where they are used.

Should it be "txo-Toto" or "txo" for BCP 47 code? I guess Developer doesn't complain about "txo-Toto".

In the .kmn file should the &ETHNOLOGUECODE line be removed?

I'll approve this as is. If you decide to fix these, dismiss the review.

DavidLRowe (approved):

LornaSIL:
- I changed the tag to `txo` since that follows CLDR. I'm not convinced Toto is the primary script in use, but it is their own script so that makes sense.
- I deleted one unnecessary graphic (bmp). I had converted that to an .ico file. 
- `toto_icon_blue_2.png` is used in the .kps file and is displayed on opening the package.
- I think this one `toto_icon_blue.png` was the original and I'll leave it there for now.
- I removed the Ethnologue code in the .kmn file.
Thanks!

DavidLRowe (approved):

PR #2161, opened 2023-03-02
[gondi_gunjala] update for new font

DavidLRowe:
No changes to keyboard.
Updated Narnoor font in shared folder.
Changed package to use shared folder.


LornaSIL (approved):
LGTM

PR #2162, opened 2023-03-04
Carian - 1st Attempt

Runeboard:
Please let me know if anything needs fixing. Thanks.


keyman-server:

LornaSIL:
Thanks for your submission. 

I had a bit of trouble getting the touch layout to compile. I'm not sure which of these changes are required to make it compile and which would just be nice:
- In the .kmn file, change the order of those two statements to:
```
begin Unicode > use(main)
group(main) using keys
```
If you don't want to output Latin characters in the blank spots, you could add these rules after `group(mail)`
```
store(nul) "QWERIOASFHXC"
+ any(nul) > nul
```
- In **Keyboard / Fonts** add Catrinity as the font for both the **On-screen keyboard** and the **Touch layout**
- In the .kps file:
  - Add the Catrinity.otf font to the files
  - On the **Keyboards** tab, add the Catrinity font to Keyboard and Display font.
  - The BCP47 tag can just be `xcr`. 
- You should delete the .kmx and .kvk files from the source folder. These are compiled versions of your source files. Our build system automatically builds those and puts them in a folder (`carian\build`).
- In the keyboard_info file:
    - where you have "languages" it should be `xcr` rather than `Carian`
    - The description should be updated to include what you have in the readme. This file is actually the first thing people see when they are looking for a keyboard so it's helpful to give a good description (all on one line). 
    - Since hardly anyone will have a font that supports Carian, I'd add some lines to the `keyboard_info` file to include support for the font. In the end, this is kind of what I'd expect the whole keyboard_info file to include:
```
{
    "license": "mit",
    "languages": { 
         "xcr": { 
              "font": { 
              "family": "Catrinity", 
              "source": "Catrinity.otf" 
              } 
         }
    },
	  "description": "This keyboard is designed for typing the Carian language, using the Carian script. "
}

```

Let me know if you have questions about any of this.

Runeboard:
It seems syncing caused the original upload to be deleted so I'm about to upload everything again. Is syncing necessary before updating files?

Runeboard:
You're welcome.

For the 2nd attempt, I've made the changes except as stated below.

When trying to add the following:
store(nul) "QWERIOASFHXC"
+ any(nul) > nul
it says that the code is then too complex to display the keyboard. Hence I've not included it. This doesn't matter anyway and indeed it could even be useful to retain access to those characters as some Latin & Greek characters were at times used alongside Carian.

In the keyboard_info file, I was a little confused about which types of brackets to use as the square brackets turn blue on screen whereas the curly brackets don't. As such, I left the blue square brackets in place but otherwise followed your instructions. Please feel free to change them if necessary.

PR #2167, opened 2023-03-13
[sil_khamti] Add touch layout

darcywong00:
This adds a touch layout to the sil_khamti keyboard (mirrors the desktop layout of default and shift layers)


LornaSIL:
@darcywong00 I think you still need to bump the version in the kmn file.

LornaSIL (approved):
LGTM

PR #2171, opened 2023-03-17
[malar_malayalam] Update to version 1.6

ramesh-k:
#### Changes in Version 1.6
* Fix `ntha` keystroke to get conjunct `n̪t̪a` `<0D28, 0D4D, 0D24>`.
* Add shift modifier to `caps` (shift lock) layer in touch layout. (double-tapping Shift to reach a 'sticky' Caps Lock layer)


keyman-server:

PR #2175, opened 2023-03-26
GFF Amharic Keyboard v3.11

dyacob:
This is a minor release that addresses auxiliary resources and not the keyboard (.kmn) itself

* I noticed on the "Shift" layer of the virtual on-screen keyboard (`.kvks` file) that two keys had bitmap images instead of text as used in other keys.  The images were replaced with text.

* The `AmharicTypingPractice.pdf` document was actually the wrong document, it was the Amharic typing manual. It has been replaced with the correct document.


keyman-server:

LornaSIL (dismissed):
@dyacob I'm approving the keyboard.
If it were me, in the future I would remove all the history from the welcome and help files so you don't have to keep updating that kind of information. The `HISTORY.md` should be sufficient.
Also, the version is `3.11`. Previously you used dot versions which would be `3.1.1` instead.
Would you prefer to change it to `3.1.1`? I won't merge it until you say you want `3.11`.

dyacob:
@LornaSIL -oh, was that an option? :-)  Yes, I totally agree, the history is definitely easier to maintain in just the one place.  I"ll make that fix.

And... yes, I did mean `3.1.1`, I've gotten rusty.  Thanks for catching that, I'll make the fix in an update.

LornaSIL:
We might have changed the rules when we got tired of making all those changes :)

DavidLRowe:
I think the change came when Marc was working on a keyboard and got tired of changing lots of files to keep them up-to-date!

dyacob:
I finally had some moments to make the corrections discussed. I hope that I didn't miss any.  A small note is that I removed the "AmharicTypingPractice.pdf" file, I'm not sure why it was there.  It is only useful as an editable file (which the companion .docx file is).

LornaSIL (approved):
LGTM

PR #2173, opened 2023-03-24
[fv_all] - Use jq to determine included keyboard versions

darcywong00:
Fixes #2143

> We may need to fixup fv_all/build.sh to read from the .kmp files.

I elected to update build.sh to use jq to parse the compiled `.keyboard_info` files to determine the updated keyboard versions for all the keyboards included in fv_all.kmp.

Also bumped the version of fv_all to 12.5.

Tagging @caforbes

PR #2180, opened 2023-04-10
[fv_all] Cleanup build.sh

darcywong00:
Addresses review comments from #2173 
* Use jq.inc.sh for cross-platform build
* Remove unnecessary assignments and logging

Not bumping fv_all version since no new functionality


mcdurdin (approved):
LGTM. A follow-up could update all other JQ references in ci.sh, help-keyman-com.sh, regression.sh scripts so that we can run keyboard builds on all CI agents.

PR #2182, opened 2023-04-12
sil-el-ethiopian-latin: add 'mul' and update .ico

rik-shaw:
I am definitely rusty, but am attempting 2 minor changes:
1. Add 'latin' as the first listed supported language. This will enable the keyboard display when using ibus (in Linux) to display as "Latin - SIL EL - Ethiopian Latin".
2. Update the .ico file to be mulit-resolution. Hopefully this works, it will be nice to see a sharper image in the keyboard switcher overlay (in Linux)


keyman-server:

DavidLRowe:
The main thing to change is to add "la" to the list of languages in .kps file.

Also welcome.htm has old date in copyright statement, but see below.

Optional changes:
- We no longer require the copyright statement and version number in the README.md file. Consider removing them so that future revisions won't need to touch this file.
- Similarly, you can remove the date from the copyright statement in the .kmn file
- You can remove the entire copyright statement (or at least the date) from readme.htm and welcome.htm.

Still needed:
- We do need the full copyright statement, including date, in the LICENSE.md file.
- We need the version number in the .kmn file (which should get picked up by the .kps file).
- It's helpful if the HISTORY.md file has the version number and date of the change.

As for the icon file, here's what the Keyman site has: https://help.keyman.com/developer/language/reference/bitmap
@mcdurdin Is that up-to-date? I'm not sure what resolution the icons are in that file. Opening with Microsoft Paint gives 16x16, opening with another viewer shows 256x256, Windows itself reports 128x128.

mcdurdin:
Re 'la', see my comment here: https://github.com/keymanapp/keyman/issues/7548#issuecomment-1506262072. It's a patch for Linux which has negative impacts elsewhere...

Re icon files, yes the data in that topic is up to date; we haven't touched this since 9.0. .ico is not the perfect format because it is very Microsoft, but it is universally supported due to favicon and there is no other multi-res format that is as well supported. There are plenty of online tools that allow you to open and edit multi-res icons, e.g. https://redketchup.io/icon-editor is pretty good.

rik-shaw:
>

rik-shaw:
> 



> The main thing to change is to add "la" to the list of languages in .kps file.
> 
> Also welcome.htm has old date in copyright statement, but see below.
> 
> Optional changes:
> 
>     * We no longer require the copyright statement and version number in the README.md file. Consider removing them so that future revisions won't need to touch this file.
> 
>     * Similarly, you can remove the date from the copyright statement in the .kmn file
> 
>     * You can remove the entire copyright statement (or at least the date) from readme.htm and welcome.htm.
> 
> 
> Still needed:
> 
>     * We do need the full copyright statement, including date, in the LICENSE.md file.
> 
>     * We need the version number in the .kmn file (which should get picked up by the .kps file).
> 
>     * It's helpful if the HISTORY.md file has the version number and date of the change.
> 
> 
> As for the icon file, here's what the Keyman site has: https://help.keyman.com/developer/language/reference/bitmap @mcdurdin Is that up-to-date? I'm not sure what resolution the icons are in that file. Opening with Microsoft Paint gives 16x16, opening with another viewer shows 256x256, Windows itself reports 128x128.

rik-shaw:
Oh my I had wrongly generated the .htm file previously in an attempt to remove the empty boxes that were showing where paragraphs wrapped due to using tidy. Corrected now. I hope it is now acceptable.

In the future re-writing the welcome.htm would be of interest, LibreOffice isn't the greatest way to generate .html :-) I do need a PDF as well, as new users typically print it out so it is for these users that I primarily target the readme.

keyman-server:

rik-shaw:
Thank you for your patience. Upon explanation and guidance from @mcdurdin I have re-readded mulltiple resolutions to the icon, and also removed `English` and added `mul (Multiple languages)` to the list of supported languages. I hope that this will now be acceptable.

DavidLRowe:
There are discussions about the possibility of having a common source in Markdown to replace the welcome.htm and the xyz.php help files. It seems to me that that same Markdown file could also be used to create a PDF for users to print. (Does that seem feasible @mcdurdin ?)

rik-shaw:
> The .keyboard_info file has "en" rather than "mul", but other than that everything looks good to me.

Thank you, now corrected.

rik-shaw:
> There are discussions about the possibility of having a common source in Markdown to replace the welcome.htm and the xyz.php help files. It seems to me that that same Markdown file could also be used to create a PDF for users to print.

Are there some "recommended examples" in other keyboards for "grid-type" keyboard charts I could look at? I do want it to be "print-first" capable, but am open to it NOT being in LibreOffice since the source behind the LO is a mess for version control :-) I think I would look to re-write the keyboard charts in future commits if that is acceptable.

LornaSIL:
>Are there some "recommended examples" in other keyboards for "grid-type" keyboard charts I could look at? I do want it to be "print-first" capable, but am open to it NOT being in LibreOffice since the source behind the LO is a mess for version control :-) I think I would look to re-write the keyboard charts in future commits if that is acceptable.

@dyacob has keyboards with a grid and tidy html: http://help.keyman.com/keyboard/gff_tigrinya_ethiopia I think the initial files were probably a lot of work, but worth it in the end.

DavidLRowe (approved):

PR #2183, opened 2023-04-14
[font_arab] Update SHL fonts to v4.000

LornaSIL:
Update to new versions of Scheherazade New, Harmattan, and Lateef.


DavidLRowe (approved):
RSLGTM

PR #2185, opened 2023-04-18
correcting version number

rik-shaw:
Unfortunately I see I missed revving the version number in the .kps and .kpj files. I am not certain if I need to modify to version 1.4.2 or if these very minor updates can be applied "as is" to version 1.4.1? Thank you again!

### DIFF
--- a/release/sil/sil_el_ethiopian_latin/sil_el_ethiopian_latin.kpj
+++ b/release/sil/sil_el_ethiopian_latin/sil_el_ethiopian_latin.kpj
@@ -11,7 +11,7 @@
       <ID>id_a99b7133501687f365aca688249b4b88</ID>
       <Filename>sil_el_ethiopian_latin.kmn</Filename>
       <Filepath>source\sil_el_ethiopian_latin.kmn</Filepath>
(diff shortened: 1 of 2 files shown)


LornaSIL:
In the .kps you have checked "Package version follows keyboard version" so this PR seems unnecessary to me. Version 1.4.1 is online.

rik-shaw:
> In the .kps you have checked "Package version follows keyboard version" so this PR seems unnecessary to me. Version 1.4.1 is online.

OK thank you! I was seeing it still showing as 1.4.0, but then with the terminal install of it I was prompted a newer version was available and it does display as 1.4.1. I think it best to just discard this PR (I could close but maybe you prefer some other way?), when I have a more substantial change in the future I will do some better checks before submitting the PR. Sorry for the chatter, thanks again!

LornaSIL:
You can close the PR and delete the branch. Yes, versioning has been a pain which is why we highly recommend that box gets ticked so you don't have to change the version anywhere except the .kmn file and HISTORY.md file. We recommend removing versioning from all the other files in order to reduce the work when we have a keyboard update.

PR #2186, opened 2023-04-19
sil_senegal_bqj_azerty and sil_senegal_knf_azerty

madskinner2:
Initial production releases.
File 'release/si/sil_senegal_cou_qwerty' created by mistake, please delete.


keyman-server:

DavidLRowe:
I deleted (as requested) the empty sil_senegal_cou_qwerty file.
I also deleted 2 .prj files.

One issue with sil_senegal_bqj_azerty
On the touch layout, press "cv" key, then press shift to go to "othercaps" layer. The ‹≈ and ≈› keys don't work. (You could replace "T_rsnglguillmet" with "U_2248_203A" as the ID field, then remove code from .kmn file for T_rsnglguillmet. Otherwise debug .kmn code to find out why rule for T_rsnglguillmet is not outputting data. Ditto for left guillmet.)

It would be helpful to change the description in the readm.htm file. Right now it has "template" which is not very helpful. 

This is what I've found so far. I'll continue tomorrow.

DavidLRowe:
Looking at sil_senegal_knf:
- Could you put a meaningful description in the readme.htm? It could be just the one line used in the .keyboard_info file. This is the first thing displayed when a keyboard is about to be installed and "generated from template" doesn't communicate much (or inspire confidence!).
- On the touch layout, "cv" key, then shift (uparrow) to get to "others" layer. Typing the `‘` key works, but not typing the `’` key. Not sure why, but consider using U_2018 and U_2019 as ID values and avoid reference in .kmn file.

madskinner2:
sil_senegal_bqj_azerty and sil_senegal_knf_azerty both raised to version 1.0.1.
bqj:Touch keyboard 'others' layer '‹≈' and '≈›' keys pointed at correct code in kmn. Place holders 'NBSP' and 'NNBSP' replaced with '≈' and '≠'. Documentation amended accordingly.
'readme.htm' updated based on description in kmn file.
knf:Place holders for  NBSP and NNBSP changed to '≈' and '≠'. 'Ĉ' and 'ĉ' added to modifier sequences. Documentaion updated.
'readme.htm' updated based on description in kmn file.

DavidLRowe (approved):

DavidLRowe:
@madskinner2 Just FYI: The change from 1.0 to 1.0.1 isn't wrong, but it wasn't really needed since you were still making changes to the original pull request.

PR #2193, opened 2023-04-26
GFF Blin v1.5

dyacob:
This is an update to the GFF Blin keyboard that adds a mobile and tablet keyboard.  Other updates include support for combining marks (gemination and vowel length), and context-aware input for period and comma.  Desktop testing has been conducted on Windows and Mac, and the mobile and tablet testing has been performed in the web-based test only -because there is a peculiar problem that _may_ stem from a Keyman or Keyman Developer bug, or I am severely overlooking something.  Some elaboration:

Importing the `gff_blin.kmp` package into the Keyman App on the following devices fails with under the following scenarios:

* Android (KM 16.0138): When the `.kmp` file is selected, nothing happens. The app behaves as if no file was selected.
* iPhone (KM 16.0.139 ): Transfer via AirDrop, the following message appears: "Error: The package does not include support for your device."
* iPad  (KM 15.0.274): Transfer via AirDrop, the following message appears: "Error: The package does not include support for your device."

One other test on an iPhone with KM  14 saw the app flash on screen and disappear with no resulting installation.

The `TARGETS` variable was set for `any` which has worked fine for other mobile keyboards, I repeated with specific targets for the `iphone`, `ipad`, etc. but the outcome did not change.

Another note, is that upon building the keyboard the following warning message appears about codes that are never fired.  This may be true of the destop keyboard, but the codes are fired from the touch layouts, so I've ignored it:

```
gff_blin.kmn (169): Hint: 109A The rule will never be matched for key '፞' because its key code is never fired.
gff_blin.kmn (170): Hint: 109A The rule will never be matched for key '፟' because its key code is never fired.
```


keyman-server:

mcdurdin:
> Importing the `gff_blin.kmp` package into the Keyman App on the following devices fails with under the following scenarios:

The issue is that build/gff_blin.js is missing from source/gff_blin.kps, because the keyboard was originally desktop-only, so there was no gff_blin.js to include :grin:

dyacob:
Thanks @mcdurdin , I've rebuilt now and have started testing on the mobile devices and will pushing in updates shortly.

dyacob:
The missing `.js` file is now included and installation works for the mobile platforms.  However... it's only now that I remember that there are some extra fonts that I wanted to add for Blin from the "Waldba" collection (the successor to "Washra").  So, I may need a few days to review them for Blin glyph coverage.  Meanwhile, the Keyman team is welcome to evaluate the keyboards (2 touch, 1 desktop).  The additional fonts that I will add will not clash with keyboard-specific files.

dyacob:
The last update avoids the `The rule will never be matched for key '፞' because its key code is never fired` reporting.  This was achieved by substituting `U_135F` for `T_1355` as the `ID` for the touch keys, then adding a mapping `[T_135F] > U+135F`.  It seems unfortunate that this extra step was needed for the compiler to register that a U+ code was being emitted from a touch layer.

release/gff/gff_blin/source/gff_blin.kmn
```diff
@@ -163,20 +163,31 @@ c ---------------------Input of Special Symbols---------------------------------
...
+c
++ [T_135D] > U+135D
++ [T_135E] > U+135E
++ [T_135F] > U+135F
 c
 c The following can happen from the touch keyboards፡
 c
-U+135E + U+135E > U+135F c 1 + 1 > 2
-U+135E + U+135F > U+135D c 1 + 2 > 3
-U+135F + U+135E > U+135D c 2 + 1 > 3
```
mcdurdin: The reason this was not working is that you have to match on the virtual key identifier, not on the output character. This allows for matching on multi-part identifiers such as `U_1234_5678`, which is not otherwise possible.

For example:

```
U+135E + [U_135E] > U+135F c 1 + 1 > 2
U+135E + [U_135F] > U+135D c 1 + 2 > 3
U+135F + [U_135E] > U+135D c 2 + 1 > 3
```

DavidLRowe:
Some extraneous files have crept into this PR (or else I don't understand what's going on, which happens quite often!): 
- /gff_blin.kmn
- /release/gff/gff_harai/gff_harai.kpj

dyacob:
@DavidLRowe thanks for catching these, I think they just got caught in the crossfire as I was switching branches back and forth on my end.  I'll remove them now.

DavidLRowe:
The following files don't seem to be referenced. Are they necessary for the keyboard? Or should they go in an "extras" folder?
- gff_blin-orig.keyman-touch-layout
- gff_blin-wtih-syllable-popups.keyman-touch-layout
- sideimage-blin-animated.png
- sideimage-blin-house.png

DavidLRowe:
Also: 
- gff_blin-jacket-red.ico
- gff_blin-orig.ico

dyacob:
A couple of those files should be removed, and then I would like to hold on to the others for possible future use or as a reference.  An "extras" folder sounds perfect and should reduce confusion over them, I can add a file of notes to the folder to describe the holdings (Extras.md), this will also help me recall why I wanted to keep them :-)

DavidLRowe:
Typically the "extras" folder is a sibling of the "source" folder.

dyacob:
I've now stashed some files into an "extras" folder and pushed in the extra fonts for Blin.  

On the fonts, they are a new family named "Waldba" which is a successor to the "WashRa" fonts that have resided in the `release/shared/fonts/geez` folder for a number of years now. The Waldba extensions add the missing Blin letters to the older WashRa fonts, and additions for Gurage are nearly complete.  Once the Gurage additions are available across the 9 typefaces, I would want to move the fonts out of the `gff_blin/fonts` folder and into the `release/shared/fonts/geez` folder, replacing their older WashRa counterparts.

LornaSIL (approved):
LGTM

PR #2196, opened 2023-04-28
GFF Harari v1.0

dyacob:
This is a Harari language keyboard designed to comply with the orthography requirements of the Harari Language Research Academy (ሐረሪ ሪትማ በሕሲዋ ቢሳሎት አካዳሚ).  A desktop, phone, and tablet keyboard are provided.

The keyboards are ready for evaluation.  There are a couple of keyboard screenshots that I will update for the documentation (README.md and gff_harari.php) later today.


keyman-server:

DavidLRowe:
It seems there's an extraneous file included:
- release/gff/gff_geez_emufi/source/gff_geez_emufi-orig.kmn

The copyright date in LICENSE.md is 1997-2022, but it seems that it should be 2023 (based on HISTORY.md).

We recommend removing the copyright statement and version line from README.md, so that the file doesn't need to be changed for each revision.

Is this extra file needed?
- gff_harari-with-popups.keyman-touch-layout

LornaSIL:
Besides the things @DavidLRowe mentioned, the `keyboard_info` file has not been updated. It is likely a copy of the one used for gff_amharic.
I noticed the name of the keyboard is different in the .kmn from what is in the .kps. Perhaps this is intentional?

dyacob:
Thanks for catching these @LornaSIL , the issues that @DavidLRowe found were addressed in an update last evening.  I got the `keyboard_info` updated now.  The name in the `.kmn` file is intended, it is how I wish the name to appear in the keyboard menus on Windows (a terse form).  In the `.kps` file, the name is longer and more complete.  There is yet another "Name" field in the `keyboard_info` file, I don't know where this gets used but I've made it the same as with the `.kmn` file.

LornaSIL:
Unless I'm missing something, I think the `LICENSE.md` file still needs the copyright date updated to just `2023`?

dyacob:
Thanks, I did set it to 2023 now, also in the PHP help file.  I changed (c) to the copyright symbol © in two places as well, but avoided doing so in the License file.  My reason for doing so is probably outdated, I'm recalling the original MIT license file were ASCII only, and then not wanting to tamper with it more than necessary.  Is it OK to change the symbol here as well?

LornaSIL:
I don't think it matters whether we use ASCII or the copyright symbol. So, I'm happy with leaving it the way you have it.

LornaSIL (approved):
LGTM

LornaSIL:
I'm not sure how (obviously not checked very well by me), but the `keyboard_info` file says it supports `am` and the .kps says `har`. You can fix the keyboard_info file without bumping the version of the keyboard.

dyacob:
@LornaSIL , it's terrific that you caught it now though, thanks for that.  I'm surprised at how often I mess this file up.  I think it is the only(?) keyboard resource file that is *not* created via Keyman Developer.  I wish was created when a `.kps` file is saved. Much of the info is already in the KPS interface and a few extra input windows could be added for the extra info that is optionally added to the `.keyboard_info` file.

mcdurdin:
We are planning to merge the data in the .kps and the .keyboard_info file in v17 of Keyman Developer. FWIW, a .keyboard_info file is created when you start a new keyboard project in Developer. But merging the .kps data and .keyboard_info should make life easier, I hope!

PR #2199, opened 2023-05-02
GFF Blin: Fix to PDF to include intended Ethiopic fonts.

dyacob:
The only change in this request is to the typing manual PDF file.  I just realized that when the PDF was created, all of the Ethiopic font styles (Abyssinica, Waldba, etc.) were substituted for Nyala.  I must have had too much faith in Word for Mac to get this right and didn't check (it seems to depend on *how* you create the PDF, "Save as..." vs "Print"). I've redone it on Word for Windows and the fonts are there as expected.

Since this is a documentation update only, I'm hoping a version number increment can be avoided.  Thanks!


keyman-server:

LornaSIL:
We can try it, but if anyone was already prompted to update their keyboard, they won't get the updated documentation.

dyacob:
Ok, that is understood.

LornaSIL (approved):
LGTM

PR #2202, opened 2023-05-05
[sil_temiar] Initial creation

darcywong00:
Attempting to add a temiar keyboard from an APLTW 2023 participant @lublinkhof. (He said he previously tried to submit, but I can't find a "temiar" submission).


LornaSIL:
#2129 was the initial submission, but it didn't follow the correct file structure.

DavidLRowe:
For the README.md file, .keyboard_info file, readme.htm file, .php file: Can we change the description?
Even something like "This keyboard is for typing the Temiar language of Malaysia." is more helpful than "created by template".

The dates in the copyright statement in the LICENSE.md file don't match what's in the .kmn file. You can leave the dates out of the copyright statement in the .kmn file, if you like (assuming the dates in the LICENSE.md file are correct).

The OSK shows only keys that have changed. Generally it's helpful to "fill from layout" to get all the keys. Doing this now will overwrite the custom display for E, I and O. Does having `e è é ë` on the key display well?

darcywong00:
> The OSK shows only keys that have changed. Generally it's helpful to "fill from layout" to get all the keys. Doing this now will overwrite the custom display for E, I and O. Does having e è é ë on the key display well?

OK I've added "fill from layout"
image

DavidLRowe (approved):
I think this is ready. @darcywong00 If you have more changes, you can dismiss this review. If not, you can go ahead and merge.

PR #2205, opened 2023-05-07
GFF Blin v1.5.1

dyacob:
This is an update to the included fonts only.  The font dimension properties have been synchronized with Abyssinica for better typography when used together.  Blin "ng" letters have also been refined in many cases for greater clarity.


keyman-server:

release/gff/gff_blin/HISTORY.md
```diff
@@ -1,5 +1,8 @@
 # ብሊን (Blin) Change History
 
+## 2023-05-07
```
darcywong00, line 3:
```suggestion
## 2023-05-07 1.51
```

Just wondering if version 1.5.1 would be better than 1.51. (In the other places on the PR too)

darcywong00:
> This is an update to the included fonts only. 

Are you intending the keyboard to use the Waldba font files? If so, we should also update:
1. kps file so the font files are included in the keyboard package, and the keyboards to use Waldba.
2. .keyboard_info file so the the keyboard metadata gets updated.

release/gff/gff_blin/HISTORY.md
mcdurdin: Yes because 1.6 < 1.51!

dyacob:
Thanks @darcywong00 for catching the version problem, @LornaSIL corrected me on that recently also, so I'm more than a little embarrassed.  I've pushed in fixes.  Regarding the fonts, on my end I see the Waldba fonts in the KPS file list (see screenshot).  They are meant for the desktop version only, so I think a change to the `keyboard_info` file is not needed, since that is for the Keyman Web as I understand it.

WaldbaFontsInGFF_Blin

LornaSIL (approved):
LGTM.

Issue #2172, opened 2023-03-23
[sil_hebrew] diacritic marks not rendering consistently cross-platform on the OSK


MakaraSok:
It has been noticed that some key caps on the default layer of the Hebrew SIL keyboard appear empty, but they do output characters.

Keyman version: 16.0.37
iOS version: 16.3.1
Keyboard: [Hebrew (SIL) keyboard](https://keyman.com/keyboards/sil_hebrew)

It is interesting though that they do show up in an older version of iOS,

| iOS 16.3.1 | iOS 15.7.3 |
| --- | --- |
| <img width=250 src="https://community.software.sil.org/uploads/default/original/2X/b/b987fb0bf249b40b67bd108c3e75acd8df10fb16.jpeg"> | <img width=250 src="https://community.software.sil.org/uploads/default/original/2X/d/d7a984dfd139b3e49fd1776520266811ea2b1e94.jpeg"> |

MakaraSok:
A discussion has already been established at https://community.software.sil.org/t/hebrew-sil-keyboard-has-some-invisible-characters-on-ios-16-3-1-obu/7273.

mcdurdin:
See also:
* https://community.software.sil.org/t/keyboard-fails-to-load-load-correctly-on-iphone-xr-11/6958/3
* https://github.com/keymanapp/keyman/issues/7914

DylanCross:
Guys thanks for working through this issue here. Is the keyboard made by SIL as well, or did a community member make this one. Would there be any possibility of keyboard owner trying the 25CC to 200F fix in the keyboard directly instead of in the keyman codebase working? 

Agreed having a functional font for Hebrew, would be ideal, but it sounds like we don't have someone to do that. It seems like changing the characters in the keyboard would be a quick fix (although would take more time and devices to test that is doesn't break elsewhere).

mcdurdin:
> Would there be any possibility of keyboard owner trying the 25CC to 200F fix in the keyboard directly instead of in the keyman codebase working?

I am pretty sure that trying a keyboard-based fix is likely to lead to significant time and effort in testing that ultimately will lead us to conclude that we need a font-based fix :grin: (Also, in this instance, the keyboard owner is the same team that works on fonts anyway.)

mcdurdin:
The first step to a resolution here is to get a full list of the characters (Unicode values) required. If there are keys with multiple glyphs on the key, then we'd need to list those sets as well.

DavidLRowe:
[...]

PR #2207, opened 2023-05-10
[tai_tham_my] Tai Tham MY keyboard

MakaraSok:
new keyboard


LornaSIL:
Thanks for your keyboard submission! I have a few comments:

- In the `LICENSE.md` file, please add the copyright year. This is really the only place we are requiring it now.
- In the .kmn file it would be useful to add the header statement `store(&VERSION) '10.0'` This is the minimum Keyman version it will work with. That reduces the warnings, plus just makes it cleaner.
- There are remnants from your first file naming. Could you delete the following files from the PR:
   - `tai_tham_my.kvks`
   - `tai_tham_my.keyman-touch-layout`
   - `tai_tham_my.ico`
- Lastly, the keyboard won't work very well on mobile devices without the font. You should update the `keyboard_info` file to include these lines:
```
{
    "license": "mit",
    "languages": { 
         "khb-Lana": { 
              "font": { 
              "family": "A Tai Tham KH New V3", 
              "source": "A-Tai-Tham-KH-New-V3.ttf" 
           } 
        }
     },
	  "description": "The Tai Tham MY (SIL) keyboard was developed for typing Tai Lue language in Tai Tham script. The keyboard is available for computers as well as mobile devices."
}
```

MakaraSok:
Hi @LornaSIL, thank you very much for the comments. I've just made and pushed those changes/updates. PTAL again. :)

release/sil/sil_tai_tham_my/sil_tai_tham_my.keyboard_info
```diff
@@ -0,0 +1,12 @@
+{
+    "license": "mit",
+    "languages": { 
+         "khb-Lana": {
```
darcywong00, line 4: > Tai Lue language in Tai Tham script

Should this be Talu instead of Lana script?

release/sil/sil_tai_tham_my/source/sil_tai_tham_my.kps
```diff
@@ -0,0 +1,123 @@
...
+  </Files>
+  <Keyboards>
+    <Keyboard>
+      <Name>Tai Tham MY (SIL)</Name>
+      <ID>sil_tai_tham_my</ID>
+      <Version>1.0</Version>
+      <OSKFont>A-Tai-Tham-KH-New-V3.ttf</OSKFont>
+      <DisplayFont>A-Tai-Tham-KH-New-V3.ttf</DisplayFont>
+      <Languages>
+        <Language ID="khb-Lana">Lü</Language>
```
darcywong00, line 118: > Tai Lue language in Tai Tham script

Should this be Talu instead of Lana script?

LornaSIL: `Talu` is "New Tai Lue" and `Lana` is "Tai Tham (Lanna)" so I think it's correct.

LornaSIL:
I think it's good. One small change would be good. The graphics for the mobile layout in the pdf should be updated to remove the mention of the previous copyright holder (on the spacebar).

MakaraSok:
Thanks for the catching that name in the pdf. I've fixed it.

One outstanding issue which could be a concern among users is how the OSK is shown in Windows. The dotted circles causes chaos. It gets double everywhere. See the screenshot below.

image

MakaraSok:
@LornaSIL, One more thing though, should the font be moved to the `shared` folder or is it OK for now?

mcdurdin:
> One outstanding issue which could be a concern among users is how the OSK is shown in Windows. The dotted circles causes chaos. It gets double everywhere. See the screenshot below.

We are working on a systemic fix for this. See also https://github.com/keymanapp/keyboards/issues/2172#issuecomment-1539180214. I would like to capture this example as another one for the same fix. (So that is, don't hold off on this, we'll fix dotted circle chaos as a separate project!)

LornaSIL (approved):
LGTM. 

I'm willing for the font to be in the local folder for now. We should get in the habit of adding fonts to the shared folder though so we don't have to add them to the next project that uses the script.

PR #2208, opened 2023-05-10
[colchis_latin] Update per client request

LornaSIL:
Primary big change was to use hachek characters instead of breve.


DavidLRowe (approved):
LGTM

PR #2212, opened 2023-05-16
[symbolic_logic] add online help and change encoding of 3 files

DavidLRowe:


DavidLRowe:
The welcome.htm had UTF-16 encoding (which I think still causes a problem on Linux and maybe Mac?); readme.htm and .kmn files had unnecessary BOM at start of file. I also removed a few unneeded lines ("generated from template", version number).

I created the .php file from the welcome.htm and tested locally. The display isn't quite the same, but is usable.

LornaSIL (approved):
LGTM

PR #2222, opened 2023-05-24
GFF Harari v1.0.1

dyacob:
This update fixes two keys that were not triggering next-layers as expected. Also missing labiovelar popup keys were added.


dyacob:
Got it, thanks for catching that!

LornaSIL (approved):
LGTM

PR #2227, opened 2023-05-30
GFF Awngi &  Xamtanga Keyboard v1.0

dyacob:
Desktop, mobile, and tablet keyboards that support the orthographies of the  Awngi and Xamtanga languages derived from the writing practices of modern newspapers in these languages. The keyboard is believed to be suitable for related Qimant languages, however, no literature corpus has been found to confirm this assumption.


keyman-server:

LornaSIL:
I don't think we should be pulling fonts from another keyboard repo. I think it would be better if you added them to the `shared` folder rather than take the fonts from the `gff_blin` folder.

The help file seems to be the one for blin rather than updated for this keyboard.

The sideimage says it's the wrong aspect ratio. You don't have to fix it, but you might want to.

I haven't checked the keyboard very well, but can you check the phone layout for the K_L key? It says to go to the `ሰ-layer` and it displays the keys for "ሰ-layer" but it outputs the proper `ለ-layer` keys. I didn't try to figure out what was going on. The tablet layer was fine.

Everything else seems okay.

LornaSIL (approved):
LGTM

PR #2230, opened 2023-05-31
Himyarit Musnad v1.1

dyacob:
This submission of version 1.1 is made on behalf of the author,  @RiyadhAlfareh .

Version 1.1 adds Arabic letters to the keycaps of the touch layout for the benefit of greater clarity and ease of typing.

### DIFF
--- a/release/h/himyarit_musnad/HISTORY.md
+++ b/release/h/himyarit_musnad/HISTORY.md
@@ -1,6 +1,10 @@
 Himyarit Musnad Change History
 ====================
 
+1.1 (2023-05-30)
+----------------
+* Arabic letters added to key caps.
+
 1.0 (2022-09-29)
 ----------------
 * Initial release of keyboard.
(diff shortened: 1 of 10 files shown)


keyman-server:

LornaSIL:
It looks like the touch layout converted the text field (in code view) to surrogate pairs (`\uD802\uDE79\u0630`). I can't remember if that is a problem or not.

dyacob:
Oh, hmm, I wonder why? Is the combination of Musnad and Arabic side-by-side that made KMD save it differently?  I could try recoding it with an offline script, but maybe it has to stay this way for KMD to read it again later?

LornaSIL:
It seems to work, so I guess I'll approve it unless you prefer I wait.

LornaSIL:
Ok, this is the thread that has made me concerned about the surrogate pairs.
https://community.software.sil.org/t/keyboard-freezing-on-ios16/7496/7

@mcdurdin should we wait to approve this until it gets fixed?

mcdurdin:
That thread was a red herring -- the actual problem in that thread is most likely a broken font or bug in webkit with a specific font.

The surrogate-pair encoding of the kvks is not ideal but should not break anything.

LornaSIL (approved):
LGTM

PR #2232, opened 2023-05-31
[esperanto_sava_eujao] New keyboard on behalf of author

LornaSIL:


DavidLRowe:
I think it would be better to change the encoding of the esperanto_sava_eujao.kmn file to UTF-8. (I tried to make the change, but I guess GitHub thinks it's a binary file and won't let me.)

mcdurdin:
I think it is already UTF-8 but GitHub is misdetecting the encoding. If you go 'view file', you can see the content of it: https://github.com/keymanapp/keyboards/blob/53f238a4b28ac9f08252031a8ed469f309ec59c7/release/e/esperanto_sava_eujao/source/esperanto_sava_eujao.kmn

DavidLRowe:
When I downloaded the file, it showed up as UTF-16. I converted it to UTF-8, but GitHub wouldn't let me replace the file (even though the "Maintainers are allowed to edit this pull request" box is ticked).

LornaSIL:
I did change it to utf8 and saved without the BOM, but Github still has an issue with it. Maybe because it's comparing it to the previous one that was utf16?

DavidLRowe (approved):

mcdurdin:
> I did change it to utf8 and saved without the BOM, but Github still has an issue with it. Maybe because it's comparing it to the previous one that was utf16?

I can view the 'diff' okay from here now on this PR.

Sorry for the misdirection on my previous comment, yes, the file was utf-16 as @DavidLRowe said and it seems GH _sometimes_ translates that internally for presentation, and sometimes gives up :grin:

PR #2236, opened 2023-06-02
[tai_tham_my] fixed the e key

MakaraSok:
All files containing e key has been updated to match the preference of having "a combination of 2 unicode characters 1A63 Tai Tham Vowel Sign AA and 1A74 Tai Tham Sign Mai Kang ᩣ + ᩴ = ᩣᩴ" on the `e` key.


LornaSIL (approved):
I updated the version number and removed the author. Hope it's correct!

PR #2237, opened 2023-06-02
Update HISTORY.md

LornaSIL:
Updated version and documented the fix.


LornaSIL:
Committed to wrong place.

PR #2240, opened 2023-06-05
GFF Blin v1.5.3

dyacob:
This update uses the shared Waldba font folder.  Additionally, an issue labiovelar entry on mobile platforms has been addressed where the letter was not replacing the 6th form letter.  This issue was common to a number of GFF keyboard updated in recent weeks and is being fixed across the board.


keyman-server:

PR #2251, opened 2023-06-08
nepali_traditional add eyelash half RA in desktop layout

sapradhan:
in nepali text, RRA (ऱ) is always followed by halanta ( ् ) for eyelash half RA as in (पऱ्यो) , hence adding RRA halanta instead of RRA only


keyman-server:

LornaSIL (approved):
LGTM

PR #2253, opened 2023-06-13
Update to 4.0.4

ind-nt:
Minor update with three more keys encoded.


keyman-server:

DavidLRowe:
Note that for "tablet" layout, there are warnings that the "rightalt" and "rightalt-shift" layers are missing Backspace and Enter keys. In the Project settings, the "Treat warnings as errors" box is unticked to allow the build to succeed.

DavidLRowe (approved):

PR #2263, opened 2023-06-23
[font_abys] Update Abyssinica SIL font

LornaSIL:


darcywong00 (approved):
lgtm

mcdurdin:
Note that this will not be released in existing versions of the keyboards -- they'd need a version bump.

LornaSIL:
> Note that this will not be released in existing versions of the keyboards -- they'd need a version bump.

Yes, and that's fine. It wasn't a big change, but it enabled @dyacob to produce the Alternate Typetuned fonts he wanted to add to gff_ethiopic which was just approved.

darcywong00:
This https://github.com/keymanapp/s.keyman.com/blob/master/font/deploy/AbyssinicaSIL-Regular.ttf also get updated?

LornaSIL:
Technically, it should be updated, but the fix to the font only affects the ability to use it on typetunerweb, so I didn't bother this time. I just wanted the versions in the repo to match the new "alternate 1" and "alternate 2" fonts that @dyacob was adding.

dyacob:
I fully concur with @LornaSIL .  The update to the font was very minor and will not impact any of the GFF keyboards or documentation, it was sort of a metadata update that allowed 2 derivative fonts to be generated correctly.

So it's nothing critical and no rush to make the update that @darcywong00 has asked about, the update can be made whenever there is a convenient opportunity.

mcdurdin:
Thanks @LornaSIL, that's helpful and means I won't be scratching our heads later re font versions :grin:

PR #2274, opened 2023-07-04
GFF Gurage v1.0

dyacob:
This update to the GFF Gurage keyboard brings it up to date with other recent revisions to GFF keyboards.


keyman-server:

LornaSIL:
@dyacob Unfortunately, the font changes (I think) being made in this PR affect the package `\release\packages\gff_gurage_and_amharic` so the build fails. You will need to update the fonts in the package in the same way as you did in the gff_gurage keyboard. Also, change the language tag from `sgw-Ethi` to `sgw` in the kps AND in `keyboard_info`. You'll need to make the changes to the package in this same PR since the package is dependent on the changes being made in this PR.

The way packages work is that you'll need to bump the version of the package in the .kps and not rely on versions in the .kmn.

dyacob:
@LornaSIL , I've created PR #2282 which migrates the Amharic & Gurage package to use the Waldba font collection.  It uses the versions in the master branch of course.  Once the new gff_gurage package is merged into master I'll make a v1.0 version of the package that uses the update as expected (mostly documentation changes).

dyacob:
The `gff_gurage_and_amharic` folder is now updated and synched.

LornaSIL (approved):
LGTM

PR #2275, opened 2023-07-04
GFF Gurage Legacy v0.5

dyacob:
This update to the GFF Gurage Legacy keyboard address a number of minor issues and brings the keyboard up to date with recent versions to GFF keyboards.


keyman-server:

PR #2276, opened 2023-07-04
GFF Ethiopic v1.0

dyacob:
This is a rewrite of the GFF Ethiopic keyboard that was last updated for Keyman 7. The keyboard supports input of all 523 symbols defined in the Unicode standard for Ethiopic.  The package supports the composition of 59 additional variant forms available in the Abyssinica SIL font.

Tablet and phone layouts have also been added.


keyman-server:

LornaSIL (approved):
LGTM

PR #2281, opened 2023-07-05
[common_devanagari] Updates

Raji1205:
We have done some changes in existing keyboard. Please give reviews/comments. Thank you


keyman-server:

LornaSIL:
This branch has too many conflicts as you can see in the message above. You may need to update your branch first and then recommit. **Before** you recommit, make sure you do NOT commit anything in the `build` folder. We do not want the compiled files as part of the pull request because those files are automatically built on our build system. 

Also, the `common_devanagari.kpj.user` file should not be committed.

Also, you have changed the structure for the welcome.htm and common_devanagari.php files to use some sort of Word formatted html. We really do not want that. We want the files to be clean html. The .php file especially needs the php header structure retained.

Once you clean up the php and htm files, I don't think you will need the `welcome_files` folder or the `common_devanagari_files` folders as that appears to be part of Word's messy html generation which we do not want.

The HISTORY.md file doesn't show what has changed in the keyboard. You should update that to indicate what the changes are.

(I don't know if you are using a different file format, but all your files appear to be completely new. Maybe if you can update your branch before doing a commit I'll see your changes properly, but do the above things first.

mcdurdin:
> (I don't know if you are using a different file format, but all your files appear to be completely new. Maybe if you can update your branch before doing a commit I'll see your changes properly, but do the above things first.

Yes, this is happening because the fork is not up to date. The fork should be brought up to date, using the "Sync Fork" button:

image

Note that it is likely that merge conflicts will arise with the common_devanagari files, so you may find it easiest to move them out of the repo temporarily before attempting to sync the fork, and then re-adding them to the repo.

PR #2282, opened 2023-07-05
GFF Amharic & Gurage Package

dyacob:
This is a simple package migration to the Waldba font collection.


LornaSIL:
I *think* these changes must be made in the same PR as the gff_gurage font
changes are being made.

dyacob:
Ahhh!  Yes, that would be best and avoids the need for an interim update which this branch is.  I'll close this PR and update the other branch.  thanks!

keyman-server:

PR #2283, opened 2023-07-06
Keyboard files edited

Raji1205:
Edited the  keyboard files, please review.


keyman-server:

LornaSIL:
This is much better! Thank you.

Can you please restore the `welcome.htm` file and the `common_devangari.php` files to their original status? The format as they are now is not acceptable.

I checked your change of content, and the only change I see is under **Conjunct characters** you have a sentence " A simple conjunct is typed by pressing x between **any** consonant" where you changed "the" to "any" so you could change that if it's significant, but the file format for welcome.htm was clean html before and now it isn't. And the `php` file will not render render without a php header and all the Microsoft cruft. Please either restore the contents of the file to what it was before, or alternatively, you could give me permission to edit your files. Right now it won't let me.

I'll attach the files that "work" in case you lost the original format.
[common_devanagari.zip](https://github.com/keymanapp/keyboards/files/11969616/common_devanagari.zip)

LornaSIL (approved):
LGTM

Raji1205:
Thank you very much for helping on keyboard update process. God bless you.

LornaSIL:
You are very welcome!

PR #2289, opened 2023-07-12
GFF Geʾez v2.0

dyacob:
This update synchs the keyboard with a number of the revisions that have gone into other GFF keyboards recently. Other changes include:

**The mobile keyboard has been removed.** I believe the touch layout was likely too esoteric to be worth maintaining. It can be returned (and updated) if there is demand for it.

**The Abyssinica SIL Alternate 1 and 2 fonts are added.**  The fonts introduced with the `gff_ethiopic` keyboard have been included with the package.  Now located under the `shared/fonts/sil/abyssinica` folder.  If I had had better foresight I would have placed them here initially.  The next version of the `gff_ethiopic` keyboard will remove the local fonts and reference the shared copies.

**Syllable composition uses lowercase-only vowels.** Trying this experimentally and may migrate this change to the other GFF keyboards for Ethiopic script. Allowing for vowel composition to be case-independent has some advantageous, it's analogous to "training wheels" to help support new typists. In some cases for languages that use the "ዐ" (Aynu-A) series frequently, it introduces an inefficiency for experienced typists.  I would like to make this a keyboard option at a future time.


keyman-server:

dyacob:
It just occurred to me that maybe the `.keyboard_info` should have the following added?

```
    "gff-gez-powerpack-7": {
      "deprecates": true
    }
```

LornaSIL:
Let me deal with that in the PR #2291

PR #2290, opened 2023-07-12
GFF Harari v1.0.4

dyacob:
This revision makes a single typing change for a critical entry (a simple and embarrassing fix).

Five screenshots of the mobile layout used in the help pages were also updated to reflect the recent revisions.


keyman-server:

LornaSIL (approved):
LGTM

Issue #2259, opened 2023-06-19
[GFF] Deprecate some of these keyboards


LornaSIL:
Need to review which of these can be "deprecated" in the `keyboard_info` file.

https://keyman.com/keyboards/gff_awn_7 > https://keyman.com/keyboards/gff_awngi_xamtanga
https://keyman.com/keyboards/gff-awn-powerpack-7 > https://keyman.com/keyboards/gff_awngi_xamtanga
https://keyman.com/keyboards/gff-tir-et-powerpack-7 > https://keyman.com/keyboards/gff_tigrinya_ethiopia
https://keyman.com/keyboards/gff-sgw-powerpack-7 > https://keyman.com/keyboards/gff_gurage and https://keyman.com/keyboards/gff_gurage_legacy

I suspect we can deprecate the above, but not the below (yet):
https://keyman.com/keyboards/gff_bcq_7
https://keyman.com/keyboards/gff-bcq-powerpack-7
https://keyman.com/keyboards/gff-mym-powerpack-7
https://keyman.com/keyboards/gff_mym_7

What do you think @dyacob

dyacob:
@LornaSIL I think this deprecation list is perfect.  

I have a new `gff_ethiopic` that is nearly complete, perhaps ready in a few days or over the next weekend.  It is meant to replace `gff_ethiopic_7` and can also be the replacement for the `bcq` and `mym` keyboards (I don't have any plan to update these).

LornaSIL:
Some of the first ones had already been deprecated. I just hadn't keep up with them.
We'll wait on the gff_ethiopic one to deprecate the others.

dyacob:
With the deprecation, can these sections be removed from the respective `.keyboard_info` files?

```
  "related": {
    "gff_sgw_7": {
      "deprecates": true
    }
```

LornaSIL:
I'm not sure what you are asking. It is that line that actually does the
deprecation. Adding DEPRECATE.md is merely informational when someone looks
at the repo and thinks about updating that keyboard.

dyacob:
I was imagining that if the deprecated keyboard was no longer discoverable via the website (due to deprecation) then the `related` section might be referring to a keyboard (`gff_sgw_7`, etc) that was no longer around to be on the receiving end of a reference. In that case, with the referenced item no longer around, it might be ok to remove the `related` section.

LornaSIL:
Ah, ok. I don't think Marc wants keyboards actually deleted,  but you can
[...]

PR #2291, opened 2023-07-12
[ethi_dep] Deprecate remaining legacy gff keyboards

LornaSIL:
Relates to #2259


LornaSIL:
My changes will conflict with @dyacob PR so we'll get his through first, then I'll modify this one.

DavidLRowe (approved):

Issue #2303, opened 2023-07-19
chore: keyboards that target touch but are missing .js in packages


mcdurdin:
```
balochi_scientific.kps - warn KM0401A: Keyboard balochi_scientific targets touch devices but corresponding balochi_scientific.js file is not in the package.
gff_amh_7.kps - warn KM0401A: Keyboard gff_amh_7 targets touch devices but corresponding gff_amh_7.js file is not in the package.
gff_amh_powerpack_7.kps - warn KM0401A: Keyboard gff_amh_7 targets touch devices but corresponding gff_amh_7.js file is not in the package.
gff_gurage_and_amharic.kps - warn KM0401A: Keyboard gff_amharic targets touch devices but corresponding gff_amharic.js file is not in the package.
gff_amh_powerpack_7.kps - warn KM0401A: Keyboard gff_ethiopic_7 targets touch devices but corresponding gff_ethiopic_7.js file is not in the package.
gff_ethiopic_7.kps - warn KM0401A: Keyboard gff_ethiopic_7 targets touch devices but corresponding gff_ethiopic_7.js file is not in the package.
gff_gurage_and_amharic.kps - warn KM0401A: Keyboard gff_gurage targets touch devices but corresponding gff_gurage.js file is not in the package.
lao_phonetic.kps - warn KM0401A: Keyboard lao_phonetic targets touch devices but corresponding lao_phonetic.js file is not in the package.
```

DavidLRowe:
#2308 fixes all but:
```
gff_amh_7.kps
gff_amh_powerpack_7.kps
gff_ethiopic_7.kps
```
which are deprecated.

mcdurdin:
> How do you want that handled? Or maybe you should not be testing deprecated keyboards?

So currently the keyboards are listed as supporting mobile on keyman.com, although they don't appear in default searches as they are deprecated.

In order to keep things neat and tidy, it'd be good to tidy up the discrepancy. It's not going to impact end users much but it helps to reduce noise in the builds and avoid confusion in the future.

Either we remove the 'any' from the &targets store, or add the .js to the .kps. If we remove the 'any' from the &targets store, we don't need to bump the version -- it will mean that the keyboards won't be listed as supporting mobile any more because the .keyboard_info is always updated with every build -- at least for the database, if not for downloads.keyman.com.

But if we add .js to the .kps, we need to bump the version, or the updated keyboard won't be published, and the data on keyman.com will still not line up with the downloads available.

[...]

PR #2308, opened 2023-07-19
[2303_missing_js] Update some of the keyboards in this issue

LornaSIL:
Partially addresses #2303 

@mcdurdin I did not address:
```
gff_amh_7.kps
gff_amh_powerpack_7.kps
gff_ethiopic_7.kps
```
because those are deprecated. 
How do you want that handled? Or maybe you should not be testing deprecated keyboards?

### DIFF
--- a/release/b/balochi_scientific/HISTORY.md
+++ b/release/b/balochi_scientific/HISTORY.md
@@ -1,6 +1,10 @@
 Balochi Scientific Keyboard Change History
 ===============================
 
+1.0.5 (19 Jul 2023)
+------------------
+* Add mobile js to package
+
 1.0.4 (1 Sep 2022)
 ------------------
 * Use bal-Latn instead of bcc-Latn
--- a/release/b/balochi_scientific/LICENSE.md
+++ b/release/b/balochi_scientific/LICENSE.md
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2017-2022 SIL International
+Copyright (c) 2017-2023 SIL International
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
--- a/release/b/balochi_scientific/balochi_scientific.kpj
+++ b/release/b/balochi_scientific/balochi_scientific.kpj
@@ -16,7 +16,7 @@
       <FileType>.kmn</FileType>
       <Details>
         <Name>Balochi Scientific</Name>
-        <Copyright>Â© SIL International</Copyright>
+        <Copyright>© SIL International</Copyright>
         <Message>The Balochi Scientific keyboard is distributed under The MIT License (MIT).</Message>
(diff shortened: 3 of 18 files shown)


dyacob:
Did Keyman 7 produce JS files?  I think it only created desktop keyboards, no web or mobile.  So they may not really be "missing" in a sense.  The packages were just not maintained to target platforms that came later -which is handled by the successor projects that deprecate these.

mcdurdin:
> Did Keyman 7 produce JS files? I think it only created desktop keyboards, no web or mobile. So they may not really be "missing" in a sense. The packages were just not maintained to target platforms that came later -which in turn is handled by the successor projects that deprecate these.

The reason these keyboards were flagged is because their `&TARGETS` system store included mobile, e.g.:

https://github.com/keymanapp/keyboards/blob/ca4b1ec545b3f88dd18b84cefe341b3a92f6717f/release/gff/gff_amh_7/source/gff_amh_7.kmn#L100

mcdurdin:
I note that gff_gurage_and_amharic has an override for platformsupport in its keyboard_info:

https://github.com/keymanapp/keyboards/blob/59c44e8384b736f5f8e51a0af39f91bc677f0cce/release/packages/gff_gurage_and_amharic/gff_gurage_and_amharic.keyboard_info#L28-L32

dyacob:
This was the original intent, that it be desktop only.  GFF Gurage _does_ have a touch layout, but it really only works in the Keyman App at this time since the Android and Apple system fonts do not yet support the Gurage letters added to Unicode 14.  The KM app can load its own font.  Android 14 *should* finally include a Noto Ethiopic font update with the 28 extra letters. Apple is still dragging its feet.

So the override would eventually be removed. Can `platformSupport` include a minimal version of Android or iOS?  A way to specific `Android >= 14` ,  `iOS >= 18.3` would be helpful when the system font updates arrive.

DavidLRowe:
@dyacob @mcdurdin Does any of this discussion need to be moved to a new issue so that it can be acted upon?

mcdurdin:
> @dyacob @mcdurdin Does any of this discussion need to be moved to a new issue so that it can be acted upon?

I think it's all good now. There's a follow-up, related, issue regarding .keyboard_info files and inconsistent platform support data that may illuminate a little in #2311.

mcdurdin:
> Can `platformSupport` include a minimal version of Android or iOS? A way to specify `Android >= 14` , `iOS >= 18.3` would be helpful when the system font updates arrive.
> 
> Checking the [.keyboard_info 1.0 Specification](https://help.keyman.com/DEVELOPER/cloud/keyboard_info/1.0/) there doesn't appear to be support for this, but it would be good to add to the `KeyboardPlatformInfo` object at some point.

This would be a "nice to have" which I will track in work around the package metadata formats in keymanapp/keyman#9351. No promises at this point! I will note that it's actually really hard to maintain this level of version metadata.

PR #2310, opened 2023-07-21
update packages of deprecated keyboards

DavidLRowe:
@mcdurdin this PR includes gff_amh_7 and gff_ethiopic_7, but I wasn't sure what to do about gff_amh_powerpack_7.


DavidLRowe:
Since gff_amh_7 had a touch layout file, I added the .js to the package. 

Since gff_ethiopic_7 had no touch information, I changed the &Targets value. I bumped the version number, though not strictly necessary.

Not sure how to handle the package (which includes both of these, I guess).

mcdurdin:
> Not sure how to handle the package (which includes both of these, I guess).

Good question! I am not sure either. I do wonder if, given both keyboards are deprecated, we should just remove the 'any' from targets and be done with them? They never worked on mobile apps anyway given the .js was missing, and on keymanweb.com we now have the new Amharic keyboards which are much nicer.

DavidLRowe:
So you'd rather set the target to 'desktop' only in the .kmn file?

mcdurdin:
> So you'd rather set the target to 'desktop' only in the .kmn file?

I think so, yes.

DavidLRowe:
gff_amh_7 changed to desktop only. The powerpack seems to have picked up those changes.

Hopefully, together with #2308 , this PR...

Fixes #2303

LornaSIL (approved):
LGTM

PR #2312, opened 2023-07-27
[fv_hulquminum_combine] keyboard with combining diacritics

xyzhelen:
new mobile keyboard to accommodate the Hul̓q̓umin̓um̓ orthography found in the 1996 book Hul̓q̓umin̓um̓ Phonics @caforbes


keyman-server:

DavidLRowe:
The copyright date in LICENSE.md should change to agree with what's in the .kmn file, which has 2015-2023.

There's a fv_hulquminum_combine.keyman-touch-layout file that has been added to the source folder under the keyboard folder for fv_hulquminum. This file is out of place and should be removed.

There's another spurious fv_hulquminum_combine.keyman-touch-layout file in the fv_hulquminum_combine folder. This file is out of place and should be removed.

The fv_hulquminum_combine.keyman-touch-layout file in the fv_hulquminum_combine/source folder is in the proper place.

The fv_hulquminum_combine.keyboard_info file includes a deprecation of fv_hulquminum_kmw. I'm not finding that keyboard, so I'm wondering about that deprecation entry (and the one in the fv_hulquminum keyboard), but will need to wait and ask until next week to ask Lorna when she's back. But meanwhile you can address the issues I've mentioned. Thanks!

LornaSIL:
Both keyboards don't need to deprecate the same keyboard. I would think this keyboard doesn't need the deprecation statement. 

I'm a little concerned about the keyboard name being the same as the name for `fv_hulquminum` as well as the description being the same. I would think it would be helpful to have a distinctive name and description if the keyboards are different.

caforbes:
Is this finished or are there further issues we have to address?

LornaSIL:
The keyboard name for both `fv_hulquminum_combine` and `fv_hulquminum` will both be **Hul’q’umi’num’**. 

Don't they need to be distinguished?

xyzhelen:
The name for the fv_hulquminum_combine keyboard should be Hul̓q̓um̓inum̓ and the name for the fv_hulquminum keyboard is Hul'q'umi'num'. One version has combining diacritics in the name and the other one doesn't. Is that enough? I could also add something like "(with combining diacritics)" to the name of fv_hulquminum_combine

LornaSIL (approved):
Ok, if that's a sufficient difference that is fine. LGTM.

PR #2317, opened 2023-07-31
[el_dari_clra] Update to release and deprecate dari_clra

LornaSIL:


DavidLRowe:
It doesn't seem that the `dari_clra.kmx` file is needed.

Is it worth changing the .php file to include the Ctrl+Alt level? (IIUC `<div id='osk'>` will only output default and shift layers.)

DavidLRowe (approved):

PR #2321, opened 2023-08-01
[fv_allkeys] Remove spurious allkeys store - nonfunctional change no …

### LINKED ISSUE #1827: [fv_*] Add NCAPS modifier to some rules to avoid inconsistent matches

The upcoming kmcmp.exe release will generate warnings about inconsistent matches needing NCAPS (keymanapp/keyman#6347)

```
```

Many of the FV keyboards also have this extraneous allkeys store which was a workaround for a very old Keyman bug and can now be removed
```
store(allkeys) 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890-=[]\;,./`~!@#$%^&*()_+{}|:"<>?' "'"
+ any(allkeys) > index(allkeys, 1)
```

This is a container issue for applicable FV keyboards. I'll associate PRs to this issue as I find the affected keyboards.

Tagging @madebybridget

LornaSIL:
…version bump

Fixes #1827

### DIFF
--- a/release/fv/fv_gitsenimx/source/fv_gitsenimx.kmn
+++ b/release/fv/fv_gitsenimx/source/fv_gitsenimx.kmn
@@ -114,7 +114,4 @@ nomatch > use(x)
 
 group(x) using keys
 
-store(allkeys) 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890-=[]\;,./`~!@#$%^&*()_+{}|:"<>?' "'"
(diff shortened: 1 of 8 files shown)


DavidLRowe (approved):

DavidLRowe:
@madebybridget @caforbes These changes to the keyboard source (.kmn) files don't change the function of the keyboards. They only remove a workaround for a Keyman bug that was fixed long ago. But if you work on any of these keyboards, you should make sure you have the latest version. There was no version number increment, so the keyboards will not be offered as an update.

PR #2327, opened 2023-08-03
Add (deprecated) note to PHP help files

DavidLRowe:


DavidLRowe:
All but two files were modified by a script which added "(deprecated)" to $pagename and also to $pagetitle (or checked that $pagetitle was set to $pagename). Two files didn't fit the pattern exactly and were changed by hand:
- galaxiegreekandhebrew.php (since the name used "Keyboards" instead of "Keyboard")
- naskeysdesktop.php (since the name had the extra word "Package")

LornaSIL (approved):
LGTM

PR #2332, opened 2023-08-07
[hcesar] Add files via upload

Tiagoink:


keyman-server:

LornaSIL:
Unfortunately I don't feel we can accept this pull request since it's removing some significant information from the keyboard such as changing the target from desktop to just Windows, removing the on-screen keyboard, adding inconsistent rules for RALT and ALT (we cannot have both). Also, since you are trying to add RALT keys we will need to have the on-screen keyboard updated as  well as the documentation. 

@sewhite would you want to see if you like his changes to the hcesar keyboard? You created the experimental one that we already have. If the changes @Tiagoink wants to make look good, we can work with Tiagoink to make the changes fit our process. If the changes wouldn't be acceptable then Tiagoink could submit a new hcesar keyboard under a different keyboard name.

LornaSIL:
@sewhite or @Tiagoink I'll need the latest version that you are both happy with. The whole package has to be updated, not just the .kmn file. I can do that if I get the correct .kmn.

LornaSIL:
Fixed by #2341.

PR #2333, opened 2023-08-07
Runeboard Version 2.0

Runeboard:
I'm aware that the 4 png files for the tablet layout are missing but there appears to be no means by which to export them from Keyman Developer. If there is, please let me know. Otherwise, how can I make them? Thanks.


keyman-server:

LornaSIL:
I'm a bit confused about this PR. Could the `Runeboard` keyboard PR be split separate from the `elfdalian` keyboard?

Is `elfdalian` actually a keyboard or a lexical model? Lexical models should be submitted in this repo: https://github.com/keymanapp/lexical-models

The graphics used in `welcome.htm` are named `LayoutU_.png`, `LayoutU_CA.png`, `LayoutU_S.png`, `LayoutU_SCA.png`. You cannot change those names without also changing `welcome.htm`. It would be easier if you retain the existing names so the welcome file doesn't need updating.

The same would be true for the naming of the graphics for the tablet layout. Those were probably just created as screenshots when testing the keyboard in a browser. If you think of a better system, that's fine, but no, we cannot export a tablet graphic from Keyman Developer at this time.

Runeboard:
Sorry for the confusion. I meant to split them but I'm still learning how everything works around here. I'll close this request and see if I can work out how to start 2 separate requests.

Thanks for the tip regarding the image file names. I had just used what Keyman produced but I've now updated them as you suggested.

Thank you too for the instruction on how to create graphics for the tablet layout.

PR #2335, opened 2023-08-09
Created a folder for Keyboard "العربية_بغير_نقاط" under release

AlMunsajimAlMar:
This keyboard is for arabic without dots, as it was originally used in the past.


LornaSIL:
Thanks for your submission. I think the folder name and filename must be
ascii with no spaces. You can use an `_` for spaces. Then, whatever name
you give it, such as `arabic_nodots` would need to go into the `a` folder
if that's what the filename starts with. That structure helps keep things
tidy and that's how our build system works.
You may name the keyboard using Arabic characters, just not the file/folder
names.

keyman-server:

LornaSIL:
I'm sorry I haven't had a chance to do a thorough review. I do see a couple issues you could work on. You have two `.kmn` files.
The one in the source folder doesn't have any rules. You should delete that file. Then copy the `.kmn` file that is in the root of `arabic_wo_dots` to `arabic_wo_dots/source` and rename it `arabic_wo_dots.kmn`. You should then reimport the onscreen keyboard as that is currently blank.

You may then need to reassociate the .kmn file with the .kpj file. I'm not sure. Make sure you have a language associating for the keyboard in the .kps and in the `keyboard_info` file.

LornaSIL:
There are a number of issues with the keyboard. I'll just list some that need resolving before I can check it.
- Make sure to open the `arabic_wo_dots.kpj` file in Keyman Developer. 
- When you go to the **Keyboards** tab you should only have `arabic_wo_dots.kmn` (from the source folder) associated with the keyboard. Currently this file doesn't have keyboard rules in it though. So, you'll need to add the rules from the other .kmn file to this one.
    - Your keyboard lists all the targets: onscreen keyboard (which is currently blank and needs to be fixed or removed as a target), and all the mobile/touch targets. The Touch layout is currently for Latin, so that either needs to have the proper layout, or just remove it as a target. It might be easiest for you to initially just have **Desktop** as a target so you aren't having to learn too much at one time. This would be the best rule and remove the onscreen keyboard and touch layout as targets.
        -  `store(&TARGETS) 'desktop'`
    - When I try to compile your keyboard there are some problems with the rules. The compiler will tell you which rules are problems. In particular, this rule at or around line 66 is a problem. The output isn't enclosed in quotes:
        - `+ [SHIFT RALT K_R] > '࣫R`
    -  Once I fixed that rule then it displays lots of rules that will never "fire" because there is another rule before it for the same keystrokes. So, it will be important for you to test the keyboard by compiling it and fixing all the issues that are found.'
- After fixing the .kmn so that it compiles, you'll want to go to the **Packaging** tab and fix up the .kps file.
  - Within the .kps, go to the **Keyboards** tab and click on **Add** for the BCP 47 tag. I guess that should be "ar" for Arabic. Without that tag the keyboard won't know which language to associate itself with on the operating system.
  - Click on the **Build** tab and see if you can compile the package. If not, follow the messages to see what the issues are with compiling. Those messages will help you make a better keyboard.
- It would be nice if you could improve the documentation in the readme.htm and welcome.htm files. We like a short description in the readme.htm file of what the keyboard does and what language it is used for. This helps the user to know whether they even want to use the keyboard. The welcome.htm file should have a description of the layout for the keyboard. If you are not able to improve the documentation at this point, then we recommend you add the keyboard to the `experimental` section of the repo rather than the `release` section. We have less rigorous requirements for keyboards submitted to `experimental`. However, you still need the keyboard to compile before we can accept it in `experimental`.

PR #2336, opened 2023-08-11
chore: preprocess .keyboard_info, whitespace and encoding consistency 🎺

mcdurdin:
Relates to keymanapp/keyman#9351.

This commit makes no material changes to the content of the files. I just loaded the .keyboard_info files into a node JSON parser and saved them out with consistent formatting. This is groundwork for eventually being able to remove the .keyboard_info files altogether, and is intended to simplify the process of validating subsequent PRs which *will* have changes to the content of the files.

Use 'hide whitespace' when reviewing to save a lot of effort.


mcdurdin:
Note @LornaSIL as you've merged #2237 into this you will find it easier to review if you view only the first two commits 😀

LornaSIL (approved):
A few curious questions.
1. What is the `en-QP` language tag? Never heard of QP and it's not a valid country code. A website indicates it could be regional for North Africa?! Should these legacy ids be checked?
2. All these platformSupport statements for Windows and macOS, should we add "Linux" or doesn't it matter at this point?
3. In the future we could remove the keyboard id and keyboard package name as those are redundant.

DavidLRowe:
QP (private use region code) was the solution we came up with for the "postmodern" English keyboards / lexical models. That orthography uses a mix of Latn and Cyrl characters to write English and we didn't want it picked up as "en".

LornaSIL:
>QP (private use region code) was the solution we came up with for the "postmodern" English keyboards / lexical models. That orthography uses a mix of Latn and Cyrl characters to write English and we didn't want it picked up as "en".

So, does that need adding to langtags?

DavidLRowe:
QP is listed in the special _conformance tag for regions, which means (IIUC) it doesn't appear / isn't used in langtags. Asking the  LDML API for en-QP returns en.xml. Thus, at the moment, langtags is ignorant of this use of QP, but I think that's fine, since QP is a private use region tag.

mcdurdin:
> * All these platformSupport statements for Windows and macOS, should we add "Linux" or doesn't it matter at this point?

If these are for legacy keyboards, we could add it I think. 

> * In the future we could remove the keyboard id and keyboard package name as those are redundant.

The purpose of the id field is so that the JSON can be understood if it doesn't have a filename, e.g. in response to a keyboard search query. Soon, this file will be completely automatically generated (for experimental/release keyboards), and then these two fields won't matter so much -- and I'll have another big PR with the changes patched in to the .kps files automatically (without bumping their version numbers). 

For legacy keyboards, we'll need to keep the file because we won't want to modify the .kmp files to add the extra metadata to them.

The early changes in these 2 PRs didn't require changes to the compiler, but the upcoming changes will only work with the v17 compiler.

PR #2339, opened 2023-08-11
[font_lateef] Add new version of Lateef

LornaSIL:


DavidLRowe (approved):

mcdurdin:
Note: this will not be deployed to existing keyboards, is that okay?

LornaSIL:
Yes, that's fine.

Issue #1703, opened 2021-12-08
[korean_rr] Space does not work at all including for combing jamos 


rc-swag:
This issue first raised on community forum here - [Korean RR Keyboard space not working](https://community.software.sil.org/t/korean-rr-keyboard-space-not-working/5663)

You can not type a space. Or when attempting to combine jamos nothing happens.
User report was for Windows version 11. Did not specify Keyman version.

Reproduced on Win 11 with Keyman 14 and 13.
Reproduced on Win 10 with version 15 alpha with common core turned off.

MakaraSok:
I've tested the keyboard with Keyman 14 Stable (14.0.283) on Windows 10 Pro 21H1. The space does not seem to do the job. In Notepad, it doesn't combine the character. In LibreOffice, it doesn't separate the clusters being typed. Below is the output of "han guk" on the two apps.

Keyman 14 Stable installed from its .kmp available in the keyboard library
<img width="139" alt="image" src="https://user-images.githubusercontent.com/28331388/145136262-7d565a94-e4c2-4320-8104-f801c49cd8eb.png">

It looks OK on KMW though, both local build and online.

Local build:
<img width="139" alt="image" src="https://user-images.githubusercontent.com/28331388/145142257-e5cb5649-7a00-4d9a-aff4-8904f83316dd.png">

Online KMW:
<img width="139" alt="image" src="https://user-images.githubusercontent.com/28331388/145142386-5310c566-f072-4877-8dba-c0344e4dd3cc.png">

rc-swag:
From my investigations, the JavaScript version of jamo_combine operates with this condition -  "if matched length is less than two, it will simply output a space".  The jamo_combine.dll does not insert a space. This is the core issue. 
jamo_combine.dll is used by 3 other keyboards so we need to be careful changing its behaviour. 

Other issues found with this keyboard
The 64-bit version of the dll needs to be built.
The following dlls do not exist.
show_ime.dll
hide_ime.dll
select_hanja.dl
scroll_hanja.dll

DavidLRowe:
@rc-swag What action (if any) do we need to take on this issue?

rc-swag:
> @rc-swag What action (if any) do we need to take on this issue?

[...]

PR #2345, opened 2023-08-18
[korean_rr] remove non-functional Windows target

DavidLRowe:


DavidLRowe:
Addresses the immediate issue raised on #1703 by removing Windows as a target.

Does not address Keyman's lack of IMX support.

mcdurdin:
> Does not address Keyman's lack of IMX support.

Keyman for Windows has IMX support. So I think the big issue here is that I haven't had space to look into this problem, because there's a bunch of history here around this keyboard, and I don't remember it. I understand the Windows target is not working but would like to try and fix that rather than simply disable it and then forget again :)

DavidLRowe:
>    Does not address Keyman's lack of IMX support.

> Keyman for Windows has IMX support. So I think the big issue here is that I haven't had space to look into this problem, because there's a bunch of history here around this keyboard, and I don't remember it. I understand the Windows target is not working but would like to try and fix that rather than simply disable it and then forget again :)

Sorry for the sloppy comment. Perhaps "Does not address this keyboard's IMX issues on Windows."? (Instead this PR removes Windows as a target.)

Issue #1703 will remain open.

DavidLRowe:
@rc-swag 
When I changed the target to web only, the .kmx file wasn't generated, which caused an error.
I removed the .kmx from the .kps list, and got an error saying it needed either .kmx or .js.
I added the .js and it built.
But the question is whether this package has everything needed to work on the web. Is it possible to build a web-only keyboard?

LornaSIL:
@darcywong00 I don't have enough knowledge to know if this should be approved or not.

release/k/korean_rr/source/korean_rr.kps
```diff
@@ -9,7 +9,6 @@
     <ReadMeFile>readme.htm</ReadMeFile>
     <MSIFileName></MSIFileName>
     <MSIOptions></MSIOptions>
-    <FollowKeyboardVersion/>
```
darcywong00, line 12: @DavidLRowe - is it possible to keep this checked?

darcywong00:
> @darcywong00 I don't have enough knowledge to know if this should be approved or not.

I think this one is fine to approve. @DavidLRowe is keeping #1703 open and I also created a related issue keymanapp/keyman#9516 for refactoring Keyman for Windows to support korean_rr.

LornaSIL:
It won't let me check it, so I think @rc-swag needs to approve it now.

release/k/korean_rr/source/korean_rr.kps
rc-swag: @LornaSIL  @DavidLRowe  I think we need to see if we can keep this checked then I am ok to approve

rc-swag (approved):
lgtm

Issue #2342, opened 2023-08-17
.keyboard_info cleanup


darcywong00:
I was looking up keyboard examples in the .keyboard_info files (while reviewing keymanapp/keyman#9459) and found two files to cleanup:

.keyboard_info files should be in root of the keyboard instead of source/
* https://github.com/keymanapp/keyboards/blob/master/release/n/northern_sierra_miwok/source/northern_sierra_miwok.keyboard_info

* https://github.com/keymanapp/keyboards/blob/master/release/fv/fv_southern_tutchone/source/fv_southern_tutchone.keyboard_info

fv_southern_tutchone may be tricky because the fv_southern_tutchone.keyboard_info is actually the touch layout used by the .kmn
https://github.com/keymanapp/keyboards/blob/595c45588b3a3a6df95597024b0c55597f5d5a62/release/fv/fv_southern_tutchone/source/fv_southern_tutchone.kmn#L26

Note: The existing fv_southern_tutchone.keyman-touch-layout has "shift" as the first layer (possibly relates to keymanapp/keyman#8718?)

mcdurdin:
Suggested solutions:

1. northern_sierra_miwok/source/northern_sierra_miwok.keyboard_info: delete, because there is one in parent folder as well.

2. fv_southern_tutchone:
  a. delete source/fv_southern_tutchone.keyman-touch-layout
  b. rename source/fv_southern_tutchone.keyboard_info to source/fv_southern_tutchone.keyman-touch-layout
  c. fixup `&layoutfile` reference in fv_southern_tutchone.kmn
  d. ask FV team to verify the touch layout is correct.

PR #2348, opened 2023-08-18
[2342] Fix issues related to Issue 2342]

LornaSIL:
@madebybridget can you do a quick check that I haven't broken this keyboard? There was some inconsistent filenaming. The content should be exactly the same, so I didn't need to do a version bump.


LornaSIL:
Fixes #2342

darcywong00:
Also tagging @caforbes to check the touch layout for fv_southern_tutchone is unchanged.

darcywong00:
Ah, I see fv_southern_tutchone has an open PR for the touch layout in #2350 .
Maybe we fix the file re-name there.

caforbes:
Just checking to see if the issue is fixed with #2350 or if you need anything further from us.

LornaSIL:
Fixed by #2351 and #2350

LornaSIL:
I think we're good.

PR #2349, opened 2023-08-19
[arabic_wo_dots] Add files via upload

AlMunsajimAlMar:


keyman-server:

release/a/arabic_wo_dots/source/arabic_wo_dots.kmn
```diff
@@ -0,0 +1,148 @@
+﻿c with name "العربية بغير نقاط"
+store(&NAME) 'العربية بغير نقاط'
+store(&COPYRIGHT) '© عبيدالرحمن'
+store(&VERSION) '10.0'
+store(&KEYBOARDVERSION) '1.0'
+store(&TARGETS) 'windows macosx linux web'
```
darcywong00, line 6: Since you listed KeymanWeb as a target, I think you should also add
```
store(&KMW_RTL) '1'
```
so the web platform will treat the keyboard as right-to-left

Reference
https://help.keyman.com/developer/language/reference/kmw_rtl

LornaSIL:
Unfortunately, this keyboard is still not ready. I believe it will need to go into the `experimental` folder, rather than `release`.
Your file structure should be:
```
experimental\a\arabic_wo_dots\arabic_wo_dots.keyboard_info
experimental\a\arabic_wo_dots\arabic_wo_dots.kpj
experimental\a\arabic_wo_dots\HISTORY.md
experimental\a\arabic_wo_dots\LICENSE.md
experimental\a\arabic_wo_dots\README.md
experimental\a\arabic_wo_dots\source
experimental\a\arabic_wo_dots\source
experimental\a\arabic_wo_dots\source
experimental\a\arabic_wo_dots\source\arabic_wo_dots.ico
experimental\a\arabic_wo_dots\source\arabic_wo_dots.kmn
experimental\a\arabic_wo_dots\source\arabic_wo_dots.kvks
experimental\a\arabic_wo_dots\source\arabic_wo_dots.kps
experimental\a\arabic_wo_dots\source\readme.htm
experimental\a\arabic_wo_dots\source\welcome.htm.
```
Currently the .kpj is in the source folder and it should be in the root of the `arabic_wo_dots` folder.
You should also delete the .kpj.user file that does not need committing.

As @darcywong00 mentioned, you need to add this rule to the .kmn: `store(&KMW_RTL) '1'`
You should actually add both these lines to your .kmn:
```
store(&VISUALKEYBOARD) 'arabic_wo_dots.kvks'
store(&KMW_RTL) '1'
```
Then, you would need to go to the **Onscreen** tab and select "fill from layout" so that the onscreen keyboard will have characters to display. It will be based on the layout you created.

Also, your .kpj file still references an old .kmn  and that should be removed.

The readme.htm and welcome.htm don't give much information about how to use the keyboard, but if you put it in `experimental` we can allow it to not be well documented. However, it's very helpful to provide information so people know how to use the keyboard.

The keyboard has rules on the RALT layer for a lot of characters in the presentation form blocks (like initial, medial, final, isolate forms of characters). Those are really not recommended for use anymore. It would be much better for your keyboard to support the characters in the "Arabic", "Arabic Supplemental", etc blocks. Fonts always support those blocks better than the presentation forms. (I'm seeing a lot of tofu on that block and in a few other places).

I still don't see that you've set the language code in the .kps file. You need to set that in the .kps / Keyboards tab. Click on "Add" and add your language tag. Otherwise, the keyboard won't work very well.

If you do the things I've suggested the keyboard will at least compile. I would be willing to accept it in the experimental folder, but not in the release folder.

LornaSIL:
@AlMunsajimAlMar since this PR hasn't been updated, I'm going to close this PR. Feel free to fix up the files and when they are ready resubmit them to `experimental`.

PR #2352, opened 2023-08-28
[Lateef] Add new version of font

LornaSIL:
Added new version, but no existing keyboards include the new characters (that I'm aware of).


darcywong00 (approved):
lgtm

Issue #2353, opened 2023-08-31
[old_hungarian]Missing decimal comma from Old Hungarian layouts


rovasiras:
When decimal numbers are used, it was the best choice to add decimal comma.

rovasiras:
@LornaSIL I try to resolve this problem, but I don't know that I could prepare it on desktop layouts. The xkeyboard-config has a merge request already.

LornaSIL:
> I try to resolve this problem, but I don't know that I could prepare it on desktop layouts. The xkeyboard-config has a merge request already.

Can you provide a link to the xkeyboard-config request?

rovasiras:
> > I try to resolve this problem, but I don't know that I could prepare it on desktop layouts. The xkeyboard-config has a merge request already.
> 
> Can you provide a link to the xkeyboard-config request?

https://gitlab.freedesktop.org/xkeyboard-config/xkeyboard-config/-/merge_requests/618/diffs

rovasiras:
https://gitlab.freedesktop.org/xkeyboard-config/xkeyboard-config/-/merge_requests/620/diffs

PR #2356, opened 2023-09-03
[old_hungarian] Added Kende with ligatures for "new ages" and added decimal comma

rovasiras:
Added Kende with ligatures for "new ages" and added decimal comma. Mainly resolves https://github.com/keymanapp/keyboards/issues/2353 issue


keyman-server:

rovasiras:
@LornaSIL Please merge this pull request.

rovasiras:
@LornaSIL Something is wrong?

LornaSIL:
@rovasiras I just need to have time to check it. It may take a few days.

LornaSIL:
I think I would prefer that you remove the `.eot` and `.svg` files from the `fonts\hung\kende\web` folder. Possibly remove the `.woff` file as well. The `.woff2` standard is the defacto standard now and there's no reason to include the other formats. (The `keyboard_info` file only references the `Kende.woff2` and `Kende.ttf` font files so there's really no need for any of the other formats.

Otherwise, I think it all looks good.

LornaSIL:
You don't need to close the issue, just remove those files.

rovasiras:
> You don't need to close the issue, just remove those files.

I can not delete these fonts. Could you do it by hand?

LornaSIL:
I was able to delete the svg one, but it won't let me delete the others. Could you try to go to the commit tab, click on `Kende.eot` and then on the 3 dots, will it let you delete? The Delete button is grayed out for me.
image
Hopefully you can do the same for `Kende.woff`

rovasiras:
> I was able to delete the svg one, but it won't let me delete the others. Could you try to go to the commit tab, click on `Kende.eot` and then on the 3 dots, will it let you delete? The Delete button is grayed out for me. image Hopefully you can do the same for `Kende.woff`

The "delete file" appears in red color. When I try delete, appears a text that it was deleted, but that is not true, fontfile not erased from the repo. I think, that cloning with this repo, and using git commands the problem could be prepared.

rovasiras:
@LornaSIL Sorry, I am in a hospital, and did not bring my Laptop to it.

LornaSIL (approved):
LGTM.

PR #2362, opened 2023-09-13
Add Old English keyboard

Runeboard:
Hopefully all is in proper working order.


keyman-server:

DavidLRowe:
There are some extraneous files included in this pull request that need to be removed:
- remove old_english.kpj.user
- remove build folder and contents

(It may seem counter-intuitive to delete the build folder and contents, but the keyboards repository contains only the source files and the compiled forms are built by the system, so we don't need or want any pre-built items.)

Adjustment to the list of files in the package:
- remove old_english.ico
- add layoutU_.png
- add layoutU_S.png
- add layoutU_RCRA.png
- add layoutU_SRCRA.png

The .ico file is compiled into the the .kmx file, so it doesn't need to appear separately in the package. The .png files are referenced by welcome.htm, so they need to appear in the package.

Additionally, the path for the font file is currently `..\..\..\..\..\Downloads\Catrinity.otf` but it needs to be `fonts\Catrinity.otf`. That is, it needs to refer to the copy of the font that is in the fonts folder, not to a copy elsewhere on your computer. You may find it easiest to remove the font from the file list, then add it again by navigating to the correct copy.

DavidLRowe:
For the desktop layout, rather than using `RCTRL RALT` and `RCTRL RALT SHIFT`, you may want to make it easier on the typist by using simply `RALT` and `RALT SHIFT`. 

In the .php help file, the reference to 
`<div id='osk' data-states='default shift ctrl-alt shift-ctrl-alt'>`
needs to change to 
`<div id='osk' data-states='default shift rightctrl-rightalt rightctrl-rightalt-shift'>`
in order to match the modifiers that you're using. 

If you decide to change to using `RALT` and `RALT SHIFT`, then you'll want
`<div id='osk' data-states='default shift rightalt rightalt-shift'>`
in the .php file.

For the touch layouts, you currently have both `tablet` and `phone` layouts. I haven't examined them in detail, but they seem to be the same. Our recommendation is to delete one of these layouts rather than trying to keep the two layouts synchronized. Keyman will use a phone layout on a tablet if there is no tablet layout available, and vice versa. In this case it seems as though you could remove the tablet layout and just maintain the phone layout.

In the .php help file, the reference to
`<div id='osk-tablet' data-states='default shift ctrl-alt'>`
needs to change. The actual layers in the touch layout are `default shift numeric` and the reference in the .php file needs to match (that is, replace `ctrl-alt` with `numeric`). In addition, if you take the advice above and delete the tablet layout, you need to change `osk-tablet` to `osk-phone`.

Let me know if anything is unclear.

Runeboard:
Thanks David. I see your logic about using the right Alt key being easier than using right Alt & right Ctrl but I was concerned that some people might encounter conflicts if I only used the Alt key rather than both.

I removed the phone layout as I hadn't realised that it was there. I only meant to create the tablet layout and it seems the phone layout hadn't pre-loaded.

DavidLRowe:
Changes look good, however the .keyboard_info file has a problem with quotation marks in the description field. I think you want to use `\"` or `'` in the string.

DavidLRowe:
You can use https://jsonformatter.curiousconcept.com/ to check the syntax of the .keyboard_info file. (It won't catch all errors, but can help find some syntax errors.)

Runeboard:
Thanks. I resolved the problem simply by removing the quote marks. Those who know what it's about will understand without them anyway. Are any other amendments needed?

DavidLRowe:
Now that the quote marks are fixed, another error is apparent. The description field is misplaced, so it ends up being part of the languages field. (Here's a case where the syntax is valid, but the resulting file doesn't match the needed structure.)

If I've counted my curly braces correctly, you should replace:
```
            ]
         }
      },
      "description":"The ... it."
   }
}
```
with:
```
            ]
         }
      }
   },
   "description":"The ... it."
}
```
at the end of the file. (NB: I've shortened the description field here to show the structure. You would need to keep your existing description.)

Runeboard:
Thanks. Think that's fixed now. Decided to take your advice to use single quote marks too.

DavidLRowe:
I think there's a comma missing or misplaced. Will look at it tomorrow, unless you get to it first!

Runeboard:
I'm not sure why the comma matters but I've fixed it to conform with what you showed before.

DavidLRowe (approved):

PR #2364, opened 2023-09-18
Elfdalian

Runeboard:
This was given me to upload a while ago. While I'm no expert at this myself, I've made an effort to substantially improve the file structure. Please let me know if anything more still needs to be done. Thanks.


keyman-server:

LornaSIL:
We do need a copyright holder which would need adding to the LICENSE.md, .kmn, and .kps. Copyright should be either 2023 or 2021-2023.

The version number is currently `1.05`. Can you change that to `1.0.5`? Using two digits sometimes confuses the versioning. 

It would be useful to remove the copyright and version statements from the welcome.htm and readme.htm. That way they won't need updating in the future.

Remove `elfdalian.ico` from the root folder and leave the one in `source`

I didn't test it, but it does seem to build so if you can fix these issues it should be okay.

Runeboard:
Thanks Lorna. Is everything in order now?

LornaSIL:
>Thanks Lorna. Is everything in order now?

I think so. I'm kicking off a build and we'll see if I missed anything. Thanks!

LornaSIL (approved):
Minimal check, LGTM.

PR #2367, opened 2023-09-19
[old_hungarian] Update mobile layout longpress buttons

### LINKED ISSUE #2361: [old_hungarian][mobile devices]mistakes with punctuation signs
@LornaSIL There are problems with long press keys. In default layer the "main" key is decimal comma instead of reverse comma, the floating keys must to be the following, left to right:

Decimal comma,  
hyphen minus, 
double exclamation mark,
low reversed 99 quotation mark, 
high reversed 99 quotation mark, 
colon,
exclamation mark,

Next line:
base single quotation mark,
base double quotation mark,
dot,
reverse question mark,
reverse semocolon,
reverse comma.

These are on shift layer, too.

I'll follow the list in ralt layers.

LornaSIL:
Fixes #2361

### DIFF
--- a/release/o/old_hungarian/HISTORY.md
+++ b/release/o/old_hungarian/HISTORY.md
@@ -1,5 +1,9 @@
 Old Hungarian Change History
 ====================
+1.7.6 (2023-09-19)
+------------------
+* Updated mobile layout
+
 1.7.5 (2023-09-03)
 ------------------
 * Added decimal comma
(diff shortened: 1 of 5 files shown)


LornaSIL:
@DavidLRowe the whole mobile layout looks "new"! Sorry about that. The only thing I changed are the 3 longpress buttons at the bottom right on the "default", "shift", "caps" layers.

mcdurdin:
> @DavidLRowe the whole mobile layout looks "new"! 

This can be mitigated somewhat in the GitHub UI by selecting the cog in the Files changed tab and choosing 'hide whitespace'.

Doing that also shows that this PR fixes something that was somewhat broken in the existing file: the existing file was using surrogate pair escapes, which is a bit ugly. I don't have an appropriate font so I get tofu on screen but storing the char is still better. My memory says this was fixed several versions ago in Keyman Developer (but I haven't searched for the relevant issue).

image

PR #2377, opened 2023-09-29
[gaahmg] New keyboard

LornaSIL:


DavidLRowe:
For the touch layout, on the default and shift layers, some of the punctuation on the full stop longpress keys are incorrect:
- `;` gives U+0308
- `\` gives U+0300

On the numeric layer, the longpress keys for `<`, `>`, `{`, `}` give other characters.

One further observation (not necessarily needing a change): For the desktop keyboard, it seems that it's not possible to type `\` or `|` if using a US-style keyboard, that is one without the extra key between the left shift and `Z` key. I understand this keyboard is based on an MSKLC layout.

DavidLRowe (approved):

PR #2388, opened 2023-10-06
[chinuk_wawa] Error correction for version 1.1

cachang:
I had an error in the coding for one long-press key in version 1.1 uploaded yesterday, corrected now in kmn file.


keyman-server:

DavidLRowe:
In order for this change to be published, the keyboard version needs to change. You'll need to change the version number in the .kmn file. (In Keyman Developer, open the project, select the .kmn file and use the Details tab .)

You should also update the HISTORY.md file. (Since you already removed the version number from README.md, you won't need to change that file.)

Since you're changing files, it would be nice if you could change the dates in the copyright statements (LICENSE.md and the .kmn file) to read: "2022-2023" rather than "2022-3" or "2022-23".  

Thanks!

cachang:
Thank you! I've updated the .kmn, HISTORY and LICENSE files and uploaded
them to the repository. I appreciate all your help navigating the process!

DavidLRowe (approved):

PR #2391, opened 2023-10-09
Add updated version of bu_phonetic

emily-roth:
Updated bu_phonetic keyboard and all relevant documentation to use the Unicode saltillo codepoint U+A78C instead of the old PUA codepoint at request of David Rowe and Lorna Evans.


keyman-server:

LornaSIL:
@emily-roth Thanks for this Pull Request!. Can you delete the two files that crept in with a tilde in front of the filename? I think those are temporary files.

emily-roth:
> @emily-roth Thanks for this Pull Request!. Can you delete the two files that crept in with a tilde in front of the filename? I think those are temporary files.

LornaSIL (approved):
LGTM. Thanks for your first PR!

PR #2397, opened 2023-10-11
[tai_tham_php] Experiment to see if font will display

LornaSIL:
The font doesn't seem to display in the help page. The font doesn't seem to have a bit set for the Tai Tham script so this is an experiment to see if specifying the font in a span will force it to display the glyphs.


DavidLRowe (approved):
RSLGTM

LornaSIL:
The glyphs are now showing for me so it seems to have worked.

PR #2398, opened 2023-10-12
Fv dene zhatie

HopsAndHops:
Built desktop keyboard, created documentation, cleaned mobile rules, centered mobile keyboard. 

Approved by @caforbes


keyman-server:

DavidLRowe:
lgBlue162x31.png is in the package file list, but doesn't exist in the keyboard files (nor is it referenced in welcome.htm). This caused the build to fail. Either remove it from the package file list, or include it in the keyboard files and reference it in welcome.htm.
BUT I see that you've already removed it.

Currently the .keyboard_info and .kps files both have "xsl-Latn". It turns out that "xsl" and "scs" are assigned to the macrolanguage "den", and furthermore "xsl" has been chosen to represent "den" in BCP 47 codes. So the preferred code is just "den". I think that the "xsl" or "xsl-Latn" ones would still work, but "den" would be a better choice long term.

You can decide next week whether you want to change that or leave it as is.

DavidLRowe:
Ah, the file name `lgBlue162x31.png` needs to be removed from the list of Package files. The build of the package fails because this file is in the list, but not present in the keyboard files.

DavidLRowe (approved):

PR #2399, opened 2023-10-14
version 1.3 upload

cachang:


keyman-server:

DavidLRowe:
FYI: Instead of using `"id": "T_0063_0313"` in the .keyman-touch-layout file with a corresponding `+ [T_0063_0313] > 'c̓'` rule in the .kmn file, you can (with Keyman Developer v15 or later) use `"id": "U_0063_0313"` in the .keyman-touch-layout file without needing a rule in the .kmn file. This is not a requirement, just a newer option that may make it easier to create and maintain the keyboard.

cachang:
Thank you, that is great to know and I'll use it moving forward!

DavidLRowe (approved):

PR #2403, opened 2023-10-17
Fv dene zhatie

HopsAndHops:
Updated BCP 47 code as per recommendation, discussed with and approved by @caforbes. Updated keyboard name to better reflect language name.


keyman-server:

DavidLRowe:
@HopsAndHops @caforbes I noticed that the "Language" field in the Package Source (.kps file) is still using "Slave (Athapascan)" as the name of the language. That's the default provided by Developer, but you don't need to keep it. I asked Marc for clarification.

From @mcdurdin : "They can put anything they like into the `<Language>` element in the package, but it doesn't get used everywhere. I would recommend using `Dene Zhatıé` in that element."

HopsAndHops:
That should do it! Thanks for the recommendation :)

DavidLRowe:
See issue #2412 The .php help file references an image file that isn't present in the help folder. 

One solution is to make a copy of the file from the welcome folder and place it in the help folder. (Make sure not to move it, since it is needed in the welcome folder.)

Alternatively, you could use the special syntax available for .php help files and replace the reference to the image file with:
```
<div id='osk' data-states='default'>
</div>
```

DavidLRowe (approved):

PR #2404, opened 2023-10-17
chore: add LICENSE.md to .kps files

mcdurdin:
* Adds LICENSE.md for each keyboard project to its .kps file.
* Cleans up some misformatted licenses without material change to content

* Note: two gff deprecated keyboards have license change from CC to MIT. Need to verify if this is acceptable with @dyacob. These two keyboards slipped past the net as we do require MIT for all keyboards in release/.

TODO:
- [x] Depends on the following PR, and so this will require a version bump for kmc once that PR lands, then this one should build:
  * keymanapp/keyman#9784


darcywong00 (approved):
rslgtm

mcdurdin:
@dyacob are you okay with gff_amh_7 and gff_ethiopic_7 keyboards being relicensed to MIT (they are both old of course, but because they are in release/, they trigger a MIT license validation in the 17.0 compiler toolchain)

dyacob:
@mcdurdin yes, it's ok with me.

mcdurdin:
Thanks @dyacob, that really helps -- much easier than any of the alternative solutions!

PR #2407, opened 2023-10-18
chore: copy keyboard_info description into package

mcdurdin:
Converts keyboard_info description from HTML to Markdown and adds into the corresponding package.


release/d/dega/source/dega.kps
```diff
@@ -21,6 +21,13 @@
     <Copyright URL="">© 2021 MSFJ</Copyright>
     <Author URL="">Y Phič Hđơ̆k</Author>
     <Version URL=""></Version>
+    <Description>Dega keyboard is created for indigenous people in the Central Highlands
+of Vietnam, to preserve their writing, this Dega keyboard can be used by
+Êđê, Jơrai, Kơ'Ho, Mơnông, Bahnar and many other indigenous ethnic
+groups.\
```
darcywong00, line 27: Does `\` work for line continuation?

darcywong00 (approved):
rslgtm

release/d/dega/source/dega.kps
mcdurdin: Unsure. It's probably going to be fine, so went ahead with it as is.

PR #2410, opened 2023-10-18
chore: move related keyboard data to kps

mcdurdin:
Based on a generated report:
[related.txt](https://github.com/keymanapp/keyboards/files/13019192/related.txt)

### DIFF
--- a/release/sil/sil_cameroon_azerty/source/sil_cameroon_azerty.kps
+++ b/release/sil/sil_cameroon_azerty/source/sil_cameroon_azerty.kps
@@ -449,4 +449,7 @@ an AZERTY (French) Keyboard.</Description>
     </Keyboard>
   </Keyboards>
   <Strings/>
+  <RelatedPackages>
+    <RelatedPackage ID="cameroon azerty unicode" Relationship="deprecates" />
+  </RelatedPackages>
 </Package>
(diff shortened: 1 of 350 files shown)


release/sil/sil_cameroon_azerty/source/sil_cameroon_azerty.kps
```diff
@@ -449,4 +449,7 @@ an AZERTY (French) Keyboard.</Description>
     </Keyboard>
   </Keyboards>
   <Strings/>
+  <RelatedPackages>
+    <RelatedPackage ID="cameroon azerty unicode" Relationship="deprecates" />
```
darcywong00, line 453: How are spaces handled in legacy package IDs?

mcdurdin: They basically just work. We don't accept them for new packages but this will be fine.

release/sil/sil_hebrew/source/sil_hebrew.kps
```diff
@@ -135,4 +135,7 @@ official Biblical Hebrew keyboard from SIL.</Description>
     </Keyboard>
   </Keyboards>
   <Strings/>
+  <RelatedPackages>
+    <RelatedPackage ID="ezrasil251" Relationship="deprecates" />
+  </RelatedPackages>
```
darcywong00, line 140: I see sil_hebrew_legacy relates to this keyboard, but not the other way around?

mcdurdin: Correct. We don't want to update older keyboards to point to the new ones that replace them, as in many cases we don't even have the old ones. The api.keyman.com database build figures out the back-references for the api responses.

Another reason: we only need the link one-way -- having two way is another opportunity for inconsistency :grin:

darcywong00 (approved):
lgtm

PR #2411, opened 2023-10-19
fv nuucaanul

HopsAndHops:
Update to single codepoints where possible, change name to better reflect language name, center mobile keyboard, clean rules.

### DIFF
--- a/release/fv/fv_nuucaanul/README.md
+++ b/release/fv/fv_nuucaanul/README.md
@@ -1,4 +1,4 @@
-Nuučaan̓uł keyboard
+nuučaan̓uł keyboard
 ======================
 
 Nuučaan̓uł keyboard layout for Unicode
--- a/release/fv/fv_nuucaanul/README.md
+++ b/release/fv/fv_nuucaanul/README.md
@@ -1,7 +1,7 @@
 nuučaan̓uł keyboard
 ======================
 
-Nuučaan̓uł keyboard layout for Unicode
+nuučaan̓uł keyboard layout for Unicode
 
 Links
 -----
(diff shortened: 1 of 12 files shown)


keyman-server:

release/fv/fv_nuucaanul/README.md
```diff
@@ -1,4 +1,4 @@
-Nuučaan̓uł keyboard
+nuučaan̓uł keyboard
 ======================
 
 Nuučaan̓uł keyboard layout for Unicode
```
caforbes, line 4: Here's another spot with a name to perhaps update.

release/fv/fv_nuucaanul/source/fv_nuucaanul.kmn
```diff
@@ -220,40 +193,40 @@ c Diacritics
 c Special Characters
 
 + [CAPS K_8] > 'Ł'
-+ [CAPS K_7] > 'Ḥ'
++ [CAPS K_7] > 'Ḥ'
 + [CAPS K_6] > 'X̣'
 + [CAPS K_5] > ''
```
caforbes, line 198: This is not rendering for me in the github view, is it one of the private use characters? If it can't be replaced with a valid unicode character yet, maybe add a comment indicating what is supposed to display here.

DavidLRowe: The CAPS K_5 line is producing U+F899, which is in the Private Use Area (PUA).

LornaSIL: In Aboriginal Sans (which I think the fv keyboards recommend) that is an upside down capital Y - with a bar on the stroke.
image

That character has been provisionally accepted to Unicode:
http://www.unicode.org/alloc/Pipeline.html
LATIN CAPITAL LETTER LAMBDA WITH STROKE

So, I think you need to keep the PUA character, but it might be useful to add a comment `c provisional Unicode assignment to U+A7DC. Codepoint should be changed once it's in Unicode and fonts support it`

Or something to that affect. It will help not to have to do the research again.

HopsAndHops: Thanks, everyone! The name in README has been changed and Lorna's suggested comment has been added. Super great that the capital lambda with stroke is on its way to being added!

PR #2414, opened 2023-10-19
chore: add fully published version of .keyboard_info for legacy packages

mcdurdin:
The legacy packages rarely have changes. The new kmc compiler does not attempt to build .keyboard_info from compiled .kmp files any more; instead it works from source files only, so it won't work with the legacy packages, which are mostly only available as compiled content. So we want to take the latest published version of the .keyboard_info files for these legacy packages, and publish it as source data, so it can simply be copied into the build folder.

This means that if we do need to modify a legacy package for any reason, we will need to manually edit the associated .keyboard_info. But this
seems to be a very infrequent event.


darcywong00 (approved):
rslgtm

PR #2415, opened 2023-10-19
chore: remove release and experimental .keyboard_info

mcdurdin:
These are 100% generated from keyboard metadata now, and are no longer used in source.


darcywong00 (approved):
rslgtm

PR #2416, opened 2023-10-19
chore: update help repo as a PR

mcdurdin:
This will open a pull request for keyboard layout documentation changes, so then the changes will go through CI, which will avoid the current issue where broken links in keyboards cause other PRs to fail CI.

It will also mean that we no longer push to master for keyboard help publishing, which will be cleaner.


rc-swag (approved):
lgtm

PR #2418, opened 2023-10-22
[malar_malayalam] Update to version 1.7

ramesh-k:
#### Changes in Version 1.7 ####
* Allow space after ZWJ (200D) to type words ending with legacy encoding chillu characters.


keyman-server:

LornaSIL:
Does the welcome file need to document this change?

ramesh-k:
> Does the welcome file need to document this change?

@LornaSIL Since all chillu characters in Malayalam have already been encoded in Unicode, this is a minor fix for compatibility. There is no need to document this change in the welcome file. Thank you.

LornaSIL (approved):
LGTM

PR #2419, opened 2023-10-23
chore: update README.md and jq version

mcdurdin:
README.md rewritten for kmc.

jq-win64.exe updated because previous version crashed on some inputs.


README.md
```diff
@@ -25,42 +28,70 @@ Keyman Developer 17.0+ is recommended to edit these files. <https://keyman.com/d
...
+  * Ensure coreutils is on the PATH: `export PATH="$HOMEBREW_PREFIX/opt/coreutils/libexec/gnubin:$PATH"`
 
 ### Build instructions
 
-`build.sh` can be used to build all the keyboards from the command line.
+`./build.sh` can be used to build all the keyboards from the command line.
 
-* The Keyman Developer compiler is included in the `tools/` folder and so Keyman Developer
-  is not required for builds.
+* `kmc`, the Keyman Developer compiler, is installed as a Node.js modul, and so
```
darcywong00, line 51:
```suggestion
* `kmc`, the Keyman Developer compiler, is installed as a Node.js module, and so
```

Do we call it a Keyman Developer CLI compiler in other places?

darcywong00 (approved):
lgtm

README.md
mcdurdin: > Do we call it a Keyman Developer CLI compiler in other places?

Yeah nah. CLI would need to be 'command line' which is okay but probably not needed everywhere.

PR #2421, opened 2023-10-23
[newa_traditional_extended] New keyboard

SvarnimN:
Newa Traditional Extended Layout is based on Newa Traditional Layout. Newa Traditional Extended layout is more accessible and easy to use. It also includes the most commonly used conjuncts and special characters.


keyman-server:

LornaSIL:
Nice Pull Request! 

Two things. Could you delete the `.gitignore` file? from the PR?

- The other thing isn't required, but I would recommend you add the `keyboards\release\shared\fonts\noto\Newa\NotoSansNew-Regular.ttf` font to the .kps files. Then, for the **Keyboards** tab you can select the font for both "Keyboard font" and "Display font". (I see the `newa_traditional` keyboard doesn't do this, but we should probably add it!) Doing this would make it easier for the users of the keyboard.

SvarnimN:
@LornaSIL .gitignore file has been removed and Noto Sans Newa Regular font has been added to the kps file.

LornaSIL (approved):
LGTM

PR #2424, opened 2023-10-24
Update to version and .keyman-touch-layout file

HopsAndHops:
Updates per @DavidLRowe 's recommendations.


keyman-server:

LornaSIL (approved):
I'm surprised that they use a dotless "i". I would imagine it's a bit of a problem with casing! But, anyway, the changes you made are fine.
LGTM.

PR #2429, opened 2023-10-25
feat: automatically deploy fonts to s.keyman.com

### LINKED ISSUE #2323: feat(infrastructure): automatic deployment of fonts to s.keyman.com
The goal was for @mcdurdin to add functionality so the fonts would automatically be copied to s.keyman.com so I wouldn't need to do that extra step. I'm not sure where we are in that process either.

_Originally posted by @LornaSIL in https://github.com/keymanapp/keyboards/issues/1942#issuecomment-1660620144_

mcdurdin:
Fixes #2323.

All fonts in release/shared/fonts will be deployed to s.keyman.com.

Furthermore, changes to s.keyman.com will now be opened as a pull request, in order to allow us to manually address any CI failures, and remove the need for direct commits to master.

Note: a further improvement in the future may be to scan all packages for web font references and include those.

Note: another question is about font versioning -- if a font is updated, s.keyman.com will continue to serve the old one for some time due to caching. How important is this to fix?

### DIFF
--- a/resources/s-keyman-com.sh
+++ b/resources/s-keyman-com.sh
@@ -154,7 +154,7 @@ function commit_and_push {
   echo "committed keyboards and fonts to '$branch'"
   git commit -m "auto: Keyboard and font deployment" || return 1
   git push origin $branch || return 1
-  echo "Push to help.keyman.com complete"
+  echo "Push to s.keyman.com complete"
 
   hub pull-request -f --no-edit -l auto || return 1
   echo "PR created"
--- a/resources/s-keyman-com.sh
+++ b/resources/s-keyman-com.sh
@@ -6,11 +6,11 @@
 SHLVL=0
 
 #
-# In this script we update the git repo and copy the keyboards over, 
+# In this script we update the git repo and copy the keyboards over,
 # for all keyboards in .cache/upload. So this should be run after ci.sh.
 #
 # The only command line parameter is -f. This should only be used if you
-# want to update all the keyboards in the s.keyman.com repo, even if they 
+# want to update all the keyboards in the s.keyman.com repo, even if they
 # already exist. This could possibly be needed if a severe compiler bug is
 # found, but rarely otherwise. Note that this would only update the file on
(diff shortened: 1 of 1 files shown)


resources/s-keyman-com.sh
```diff
@@ -106,18 +125,45 @@ function upload_keyboards {
...
+    return 0
+  }
+  echo "changes added to cache...>>>"
+
+  echo "creating new branch '$branch'"
+  git switch -c $branch
+  echo "committed keyboards and fonts to '$branch'"
+  git commit -m "auto: Keyboard and font deployment" || return 1
+  git push origin $branch || return 1
+  echo "Push to help.keyman.com complete"
```
darcywong00, line 157: copy-paste?

```suggestion
  echo "Push to s.keyman.com complete"
```

darcywong00 (approved):
lgtm

darcywong00 (approved):
lgtm

LornaSIL:
This sounds good! I assume this means I no longer need to commit to s.keyman.com. 
What happens if someone has a font in the **keyboard** folder (not shared)? Do we need to make sure it ends up in the shared folder? Also, `experimental` doesn't have a `shared` folder, We just have been putting fonts in the keyboard folder there. Do we need to make changes to how we are handling those?

mcdurdin:
> This sounds good! I assume this means I no longer need to commit to s.keyman.com.

Correct.

> What happens if someone has a font in the **keyboard** folder (not shared)? Do we need to make sure it ends up in the shared folder? Also, `experimental` doesn't have a `shared` folder, We just have been putting fonts in the keyboard folder there. Do we need to make changes to how we are handling those?

Currently, the deployment only takes files from release/shared/fonts. A future update will hopefully scan all the .kps files and take all the DisplayFont, OSKFont, WebDisplayFonts and WebOSKFonts entries from those files. But a problem here is that it is quite possible for two different packages to have a font file with the same name (and potentially different versions), so I haven't hurried into this.

PR #2432, opened 2023-10-25
New Burushaski Girminas for Perso Arabic script

quwatsunny:
New Burushaski Keyboard developed to write the Burushaski Perso Arabic Script


quwatsunny:
Also documentation and fonts have been added for the Burushaski Girminas latin Keyboard

keyman-server:

LornaSIL:
Since one of the things you are committing is a new keyboard, I would prefer you split the Pull Request into 2. One for the existing keyboard and one for the new keyboard.

A few comments.
- Your master is out of date with the Pull Request. Please update from master.
- You committed a change from a different keyboard `rac_burushaski.kpj`. I imagine you started by looking at that keyboard and it got modified. Please make sure that file does not get committed.
- Make sure to delete all the `*.bak` files that have crept in.

For the `burushaski_girminas` keyboard:
- If you have cloned the whole repo, it would be preferable for you to use the DejaVu fonts from the shared folder rather than adding yet another copy of the font to the repo. Those are located here: `keyboards\release\shared\fonts\dejavu`. If you don't have a clone of the whole repo that may make it difficult for you to use the shared fonts.

For the `burushaski_girminas_perso_arabic` keyboard:
- I haven't done any testing, so these are just initial comments.
- The welcome file should be named `welcome.htm` and not `welcome.html`. (My experience is that it won't show up when the package is installed if the suffix is `html`).
- And then a font question about `Burusho_Quwatz_Noto_Nastaliq.ttf`. I assume you have modified that font from `Noto Nastaliq Urdu`? Since the font is under the OFL license, that is fine. However, the internal font NAME is `Noto Nastaliqu`. That is the name people will see when they install the font (not **Burusho Quwatz Noto Nastaliq**. I would recommend you work on the font a bit more to make sure the internal font name gets set properly.  I don't know what font tool you are using to make the changes and still preserve the OpenType, but I opened the font in FontLab and I'll show you my concerns about the naming.

It would be the **family name** that needs fixing:
image

In addition **Burusho Quwatz Noto Nastaliq** is VERY long. If I change the **family name** to that, and then click on **Build Names**, I get this:
image

Where the **Full name** and **PostScript:** name get truncated. If that works for you, I guess it's fine, but I'd recommend a shorter name such as **Burusho Quwatz Nastaliq**. Then you end up with these names which seem appropriate:
image

My understanding is the Noto fonts allow you to change the name like that since they are under the OFL license.

quwatsunny:
Thanks a lot Lorna.  I appreciate your attention to detail. I will start working on these points one by one Have a wonderful day. Sent from my iPadOn Oct 25, 2023, at 10:26 AM, Lorna Evans ***@***.***> wrote:﻿
Since one of the things you are committing is a new keyboard, I would prefer you split the Pull Request into 2. One for the existing keyboard and one for the new keyboard.
A few comments.

Your master is out of date with the Pull Request. Please update from master.
You committed a change from a different keyboard rac_burushaski.kpj. I imagine you started by looking at that keyboard and it got modified. Please make sure that file does not get committed.
Make sure to delete all the *.bak files that have crept in.

For the burushaski_girminas keyboard:

If you have cloned the whole repo, it would be preferable for you to use the DejaVu fonts from the shared folder rather than adding yet another copy of the font to the repo. Those are located here: keyboards\release\shared\fonts\dejavu. If you don't have a clone of the whole repo that may make it difficult for you to use the shared fonts.

For the burushaski_girminas_perso_arabic keyboard:

I haven't done any testing, so these are just initial comments.
The welcome file should be named welcome.htm and not welcome.html. (My experience is that it won't show up when the package is installed if the suffix is html).
And then a font question about Burusho_Quwatz_Noto_Nastaliq.ttf. I assume you have modified that font from Noto Nastaliq Urdu? Since the font is under the OFL license, that is fine. However, the internal font NAME is Noto Nastaliqu. That is the name people will see when they install the font (not Burusho Quwatz Noto Nastaliq. I would recommend you work on the font a bit more to make sure the internal font name gets set properly.  I don't know what font tool you are using to make the changes and still preserve the OpenType, but I opened the font in FontLab and I'll show you my concerns about the naming.

It would be the family name that needs fixing:

In addition Burusho Quwatz Noto Nastaliq is VERY long. If I change the family name to that, and then click on Build Names, I get this:

Where the Full name and PostScript: name get truncated. If that works for you, I guess it's fine, but I'd recommend a shorter name such as Burusho Quwatz Nastaliq. Then you end up with these names which seem appropriate:

My understanding is the Noto fonts allow you to change the name like that since they are under the OFL license.

PR #2440, opened 2023-10-29
[newa_traditional_extended] show texts for the invisible character keys on OSK

SvarnimN:
Show texts for the invisible character keys i.e. zwnj, zwj and dead keys on OSK, which were missing in the previous version.


keyman-server:

LornaSIL (approved):
LGTM

PR #2441, opened 2023-10-29
[burushaski_girminas_perso_arabic] New keyboard

quwatsunny:
I would appreciate if you could merge this new Burushaski Girminas Keyboard for Perso Arabic script of the Burushaski Language. 
For the  help page, I would appreciate if you could ensure the embedded font to work, else the text in the said script will not display properly. 

Also, I was not able make the font work in the welcome.htm page, within the frame of the Keyman. I am not sure how that works. Ideally I would have loved to make the css work for the embedded fonts in the help php file and welcome.html. It works fine on browser except it does no Let me know if there is any work around. 

Otherwise then that, the keyboard is ready to be published.


keyman-server:

LornaSIL:
Thanks for the PR. There are a number of things that need to be addressed before we can merge it. I hope I can make it clear.

- In the `keyboard_info` file itself, the whole text in the description must be in one line. Otherwise the code will not validate. You cannot have a linebreak in that text. However, you can add this `\r\n\r\n` where you have the linebreak. So, something like this `"description": "Burushaski Girminas ... Burushaski.\r\n\r\nThis particular orthography ... and Yasin)."` would give you the linebreak on the home page of the keyboard.
- You should delete this file in the source folder `burushaski_girminas_perso_arabic-code.js`. We don't need any compiled files in the repo.
- In the .kmn file:
   - On line 74 you have this rule: `+ [SHIFT K_Y] > nul`
   - On line 130 you have this rule: `+ [SHIFT K_Y] >  U+06D2 U+08F7 c On screen keyboard` 
   - These rules conflict and line 130 rule will never "fire". You should decide which you want. It seems you want something on the OSK, but rules don't work for the OSK. You need to write the rules for the keyboard.
- In the .kps file:
   - When I try to build I get this message `burushaski_girminas_perso_arabic\build\icon_burushaski_keyboard_package.png does not exist` 
   - That file is not in the pull request and anyway, you shouldn't be pulling an image from the `build` folder since that is a temporary folder that is not committed. If you need that image, please put it in the `source` folder.  
   - Also, please remove the .kmn file from the list of files in the .kps. We just need the .kmx in the package (which you have included)
   - Also, please remove the .woff and .woff2 files  from the list of files in the .kps. (It's correct to leave them in the Pull Request though because we need those for displaying the online help.)
- For the help folder, you should delete the `fonts` folder. We do not want the fonts in the repository twice. It is very awkward to keep fonts updated to the correct version when there are multiple copies. Our build system will put the fonts in a place available for the online help.
- Also, in the help folder, the `welcome.php` file should be renamed `burushaski_girminas_perso_arabic.php`
- If you could try this for the header, I think it would be best:
```
<?php 
  $pagename = 'Burushaski Perso Arabic Keyboard Help';
  $pagetitle = $pagename;
  require_once('header.php');
?>

<head>
	<link rel="stylesheet" href="stylesheet.css"/>
</head>
```

Once these things are fixed we will review to see if there are other issues.

quwatsunny:
@LornaSIL @DavidLRowe  I hope you are doing well. Kindly take a look at this pull if possible.

LornaSIL (approved):
LGTM

LornaSIL:
Ok, I've approved it. I was unavailable last week. Sorry to be slow. Could you please tell me how this keyboard compares with https://keyman.com/keyboards/rac_burushaski. Is it similar? I didn't compare the character set, but I'm just wondering if yours could be considered better or similar or how they compare.

LornaSIL:
Also, the font will not get deployed to the online help yet. I'm still checking on what needs to happen since we've changed a few things in the build system.

quwatsunny:
@LornaSIL  thank you very much.  Since there was a silence, i was concerned about your wellbeing more than getting the keyboard approved. I am glad that you are well. 

To your question about comparing the two keyboards - these are two different orthographies.  One promoted by Burushaski Research Academy that promotes books and research by Late Nasir Uddin Hunzai a Burushaski speaker.  The organization is now run by his family members and friends and primarily representative of religious Burushos of Hunza.  While the language is spoken in Hunza, Nagir, Yasin and kashmir, In late, 2020, linguists, researchers, men of letters, and language enthusiasts from all these regions came together and approved a common orthography which required a rekeyboard.  These people have an organizaiton celled Burusho Mahraka which representative of all three regions.  The orthography has a Primer approved by the government to be included in syllabus. 

People of Burushaski Research Academy were not ready to give up their orthography.  Obviously there are differences between the two orthographies and hence required a new keyboard for the common Orthography of all three regions.  I would not be able to say which one is better because there two different orthographies with different glyphs.

LornaSIL:
Great! Thanks for the explanation.

PR #2443, opened 2023-10-30
[bjcree] Update keyboards which include the bjcree fonts

LornaSIL:
There was an issue with font fallback in Word because the font did not have the Cans Unicode bit set. This version of the font has been tweaked to have the bit set. Permission granted by copyright holder.
All the keyboards that use the font have had to be updated because the filename(s) of the font also changed.


LornaSIL:
Before approving we are waiting for the final okay that the font tweak worked.

LornaSIL:
Relates to: https://community.software.sil.org/t/font-changing-while-typing/7982

LornaSIL:
>I heard back from the Cree School Technicians and the new font fixed the problem!

This PR is now ready for review as the feedback is that the new font fixes the issue.

mcdurdin (commented):
One problem with having fonts with same family name but different filename is that both can be installed on the same system. In your opinion, is this likely to cause trouble for end users?

DavidLRowe (approved):
LGTM

DavidLRowe:
@LornaSIL I'll let you think about Marc's question and merge if you think the PR is ready.

LornaSIL:
#2455 replaces this one.

PR #2444, opened 2023-10-30
[Fv dane zaa zaage] Created mobile rules

HopsAndHops:
- Changed combining keys to dead keys
- Created new set of mobile rules
- Added return to default layer to mobile layout
- Added stacking prevention


keyman-server:

PR #2447, opened 2023-11-01
Change deseret.kmn to UTF-8

DavidLRowe:
No change in content. Change from UTF-16 LE (with BOM) to UTF-8 (no BOM).


mcdurdin (approved):
LGTM

PR #2449, opened 2023-11-02
[Yorùbá Ìrọ̀rùn] Initial submission of Yorùbá Ìrọ̀rùn Keyboard

w3techy:


keyman-server:

DavidLRowe:
Thanks for your contribution.

There is a second yo.kps file in the yo folder that should be removed. (The yo.kps file in the source folder is properly placed and needs to be kept.)

On the Details tab for the .kmn file, there's a "Targets" item. Note that "any" implies all the others. 
- If you are supporting a desktop layout and a touch layout, you should select "any" and deselect all the others.
- If you are not supporting a touch layout, you should deselect "any" and select only "desktop" and "web". And in addition, you should use the "Details" tab to remove the Touch Layout.

There are a number of rules in the .knm file that conflict. See the compiler output for details.

In the Package Files section, there are two source files that do not need to be in the package. These should be removed:
- yo.kpj
- yo.kps

Any corrections you make to the "yoruba" branch of your fork of the keyboards repository will be added to this pull request. When the pull request is ready, we can accept the keyboard.

DavidLRowe:
Some further notes on the touch layout:
Default layer: Keys in top row have no labels even though they produce values ( 1 through 9, 0, - = ).
Leftmost key in third row has no label but produces backtick character.
Shift key doesn't activate Shift layer.

Version numbers in Keyman are x.y or x.y.z (so your 1.0.1 is fine). You're welcome to use four part numbers internally, but using three part version numbers will keep our build system happy!

DavidLRowe:
@w3techy Just checking the status of this PR. There are a few items I mentioned that you haven't addressed. Let me know if you need more details.

LornaSIL:
@w3techy do you need assistance with making the changes that David has requested?

LornaSIL:
@w3techy this keyboard has some implementation issues that will give your users a lot of problems. It has almost the same keystrokes on LALT as on RALT. We do not recommend even putting keys on LALT because many/most applications use LALT as a hotkey.
Since you are having computer issues, I'm willing to fix this up to address that issue and we could then approve it. Then you could fix any changes we made that you didn't like. Since this PR has been sitting here awhile I'd like to get it fixed and approved if possible.

LornaSIL:
@w3techy We'll be closing this PR in a few days if we don't hear back from you.

w3techy:
@LornaSIL I will be needing assistance with making the changes requested.

DavidLRowe:
Replaced by #2559

LornaSIL:
Your keyboard is now online: https://keyman.com/keyboards/yo
I hope it's how you want it. I had to do quite a bit of work on the mobile layout as it was not actually functional.
The online help file isn't quite right, but we're working on it and maybe by the time you look it will be fine.

PR #2454, opened 2023-11-08
Responsive Page

quwatsunny:
There was a charter missing in the image illustration. Also css file white-space nowrap was causing the page to be non-responsive.  I have removed it to resolve the issue of responsiveness in the help page.


LornaSIL (approved):
LGTM

PR #2456, opened 2023-11-09
[Fv dane zaa zaage] Updated mobile layout

HopsAndHops:
Updated mobile layout, cleaned existing rules and created new ones, added stacking prevention rules, reverted change that was pushed mistakenly.


keyman-server:

PR #2460, opened 2023-11-14
[fv_uwikala] changed keyboard name

HopsAndHops:
cleaned up mobile rules, added rule to backspace accents instead of whole character, centered mobile layout


keyman-server:

LornaSIL (approved):
LGTM

PR #2462, opened 2023-11-16
[kayan] Re-order punctuation on long-press of period key

### LINKED ISSUE #2402: [kayan] Request to reorder punctuation longpress keys
Request from User
"I am wondering who I should initially ask if I want to adjust the Keyman
keyboard for Kayan for Android?  The main need is to change the order of
the symbols that appear on long press for the period/full stop so that
the more commonly used symbols are closer to the period and the less
used ones are farther away. The rest of the keyboard, at this point, is
to remain unchanged."

darcywong00:
Fixes #2402 

> [punctuation] would end up being
> : ; / ' " ! , ?

This re-arranges the long-press punctuation keys on the period key. The user requested the common ones be closer to the base period button (towards the right of the long-press options).

User also preferred forward slash for websites


LornaSIL (approved):
LGTM

PR #2470, opened 2023-11-28
[dazaga_gourane] Initial submission of Dazaga Gourane touch keyboard

ZvSHUB:


DavidLRowe:
Thanks for your submission. 

In the README.md, dazaga_gourane.keyboard_info, readme.htm and welcome.htm files you have left the placeholder text: "dazaga_gourane generated from template". This should be replaced by a more meaningful description that will help people searching for the keyboard. It could be as simple as "The Dazaga Gourane keyboard is designed for typing the Dazaga language."

You currently have "dzg-Latn-TD" in the .keyboard_info file and the .kps file. The preferred tag would be simply "dzg".

Are your "tablet" and "phone" layouts the same? If so, I'd suggest removing one of them. (If there's no phone layout, Keyman will use the tablet one, and vice versa.) The only reason to have both layouts is if they are different (for example, using four rows instead of five on the phone layout). If they're the same, it's just extra work keeping the two layouts in sync. So, I'd recommend removing the "phone" layout.

We normally ask for a desktop layout (and a corresponding on-screen keyboard). Is that something you could add?

If anything is unclear, please let us know. Thanks!

LornaSIL:
@ZvSHUB We'll be closing this PR in a few days if we don't hear back from you.

LornaSIL:
@ZvSHUB I'm closing this Pull Request. If you wish to fix the issues listed above, we will be happy for you to open a new Pull Request or reopen this one.

PR #2471, opened 2023-11-29
[shared_fonts_three] Used shared fonts rather than local

LornaSIL:
These are the rest of the inscript keyboards and also the buhid keyboard.


DavidLRowe (approved):
Thanks for all this work!

PR #2472, opened 2023-11-29
[shared_fonts_four] Used shared fonts rather than local

LornaSIL:


LornaSIL:
For sxava I removed two fonts that weren't actually needed for the keyboard and the license was unclear.

PR #2477, opened 2023-12-01
chore: convert to UTF-8

DavidLRowe:
Additional search found two .php files with non-UTF-8 encoding:
release/l/lahu/source/help/lahu.php "ANSI"
release/rac/rac_brahui_latin/source/help/rac_brahui_latin.php "UTF-16 LE BOM"


LornaSIL (approved):
LGTM

PR #2479, opened 2023-12-01
[font_licenses] Use fonts with a clear license

LornaSIL:
I've tried to review all the fonts in the shared repo for the licenses. Where possible, I have changed to use a Noto font.

The `kbdsn1` keyboard is deprecated, but I couldn't delete the font without updating that keyboard. Since it is deprecated, I simply removed the font rather than adding in a Noto font. I had to bump the version.


DavidLRowe (approved):
Thanks for all your work on these!

DavidLRowe (approved):
tainua.kpj  was updated in #2478. I resolved the conflict by using the version number (1.0.5) from that PR (which is already in master).

Reapproving

PR #2485, opened 2023-12-05
[baybayin] removed r and R from nul store

Meng-Heng:
Fixed a character:
170D ᜍ TAGALOG LETTER RA

Report on App Store (by Rymarpinoy – Nov 22, 2023), quoted below:
> The app is cool and all with the Ancient Filipino Scripts but here’s the thing, 1 The Baybayin script keyboard itself The weird things is I can’t backspace the Pamudpod symbol and I can’t type the symbol “Ra” 2 The keyboard won’t load for me on the keyboard itself

### DIFF
--- a/release/b/baybayin/HISTORY.md
+++ b/release/b/baybayin/HISTORY.md
@@ -1,5 +1,10 @@
 Baybayin Keyboard Changelog
 
+1.2 (2023-12-04)
+----------------
+* Fixed a character:
+170D ᜍ TAGALOG LETTER RA
+
 1.1.1 (2023-11-29)
 ----------------
 * Used shared font instead of local
--- a/release/b/baybayin/baybayin.kpj
+++ b/release/b/baybayin/baybayin.kpj
@@ -3,14 +3,16 @@
   <Options>
     <BuildPath>$PROJECTPATH\build</BuildPath>
     <CompilerWarningsAsErrors>True</CompilerWarningsAsErrors>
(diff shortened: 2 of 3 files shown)


keyman-server:

Meng-Heng:
> I do not think "R" should have been removed from the nul store because there is no rule for the capital letter, there is just a rule for the lowercase [K_R].

```
+ [SHIFT K_R] > 'ᜟ'
```
I found the key is mapped to a character as shown above (line 33).

> Also, would you like to regenerate the graphic (for the default layer) to go in the welcome.htm file so that the user will see that r will output 170D ᜍ TAGALOG LETTER RA?

I have regenerated two images of the keyboard layout and included them in the next commit.

LornaSIL:
> I found the key is mapped to a character as shown above (line 33).
You are correct. I missed it when I looked at the source

LornaSIL:
I can accept this PR "as is" or you could change to using the Catrinity font which would address this issue #2489 
Catrinity is in this folder: `\release\shared\fonts\catrinity`

I'll go ahead and approve it, but not merge. Then if you want to go ahead and merge we can address the font issue later.

LornaSIL (approved):
Approving, but let @Meng-Heng decide whether to merge or to change fonts.

LornaSIL:
Actually, I'm wrong. The newest version of the Noto Sans Tagalog font does include those characters. Let me update the font in the repo before we merge this PR.

PR #2499, opened 2023-12-13
Laz Keyboard

okandale:
Laz keyboard for Laz speakers in Turkey including a dictionary.


keyman-server:

LornaSIL:
We prefer that keyboards not just be a mobile layout. Is there a reason this wouldn't work as a desktop keyboard as well?

okandale:
Hi Lorna,

This keyboard was specifically designed for smartphones; it has up to 5 layers and the keys have been placed in a way that makes sense on phones.

Transferring it to PC as is would make it incredibly inconvenient and I believe nobody would use it that way; I am planning on creating a PC keyboard as well but its layout will differ vastly (i.e. creating a dead-key Caron for all the Laz letters, putting ʒ where the Turkish keyboard usually has ü (and make ü available on the alt-layer).
How would you proceed? Should I create the Laz keyboard for PC in this as well or create a separate project? I really don't think it makes sense to use the phone layout for PC.

Best wishes,

Okan





Sent from Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

LornaSIL:
>How would you proceed? Should I create the Laz keyboard for PC in this as well or create a separate project? I really don't think it makes sense to use the phone layout for PC.

Many of the keyboards in our repo are quite different between desktop, mobile-phone and mobile-tablet. But the character set that the different layouts produce should be the same. My preference would be you tie it together in one keyboard. Then, it wouldn't be laz_android. It could be just `laz`.

mcdurdin:
One point of consideration -- it's not uncommon for Android users to have an external bluetooth keyboard, in which case they need a desktop layout to use with it.

okandale:
I managed to create the PC keyboard version of my keyboard (and it only took me 4 hours, yay!) But I'm not sure how to update my pull request. Do I just create a new one?

Sent from Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

LornaSIL:
You changed the folder name to `laz`, but the project name (files) will also need changing to `laz`. It's a bit complicated because the .kpj, .kmn, kps, all reference the filenames. Do you want me to pull and make the changes for you?

okandale:
Hi Lorna, I am very sorry about that! Could you please change the names to Laz?

Thank you!

Best wishes,
Okan

Sent from Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

LornaSIL:
I'll try to do it today, but someone else will have to approve it. He isn't available until next week.

LornaSIL:
>Thanks, Lorna! Would I have to manually change the names on my computer as well so whenever I update it you won't have to change the name? I already spotted a mistake in my phone keyboard (the - (dash) key's ID is incorrect). Thanks!

Yes, you will need to update your files to match what is on the keyman keyboard repo. For now, go ahead and do any of the updates you see. I'll start working on the revision in about an hour. After that, you would then need to wait until the keyboard is "live" to do further updates.

LornaSIL:
Replaced by #2503

PR #2501, opened 2023-12-14
Laz Keyboard

okandale:
Previously only for mobile devices but now with an updated version supporting pc keyboards.


keyman-server:

LornaSIL:
We are going to need to be consistent with your language tag. In one file you have used `lzz` and in another file you used `lzz-TR`. Since the language is primarily used in Turkey, the `TR` is not needed. I would recommend going with `lzz`. This would need fixing in your lexical-model as well.

LornaSIL:
Replaced by #2503

PR #2503, opened 2023-12-15
[laz] Commit renamed folder/files for okandale

LornaSIL:
Mostly I just renamed the filenames, but I did have to fix a few of the touch layout rules as well as the desktop rules for producing ")".
Also, added a bit of documentation in the welcome.htm file and added a help fil.


okandale:
Hi Lorna, you asked me to change the region code to lzz instead of lzz-tr. I am happy to change it but I'm unable to see the region code option inside keyman developer. It only appears when creating a new project, so I am also not able to see which files were labelled as lzz instead of lzz-tr.

Thank you for your help!

LornaSIL:
@okandale this PR uses just `lzz` so once it's approved it will be available. That means your lexical model should just use `lzz`.

DavidLRowe (approved):

LornaSIL:
I've submitted a new PR on your behalf: https://github.com/keymanapp/keyboards/pull/2508

For the future you should get set up to be able to pull changes so you have the most up-to-date files. Then, if we have to edit the keyboard you can still get the updated files and continue further development. Start here:
https://help.keyman.com/developer/keyboards/github/
Then if you have questions you can use our [community site](https://community.software.sil.org/c/keyman/19) to get assistance.

Issue #2502, opened 2023-12-14
[fv_all] - Some keyboards are not the latest


darcywong00:
@caforbes reported the latest 16.0 FV app doesn't include the latest versions for some keyboards:
* Southern Tutchone (updated 4mo ago) - on my Android phone it's listed as 9.2.1 but I see a plain qwerty layout (9.2)
* For Dane-Zaa Zaage (updated 2mo ago) I can only download 9.2, not 9.3
* we don't see a recently added keyboard in the list anywhere to download (Hul̓q̓umin̓um̓/hulquminum_combine

-----

Some other things to investigate:
* The latest [fv_all.kmp](https://downloads.keyman.com/keyboards/fv_all/12.5/) package (12.5) has a Last-modified timestamp 2023-04-03 while the folder is 2023-12-12
* fv_all (12.5) is including fv_dane_zaa_zaage 9.2

darcywong00:
> The latest fv_all.kmp package (12.5) ....

Ah right, I forgot this is where we increment the fv_all.kmp version
https://github.com/keymanapp/keyboards/blob/1a34852801783764f475743c6c515068b0910594/release/packages/fv_all/source/fv_all.kps.in#L21

darcywong00:
We might also need to update [keyboards.csv](https://github.com/keymanapp/keyman/blob/master/oem/firstvoices/keyboards.csv) in the main repo so the FV apps can associate the BC region with fv_hulquminum_combine

DavidLRowe:
@darcywong00 Does #2504 fix this issue?

darcywong00:
> Does https://github.com/keymanapp/keyboards/pull/2504 fix this issue?

That plus keymanapp/keyman#10269 which associates the BC Coast region for fv_hulquminum_combine will fix the issue.

darcywong00:
Fixed by the 3 associated PRs.
keymanapp/keyman#10285 has the fix for 16.0

PR #2350, opened 2023-08-24
Fv southern tutchone

HopsAndHops:
- Updated to single codepoint outputs where possible
- Updated documentation
- Centered keyboard
- Removed duplicate punctuation

Approved by @caforbes

PR #2504, opened 2023-12-16
[fv_all] Bump versions to publish earlier fixes

darcywong00:
Addresses these points on  #2502 
* Southern Tutchone (updated 4mo ago) - on my Android phone it's listed as 9.2.1 but I see a plain qwerty layout (9.2)
* we don't see a recently added keyboard in the list anywhere to download (Hul̓q̓umin̓um̓/hulquminum_combine
* The latest [fv_all.kmp](https://downloads.keyman.com/keyboards/fv_all/12.5/) package (12.5) has a Last-modified timestamp 2023-04-03 while the folder is 2023-12-12
* fv_all (12.5) is including fv_dane_zaa_zaage 9.2

Changes:
* The keyboard version in fv_southern_tutchone.kmn should be 9.2.2 to to publish the updates from #2350.
* The fv_all keyboard package is maintained in fv_all.kps.in so it needs to be bumped to 15.6 to publish updates since April 2023.
* Bumping fv_all version will include hulquminum_combine

I still need to update the keyboards.csv file in the Keyman source repo for the mobile apps to associate Region and Language ID for ~fv_southern_tutchone~ fv_hulquminum_combine.

Tagging @caforbes


caforbes:
Looks good, thanks for looking into this. For adding the new keyboard, I assume you meant `fv_hulquminum_combine`, which can have the same values as `fv_hulquminum`.

Are there any other steps we might need to take when performing keyboard updates in future? We have more planned.

FYI I will be away until Jan 4.

darcywong00:
Sorry will have to investigate why hulquminum_combine isn't getting included

darcywong00:
Thanks all. I'll merge this first, so that the new fv_all.kmp keyboard package will be available to test #10269.

PR #2506, opened 2023-12-18
[dega] change = to ! on numeric and more layout

Nnyny:
change = to ! on Dega keyboard, phone platform and change images on welcome.html


keyman-server:

LornaSIL:
Thanks for your submission @Nnyny ! When we update a keyboard, we always need to update the version number in the .kmn file.
Also, the HISTORY.md file should be updated with the version number and a very short description of what changed.

Also, if the keyboard hasn't been changed, it may be that the copyright year needs to be updated to 2023 in LICENSE.md. However, I believe this keyboard has been updated recently so for this time, the copyright dates probably do not need updating.

Also, while you are updating the .kmn and the HISTORY.md file, could you remove the version number and the copyright year from the README.md file? That way the README.md file doesn't have old information and wouldn't need to be updated again.

FYI: if the version number is not changed, then our build system would not know it needs to serve up a new keyboard package to people.

Nnyny:
Thank you @LornaSIL! I will work on that.

LornaSIL (approved):
LGTM

LornaSIL (changes requested):
Actually, I'm sorry, but there are still issues with this PR.

You have accidentally saved the `phone_numeric` and `phone_symbol` files as .jpg files instead of .png files. That means the .png files are still in the repo _and_ also the `.kps` file is still referencing the .png files. So, you can either:
- delete the .png files from the repo and change the .kps file to reference the .jpg files
or
- rename the .jpg files to .png and make sure the welcome.htm file references the .png files again.

I'm sorry I did not catch this earlier.

LornaSIL (approved):
LGTM

PR #2511, opened 2023-12-27
Propose IDC Deseret keyboard layout

sigilante:
IDC Deseret is designed to facilitate transcription of written manuscripts.  It does not rely on QWERTY mnemonics but introduces a two-row pattern of unvoiced/voiced for the consonants.  It only provides combining keystrokes for the regular two ligatures, and does not rely on long-press or combining keystrokes otherwise.  (The price is losing quick access to the number row, but this is not commonly needed in most transcription work.)

```
𐐎𐐑𐐓𐐕𐐗𐐙𐐛𐐢𐐤𐐝𐐞-=
𐐒𐐔𐐖𐐘𐐚𐐜𐐡𐐥𐐟𐐠[]\
𐐀𐐁𐐂𐐃𐐄𐐅𐐆𐐇𐐈;'
𐐉𐐊𐐋𐐌𐐍𐐐𐐣,.𐐏
```

- [SIL Forum discussion](https://community.software.sil.org/t/can-someone-make-the-deseret-keyboard-work-on-mobile/6738/7)
- [Illinois Deseret Consortium](http://faculty.las.illinois.edu/rshosted/deseret.html)


keyman-server:

DavidLRowe:
@sigilante Thanks for your submission. We plan to resume reviews in January.

LornaSIL:
Thanks for your submission! I have a few comments. I haven't done a thorough test, but it would be good to resolve these issues first before we do a more thorough review.

You have put the keyboard in a folder called `deseret_ipc`. That should probably be `idc`. However, the files within that folder are `idc_deseret` which makes more sense. That means the folder name should be changed to `idc_deseret` and it should be placed in the `i` folder, not the `d` folder. All of this is part of consistency for our build system.

Next, the .kpj files references two different .kmn files. Only one of them is part of this pull request. The one you have included, does not have a header statement for the `.kvks` file which is what is used for compiling into Web format `.kvk`. So, I'm not sure if you have committed the correct .kmn file. Maybe you should double-check. Otherwise, if the .kmn file is correct, maybe you should just add the header statement:
`store(&VISUALKEYBOARD) 'idc_deseret.kvks'`

Also, your Pull Request includes `idc_deseret.keyman-touch-layout` . However, that file looks to be for a Latin keyboard and not a Deseret keyboard. You'll either need to enhance the touch layout to be for Deseret or remove that file...and in that case the "targets" in the .kmn should not be for "any" it should be for "desktop web". If you do want a touch layout, you also need to add a header statement for that file: `store(&LAYOUTFILE) 'idc_deseret.keyman-touch-layout'`

In the .kps file you have the language tag of `en-Dsrt` which I believe is correct. In the `keyboard_info` file you've included 4 language tags. If Hopi uses/used Deseret, it is okay to include it, but you should also add that tag to the .kps. However, pllease delete the `en` and `hop` language tags as those indicate support for the Latin script, not Deseret.

Also in the `keyboard_info` file you should delete the final comma. In the end, I believe this is what your file should include:
```
{
    "license": "mit",
    "languages": [
        "en-Dsrt",
        "hop-Dsrt"
    ],
	  "description": "Keyboard for the 38-letter Deseret alphabet",
      "authorName": "N. E. Davis"
}
```

There are other issues, but I'd like to see these resolved before we review it further.

sigilante:
Thanks for the check.

1. `/d/deseret_ipc` → `/i/idc_deseret`
2. Investigating the `kvks` situation, watch for another commit on that.
3. Investigating the touch layout situation, watch for another commit.
4. Removed `en` and `hop`; retaining `en-Dsrt` and `hop-Dsrt`.

sigilante:
OK, those precursor issues are I think resolved.  (I'm using Wine access to Keyman which is not particularly stable so I do mistrust some of the things it tells me.)  But I think the sources are now consistent.

I do want to roll out a mobile/touch keyboard but I'm waiting on that until this is correct.

LornaSIL:
This looks pretty good. A few more requests.
- When I open the .kpj file and go to the Keyboards tab, I see two .kmn files listed:
  - image
   - Could you please click on the SECOND one and select "Remove from project"
   - image
- Then open the other .kmn file (the correct one) and if you could add this header statement it would reduce the notifications when we compile:
 `store(&VERSION) '10.0'`
- Then open the .kps file. You need to add the `LayoutU_.png` file to the list of files so that when the welcome.htm file is opened the graphic will be available.
- In the `welcome.htm` file you could remove this paragraph: IDC Deseret 1.0 generated from template. 
- In the `readme.htm` you could remove the version: Version 1.0 (that way you don't have to update that file when you update the keyboard.)
- Lastly, for any keyboard going into the `release` folder we require an online help file. You need to create a `help` folder in the `source` folder. Then, put a file called `idc_deseret.php` in that `help` folder. I have created the file for you, so you just need to add it to the pull request. (The keyboard layouts will get generated so you don't need the graphic in the help folder.)
[help.zip](https://github.com/keymanapp/keyboards/files/13835564/help.zip)

LornaSIL:
Since we haven't heard back from you, I will try to make these changes myself and submit the help file in a separate PR.

LornaSIL (approved):
With these fixes it should be fine. I'll do a separate PR for the online help.

LornaSIL:
We did go ahead and approve this keyboard and it is now online. https://keyman.com/keyboards/idc_deseret I realized after I approved it that the RALT on-screen keyboard doesn't display any of the letters, so next time you update the keyboard you should try to fix that infilicity as no one will know what to type.

sigilante:
Thank you.  I will look at that next time I can block out time on this
project, hopefully within a couple of weeks.

PR #2518, opened 2024-01-04
Add PHP header to help file

DavidLRowe:


DavidLRowe:
It'll be another day before this can be reviewed and online. 

Note: Because this change is just to the online help, no change to the version number is needed. If you end up revising the keyboard, you'll need to update the version number (in the .kmn file, plus in the HISTORY.md file). That way the updated keyboard will be offered to users.

DavidLRowe:
I removed the `<html>` and `</html>` tags and also four spaces of indent from the whole file. If you hide whitespace changes, you'll see that the changes are in the PHP header and some of the head element data.

@anvalon It'll be another day before this change gets reviewed and up on the Keyman help site. Sorry I didn't catch it earlier.

anvalon:
@DavidLRowe Ok, thanks.

LornaSIL (approved):
LGTM

PR #2522, opened 2024-01-22
[yezidi] Adjust files for author'

LornaSIL:
Updated files for author.
- Put Yezidi fonts in shared folder and adjusted .kps to reference those fonts
- Changed language tag from `und` to kmr and `ku`
- Updated and added some documentation files
-


LornaSIL:
Ready for review @DavidLRowe

DavidLRowe (approved):

LornaSIL:
@Meng-Heng the keyboard is now available here: https://keyman.com/keyboards/yezidi
Since there were changes to the files, if the requester (or you) wants to update the keyboard, make sure that they start with the files in the repo and not the version that I was sent.

PR #2523, opened 2024-01-23
[sil_tagdal] Initial creation

darcywong00:
Initial keyboard submission on behalf of a user who requested a desktop and mobile Tagdal keyboard using Latin script (tda-Latn).

The pre-existing Tagdal keyboard uses a different script.
https://keyman.com/keyboards/basic_kbdtifi2

The user also requested a few IPA characters for transcription purposes (most accessed on right-alt).

The mobile keyboard is primarily intended to use alongside an external keyboard.

I'll start this in draft and see if the user has more testing feedback.


darcywong00:
User said the keyboard is working fine.

DavidLRowe:
Just a very minor observation that the .png images seem to have a mix of fonts (serif font "C" vs non-serif font "Ɣ", for example). I'm happy to approve the keyboard as is, but thought I'd point it out in case you want to make a change.

"{" and "}" on longpress keys on numeric layer of touch layout give different output (U+014A and U+030C) since they're picking up the .kmn rules. (There may be other keys that do that; I didn't exhaustively test the keyboard.)

I'm not seeing how to get to the rightalt layer of touch layout. Is it a longpress on the shift key? That doesn't seem to work in the browser tab opened from Developer's "Open in browser" button. That is, I see a dot on that key, but longpress doesn't open a menu. Is it working for you when tested on a real device? Or maybe I need to try a different browser?

darcywong00:
Thanks for the feedback

>  the .png images seem to have a mix of fonts (serif font "C" vs non-serif font "Ɣ", for example).

I originally used png images as locally rendered by help.keyman.com. I've updated the desktop images to use the layouts from Keyman Developer.

> "{" and "}" on longpress keys on numeric layer of touch layout give different output (U+014A and U+030C) since they're picking up the .kmn rules. (There may be other keys that do that; I didn't exhaustively test the keyboard.)

Good catch! I've updated some of the numeric keys to use default Unicode codepoints, so they should work now

> I'm not seeing how to get to the rightalt layer of touch layout. Is it a longpress on the shift key? That doesn't seem to work in the browser tab opened from Developer's "Open in browser" button. That is, I see a dot on that key, but longpress doesn't open a menu. Is it working for you when tested on a real device? Or maybe I need to try a different browser?

When debugging from Developer, I had to set the "Device" dropdown to one of the mobile devices (phone / tablet). Then, the OSK shows a longpress option on the default Shift key to get to Right-Alt.

image

DavidLRowe (approved):

PR #2524, opened 2024-01-24
Soumission Premiere, clavier Ife

MattGyverLee:
This is a first submission of the Ife Keyboard, submitted to experimental without documentation.


LornaSIL:
Thanks for your submission!

We don't really like having the word "Unicode" in a keyboard name. I'd rather it was just "Ifè" or "Ifè (SIL)". Since we don't accept non-Unicode keyboards anymore it is rather redundant. Just like I don't like the word "keyboard" in the name.

Otherwise, there is a small typo in the .kmn which causes the keyboard not to compile. The first line `store(&BITMAP) 'qtbb6u1_ife.ico'` has a spurious `q` in it so it's not finding the icon file.

You'll also need to add a version to the keyboard, and while you are at it can you also add the minimum keyman version. So, these 3 lines could be added or fixed:
```
store(&VERSION) '10.0'
store(&KEYBOARDVERSION) '1.0'
store(&BITMAP) 'tbb6u1_ife.ico'
```

The filename/foldername is fine with our build system. However, it does seem like a strange name to me. I assume it was based on the tbb keyboard at some point. We don't like to change filename/folder names (we **won't** change them in `release` folder, but we might consider it in `experimental` so make sure you are using the foldername/filenames you want so that the URL remains stable for this keyboard.

DavidLRowe:
No need to keep "tbb6u1" in the name. I'd suggest you go with "ife" as the keyboard name.

IIRC this was originally designed for US English keyboards and the &mnemoniclayout was added later, but it might make more sense to have a second layout tailored for a French keyboard (especially if touch layout will be added at some point). The original goal was to be able to type any character-diacritic combination with just two keystrokes. In addition, the existing keyboarding usage put the diacritics before the base character. So, for example, ɛ̃̀ is typed with "~" (for the two diacritics) and "[" for the epsilon.

MattGyverLee:
> store(&VERSION) '10.0'
> store(&KEYBOARDVERSION) '1.0'
> store(&BITMAP) 'tbb6u1_ife.ico'

These comments make sense. I'll work on renaming.

I'm a bit surprised that you want me to add &VERSION. I like leaving it open. If I omit it, it generates 9.0 for Windows and 10.0 for web based on the features I use. If I add it, both are forced to 10. This allows me to know what versions are the minimum,

I got this working (with Eric's suggestion of using the Togo flag icon). I have also generated a Positional AZERTY layout that works for this keyboard (even if some of the deadkeys are currently awkward. Eric wanted the working Qwerty for now, and said the translators would be "troubled" to start using an AZERTY one now. *sigh* This is the sad history of keyboards.

Based on Eric's feedback, I will leave this one as sil_ife_mn (mnemonic) and "reimagine" sil_ife_qw and sil_ife_az as "efficient" letter-diacritic keyboards, which I prefer. Then they'll have the option.  I have a 3-step process for positionalizing a keyboard.

LornaSIL:
@MattGyverLee Were you going to do an update or were you waiting on us? The `store(&VERSION) '10.0'` change isn't required. It just quiets the messages. However, it really would be good to get people off Keyman 9.0!

keyman-server:

keyman-server:

LornaSIL:
Thanks @MattGyverLee . It mostly looks good. Since it's experimental we don't require more description, but if you want to add a little description it might help users know why they would want the keyboard. (README.md, keyboard_info, readme.htm, and/or welcome.htm.

The .kmn can list just desktop and web. It doesn't need the specific operating systems listed.

The .kps file could have the same title as the .kmn if you want: **Ifè Mnemonic** (instead of the filename).

The above things aren't required changes, but this one is: The .kps file should *not* include the .js file. That is just needed for touch layouts which this keyboard does not currently include.

keyman-server:

LornaSIL (approved):
Looks good. I was able to remove the two files that shouldn't be in the .kps file.

PR #2532, opened 2024-02-05
[FV plains cree] New desktop keyboard, updated mobile keyboard

HopsAndHops:
Created desktop keyboard, updated mobile keyboard, updated documentation to reflect changes.


keyman-server:

LornaSIL:
Thanks for this PR. A few comments/requests:
- Could you delete the file `.DS_Store` in the root of the keyboard?
- I don't think `keyman.png` in the `welcome` folder is required. It would be good to delete that and remove it from the list of files in the kps.
- In the .kmn header statement you have selected "any", can you remove "mobile" and "tablet" as those are redundant.
- These characters don't have rules `qudfgjzvbQWERTYUIOPASDFGHJKLZXCVBNM` and so the Latin is output. We don't much like to mix scripts in keyboards. If you could write a rule to not output the Latin characters that would be appreciate. Something like this would accomplish that:

```
store(nul) "qudfgjzvbQWERTYUIOPASDFGHJKLZXCVBNM"
+ any(nul) > nul
```

Otherwise, I think it looks good.

HopsAndHops:
Thanks for the comment, Lorna! I'll make these changes, however I realized I pushed the wrong Cree keyboard. I'll close this PR and create the correct one.

PR #2554, opened 2024-02-14
[rawang] Optimize touch layout

### LINKED ISSUE #2551: [rawang] Optimizations for touch layout
Follow-on to #1338

User requested the following optimizations for the Rawang touch keyboard

## Default layer ##
* Move punctuation to longpresses on the period key (move to the right of spacebar)
* Keep numerics on top row
* Add long presses to vowels (a, e, i, o, u, ø) and v
* For the long press hint, use the character with high tone
* Next to <kbd>m</kbd>, add <kbd>:</kbd> key then Backspace

## Shift layer ##
* Uppercase of default layer, each key returns to default layer

No need for numeric or symbolic layer

darcywong00:
Fixes #2551 by optimizing the touch layout:
* Adds high, low, middle tone longpresses to vowels (a, e, i, o, e, ø) and "v"
* Add numerals as longpresses to the top row
* Add numeric layer
* Add start of sentence detection

I didn't use special characters for the hints.

~PR in draft while user tests it out~


darcywong00:
Some user feedback for me to address...

> Let's remove the colon from the main keyboard layout and keep only the one that's accessed via long-pressing the period key. That would create some extra space in that row, right? If so, would it work to widen the shift and backspace keys?
>
> And I forget if it's possible to make it so that hitting shift twice in rapid succession will make it caps lock...

darcywong00:
Additional round of feedback from user testing

He likes the spacing of the numeric layer (only has 4 rows instead of 5), so
* Move the top row of numbers and punctuation as long presses on the QWERTY row (far right of the options)

* for the <kbd>.</kbd> longpress options, can remove <kbd>,</kbd> and maybe add <kbd>(</kbd> and <kbd>)</kbd>

darcywong00:
User said this is ready to review now

LornaSIL (approved):
LGTM

PR #2556, opened 2024-02-14
moved hcesar to release

FranciscoMatos1:


keyman-server:

LornaSIL:
This PR looks like it's still in draft mode AND the files all look like they are brand new, so I don't think I'd want to approve it with the extra things I'm seeing like creating "e" and "fggh", etc. Do you want to fix it up so it's a clean PR or should we just wait for moving to release at a later time? There's no real difference in leaving it where it is, the url will remain the same.

PR #2557, opened 2024-02-14
[starling_sxava_eo] Initial submission of the keyboard

tomasbm01:


keyman-server:

LornaSIL:
I haven't checked the keyboard yet, but we don't like fonts without a clear and open source license. So, since you aren't actually using the `esl_gothic_unicode_1.ttf` font, I'd like that removed.

Then, we do prefer to use fonts in the `shared` folder when possible. **Inter Alia** and **Ormin** are available here: `keyboards\release\shared\fonts\shaw\ so if you could reference the fonts there and not a local version, that would be great. We try to keep those fonts updated to the newest versions. When a font is in a keyboard folder it's harder to keep the versions of the fonts updated.

keyman-server:

tomasbm01:
Thank you, Lorna. I just removed the font ESL_Gothic_Unicode. I really like the idea about the online fonts, but I don't know how to reference them, sorry. This is my first keyboard and my first project in GitHub, and I don't know how to use this platform.

keyman-server:

LornaSIL:
- The font file `esl_gothic_unicode_1.ttf` still needs removing from the Pull Request. You did remove the reference to it from the .kps file.
- There are two keyboards in this folder. Our build system requires one keyboard per folder.
   - That means you will need to remove the files related to `starling_sxava_eo_tel` and put them in a separate folder called `starling_sxava_eo_tel`. 
   - Then you'll need to create a .kpj and .kps related to that keyboard. 
   - Remove all references to `starling_sxava_eo_tel` in the `starling_sxava_eo` keyboard.
   - You'll need to adjust the welcome.htm (and maybe readme.htm) to just reference the specific keyboard.
- Delete `defaultpackage.gif` from the root of `starling_sxava_eo`
- `HISTORY.md` has a 2023 year. I think you meant 2024
- `README.md` - if you remove the version number you won't need to update that file each time you update the keyboard.
- `starling_sxava_eo.keyboard_info` needs the language `eo-Shaw` added.
- `starling_sxava_eo.kps` needs the language tag to be `eo-Shaw`
- The `help` folder seems to be a duplicate of what's in the `welcome` folder.
   - You'll need to renamed `welcome.php` to `starling_sxava_eo.php` and fix up the header.  (Remove `<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">`
   - You can delete `readme.md`
   - It feels like there are a lot of files in the `help` and `welcome` folder that aren't actually required. Can you double-check that?

- Lastly, I have a lot of concerns about this keyboard and pull request. The biggest issue is that you are trying to support both the Latin script and the Shavian script in the same keyboard. We don't allow that because keyboards are tied to a language tag it just won't work properly for the users. You should create a keyboard that just supports Shavian (langtag would be `eo-Shaw`) and a separate keyboard that just supports Latin (langtag would be `eo`). (I even see Greek in the keyboard and I'm not sure how that is being used, but _usually_ (not always) a Latin orthography shouldn't be using characters from the Greek block.)

Since there are so many issues here, I'm going to close the Pull Request. I think it will be easier for you to adjust the keyboards based on my feedback and then resubmit as a separate Pull Request. I suppose it might turn into 4 separate keyboards.

PR #2559, opened 2024-02-15
[yo] Yoruba keyboard

LornaSIL:
On behalf of W3techy


LornaSIL:
If this works it replaces #2449.

DavidLRowe (approved):

PR #2564, opened 2024-02-20
[bj_cree_east] Fix bug related to previous change to NFD

LornaSIL:
This was the report

> this is what she types: waau, twaau, chwaau, swaau
> this is what she gets: ᐙᐅ, ᑣᐅ, ᒞᐅ, ᔂᐅ
> She does NOT want that… and she does not want this either: ᐙᐧ, ᑣᐧ, ᒞᐧ, ᔂᐧ
> 
> But this is what she wants: ᐙᐤ, ᑣᐤ, ᒞᐤ, ᔂᐤ
> 
> (the final u typed should be a U+1424, not a U+1405 or U+1427)

I discovered the same issue arose with wiiu and wuuu. So, I wrote rules to fix that issue. It's related to when I updated the keyboard to support NFC by default and NFD with an option key.

I have asked if the problem exists elsewhere.


LornaSIL:
I realized the same bug existed in the bj_oji_cree keyboard as well since I updated them to use NFC at the same time.

DavidLRowe:
I see that bj_cree_east.kps has
```
<Version>9.1.1</Version>
```
whereas the updated keyboard version is 9.2.0. I'm assuming the `<FollowKeyboardVersion/>` will override what's in the .kps file with the 9.2.0 from the .kmn file. If not, you can dismiss my review and change it.

DavidLRowe (approved):
LGTM (see comment)

LornaSIL:
Technically it's supposed to be fine, so I will go ahead and merge and double-check :)

PR #2567, opened 2024-02-23
[manga][manga_ajami] Add keyboards for author

DavidLRowe:
Eric Fields provided these keyboards for review. I've put them in this pull request. If he decides not to submit them, we can close the PR without merging it.

Both keyboards are for the same language, Manga (kby). manga_ajami (kby) is Arabic script. manga (kby-Latn) is Latin script.

Issues:
- Neither keyboard has a touch layout (for use on mobile phones), other than the default US English one that Keyman Developer provides. I'll check with the author about how to proceed.
- There is no documentation (that is, the readme.htm and welcome.htm files need to be filled in, and the corresponding PHP help files created). This is temorarily acceptable for the experimental section of the keyboards repository.
- For the Latin keyboard, we should think about normalization. For example, `a` followed by `RALT+/` produces `á` (U+0061 U+0301) and `RALT+q` produces `á` (U+00E1). Although the two forms are equivalent, one is stored as two characters and the other as a single character. This would show up when you try to backspace to delete the character. The first form needs two backspaces; the second only one. The choice of normalization form should probably match whatever NFD/NFC normalization setting is used in Paratext.

Additional question: The icon for the Manga Latin keyboard is the flag of Niger, but I'm curious about the icon used for the Manga Ajami keyboard. Is it a flag of some sort?


LornaSIL:
Replaced by #3158

PR #2587, opened 2024-03-04
[sil_ipa] Add subscript numerals to `_` rota

### LINKED ISSUE #2401: [sil_ipa] request to add subscript numerals
We know how to type superscript numerals, but a user wants to be able to type subscript numerals into FLEx.  Bu_phonetic seems to have this feature.

darcywong00:
Fixes #2401

This adds subscript numerals to the underscore <kbd>_</kbd> rota

e.g.

| Character | Sequence |
|-----------|-------------|
|  ₃             | 3_              |
|  3̠            | 3__             |
|  3̞            | 3___           |
|  3̙            | 3_____        |
| 3̧             | 3______      |
| 3̠             | 3_______     | 

(update)
When options are set to 'before', the sequence
| Character | Sequence |
|------------|------------| 
| ₃               | _3             |

Rota still uses "_" after.


darcywong00:
> So, it looks like you intentionally didn't try to make it work differently when the `options` are changed.

At first I was going to follow the 'before' and '' options, but then it seemed odd for a rota.
I can try going back to options

darcywong00:
OK, I've updated the options rules to convert subscript numerals with <kbd>U+0320</kbd> since underscore <kbd>_</kbd> automatically converts to combining minus sign below U+0320.

For 'before' options, `_3` outputs `₃`; and type `3_` to get to the rota `3̠`.

For 'after' options, `3_` outputs `₃`, `₃_` gets to the rota `3̠`

LornaSIL (commented):
I don't think we want the `sil_ipa.php` file in the `welcome` folder. The one in the `help` is the correct location.

LornaSIL (approved):
LGTM

PR #2592, opened 2024-03-05
[rac_dogri] chore: Simplify pathing in rac_dogri.kps

darcywong00:
I was trying out the kmc compiler on staging-17.0 branch and had to replace a lot of slashes in .kps files to build on Linux.
That revealed the pathing in rac_dogri.kps can be simplified.

Not bumping versions since no rebuild necessary.

### DIFF
--- a/release/rac/rac_dogri/source/rac_dogri.kps
+++ b/release/rac/rac_dogri/source/rac_dogri.kps
@@ -37,19 +37,19 @@
       <FileType>.htm</FileType>
     </File>
     <File>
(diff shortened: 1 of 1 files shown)


mcdurdin:
If it builds with kmcomp then it should also build with kmc. Can you open
an issue?

mcdurdin (approved):
Yes, this is a good cleanup. LGTM

mcdurdin:
More generally, path changing issue tracked at keymanapp/keyman#10942. Let's avoid changing keyboard sources for it as it needs to be fixed in kmc

Issue #2213, opened 2023-05-17
[sil_ipa] no charts for mobile keyboard


mcdurdin:
Today I assisted a user trying to find ˀ on their mobile keyboard -- and it took us some time to find it under <kbd>q</kbd>; it makes sense but wasn't obvious (no separate question mark key which was what we looked for first). I checked the help docs and realised that we have no documentation on the mappings, so this might be something to consider.

Note: I want to improve the mobile keyboard automatic documentation to have it generate charts of all long-press keys which would simplify this issue sigificantly, so treat this as low-priority, just documenting the issue.

darcywong00:
Documenting sil_ipa 1.8.7 gesture chart below (does not include gesture to be added in #2521

---
title: sil_ipa Gesture Table
---

Gestures for the tablet Keyboard layers are described in the tables below:

### default Layer
#### Row 1
|Base Key | Longpress |
|---------|-----------|
|   q   |   ʔ    ʕ    ʡ    ʢ    ˀ    ˤ   |
|   w   |   ɰ    ʍ    ʷ   |
|   e   |   ɘ    ə    ɛ    ɜ    œ    ɶ   |
|   r   |   ɹ    ɻ    ɽ    ɾ    ʀ    ʁ   |
|   t   |   ʈ    θ   |
|   y   |   ʏ    ʸ   |
|   u   |   ɯ    ʉ    ʊ    ʌ   |
|   i   |   ɨ    ɪ   |
|   o   |   ø    ɒ    ɔ    ɞ    ɤ    ɵ   |
|   p   |   ʘ    ǀ    ǁ    ǂ    ǃ   |
#### Row 2
|Base Key | Longpress |
|---------|-----------|
|   a   |   æ    ɐ    ɑ   |
|   s   |   ʂ    ʃ   |
|   d   |   ð    ɖ    ɗ   |
|   f   |   ɸ   |
|   g   |   ɠ    ɡ    ɢ    ɣ    ʛ    ˠ   |
|   h   |   ħ    ɥ    ɦ    ɧ    ʜ    ʰ   |
|   j   |   ɟ    ʄ    ʝ    ʲ   |
|   k   |  |
|   l   |   ɫ    ɬ    ɭ    ɮ    ɺ    ʎ    ʟ    ˡ   |
#### Row 3
|Base Key | Longpress |
|---------|-----------|
|   `*Shift*`   |  |
|   z   |   ʐ    ʑ    ʒ   |
[...]

PR #2594, opened 2024-03-07
feat: Add tool to generate gesture table

darcywong00:
Addresses #2213

This adds a gestureChart tool to document gestures in a touch layout as a Markdown table. Future TODO is incorporating the Markdown table into the welcome.htm / .php help docs. If we sync the .md file along with the .php help, the site could render it as
image

As of 17.0, gestures can be:
* longpress
* flicks
* multitap

## Usage
```
node gestureChart.js -k <keyboard>
    -k, --keyboard <full path to keyboard directory>
```

Sample outputs:
Current sil_ipa gestures (on master)
[sil_ipa-tablet (current).md](https://github.com/keymanapp/keyboards/files/14517501/sil_ipa-tablet.current.md)

sil_ipa incorporating gestures from #2521
[sil_ipa-tablet.md](https://github.com/keymanapp/keyboards/files/14518490/sil_ipa-tablet.md)

rawang from #2554
[rawang-tablet.md](https://github.com/keymanapp/keyboards/files/14518469/rawang-tablet.md)

### DIFF
--- a/tools/gestureChart/gestureChart.ts
+++ b/tools/gestureChart/gestureChart.ts
@@ -81,8 +81,8 @@ function processKeyboard(dir: string) {
     l => l.row.some(r => r.key.some(k =>
       k.sk || k.flick || k.multitap)));
   if (filteredKeys && filteredKeys.length > 0) {
-    let gestureTable = `## ${name} Gesture Table\n`;
-    gestureTable += generateTable(formFactor, filteredKeys);
+    let gestureTable = `${newTitle(name)}${newPreamble(formFactor)}`;
+    gestureTable += generateTable(filteredKeys);
 
     // Write Gesture Table to file
     if (gestureTable) {
@@ -97,12 +97,11 @@ function processKeyboard(dir: string) {
 
 /**
  * Create Markdown table
- * @param formFactor phone or tablet
  * @param filteredKeys Filtered touch layout object containing all the gestures
  * @returns Markdown table as string
  */
-function generateTable(formFactor: formFactorType, filteredKeys) : string {
-  let table = `### ${formFactor} Layout\n`;
+function generateTable(filteredKeys) : string {
+  let table = '';
   for (let l in filteredKeys) {
     let layer = filteredKeys[l];
     let layerID = layer.id;
@@ -152,6 +151,16 @@ function generateTable(formFactor: formFactorType, filteredKeys) : string {
   }
 
   return table;
+}
+
+function newTitle(name: string): string {
+  return `---\n` +
+         `title: ${name} Gesture Table\n` +
+         `---\n\n`;
+}
+
+function newPreamble(formFactor: formFactorType) : string {
+  return `Gestures for the ${formFactor} Keyboard layers are described in the tables below:\n\n`;
 }
 
 function newTableHeader(): string {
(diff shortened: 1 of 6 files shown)


tools/gestureChart/gestureChart.ts
```diff
@@ -0,0 +1,208 @@
...
+    codes = id.substring(2).split('_');
+    for(let c in codes) {
+      text += `&#x${codes[c]};`;
+    }
+  } else if (id.startsWith("K_")) {
+    // Parse K_ID
+    text = id.substring(2);
+  } else {
+    text = `[${id}]`;
+    console.warn(`id ${id} doesn't start with U_`);
```
darcywong00, line 205: Looking up text from a [T_XXXX] key in the .js keyboard file is an extra complexity

tools/gestureChart/gestureChart.ts
```diff
@@ -0,0 +1,217 @@
...
+ * Extra formatting for special Markdown characters
+ * @param key
+ * @returns string
+ */
+function getTextFromKey(key) : string {
+  let text = "";
+  if (key.text) {
+    if (key.text.startsWith('|')) {
+      // Handle vertical line character
+      text += key.text.replaceAll('|', '&#124;');
```
mcdurdin, line 182: This will not work with `x|` or things like that. Nor `|*`. We need to do replacements on the whole string. There's actually a `escapeMarkdownChar` function in Developer which should do what is needed:

https://github.com/keymanapp/keyman/blob/5062bbad984760a40e5ee3d012b65d54d805db1d/developer/src/common/web/utils/src/markdown.ts#L2-L21

tools/gestureChart/gestureChart.ts
mcdurdin: It's also not needed -- we don't support inferring text for `T_` keys in the touch layout :grin: -- so this is the correct approach I think

mcdurdin (changes requested):
This is a useful tool. However, we don't want to start putting tooling into the keyboards repository, because this is really a Developer tool.

So, we need to move the tool to keymanapp/keyman, and ideally we will integrate this into kmc as a kmc module. Then we could merge it in 18.0.

If that sounds like too much work for now, then feel free to put it into its own repo -- just we shouldn't keep it in keyboards repo.

One other challenge is that while this tool generates Markdown, we don't currently support Markdown in keyboard documentation welcome.htm, which is where this content belongs (not just on help.keyman.com).

The whole Markdown documentation proposal can be followed in https://github.com/keymanapp/keyman/issues/9477

darcywong00:
> This is a useful tool. However, we don't want to start putting tooling into the keyboards repository, because this is really a Developer tool.
> 
> So, we need to move the tool to keymanapp/keyman, and ideally we will integrate this into kmc as a kmc module. Then we could merge it in 18.0.
> 
> If that sounds like too much work for now, then feel free to put it into its own repo -- just we shouldn't keep it in keyboards repo.

Understood. For now, I'll move this to its own repo (so I can continue to use on an individual keyboard basis).
I'll then incorporate all these review comments as keymanapp/keyman#10965

darcywong00:
This is moved to a separate repo (keymanapp/gestureChart) and comments to be resolved in keymanapp/keyman#10965

mcdurdin:
Sounds great! I like this tool, and it's going to be helpful in our updated documentation approach in the future too

PR #2596, opened 2024-03-10
Add keyboard layouts from KreativeKorp

RebeccaRGB:


keyman-server:

keyman-server:

LornaSIL:
Thanks for the PR! We probably won't be able to check keyboards this week,
but hopefully soon after.

keyman-server:

keyman-server:

LornaSIL:
@RebeccaRGB Thanks for submitting these keyboards. It will be really good to get these into our repo. I'm in meetings this week and cannot do more review now, but I wanted to give you some initial feedback so that when I'm actually able to check keyboards again, you'll be able to address these.
- First of all, I find it is much easier (to the keyboard approver) to address each keyboard in a separate Pull Request. That way we can deal with any issues in one keyboard and learn from the process so we don't have the same issues in the next keyboard.
- If a keyboard has ANY characters that are either in the PUA or in the Unicode Pipeline (or roadmap) we ask that the keyboard be put in the `experimental` section rather than `release`. As soon as the characters are officially in Unicode we are happy to move it to `release`. 
- A lot of the keyboards use the decimal (`d69`) syntax which seems unusual in these days, but perhaps these keyboards are older and that's why that syntax is being used. You can leave it that way if it's not using "hacked" or PUA encoding.
- However, I did notice that at least one of the keyboards has messages in the Message window which you would probably want to address. For example,
`kreative_superlatin.kmn (220): Hint: 109A The rule will never be matched for key ' ' because its key code is never fired.` Since that rule, isn't being fired (and it relates to most of the deadkey rules) that keyboard could be improved by making sure to address those issues.

Hope this helps. Maybe submit one keyboard with these issues fixed up and we'll see if any other issues arise.

RebeccaRGB:
Is it OK to ask for help in a PR in this repository?

1) How do I put each keyboard in a separate PR? Every time I push to my repository it just adds to the existing PR.

2) The only one using PUA is the sitelen pona one. I'll make sure to move that to experimental then.

3) If the decimal syntax is disfavored, which syntax should I use? Especially for non-printable characters like zero width joiners and the like?

4) This message confuses me because I know for certain that all of the inputs to the dead key rules are being triggered by at least one key. What am I missing?

DavidLRowe:
Absolutely fine to ask questions here! (We may be slow to respond this week since many of us are involved in a conference.)
1. Because you created your PR off the 'master' branch, all changes to the master branch become part of the PR. Create a branch for each keyboard (you can even use the keyboard name as the name of the branch), then commit your files to that branch and create a PR for each one. Because you currently have changes to the master branch, any new branches you create from the master branch will have those changes as well, so that will pose a problem. Let us know if you need more information.
2. So instead of `release/kreative/...` you would target  `experimental/kreative/...`
3. Typically a keyboard will use either the actual character or the hexadecimal representation of the Unicode value. So U+0045 instead of d69. See https://help.keyman.com/developer/language/guide/strings for more details.
4. I haven't looked at the keyboard (as Lorna has), so I can't give you specifics. Typically the warning message gives you two line numbers which the compiler thinks have the same value on the left side of the rule. 

You'll save yourself some time if you take a single keyboard through the whole process, then apply what you learn to getting all the other keyboards submitted. 

Thanks again for your submission. Keep asking questions

LornaSIL:
Hi again. The keyboard I looking at was `kreative_superlatin`. The rules it is having trouble with are because the store for input uses `d160` (U+00A0), but there is no way to access that key on a standard keyboard. So, I think it's just giving the warning for that one character. There are a couple of ways you could deal with this. If you really need 00A0 then change the first store to reference the keystroke. So, this would compile without a warning message:
```
store(U02CBin) d65 d97 d69 d101 d73 d105 d78 d110 d79 d111 d85 d117 d87 d119 d89 d121 [SHIFT RALT K_SPACE] d32
store(U02CBout) d192 d224 d200 d232 d204 d236 d504 d505 d210 d242 d217 d249 d7808 d7809 d7922 d7923 d768 d715
dk(U02CB) + any(U02CBin) > index(U02CBout, 2)
```

The second way is to create another group. Just after this rule: `+ [K_SPACE] > d32` add:
```
match > use(normalise)

group(normalise)
```
Then remove the `+` in the following rules:
```
dk(U02C6) any(U02C6in) > index(U02C6out, 2)
```
You would have to test this since I'm not sure if your `U02C6in` store is actually output from the above rules or not.

ZWJ - U+200D
ZWNJ - U+200C
On the On-screen keyboard I sometimes put the text on the keycap. I just type it into the text field:
image

It doesn't overwrite any of the rules in the keyboard. You can see this keyboard has a lot of that on the shift layer: 
https://help.keyman.com/keyboard/kashmiri_phonetic/1.0.1/kashmiri_phonetic

You can do the same thing on the mobile layout. However, Keyman developer does not have a special code for ZWNJ. In the Keycap value, instead of text, you can select ` *ZWNJ*`
image
and it shows up like this:
image

I don't think there is an equivalent for `ZWJ`.

As for github, I'm not an expert, but I work on the commandline and use commands, such as
` git checkout -b kreative_superlatin`

do my work.

`git status` to see that no extraneous files have crept in.

`git add .` (if I want all the files committed)

`git commit -m "[kreative_superlatin] New keyboard submission"`

`git push origin kreative_superlatin`

Then I usually have to go to my Github repo and actually create the Pull Request so it goes to the Keyman/keyboards repo.

I hope this is helpful.

RebeccaRGB:
Thank you David and Lorne for answering my questions. This is all very helpful. I'm going to close this PR and resubmit the keyboards individually once I've fixed them up according to your advice.

PR #2619, opened 2024-03-21
[kreative_supercyrillic] New keyboard submission

RebeccaRGB:


keyman-server:

keyman-server:

LornaSIL:
The language tags for 
- `az` should be `az-Cyrl`
- `bs` should be `bs-Cyrl`
- `uz` should be `uz-Cyrl`
All of these reference the Latin script so they require the script code.
These should be changed in the .kps and in keyboard_info.

You have a tablet and a phone layout. I didn't look carefully, but they seem to be the same layout. When the layout is the same, we recommend just using tablet (or phone) so that only one layout needs to be maintained. (If there is only a tablet layout, it will also work on a phone and vice versa.)

Otherwise it looks good. The two language tags must be changed. The mobile layout doesn't have to have one deleted, but I would recommend it.

RebeccaRGB:
I get this message when using the code az-Cyrl:

kreative_supercyrillic.kps: Keyboard with ID 'kreative_supercyrillic' has a BCP 47 warning: 'az-Cyrl' is a valid tag but is not canonical: it should be 'az-RU'.

Should I change it to az-RU?

LornaSIL:
Oh, yes, that's good.

keyman-server:

LornaSIL (approved):
LGTM.

PR #2620, opened 2024-03-21
[kreative_superipa] New keyboard submission

RebeccaRGB:


keyman-server:

LornaSIL:
It looks pretty good.

For our IPA keyboards we have used `und-Latin` for the `.kps` and `und-fonipa` in `keyboard_info`. I think there's a good reason, but I can't remember why, maybe because of something in how Windows handles language tagging. Even that isn't ideal. Is that still our recommendation @mcdurdin ?

I will again make the comment that you have a tablet and phone layout that look the same. If that is true, it would be easier to maintain just one and delete the other one. 

Otherwise it looks good.

keyman-server:

mcdurdin:
> I think there's a good reason, but I can't remember why, maybe because of something in how Windows handles language tagging.

There's a bug in .NET framework relating to `und` language tag which causes any .NET application to instantly crash when that language is selected: https://github.com/keymanapp/keyman/issues/10727. Microsoft are aware of the issue but I am not sure how quickly it will be addressed. We may try and submit a pull request to them at some point, but there are challenges with doing that too.

`und-fonipa` won't even install in Windows: https://github.com/keymanapp/keyman/issues/10754.

> Is that still our recommendation @mcdurdin ?

So yes, the split setup is still our recommendation, I think, but beware of the above crash scenario. The alternative tags seem generally worse, perhaps `mul-Latn` is borderline better, although I haven't tested that thoroughly.

keyman-server:

LornaSIL (approved):
LGTM

PR #2625, opened 2024-03-22
[sil_tanguat] Initial creation

### LINKED ISSUE #2624: Tanguat: Design for new keyboard (based on Ukelele)
A Ukelele keyboard user emailed me requesting a keyboard to support the Tanguat language

> I have attached the .keylayout that I’ve set for our Tanguat keyboard that we use on our Macs... The Ethnologue code is TBS. And the people group (and one of the villages) name is Tanguat.
>
>It’s just based on the English keyboard and we re-mapped these four keys (and their capital versions).
>
>f = ə
>F = Ə
>c = ñ
>C = Ñ
>[ = f
>{ = F
>] = c
>} = C
>
>Again, I’m looking to have similar setup on iOS…then that would help us type better on our mobile devices, too. I was looking into the Keyman instructions but had more trouble figuring it out than I did with the Ukelele.
>
>If you can give me some pointers, that would be very helpful.

---

I'll skip attaching the .keylayout file since it should be straightforward to add the four keys. Additional changes:
* Use Right-Alt layer for <kbd>[</kbd>, <kbd>]</kbd>,<kbd>{</kbd>,<kbd>}</kbd>since they're rarely used
* mobile layout would just have the latin characters as long-press options

darcywong00:
Fixes #2624 

Initial creation of keyboard for Tanguat language based on Ukelele layout.

### key mapping
f = ə
F = Ə
c = ñ
C = Ñ
[ = f
{ = F
] = c
} = C

### Additional changes:
* brackets and braces accessible via rightalt and rightalt-shift layers
* add mobile layout with latin letters on longpress
* add basic start of sentence detection


DavidLRowe (approved):

darcywong00:
My bad - this needs a .keyboard_info file for the keyboard to deploy. (I was using Developer 17.0 beta 😄 )
Will rectify...

PR #2626, opened 2024-03-22
[sil_tanguat] Build for 16.0

darcywong00:
Follow-on to #2625 which failed to deploy because the .keyboard_info file was missing. Also because kmcomp in the CI build doesn't handle .kpj 2.0 format.

* Downgrade .kpj format to 1.0
* Add .keyboard_info file for keyboard deployment


LornaSIL (approved):
LGTM

PR #2639, opened 2024-03-26
[FV Southern Carrier]

HopsAndHops:
- New desktop layout
- Updated mobile layout
- Updated documentation to match addition and changes


keyman-server:

keyman-server:

LornaSIL:
I see you've changed the fonts included in the package, but rather than using the fonts in the `shared` folder it is looking in your `Downloads` folder. It was:
`<Name>..\..\..\shared\fonts\canadian_aboriginal\AboriginalSerifREGULAR.ttf</Name>`
and now it is `<Name>..\..\..\..\..\..\..\Downloads\AboriginalSans\AboriginalSansREGULAR.ttf</Name>`
Can you change that to use the `shared` fonts since that is what our build system requires?

In the `keyman-touch-layout` there's something strange going on. It is now using this syntax `"text": "\u140A",` rather than the glyph. Could you try to make a small change in the `keyman-touch-layout` (within Developer) and then save the file. See if the source changes back to using the glyph rather than this syntax `"text": "\u140A",`? I've seen this happen before, but I haven't figured out what causes it. It wasn't like that in your first commit, but now it is.

HopsAndHops:
Hi @LornaSIL ! Made those changes, it looks like the touch-layout should be alright now. Gotta love those fun bugs, that actually reminded me that I needed to report another one (sorry!)

keyman-server:

HopsAndHops:
Hi @LornaSIL , I just went and looked at the updated webpage and it looks like the PDF I included can't be downloaded from the site. I referenced the .php for fv_haisla and the download works there - do you have any idea why this might not be working for this keyboard?

I also just noticed that the link isn't working on Windows for either keyboard. I've tried repackaging with the pdf in the source, help, and welcome folders and I haven't been able to make it work. It works fine on Mac so I'm a bit lost on what might be going on here. Thanks for your help!

LornaSIL:
Look at the syntax for the one that works. I think you are using `file:` on
the one not working. You can just update the php file with no version bump
if that's all it is.

However, if the link isn't working from the welcome.htm file then when you
fix that, you _will_ need a version bump because that is modifying the
contents of the kmp keyboard package.

DavidLRowe:
As Lorna says, change `<a href="file:fv_SouthernCarrierGuide.pdf">` to `<a href="fv_SouthernCarrierGuide.pdf">` in the .php file.

A similar change is needed in the welcome.htm file. Changing that one (as Lorna said) will require incrementing the version number.

HopsAndHops:
ah, thank you both so much! I'll make those changes and make new PRs with the version bump.

PR #2641, opened 2024-03-26
[noto_fonts] Add Noto fonts for potential keyboard usage

LornaSIL:
Added a bunch of Noto fonts to the shared folder.


DavidLRowe (approved):
RSLGTM

PR #2644, opened 2024-03-27
Nguphingsecondsubmission

Cathaylaboratory:


LornaSIL:
Thanks for your submission! Is the "1" in the folder/filenames significant? If it's intended to be a versioning number, that should be removed. Versioning is handled in github and not by filename. We wouldn't ever want to change the keyboard `id` after it's been accepted.

keyman-server:

Cathaylaboratory:
> Thanks for your submission! Is the "1" in the folder/filenames significant? If it's intended to be a versioning number, that should be removed. Versioning is handled in github and not by filename. We wouldn't ever want to change the keyboard `id` after it's been accepted.

This number 1 after nguphing is a mark for my inner test on my own computer.I have once tried to change all nguphing1 in all files into nguphing but they can not be compiled after that, maybe because of the issue of the location of file or some quoting or other unexpectable reasons. So in the read me page and the name of the keyboard on the space icon, the name of the keyboard is still nguphing but inside the file the name is nguphing1. Is it a serious issue?

Cathaylaboratory:
> Thanks for your submission! Is the "1" in the folder/filenames significant? If it's intended to be a versioning number, that should be removed. Versioning is handled in github and not by filename. We wouldn't ever want to change the keyboard `id` after it's been accepted.

If you are convenient please change all nguphing1 into nguphing. Maybe this process will be labour intensive. If you are not convenient to do so I will try to change them by myself later.

Cathaylaboratory:
> Thanks for your submission! Is the "1" in the folder/filenames significant? If it's intended to be a versioning number, that should be removed. Versioning is handled in github and not by filename. We wouldn't ever want to change the keyboard `id` after it's been accepted.

I am sorry for that I can not overcome this issue and change my file names into nguphing. If I do that it will cause my file package can not work on my iphone. It makes changing the file name in mid- way almost impossible. Sometimes even build a new file then copy all the original codes into the new file then make it from scratch maybe can solve the issue. My recent submit taiwanese keyboard also meet this issue, I will submit the file with my inner test mark later because if I change the name my package can not work on iphone. so sorry about that

LornaSIL:
Replaced by #2646

PR #2645, opened 2024-03-27
[Fv cree latin] Updated README

HopsAndHops:


keyman-server:

LornaSIL (dismissed):
Technically you do not need to bump the version if all you are doing is updating the README.md file. This is because that particular file is not distributed in the keyboard package, it is just informational in the repo. However, I'm approving this PR.

LornaSIL:
You must have manually bumped the version in the `.kvks` file. That should still be version `10.0` as it seems to be the version of Keyman Developer and NOT the version of the keyboard. So, please change that back to `10.0`.

keyman-server:

PR #2651, opened 2024-03-29
[kreative_superpet] New keyboard submission

RebeccaRGB:


keyman-server:

LornaSIL:
@RebeccaRGB @mcdurdin I'm a bit concerned about this keyboard for several reasons. 

- We don't really like mixing scripts in a keyboard since computers need one or the other for a range of text. The main layers are Latin script and the languages listed appear to be Latin script. Then there's one code for `zxx-Zsym` and the RALT layers support all the box drawing and **Symbols for Legacy Computing**, etc.  Is this acceptable @mcdurdin or should the keyboard be specifically targetted at the Symbols? Or should the `zxx-Zsym` langtag be removed?
- The other issue is that some of the characters are still in the pipeline (for 16.0). Normally, I would say they should go in `experimental` except that I believe these codepoints are all stable now since 16.0 will go into beta very soon. (**Update**: I've just reviewed this again and the pipelined characters are apparently in the font only, not in the keyboard, so that point is probably moot). There is also one PUA character keystroke for U+F821. @mcdurdin opinions?

mcdurdin:
Thanks for the submission @RebeccaRGB! I think the keyboard is pretty much good to go. A few thoughts:

* It's not a language-related keyboard, as its purpose is to type symbols from legacy computing devices. So I would ask that we please remove all the language associations except zxx-Zsym -- even if it is possible to type other languages with this keyboard, it's outside the core purpose of the keyboard. 

* I would like the description in the .keyboard_info to be clearer that this is a symbol keyboard. Including "This keyboard is designed for typing symbols" as the first sentence, and the description from welcome.htm would probably be sufficient:

> Kreative SuperPET keyboard layout is based on the PETSCII character set and keyboard from the Commodore PET and its descendants, including David Murray's Commander X16. SuperPET allows you to type the equivalent Unicode characters using a similar mapping.

* Given the keyboard's purpose, and the documented use of the PUA character, I don't think that that is an issue to have the PUA character included.

RebeccaRGB:
Thanks for the feedback. I've made the requested changes to the language codes and description.

keyman-server:

LornaSIL (approved):
Thanks! LGTM.

PR #2659, opened 2024-04-05
Update taigi_telex.kvks

Tranhaiming:


keyman-server:

LornaSIL:
We cannot take a commit for just one file in a keyboard. They all need to be in one commit. We need resolution on the differences between `taigi_telex` and `taigi_telex2` before we make more progress.

PR #2689, opened 2024-04-11
[noto_fonts] Add more Noto fonts for unsupported keyboards

LornaSIL:
Adding Noto fonts for scripts where keyboards are being submitted.


DavidLRowe (approved):
RSLGTM

PR #2697, opened 2024-04-14
[ausephon] updated keyboardAdd files via upload

benaustralia:
[ausephon] updated keyboard


keyman-server:

LornaSIL:
Thanks for your keyboard update! We **do** want to keep using the "shared" fonts because that way we can control having the latest fonts in a keyboard package.

[Charisrelease.zip](https://github.com/keymanapp/keyboards/files/14981059/Charisrelease.zip)
 
- If you don't have the whole repo cloned, you can add this folder under `release/shared/fonts/sil/charis`
- Then make sure you grab the fonts from there. Since the fonts are already in that folder you do not need to submit the fonts.
- Please delete the fonts from your `ausephon` keyboard folder. This is probably a change we made subsequent to your original submission. We are trying to make sure we don't have to track different versions of the same font.
- It appears that the font information has disappeared in your .kps file on the **Keyboards** tab. This happens when you remove or change the fonts in the file section. So make sure you add the fonts back from the `shared` folder and then add the **OSKFont** and **DisplayFont** back in.

DavidLRowe:
To expand on what Lorna wrote (and get started on documenting this for the Keyman site):

You'll need to create a folder structure on your computer with:
- release/a/ausephon - where the "ausephon" folder contains your existing keyboard
- release/shared/fonts/sil/charis - where the "charis" folder contains the files for Charis SIL

The Charisrelease.zip file that Lorna attached already has the release/shared/fonts/sil/charis folder structure, so you could unzip that one first, then under the "release" folder, create an "a" folder, under which you can move your "ausephon" folder. 

The purpose of all this is to get the proper path when referencing these "shared" font files when you include them in your keyboard package.

In summary, you should 
- set up the folder structure with the "charis" folder 
- delete the fonts from your the ausephon/source folder 
- on the Packaging tab, remove the existing font files from the list of files, then add the font files from the "charis" folder under the "shared" folder (so that the correct path will be in the .kps file)

At this point you have two options. 
(a) You can update this pull request with the new files
(b) You can close this pull request and make a new PR using just the files in the "ausephon" folder just like you did to create this PR.

I'd suggest option (b). I would suggest using a different branch name than the one you used to create this PR. (And, in answer to your question on the Keyman forum, the branch name can be anything that GitHub accepts as a branch name.) So maybe ausephon11a?

benaustralia:
> Thanks for your keyboard update! We **do** want to keep using the "shared" fonts because that way we can control having the latest fonts in a keyboard package.
> 
> [Charisrelease.zip](https://github.com/keymanapp/keyboards/files/14981059/Charisrelease.zip)
> 
> * If you don't have the whole repo cloned, you can add this folder under `release/shared/fonts/sil/charis`
> * Then make sure you grab the fonts from there. Since the fonts are already in that folder you do not need to submit the fonts.
> * Please delete the fonts from your `ausephon` keyboard folder. This is probably a change we made subsequent to your original submission. We are trying to make sure we don't have to track different versions of the same font.
> * It appears that the font information has disappeared in your .kps file on the **Keyboards** tab. This happens when you remove or change the fonts in the file section. So make sure you add the fonts back from the `shared` folder and then add the **OSKFont** and **DisplayFont** back in.

Thanks Lorna for your response - much appreciated. I'd done what you suggested. 

Should I then edit my KPS file in a text editor and change the paths?

Keyman creates paths like this:
C:\Users\Ben\Desktop\release\a\ausephon\source\..\..\..\shared\fonts\sil\charis\CharisSIL-BoldItalic.ttf

Should they look more like this:

\release\shared\fonts\sil\charis\CharisSIL-BoldItalic.ttf

benaustralia:
Thanks David for offering me even more detail - for stepping into my new user world. I'll have a go at Option B!

DavidLRowe:
@benaustralia You probably won't need the following information if you create the shared folder structure and let Developer get the proper path when creating the package information. But to answer your question, it is possible to edit the paths in the .kps file directly (though this can be error-prone). You would end up with something like:
```
..\..\..\shared\fonts\sil\charis\CharisSIL-BoldItalic.ttf
```
where the `..\` indicates moving up one level in the folder hierarcy. So, starting in `release\a\ausephon\source`, the three `..\` bits would move you up three levels to the `release` folder, then `shared\fonts\sil\charis\` would take you down through those folders to where the font file is located.

benaustralia:
I see, so having followed the instructions above, Keyman app 'build' will package those links appropriately. Good to know how to do it manually though! Thanks again!
B

PR #2705, opened 2024-04-19
Add files via upload

benaustralia:
One more time!


benaustralia:
Thanks again Lorna and David for your patience. I hope I've got this right. If not, I think I'll read up on how to use the Git CLI because there is something about a cmd/terminal window which is more straight forward than a web interface.

keyman-server:

DavidLRowe:
We should really have the date range 2023-2024 in the copyright statement in the LICENSE.md file. Since that file is not packaged with the keyboard, I'll make a separate pull request to update that (which won't require updating the version number).

I wasn't looking for typos, but did notice a minor one that you can correct next time:
ausephon.pdf has apparent typo: 
"...symbol which is IS similar..." (remove first "is")
(cheatsheet.pdf has the same typo)

Congratulations and thanks for feedback on my draft documentation.

DavidLRowe (approved):

PR #2706, opened 2024-04-20
[ausephon] Update LICENSE.md

DavidLRowe:
The date in the LICENSE.md copyright statement should be the range 2023-2024 (after the update in PR #2705 ).

Since this file is not included in the keyboard package, there's no need to update any version number information.

### DIFF
--- a/release/a/ausephon/LICENSE.md
+++ b/release/a/ausephon/LICENSE.md
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-© 2023 Kym Barron & Ben Hinton
+© 2023-2024 Kym Barron & Ben Hinton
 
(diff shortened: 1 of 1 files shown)


benaustralia:
> The date in the LICENSE.md copyright statement should be the range 2023-2024 (after the update in PR #2705 ).
> 
> Since this file is not included in the keyboard package, there's no need to update any version number information.

Thanks for that change!

LornaSIL (approved):
LGTM

benaustralia:
Thanks Lorna. Must get hectic at keymanapp/keyboards project - literally no way to predict work load!

LornaSIL:
You are right. Thanks for the encouraging word!

PR #2707, opened 2024-04-21
Ausephon11c

benaustralia:
Hope this isn't troublesome - but just updated License.md with years 2023-2024 as suggested. Also fixed up my 'is IS' typo. Thanks for that. 

...and now wondering if 'is IS' might be a good name for a musical band. LOL.


keyman-server:

DavidLRowe:
I made the change to LICENSE.md in PR #2706 but it is awaiting approval before it can be merged.

The update you're making to the PDF file won't be offered to users unless you update the version number (to 11.1, for example). Keyman relies on the version number to decide whether a file is newer than the one it already has.

You could leave this PR as is (removing the LICENSE.md change since it will be done elsewhere) and recogise that the updated PDF won't be available until you next update the keyboard. BTW does ausephon.pdf need updating as well?

Another possiblity is to close this PR and include the updated PDF the next time you update the keyboard (and include a version change then).

Or you could add version number changes (HISTORY.md, ausephon.kmn) so that you create a new version of the keyboard that includes the updated PDF.

benaustralia:
> I made the change to LICENSE.md in PR #2706 but it is awaiting approval before it can be merged.
> 
> The update you're making to the PDF file won't be offered to users unless you update the version number (to 11.1, for example). Keyman relies on the version number to decide whether a file is newer than the one it already has.
> 
> You could leave this PR as is (removing the LICENSE.md change since it will be done elsewhere) and recogise that the updated PDF won't be available until you next update the keyboard. BTW does ausephon.pdf need updating as well?
> 
> Another possiblity is to close this PR and include the updated PDF the next time you update the keyboard (and include a version change then).
> 
> Or you could add version number changes (HISTORY.md, ausephon.kmn) so that you create a new version of the keyboard that includes the updated PDF.

Thanks again David! I think I'll let the update to HISTORY.md go through as you suggest - that way the new version is up!
And I'll close this request - making the PDF documentation updates next version.

For some strange reason I can't see ausephon.pdf in my file structure - but I can see cheatsheet.pdf. These are likely the same document. 


“Curiouser and curiouser!” - Alice in Wonderland

DavidLRowe:
I expect that you are correct in thinking the ausephon.pdf and cheatsheet.pdf are the same document (or at least different versions of the same document). I'm guessing that you renamed ausephon.pdf to cheatsheet.pdf at some point, so when you added the new version to the keyboard, the existing file was preserved.

When using the GitHub interface to update an existing keyboard, it's easy to add or replace files by dragging the new ones in. Deleting an existing file using the GitHub site needs action on the site to remove the file from the keyboard. 

I'd suggest that you make a note to delete the old ausephon.pdf file when you next update the keyboard. Thanks!

PR #2749, opened 2024-05-13
[obolo_chwerty] Remove extra obolo_chwerty folder and remove local font

LornaSIL:
On behalf of katelem24


DavidLRowe:
Should we wait until Keyman v17 is on the build system? Should the .kmn file require Keyman version 17?

mcdurdin:
> Should we wait until Keyman v17 is on the build system? 

If there is no rush on this, then I would wait.

> Should the .kmn file require Keyman version 17?

The kmn file should only require v17 if the compiled keyboard uses v17 features. The compiler normally can figure this out (and will tell you if it can't). And I just realised that this is not documented!

https://github.com/keymanapp/help.keyman.com/pull/1262 updates the documentation around the `&version` store. Better late than never?

LornaSIL:
This PR had too many conflicts that I didn't know how to resolve, so #2804 replaces it.

PR #2753, opened 2024-05-13
fv_sencoten - centered mobile layout

xyzhelen:


keyman-server:

LornaSIL:
For the record, there are a lot of differences in the touch layout, but this is related to the previous bug where it converted codepoints from something like `"O̱",` to `"O\u0331",`. In this case, it's reverting to the correct form from a previous change.

LornaSIL (approved):
LGTM

PR #2754, opened 2024-05-14
[fv nlakapamuxcheen] Second attempt at correcting name

HopsAndHops:
Second attempt to remove "fv_" from keyboard name on the Keyman page. I believe the change made in the .kps and .kpj should fix that!


keyman-server:

PR #2766, opened 2024-05-17
[fv_plains_cree] Updated terminology used in documentation, bumped version and year

HopsAndHops:
Updated documentation to include Spirit Markers alongside syllabics, updated history, bumped version number and year.


keyman-server:

LornaSIL (approved):
LGTM
Just as an FYI, changing the name in `Language ID` probably won't give you what you are wanting. I think those names are built off our official langtags file. I think the only way to get that name updated would be to get ISO 639-3 or the Ethnologue to change the name. Adding the name in other locations, as you've done, will assist in the search.

PR #2797, opened 2024-06-03
fix: restore color variables and die() function to util.inc.sh

mcdurdin:
These were removed as part of the builder script implementation of build.sh in the staging-17.0 branch, but I neglected to check other usages. Safer to put them back (copied out of util.sh pre-merge), and then we can tidily remove them again in the future.


rc-swag (approved):
lgtm

PR #2802, opened 2024-06-05
FV_lekwungen

xyzhelen:
new mobile keyboard for Lekwungen (Lək̓ʷəŋən)


LornaSIL:
I'm still not sure where we are with approving keyboards, but here's a bit of feedback.

- Please make sure to merge all the changes we've made into your clone of the repo because every keyboard changed. One change is removing `keyboard_info` and making sure that information is now in the .kps. Don't make changes to keyboards without starting from this version.
- We're moving to using Keyman 17 format, so change to that format which includes the description in the .kps now. Also, for other keyboards, where a font was included in `keyboard_info` make sure to include the font information in the .kps instead. We do not need/want you to commit the `keyboard_info` file anymore.
- We generally do not accept mobile-only keyboards anymore. In the far past we accepted just web or just mobile keyboards without a desktop keyboard and so the fv mobile-only keyboards were grandfathered in. If you could make sure new keyboards include a desktop layout (including this one) that would be best.

xyzhelen:
Thanks! I'll get to work on those changes.

keyman-server:

keyman-server:

PR #2808, opened 2024-06-10
Delete release/u/uma_phonetic directory

jackrittenberry:
This keyboard is no longer necessary and I would like to remove it from the public keyman repository


LornaSIL:
Hi @jackrittenberry . We don't allow keyboards to be removed from the release folder once they are added. This is for url stability. If there is another keyboard in our repository that would be a suitable replacement, we would be willing to "deprecate" this keyboard and recommend the other one instead of this one. Let me know if there is another keyboard that meets the same needs for the Houma language.

jackrittenberry:
Please deprecate the Uma phonetic keyboard and replace it with the Uma Graphic keyboard

LornaSIL:
Ok, I'll try to do that tomorrow.

LornaSIL:
Instead of deleting, we will deprecate. #2809

LornaSIL:
Take a look and see how it's been handled:
https://keyman.com/keyboards/uma_phonetic (it points you to https://keyman.com/keyboards/uma_graphic)

image

And Uma Graphic shows the Uma Phonetic keyboard has been deprecated.
image

Hopefully this solution will meet your needs.

PR #2820, opened 2024-06-12
[sil_greek_polytonic] Change targets to 'any'

### LINKED ISSUE #2818: [sil_greek_polytonic] add support for mobile
Copied from: https://community.software.sil.org/t/ancient-greek-ios-support/8820/3

---

I think what is happening here is that we resolved a problem in the deployment of keyboards, where some keyboards were being made available to mobile platforms, even if they weren't listed as supporting them (and in some cases didn't work at all on mobile).

The SIL Polytonic Greek keyboard is listed as supporting only ['web' and 'desktop'](https://github.com/keymanapp/keyboards/blob/f63cd02e14f9a9953bc4b72068056e205b3e3e48/release/sil/sil_greek_polytonic/source/sil_greek_polytonic.kmn#L38C18-L38C29) platforms. So OP were effectively using a desktop keyboard on the phone -- it works, but the layout is not optimal.

Obviously, as OP are successfully using the keyboard on their phone, it is not optimal to take it away.

So, I think we'll do this:
- [x] We will patch the keyboard to make it available to mobile platforms again, but without any other changes :link: keymanapp/keyboards#2820
- [ ] Someone (us, OP, anyone!) builds a proper optimized mobile layout for the keyboard

darcywong00:
Addresses the first point of #2818 (change target to ~'all'~ 'any') so community site user can continue to use the keyboard on mobile devices.

* Also added mobile font to .kps file


LornaSIL:
We need a touch layout actually added. So, the easiest would be to add a tablet layout that is completely imported from the desktop. Then after compiling, add the .js file to the .kps. I think that would accomplish what is needed.

Later a phone layout could be added that would make it easier to type on a phone.

darcywong00:
ah, I updated the PR description. This doesn't completely fix, Just addresses the first point of making the "web" keyboard available to previous mobile users. The issue will remain open for someone else to optimize a touch layout.

LornaSIL (approved):
Ok, I didn't know this would work.

PR #2824, opened 2024-06-14
[samaritan] New keyboard

LornaSIL:


release/t/takri_inscript/HISTORY.md
```diff
@@ -3,4 +3,4 @@ Takri - Inscript Change History
 
 1.0 (2024-05-21)
 ----------------
-* Created by 
+* Created by Lorna Evans
```
darcywong00, line 6: Does this belong on a separate PR?

release/h/hanifi_rohingya/HISTORY.md
```diff
@@ -3,4 +3,4 @@ Hanifi Rohingya Change History
 
 1.0 (2024-05-03)
 ----------------
-* Created by 
+* Created by Lorna Evans
```
darcywong00, line 6: Does this belong on a separate PR?

release/t/takri_inscript/HISTORY.md
DavidLRowe: Ideally perhaps Yes, but it seems fine since it doesn't affect anything in the keyboard package.

release/h/hanifi_rohingya/HISTORY.md
DavidLRowe: See previous comment.

DavidLRowe (approved):
Image for shift layer has a square box on the 4/$ key. Is this intended? If not, dismiss the review and change.

LornaSIL:
> Image for shift layer has a square box on the 4/$ key. Is this intended?

Not exactly, but I couldn't find a font with the script and that symbol. I think we can let it go for now.

PR #2828, opened 2024-06-18
fix: support build of just fv_all from build.sh

mcdurdin:
Relates to keymanapp/keyman#11811.

Build script would fail if `-k release/packages/fv_all` were the only command line parameters, as `grep` would return an error while preparing the list of files to build.


rc-swag (approved):
lgtm

PR #2831, opened 2024-06-18
fix(cs-pinyin): use published languages format in .keyboard_info

mcdurdin:
Fixes: keymanapp/keyman.com#458
Fixes: KEYMAN-COM-12V


LornaSIL (approved):
LGTM

PR #2838, opened 2024-06-21
[template] update template files

DavidLRowe:
These template files may no longer be used, but this updates them to match Keyman v17 expectations.


LornaSIL (approved):
LGTM

PR #2842, opened 2024-06-25
Wayana release - Moved help folder into source

sku21:


DavidLRowe (approved):
Change approved, but holding off on merge so that the developers can see the error.

darcywong00:
btw, I noticed the mobile layout has a "numeric" layer. You can include that in the online help by modifying
https://github.com/keymanapp/keyboards/blob/de1331eec273fd463e54b589c2288a478eb89e03/release/w/way/source/help/way.php#L79-L80

to
```xml
    <div id='osk-mobi-container'>
      <h3>Keyboard Layout
      <div id='osk-phone' data-states='default shift numeric'></div>
    </div>
```

DavidLRowe:
Good catch, @darcywong00 

@sku21 Adding "numeric" as Darcy suggests will display that layer in the online help. Since we haven't yet merged this pull request, you can make the change in this branch and it will become part of the pull request.

PR #2863, opened 2024-06-27
[naijatype] update 61 - add some missing underdots, hooks, nasals to touch

rowbory:
Updated documentation from the coverage checker to find what languages are officially supported - now over 102!


keyman-server:

keyman-server:

keyman-server:

LornaSIL:
You can test the touch layout before it goes live. Are you familiar with that? In the Keyman Developer 17.0 you can compile the keyboard, click on the **Build** tab, Click on **Test keyboard on web**, click on one of the urls and "Open in browser". Then, click on **Keyboard** and select NaijaType. Then select your device, either the Desktop or a mobile layout. It's very handy.

experimental/n/naijatype/source/naijatype.kmn
```diff
@@ -352,7 +352,16 @@ c 14: These named letters and combinations are used by the touch keyboard
 + [T_u_underdot] > 'ụ'
 + [T_n_underdot] > 'ṇ'
 + [T_s_underdot] > 's' U+0323
++ [T_z_underdot] > 'z' U+0323
```
darcywong00, line 355: l. 144
```
store(&KEYBOARDVERSION) '60'
```
should be 61

DavidLRowe: It is also possible to change the "id" field in the keyman-touch-layout file from "T_z_underdot" to "U_007A_0323", then remove the "[T_z_underdot]" rule from the .kmn file. Then the touch keyboard will generate the characters directly without needing the .kmn file. You will, however, need:
store(&VERSION) "15.0"
since the compiler (in the absense of the &VERSION statement) determines the needed level before reading the .keyman-touch-layout file and will default to a lower version.

keyman-server:

keyman-server:

keyman-server:

rowbory:
> You can test the touch layout before it goes live. Are you familiar with that? In the Keyman Developer 17.0 you can compile the keyboard, click on the **Build** tab, Click on **Test keyboard on web**, click on one of the urls and "Open in browser". Then, click on **Keyboard** and select NaijaType. Then select your device, either the Desktop or a mobile layout. It's very handy.

Thanks. Yes, I have used that extensively, and especially also to try tweaking the CSS, though it hasn't quite got the same behaviour as iPhone and Android always in relation t the popups.

experimental/n/naijatype/source/naijatype.kmn
rowbory: Thanks. That's helpful advice. Yes, in connection with a previous edit suggestion to remove the &VERSION because it can be worked out automatically, Marc had mentioned that was fine because I didn't use the double Unicode feature, and in fact that was the first time I was aware that was possible. It would let me considerably reduce the .kmn file. Would that be advantageous? The only loss would be a more semantic description of what the symbols represent.

rowbory:
> It seems like you are making a lot of changes to this keyboard. Do you want us to wait on checking this PR? If so, let us know whether you are ready for a review or not.

Yes, I'm sorry @LornaSIL I had a little window of opportunity (between work with translation teams) to try to make some improvements, but that window's come to an end now and I think I'm done, now that I've been testing it for some days. The only thing that remains is for me to investigate what it would take to convert the T_ constants to U_####_#### format as @DavidLRowe suggested. But that's a non-trivial set of changes.

keyman-server:

LornaSIL (approved):
LGTM.

FYI, unless you are distributing the in-between versions, there is no need to bump the version each time you commit, just once per PR. For example, we will never display v60 since we didn't accept the PR at that point.

experimental/n/naijatype/source/naijatype.kmn
mcdurdin: It's busy-work I guess -- no functional change to the keyboard but does make maintenance a bit easier. I probably wouldn't mark it as a high priority.

Issue #2484, opened 2023-12-05
feat: add an experimental LDML keyboard to validate CI (on staging branch)


mcdurdin:
Suggest adding the CLDR bengali keyboard, based on SIL Bengali Keyman keyboard

darcywong00:
Is @srl295 doing this in #2446?

mcdurdin:
I think we should aim to make that our experiment, yes. I don't know who will add it to the keyboards repo though.

mcdurdin:
This is being done in #2865.

mcdurdin:
We have a real keyboard imperial_aramaic added in #2993. This can be considered complete.

PR #2864, opened 2024-06-28
[ldml_test] add experimental/sil/ldml_test

srl295:
- test out LDML keyboards in repo
- https://github.com/keymanapp/keyman/issues/10505

see #2865

PR #2865, opened 2024-06-28
[ldml_test] add experimental/sil/ldml_test

srl295:
- test out LDML keyboards in repo
- https://github.com/keymanapp/keyman/issues/10505

same as #2864  but in same repo

Fixes: #2484.


DavidLRowe:
The convention for the keyboards repository is to store the keyboard in a folder based on the first letter or prefix, so:
release/s/sample
release/sil/sil_ipa

Following that convention, I'd expect
experimental/l/ldml_test OR
experimental/sil/sil_ldml_test

srl295:
The main point was to test ldml.  I expected there may be updates needed for local conventions.

Would sil_ldml_test make sense? I can rename it.

DavidLRowe:
sil/sil_ldml_test would fit the pattern. 

Is this intended to be an example? If so, sil_ldml_example maybe?
Or is this the first of a suite of test keyboards? maybe a new folder "test": test/test_ldml (with test_ldml_xxx to follow)?

The concern is that current policy is to not change keyboard folder/project names once they've been created.

srl295:
It's more of a test than an example.

experimental/sil/ldml_test/source/ldml_test.kps
```diff
@@ -0,0 +1,71 @@
...
+  </Options>
+  <StartMenu>
+    <Folder></Folder>
+    <Items/>
+  </StartMenu>
+  <Info>
+    <Name URL="">LDML Test</Name>
+    <Copyright URL="">Copyright © Steven R. Loomis</Copyright>
+    <Author URL="">Steven R. Loomis</Author>
+    <Description URL="">Test Keyboard</Description>
```
mcdurdin, line 23: This is inadequate detail because this is what will show in the keyboard details on keyman.com. We need this to be clear for end users who encounter it on keyman.com.

experimental/sil/ldml_test/README.md
```diff
@@ -0,0 +1,98 @@
+LDML Test keyboard
+==============
+
+Description
+-----------
+Test Keyboard for LDML
+
+
+Trying it Out
```
mcdurdin, line 9: Most of this information would be better placed in welcome.htm -- then it is available when the keyboard is installed, and not just in the source.

experimental/sil/ldml_test/.gitignore
```diff
@@ -0,0 +1,2 @@
+/build
+/*.kpj.user
```
mcdurdin, line 2: This file should not be needed as both of these are in the root-level .gitignore

mcdurdin (commented):
We could consider marking this keyboard as deprecated just to avoid it coming up for end users in the keyboard search?

As @DavidLRowe noted, it needs to be called either experimental/sil/sil_ldml_test or experimental/l/ldml_test.

Looking good though. Seems to be putting CI through its paces

mcdurdin:
From the build log:

```
10:56:14   ldml_test.kpj - info KM05002: Building experimental/sil/ldml_test/ldml_test.kpj
10:56:14   ldml_test.xml - info KM05002: Building experimental/sil/ldml_test/source/ldml_test.xml
10:56:14   ldml_test.xml - info KM05006: experimental/sil/ldml_test/source/ldml_test.xml built successfully.
10:56:14   ldml_test.kps - info KM05002: Building experimental/sil/ldml_test/source/ldml_test.kps
10:56:14   ldml_test.kps - info KM05006: experimental/sil/ldml_test/source/ldml_test.kps built successfully.
10:56:14   ldml_test.kpj - info KM05002: Building experimental/sil/ldml_test/ldml_test.kpj
10:56:14   ldml_test.kpj - info KM05006: experimental/sil/ldml_test/ldml_test.kpj built successfully.
10:56:14   ldml_test.kpj - info KM0500B: Project experimental/sil/ldml_test/ldml_test.kpj built successfully.
```

srl295:
I like the deprecated idea. Will update this.

srl295:
How do we mark it deprecated?

DavidLRowe:
The usual way this is done is for a keyboard to mark another keyboard as deprecated. In Keyman Developer (v17+) this can be done in the Packaging section, Details tab:
image
Use Add to add a related package and tick the Deprecated box to indicate that this related package should be deprecated.

@mcdurdin Is is possible for a package to deprecate itself?

mcdurdin:
> [srl295](https://github.com/srl295) requested a review from [mcdurdin](https://github.com/mcdurdin) [1 hour ago](https://github.com/keymanapp/keyboards/pull/2865#event-13398666893)

Still got some changes requested (plus this is in draft, so not clear if you intended to ask for review?)

> @mcdurdin Is is possible for a package to deprecate itself?

I don't know! It'd be weird, and could have other side-effects, so I don't think I'm keen on doing that. So... despite my brilliant idea, I don't think we can go this way for now.

That is a weakness in the package-deprecation pattern, but it's not an easy one to solve without repetition of data... (DEPRECATED.md is informative only at this point, but we could test for its as a possibility?) Would add complexity to the keyboards build.

mcdurdin:
> That is a weakness in the package-deprecation pattern, but it's not an easy one to solve without repetition of data... (DEPRECATED.md is informative only at this point, but we could test for its as a possibility?) Would add complexity to the keyboards build.

* #2893

srl295:
Yes apologies I missed the remaining items

mcdurdin:
I think we can probably close this as redundant given #2993 is a real LDML keyboard, so I think we don't need a test keyboard in the keyboards repo.

LornaSIL:
@srl295 can this be closed or are you still needing this open?

srl295:
> @srl295 can this be closed or are you still needing this open?

I think it's not needed.

PR #2868, opened 2024-06-28
[sil_mende_kikakui] Update to add a KeymanWeb help file

LornaSIL:
I've never done one of these before, but this keyboard especially needs it since the OSK is worthless.

I only found 2 keyboards where this file is actually working: [naijatype ](https://keymanweb.com/#ha,Keyboard_naijatype)and [sil_euro_latin](https://keymanweb.com/#aae,Keyboard_sil_euro_latin). No other keyboards have a working file. So, I tried to follow the example from these two keyboards where it does work. 

My help file is much longer than the other two, so I'm not sure how it will display, but I thought I'd try it.

### DIFF
--- a/release/sil/sil_mende_kikakui/HISTORY.md
+++ b/release/sil/sil_mende_kikakui/HISTORY.md
@@ -1,6 +1,10 @@
 Mende Kikakui (SIL) Change History
 ====================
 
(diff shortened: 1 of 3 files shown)


DavidLRowe (approved):
AFAICT LGTM

PR #2870, opened 2024-06-28
[FV diitiidatx] Minor Updates

HopsAndHops:
- Tidied rules and stores
- Normalized outputs
- Standardized name throughout documentation
- Added further information about mobile layout
- Tided documentation and code


keyman-server:

LornaSIL:
This line in the .kmn looks like you didn't change the codepoint from PUA. You just made the comment:
```
[NCAPS SHIFT K_5] > '' c new char, U+A7DC
```

There's also this line which you may wish to review and see if the character is in Unicode:
```
+ [CAPS K_5] > ''
```
Today I was looking at the help files. As far as I can tell, none of the fv help files are displaying in KeymanWeb correctly: fv_diitiidatx_kmw_help.htm

I would expect them to show up here and they aren't: https://keymanweb.com/#dtd-latn,Keyboard_fv_diitiidatx

So, I went and looked at the syntax page: https://help.keyman.com/developer/language/reference/kmw_helpfile

And it indicates that "This HTML file should not include `<head>` or `<body>` tags, as will be a snippet only. On a desktop browser, the contents of this file will replace the On Screen Keyboard."

I only found 2 keyboards on our system where the help actually displays properly. This is one:
https://keymanweb.com/#ha,Keyboard_naijatype

The source for that file:
https://github.com/keymanapp/keyboards/blob/master/experimental/n/naijatype/source/naijatype-help.htm

You can see where the `<head>` and `<body>` sections are completely stripped out. It's worth trying, to see if it fixes the issue. I haven't ever looked into this before today, so I'm not 100% sure if that's the only problem.

HopsAndHops:
Good catch on updating the character, I meant to swap that and it slipped my mind.

I'll give that a try, thanks for the suggestion!

HopsAndHops:
Updated to the new capital barred lambda (releasing with Unicode 16.0) and stripped the web help. Here's hoping that this works!!

keyman-server:

LornaSIL (approved):
LGTM

HopsAndHops:
Hi @LornaSIL I just took a look and I don't believe stripping out the head info worked. It was a good idea and worth the try! I'll look more into the syntax page and keep trying.

LornaSIL:
Yes, I agree. I need to wait until Marc is available to see what we are missing.

mcdurdin:
> Yes, I agree. I need to wait until Marc is available to see what we are missing.

It is due to the presence of `store(&visualkeyboard)` -- this is overriding `store(&kmw_helpfile)`.

Issue #1936, opened 2022-07-31
feat(sil_greek_polytonic): more control over sigma and smooth breathing


mcdurdin:
This is a report from Anna via private email to me.

- Sometimes the wrong sigma occurs and can't be fixed without some gymnastics. Could have a forced final-sigma and forced middle-sigma key.

- Putting rough breathing on is very intuitive in Keyman, but I think it's actually easier to type a key each time you want smooth breathing rather than having it automatically added when you don't want it.  Maybe repeat the vowel? eg "aa" -> ἀ.

LornaSIL:
@mcdurdin I don't mind adding a final-sigma (`@s` ?) and middle-sigma (`@S` ?) key. 
However, I don't think we should modify the rough breathing behavior. This keyboard has been around 20 years and that kind of change would be problematical for an unknown number of people.

mcdurdin:
Yes, modifying an existing keyboard is a recipe for trouble. I noted these as issues or features. If we wanted to control rough breathing behaviour, we'd need to have it as an option in the keyboard rather than changing existing behaviour. Not sure if that's a sensible thing to do at this time.

DavidLRowe:
@mcdurdin @LornaSIL Any action needed?

mcdurdin:
I think we should consider creating a new keyboard if this is a real issue, or adding it as a keyboard option in a future keyboard update. I should ask the OP if they still consider this to be a problem.

mcdurdin:
Okay, I've asked the OP

mcdurdin:
Feedback from Anna:

## Sigmas
The reason I suggested a forced σ and forced ς is for editing.  These problems occur:
* If I write ἐστιν (for example), then delete the τ, the σ turns into a ς
* If I go to add a σ into the middle of the word, it adds a ς instead. 

I make mistakes often enough for this to be annoying!

If those behaviours could be fixed, it would be great; otherwise I'd want to use the forced sigmas routinely.

## Breathing:

Problems occur when writing lecture notes and I want to write a vowel in isolation, or write a suffix in isolation (etc), and it automatically puts on a breathing mark.  If there was a way to force a no-breathing vowel, that may be good enough.

PR #2871, opened 2024-06-30
feat(sil_greek_polytonic): alternate keying methods for sigma and isolated vowels

mcdurdin:
v1.7:
* Add AltGr+S for enforced medial sigma
* Add AltGr plus A, E, I, O, U, J, or W for isolated vowels/rho without breathing marks

Fixes: #1936


LornaSIL:
Would you want to update the help file in the same way as you've done the welcome.htm file?
If you do, you could also remove the version number in the `README.md` file. Then that file doesn't need to be updated each time.

mcdurdin:
Righto, done! Thanks for picking up on those points.

LornaSIL (approved):
LGTM. And thanks.

PR #2258, opened 2023-06-18
feat: support for KbdXxxx fonts for khmer_angkor and sil_hebrew

mcdurdin:
Uses Keyman 17.0 displayMap to transform OSK and touch layout for the khmer_angkor and sil_hebrew keyboards, solving the dotted circle and missing diacritic problem for these keyboards.

In draft until Keyman 17.0 compiler toolchain is in place for this repo.

PR #2879, opened 2024-07-02
[sil_hebrew] Update keyboard to use new OSK font

LornaSIL:
Also updated touch layout to clean up `T_` rules and use `U_` rules.

If this works, I will create another PR to replace the khmer_angkor part of issue #2258


DavidLRowe (approved):
LGTM

PR #2885, opened 2024-07-02
[khmer_angkor] Update keyboard to use KbdKhmr.ttf

LornaSIL:
I need to double-check this before it is approved.


LornaSIL:
Need to verify that changes made to v1.4 did not get overridden.

LornaSIL:
I think this is ready.
Closes #2258

DavidLRowe (approved):
LGTM

mcdurdin (approved):
This looks good. Have you verified the OSK and touch layout visually in v16 or just v17? AFAICT, it should 'just work' in both.

LornaSIL:
I'm not sure. I need to test more. The OSK for sil_hebrew isn't displaying properly on my Windows 11 Home computer. It was displaying fine on my development machine, so I need to do more testing.

LornaSIL:
I *think* this is okay. I'm not 100% sure since I have development versions on my computer, but I did test on a different computer with Keyman 16.0

Nnyny:
@LornaSIL, is it possible to revert to use the old font? I test the new font with hint and it's not looking good. It becomes bold.

This is what it looks like on Android. 

new font:
photo_6203738775268475274_y

old font: 
photo_6203738775268475273_y

Another concern with the new font is the extra circle that appears on some dependent vowels on OSK. 
image

mcdurdin:
> Another concern with the new font is the extra circle that appears on some dependent vowels on OSK.

Is this in Keyman for Windows or Keyman Developer? Can you provide version details?

Nnyny:
It appears on both Keyman v17.0.327 and Keyman Developer v17.0.328.

mcdurdin:
Here's what I see for Khmer Angkor v1.5 on Keyman for Windows v17.0.328:

image

Nnyny:
Please check the shift layer, @mcdurdin.

mcdurdin:
@Nnyny

image

Nnyny:
I just reinstall the keyboard and now the dotted circles have disappeared from Keyman OSK but I can still see it in Keyman Developer.

mcdurdin:
> I just reinstall the keyboard and now the dotted circles have disappeared from Keyman OSK but I can still see it in Keyman Developer.

OK, that would be expected in Developer, because the displayMap fix only applies during the compile phase. It's a little annoying -- we could add a feature to show the end-result in Developer I guess!

PR #2887, opened 2024-07-03
[bcsans] Add BC Sans font for future use by FV

LornaSIL:
These are just fonts for future use. These fonts support both Latn and Cans.


DavidLRowe (approved):
RSLGTM

PR #2900, opened 2024-07-06
[fv_all] Add fv_ojibwa_ifinal and fv_ojibwa_ifinal_rdot

darcywong00:
Add two new keyboards to fv_all.kmp

* fv_ojibwa_ifinal #2890
* fv_ojibwa_ifinal_rdot #2891

PR #2902, opened 2024-07-06
[fv_all] Add BC Sans Regular font

darcywong00:
Follows #2900 
This adds BC Sans Regular font to the fv_all.kmp package.

The font template is pretty long - do I also need to include all the italics/bold variants of BC Sans?

https://github.com/keymanapp/keyboards/blob/01392938d6fe10ebaf72c304297b717969147873/release/packages/fv_all/source/fv_all.kps.in#L32-L85


LornaSIL:
Well, only the regular is ACTUALLY needed for the keyboard, but since this is kind of being used as a distribution method for people who want to use the fonts, it seems appropriate to include all the faces. Otherwise, when bold or italic are applied the user gets a fake bold or fake italic which isn't a pleasant user experience.

LornaSIL (approved):
RSLGTM

PR #2917, opened 2024-07-09
[sil_hebrew] Changed font for OSK to use KbdHebr instead

LornaSIL:
When I tested the online one on my home computer it was seeing square boxes. So, I realized I hadn't set the new font properly.

### DIFF
--- a/release/sil/sil_hebrew/HISTORY.md
+++ b/release/sil/sil_hebrew/HISTORY.md
@@ -1,6 +1,10 @@
 Hebrew (SIL) Keyboard Change History
 =======================
 
+2.0.1 (8 Jul 2024)
+-----------------
+* Changed display font for .kvks
+
 2.0 (2 Jul 2024)
 -----------------
 * Update keyboard to support OSK font for better display
--- a/release/sil/sil_hebrew/source/sil_hebrew.kmn
+++ b/release/sil/sil_hebrew/source/sil_hebrew.kmn
@@ -71,7 +71,7 @@ store(&COPYRIGHT) 'Copyright © SIL International'
 c store( &Message ) "http://scripts.sil.org/"
 store(&VISUALKEYBOARD) 'sil_hebrew.kvks'
 c store(&ETHNOLOGUECODE) 'hbo'
-store(&KEYBOARDVERSION) '2.0'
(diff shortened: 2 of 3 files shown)


DavidLRowe (approved):
LGTM

DavidLRowe:
Would this be an accurate summary of what needs to happen when changing a keyboard to use the special OSK font?

HISTORY.md - include new version number, and description:
Update keyboard to use special OSK font for better display

LICENSE.md - update date if needed

xyz.kmn
 - Update version number;
 - add store(&DISPLAYMAP) '../../..shared/fonts/kbd/kbdxxxx/KbdXxxx.json'

xyz.kps
 - add ../../../shared/fonts/kbd/kbdxxxx/KbdXxxx.ttf to list of fonts
 - set OSKFont to use the KbdXxxx font

xyz.kvks
 - set to use KbdXxxx font (in "encoding" element).

Needed only for the first keyboard using the Xxxx script: in release/shared/fonts/kbd/kbdxxxx/ folder, add KbdXxxx.json and KbdXxxx.ttf

LornaSIL:
I think that is right, but now I want to test the process with another Hebrew keyboard. I want to look at the OSK better with how it is now, somehow do a screenshot which is difficult to screenshot an OSK for some reason, and then make the changes and do another screenshot for comparison.

DavidLRowe:
I've had some success with something like: start a zoom session, start recording (locally), share the screen and perform the operation, then end recording. You end up with a video (.mp4 IIRC) rather than a screenshot. You can even narrate if you like.

mcdurdin:
<kbd>PrtScn</kbd> key copies entire screen to the clipboard. Works well unless you want to capture shift or other layer. For that, I use the Windows' built-in Snipping tool in timer mode for the whole screen.

mcdurdin:
> Would this be an accurate summary of what needs to happen when changing a keyboard to use the special OSK font?

...

> xyz.kvks
> 
> * set to use KbdXxxx font (in "encoding" element).

This is hopefully not needed -- this is legacy data that we are trying to scrub out (relying instead on the data in the .kps). If we find it is still being used, we should fix that!

Apart from that, that looks good.

PR #2926, opened 2024-07-10
(hcesar) mobile layout added

mareuo:


keyman-server:

DavidLRowe:
Generally looks good. There's one issue with the touch layout:
- The shift (uparrow) key doesn't work to change to the shift layer.
- You'll also need to make sure the shift layer allows you to return to the default layer.

In addition, you might consider some changes (depending on how closely you want to match the desktop layout)
- Consider moving the dieresis deadkey from the rightalt layer to the default layer (possibly remove the padding to the left of the hyphen, making room for the dieresis key on the right hand end of that row).
- Consider adding longpress keys for accented letters to the corresponding base key (so "a" would have longpress keys for a+acute, a+tilde, etc).

mareuo:
added this letters to longpress keys, 
image

keyman-server:

mareuo:
"Consider moving the dieresis deadkey", dieresis is not used.                                
"Consider adding longpress keys", added áéíóú, âêô, ãõ and à.
Are shift layers correct?

DavidLRowe:
If dieresis isn't used, then it's fine to leave that key on the rightalt layer.

DavidLRowe:
In Keyman Developer, on the Touch Layout tab, with the "default" layer displayed, click on the Shift key. You can then adjust what happens when that key is selected:
image
The main thing is to set the Next Layer field to "shift" so that the keyboard changes to the "shift" layer. (The Modifier field should probably be set to "(current layer)". I don't think it will make a difference, but it's not necessary. Similarly, the Modifier field of the "rightalt" key in the lower left of the default layer should probably be set to "(current layer").)

With the "shift" layer displayed, the Shift key should have the Next Layer set to "default" so that clicking this key on the shift layer return the user to the default layer. Normally the Key Type field is set to "Special (active)" to show that it has been activated. Again I'd suggest changing the Modifer field to "(current layer)". The "rightalt" key has the Next Layer set properly to "rightalt", but I'd suggest changing the Modifer field to "(current layer)" instead of the current reference to a non-existent rightalt-shift layer. 

With the "rightalt" layer displayed, the Shift key should have the Next Layer field set to "shift". For the key in the lower left (now labeled "default"), the Next Layer field should be set to "default" so that activating that key takes the user back to the default layer.

The main point is that, on each of the three layers, for those two keys used to switch layers, the Next Layer field should be set to indicate what the next layer should be.
and the Modifier field set to

keyman-server:

DavidLRowe:
@mareuo Your updates to correct the layer changes look good. Now you need to update the package to include the "hcesar.js" file from the "build" folder (use the Files tab for the package (.kps)). That is needed to get your new touch layout into the keyboard package.

keyman-server:

DavidLRowe (approved):

PR #2927, opened 2024-07-10
Delete release/t/taigi_sdwxv directory

Cathaylaboratory:


keyman-server:

DavidLRowe:
Our policy is not to delete keyboards once they have been submitted to the repository. You are welcome to update this keyboard, or to submit a new keyboard which could deprecate the existing keyboard (that is, mark the existing keyboard as obsolete so that it would not be offered to users).

Cathaylaboratory:
OK I understand your policy. So May I change the keyboards name from Taigy SDWXV to Taigi Viet Telex? Or I can upload my Taigi Viet Telex as a brand new program
Thank You

Sent with [Proton Mail](https://proton.me/) secure email.

DavidLRowe:
I see two possibilities:
- Revise the contents of the files in this (taigy_sdwxv) keyboard. All the filenames would retain the "taigy_sdwxv" name, but the contents would change to reflect the new name, in particular you would change the name that is displayed to users by changing the `store(&NAME) 'Taigi SDWXV'` line in the .kmn file to `store(&NAME) 'Taigi Viet Telex'`. You would also want to change any name references in documentation. That would be the simplest course of action. 
- The second option would be to submit a new keyboard (taigi_viet_telex) with your new files. In the packaging area, you would reference the taigi_sdwxv keyboard as a releated keyboard that is deprecated by the new one. With that deprecation, the taigi_sdwxv keyboard would not be offered to users unless they explicitly referenced it.

I don't know the details of what you wish to change, but I think the first option would be the easiest. You could call it version 2.0 to indicate that you've made a major change.

Please write again if you have further questions.

Cathaylaboratory:
I want to submit taigi viet telex keyboard as a brand new keyboard thank you

LornaSIL:
It might be wise to submit your keyboards to `experimental` in the future. That way if you change your mind about a keyboard we might be willing to delete it if it is in the `experimental` folder.

PR #2932, opened 2024-07-11
[runeboard] Version 3 updates

Runeboard:
* Removed duplicate character ᛅ.
* Moved ZWJ to shift+space & repositioned various keys.
* Added non-breaking hyphen, horizontal bar, guillemets (double & single), six vertical dots, & both trade mark symbols.
* On Touch Layout, added 0 on number layer & repositioned various keys.


keyman-server:

DavidLRowe:
In Package Details, the Welcome file and Readme file fields are empty. These should be set to the welcome.htm and readme.htm files that are in the package file list.

I think everything else looks fine (though I didn't check any of the keyboard outputs).

keyman-server:

Runeboard:
Thanks David. I've corrected this in my own branch. Was that successful?

For the keys, the desktop layout should be fine. I'm a little unsure about some of the longpress keys on the mobile layout though.

Could you please test whether the longpress option of ᚠ outputs the bindrune ᚠ‍ᚠ? (If your device can't display bindrunes, it will show as ᚠᚠ.) If that one works then the rest should be fine.

Previously, I recall there being some issue to do with the naming of longpress keys but I don't remember exactly what it was. However, I'm also aware Keyman has performed a major update since then, which could mean that doesn't matter anymore. Still, if there's anything which may need adjusting, please just let me know.

DavidLRowe:
> Could you please test whether the longpress option of ᚠ outputs the bindrune ᚠ‍ᚠ?
You are correct that there is a problem with this longpress key. Currently the ID field for the key contains "T_new_149". This is the keystroke value that will be passed to the .kmn file, so the .kmn file needs a rule to produce the output. (Currently there is no rule so that longpress key produces no output.) In the Text Unicode field I see U+16A0 U+200D U+16A0, so I expect you would want a rule like:
+ [T_new_149] > U+16A0 U+200D U+16A0

Note that the T_new_149 is an arbitrary name that Keyman Developer generated. You could change it to T_bindrune or whatever makes sense to you as the keyboard author.

Additionally, there is an option to use an ID field of U_16A0_200D_16A0 for the longpress key, which avoids the need to add the rule in the .kmn file to generate the actual Unicode characters. (This requires including `store(&VERSION) '15.0'` in the .kmn file, but you already have that, so you should be fine if you want to use this shortcut.)

I didn't test every key, but the rest of the keyboard seems to work as expected.

Thanks for making those changes to the package (.kps) file. One more very minor change would be to remove "runeboard.ico" file from the list of files included in the package. (The reason for this is that the .ico file has already been compiled into the runeboard.kmx file, so the original source file isn't needed in the package.)

DavidLRowe (approved):

DavidLRowe:
IDs for longpress keys on the Shift key should not matter IIUC. Seems fine when tested.

DavidLRowe:
BTW Consider changing these BCP 47 codes in a future version:
- change 'non-Runr' to 'non'
- change 'sxu-Runr' to 'sxu'

It seems that Runr is the default script for those two languages, so adding the script is not necessary.

Runeboard:
> BTW Consider changing these BCP 47 codes in a future version:
> 
>     * change 'non-Runr' to 'non'
> 
>     * change 'sxu-Runr' to 'sxu'
> 
> 
> It seems that Runr is the default script for those two languages, so adding the script is not necessary.

Thanks for mentioning that. I tend to think it best to leave it anyway as both languages are not infrequently latinised by scholars.

Thanks for the pull too.

PR #2737, opened 2024-05-06
[idoma] Version 1.0 of the Idoma language keyboard

Like-why:

PR #2985, opened 2024-07-23
[idoma] New keyboard

LornaSIL:
Replaces #2737


DavidLRowe (approved):

DavidLRowe:
@Like-why Here's a zip file with your keyboard
[idoma.zip](https://github.com/user-attachments/files/16352024/idoma.zip)
Any changes to the keyboard should be based on these files.

PR #2988, opened 2024-07-24
[Fv_lekwungen] - new desktop and mobile kb

xyzhelen:
- new desktop and mobile keyboard layouts


keyman-server:

xyzhelen:
> * new desktop and mobile keyboard layouts

The language bcp tag is activating the spelling suggestions from a dictionary for a neighboring language. The two languages have different orthography systems. Is there a subtag I can add (script or region) so this dictionary won't be used for Lekwungen? @LornaSIL

LornaSIL:
There's a merge conflict. I see you've committed a ton of files not related to this keyboard, so I suspect your master wasn't up-to-date. Could you create the PR from an up-to-date repo? I suspect you started this PR before all the work to change every file in the repo to Keyman 17.

xyzhelen:
> There's a merge conflict. I see you've committed a ton of files not related to this keyboard, so I suspect your master wasn't up-to-date. Could you create the PR from an up-to-date repo? I suspect you started this PR before all the work to change every file in the repo to Keyman 17.

I was wondering about all that :/ I will try again.

LornaSIL:
In addition, you have an fv_lkwungen.kpj in the root and in source. The one in source shouldn't be there.

So, you have `str` as the language tag. Are you saying this is `str` as well as the neighboring group? Then, is this a dialect of `str`?

xyzhelen:
> In addition, you have an fv_lkwungen.kpj in the root and in source. The one in source shouldn't be there.
> 
> So, you have `str` as the language tag. Are you saying this is `str` as well as the neighboring group? Then, is this a dialect of `str`?

Yes, the other language is SENĆOŦEN. Both would be considered dialects of "str" aka North Straits Salish.

LornaSIL:
Hm, my understanding is that Windows doesn't currently support subtags. You could try `str-x-lekwungn` (no final e because it must be 5-8 letters), but I suspect you will crash or confuse Windows. What do you think @mcdurdin ? Any other options?

DavidLRowe:
The "post-modern English" lexical model and keyboards use "en-QP" (where QP is a private use tag, and they chose "P" for "post-modern") to avoid picking up lexical models for English "en" (and vice versa). I don't know if that would be an appropriate workaround here or not.

mcdurdin:
Improving how we handle BCP 47 subtags is something that is in active discussion. In fact, we will be discussing this while in the lands of `str` (North Straits Salish) in September, alongside visit to FPCF.

https://github.com/keymanapp/keyman/labels/m%3Abcp47 lists the related issues.

For now, @DavidLRowe's suggestion of using `-Qx`, where `x` is a letter of your choice, is probably the easiest way forward, and then we can revisit after v18 is released. (I will reference this discussion in our BCP47 project work)

PR #2993, opened 2024-07-29
[imperial_aramaic] New keyboard using LDML syntax

LornaSIL:
- I guess the .kvk file is autogenerated from the xml source. However, when I install the keyboard nothing is showing when I turn on OSK. Is that something I've done, or is the tool not generating the right thing?

Otherwise, the keyboard works, but I would definitely need the help file to see what characters to type.


LornaSIL:
Also, is there something in the xml to set RTL text or does that just happen automatically?

srl295:
Nothing for rtl.

May not be osk with xml now.

LornaSIL:
https://github.com/keymanapp/keyman/issues/12056

DavidLRowe:
There seems to be a "ctrl" layer that duplicates the "default" layer, for example, pressing Ctrl+C produces the same character as pressing the "c" key (and so on for other keys on the layer). Is this intended? Is it a keyboard issue or a compiler issue?

DavidLRowe:
@srl295 Ctrl+C, Shift+Ctrl+C, RAlt+C, Shift+RAlt+C all produce the same output as "c". Is this a compiler bug?

srl295:
> @srl295 Ctrl+C, Shift+Ctrl+C, RAlt+C, Shift+RAlt+C all produce the same output as "c". Is this a compiler bug?

Not sure can you file an issue?

srl295:
Does this need a rebuild/recheck then with updated tools?

srl295 (approved):
ldml looks good

mcdurdin:
Please don't merge this yet

LornaSIL:
>Please don't merge this yet

Is there a tool where this can be tested now? Does Developer 18.0 do this? I haven't been notified of a new update for 17.0.

mcdurdin:
> Is there a tool where this can be tested now? Does Developer 18.0 do this? I haven't been notified of a new update for 17.0.

We're about to press the Go button on a 17.0 update that addresses the two LDML-keyboard related bugs. Have been swamped which is why it hasn't happened yet.

Recent 18.0 alpha e.g. 18.0.105 should work too (Keyman Developer, Keyman for Windows, Keyman for mac, Keyman for Linux).

mcdurdin:
OK, we released 17.0.329 yesterday, which I believe fixes the issues blocking this keyboard from merging. Looking forward to your feedback, do things work as you expect now @LornaSIL?

LornaSIL:
I just installed the .kmp. I had to update Keyman for Windows (and reboot) before the keyboard started to work. Then, the OSK doesn't seem to be there. And because of that, the help file doesn't display the layout! Otherwise, the keyboard works well in LibreOffice. In Word when I select the font it types one letter and then it looses the font and just displays square boxes and when I select the text and attempt to apply a font it doesn't display with the font.
@mcdurdin should we just quickly remove this keyboard or update it to use .kmn rather than xml?

DavidLRowe:
Similar experience: updated Keyman for Windows to 17.0.329, rebooted, loaded keyboard. In Notepad, text is rendered LTR. In LibreOffice Writer, text is rendered RTL (though in a left-justified paragraph). 

Word 2019 initially displayed US English, then displayed square box with "?". Note at the top: "Missing Proofing Tools - This document contains tex in Unknown Locale (arc-Armi) (arc-Armi-IR) which isn't being proofed. You may be able to get proofing tools for this language." (Out of curiosity I clicked the Download button and ended up on a Microsoft page offering Language Accessory Packs, but the list didn't include Aramaic!)

Online help doesn't display the keyboard images. OSK is blank.

mcdurdin:
@LornaSIL and @DavidLRowe -- there are a few different things going on here. I need to get myself organized to review this as well (currently my machine has mixed up versions of Keyman installed), but initial thoughts:

1. Word is going to struggle because Aramaic is not a supported language in Office, so it mucks up fonts. A possible fallback could be to use Syriac as a language code?
2. LTR vs RTL text: perhaps `arc-Armi` is not being treated as an RTL language/script in Windows? Same fallback may be helpful?
3. Online (web-based) help: we don't yet have support for LDML keyboards in KeymanWeb, so the OSK in documentation won't work yet. We could update the help to use images of the OSK for now.
4. OSK on Windows: I need to investigate further on this one.

So, TL;DR: let's stick with the LDML keyboard format and see if we can iron out the wrinkles. Am about to do the deeper investigation.

Issue #2589, opened 2024-03-04
feat [yiddish_pasekh]: independent keys for letters אױ and אײַ 


Meng-Heng:
@LornaSIL, a User contacted Keyman via Email and is looking for this implementation. Here is what they said:

_The full context translated from German using Google Translate:_
> Regarding the two missing letters אײַ / אױ: I know that I can "make" them. But I don't understand why these two missing letters don't have their own key assignment when the remaining letters, which also require a א at the beginning of the word, do have their own key. Especially since there are still unused keys (see below for the shifted keys) on the keyboard. It would just be easier to write.
> 
> Thank you again and greetings from Vienna!
> 
> ![image](https://github.com/keymanapp/lexical-models/assets/90595388/c2a3c7a0-d116-45e5-90cc-18cee71aa8e8)

Keyboard - https://keymanweb.com/#ydd-hebr,Keyboard_yiddish_pasekh
Keyboard help document - https://help.keyman.com/keyboard/yiddish_pasekh/1.9/yiddish_pasekh

The other keys that the User mentioned are **או, אי, אײ**...have their independent keys and combination keys to output them. But not for **אױ** and **אײַ** (the keys that the User wants as independent keys too).

To produce the אױ, press <kbd>Shift</kbd> + <kbd>a</kbd> and <kbd>Shift</kbd> + <kbd>o</kbd>
To produce the אײַ, press <kbd>Shift</kbd> + <kbd>a</kbd> and <kbd>Shift</kbd> + <kbd>y</kbd>

Would it be ideal to add these two characters to the SHIFTED layer or is there another reason that this is not recommended?

LornaSIL:
I can write the keyboard author to see what he thinks.

zsigrigy:
You can simply type ay and oy to get אײַ and אױ at the beginning of a word. Otherwise, you don't really need a shtumer alef before them.

Meng-Heng:
Hello @zsigrigy, can we add אײַ and אױ on the <kbd>SHIFT R</kbd> and <kbd>SHIFT D</kbd>?

EvafromVienna:
Die von @zsigrigy  angeführte Lösung, nämlich dass der shtumer alef automatisch am Wortanfang eingefügt wird, wäre natürlich die beste Lösung. Derzeit funktioniert das jedoch nur bei i, u und ey, nicht jedoch bei ay und oy.

EvafromVienna:
[...]

PR #3020, opened 2024-08-08
[yiddish_pasekh] adds point dagesh to shift layer

Meng-Heng:
Can this be updated? I did not contact the author for this change but a User has requested via Gmail. Please let me know. Thanks!


keyman-server:

DavidLRowe:
I'll defer to @lornasil (currently away) or @mcdurdin on the question of contacting the author.

I do note that this pull request has a change to `sil_euro_latin.kpj` which probably should not be included in this pull request.

mcdurdin:
It's always polite to try and reach out to the original author -- and then if they are unavailable we can proceed as we think best.

LornaSIL:
I think we need to understand the need better. The last time we reached out to Gyula (https://github.com/keymanapp/keyboards/issues/2589) the keyboard already produced what the user needed. For example, at the bottom of the welcome.htm file there is a table "Hebrew-specific letters are typed with shifted keys, too:" and many of those keystrokes output U+05BC in the correct context. If you want to forward me the email I can see if the keyboard already supports what they need. It's possible the documentation needs enhancing rather than the keyboard.

Meng-Heng:
Thank you everyone! I'll forward our conversation to you, @LornaSIL.

LornaSIL:
I've updated the keyboard in a different way (#3047). After thoroughly reviewing the keyboard, I added the keystroke to `\U` which seems consistent with the author's way of typing these things. I updated the documentation as well so hopefully it will help users know what to type.

Meng-Heng:
Thank you for implementing this @LornaSIL.

PR #3017, opened 2024-08-07
Add files via upload

Cathaylaboratory:

PR #3022, opened 2024-08-09
[taigi_viet_telex] update PHP help file from welcome.htm

DavidLRowe:
The information from the welcome.htm was copied to the taigi_viet_telex.php file, so that the online help will have the same information as the help packaged with the keyboard.


DavidLRowe:
Reference: welcome.htm updated in #3017

LornaSIL (approved):
LGTM

PR #3025, opened 2024-08-12
[ife_mn] add accents+tilde over upper case

DavidLRowe:
A few months ago, @MattGyverLee created this keyboard from an old Keyman file that I had previously worked on. That file didn't support stacked diacritics on upper case A, Epsilon, I, open-O and U. Now those characters are being requested. This adds ten additional lines to the .kmn file. 

Additional changes:
- removed some history and unneeded lines from the .kmn file
- changed the order to "latest on top" in HISTORY.md 
- updated to v17 format


LornaSIL (approved):
LGTM

PR #3033, opened 2024-08-15
chore(foochow): restore missing FollowKeyboardVersion tag

mcdurdin:
See also https://github.com/keymanapp/keyman/issues/12193


LornaSIL (approved):
Interesting! I wonder how that got deleted! I'll have to watch more carefully.

Issue #3039, opened 2024-08-20
bug(naijatype): custom CSS is too broad and impacts sites badly


mcdurdin:
The `naijatype` keyboard has the following in its associated CSS:

```css
body {
    background: rgb(193, 194, 193);
    font-family: "Andika","DejaVu Sans","AndikaAfr","Andika Afr","Andika New Basic Compact";
}
```

https://github.com/keymanapp/keyboards/blob/03b932e9b9558b6885534d31934084d00e88a6ef/experimental/n/naijatype/source/naijatype.css#L6-L9

_Originally posted by @jahorton in https://github.com/keymanapp/help.keyman.com/issues/1272#issuecomment-2298050379_

The CSS rules need to be restricted to the keyboard elements. 

(Note, a future update of KeymanWeb may use 'shadow dom' to prevent this kind of style leakage, see keymanapp/keyman#4881)

jahorton:
There is another issue with the keyboard's contents:  https://github.com/keymanapp/keymanweb.com/issues/121

mcdurdin:
> I notice it looks somewhat different, but it's still ugly.

I'm not sure what you are looking at? The two websites look fine now?

Help site:

![image](https://github.com/user-attachments/assets/6eee7be7-d561-44ae-9622-4d6b9a707e5b)

Search result:

![image](https://github.com/user-attachments/assets/71091f0e-320a-4fac-8965-ec1bbae0b36a)

LornaSIL:
Ok, I was looking at the keymanweb page.

mcdurdin:
On keymanweb.com:

![image](https://github.com/user-attachments/assets/19d42410-fec0-4ee4-a889-ebf3ddfe9aee)

LornaSIL:
I don't like the bullets leaking off the left side of the page. But, most other keyboards using this feature look the same.
Is this issue fixed? Can we close it?

mcdurdin:
> Is this issue fixed? Can we close it?

Yes. I am tracking the formatting of the hosted kmw_helpfile content separately in keymanapp/keyman#12292

PR #3041, opened 2024-08-20
[naijatype] Update keyboard to tidy code and address issues

LornaSIL:
- v62 was never actually build since the .kmn version wasn't updated.
- Removed unused AndikaAfr.ttf file
- Attempt to address: #3039
- Attempt to address: https://github.com/keymanapp/keyboards/issues/3043

@mcdurdin since I haven't done much with the css on Keyman can you confirm that I've done the right thing?


experimental/n/naijatype/source/naijatype.css
mcdurdin, line 1: The `h1` rule also needs to be removed

experimental/n/naijatype/source/naijatype.css
```diff
@@ -3,10 +3,6 @@ font-size: 250%;
 color: rgb(200,10,10);
     }
 
-body {
-    background: rgb(193, 194, 193);
-    font-family: "Andika","DejaVu Sans","AndikaAfr","Andika Afr","Andika New Basic Compact";
-}
 .kmw-key-row {
 }
 /* easier selecting from below *//*
```
mcdurdin, line 8: We could probably remove lines 8-14 as they are commented out

mcdurdin (dismissed):
I think looks okay, but some more adjustment to CSS needed

DavidLRowe (approved):

LornaSIL:
@rowbory We had to make a change to some of the code in your keyboard because the css was making the site look bad. Before you make changes to your version of the keyboard you should make sure you have the up-to-date files from the repo.
Thanks!

PR #3053, opened 2024-08-27
[vithkuqi] Fix online examples

### LINKED ISSUE #2788: [staging-17.0] Audit examples
As examples got migrated from .keyboard_info to .kps files in the staging-17.0 chain, there was some uncertainty on 
https://github.com/keymanapp/keyboards/commit/f73adec0c7ce4ef744531e00783e7056199e3721

where keyboard_info had
```
"text": "𐖑𐖥𐖵𐖣''𐖟",
```

Does that output a single quote?

darcywong00:
Fixes #2788 where the online example outputs one single-quote.


mcdurdin (approved):
I cannot find this script name attested anywhere online. Closest is https://en.wikipedia.org/wiki/Vithkuqi_alphabet which has the completely different 𐕲𐖻𐖶𐖨𐖷𐖰𐖥 which apparently comes from the [Vithkuqi Unicode proposal](https://www.unicode.org/L2/L2020/20187r2-n5138r2-vithkuqi.pdf) (see refs) but I couldn't spot it.

But looking at other text in the Unicode proposal I see only a single quote used and that makes more orthographic sense to me, so LGMT

Issue #2793, opened 2024-06-01
chore: improve targets for various keyboards with v17


mcdurdin:
Some keyboards are listed with different platform support. Keyman Developer 16 did not always report this consistently, as it 'guessed' on mobile keyboards when a .js was being produced. Also, a few keyboards had inconsistent information in their .keyboard_info, e.g. kbdsn1. v17 is more consistent on matching the &TARGETS store. These 122 keyboards are impacted:

Keyboard | Platform | Add/Remove | Level
---------|----------|------------|--------
aramaic_hebrew | android  | remove | full
aramaic_hebrew | ios  | remove | full
aramaic_hebrew | mobileWeb  | remove | full
athinkra_vai | mobileWeb  | remove | full
athinkra_vai_typewriter | mobileWeb  | remove | full
basic_kbdcherp | android  | remove | full
basic_kbdcherp | ios  | remove | full
basic_kbdcherp | mobileWeb  | remove | full
basic_kbdinuk2 | mobileWeb  | remove | full
basic_kbdiulat | mobileWeb  | remove | full
basic_kbdkhmr | mobileWeb  | remove | full
basic_kbdlt2 | mobileWeb  | remove | full
basic_kbdsors1 | mobileWeb  | remove | full
basic_kbdsw09 | mobileWeb  | remove | full
bj_cree_east_james_bay | mobileWeb  | remove | full
bj_cree_woods | mobileWeb  | remove | full
bj_mista_wasaha_cree | mobileWeb  | remove | full
bj_naskapi_classic | mobileWeb  | remove | full
coptic_greek | android  | remove | full
coptic_greek | ios  | remove | full
coptic_greek | mobileWeb  | remove | full
dene | mobileWeb  | remove | full
el_dari_clra | mobileWeb  | remove | full
el_harari_latin | mobileWeb  | remove | full
el_osmanya | mobileWeb  | remove | full
el_pan_sahelian | mobileWeb  | remove | full
engram | mobileWeb  | remove | full
esperuni | mobileWeb  | remove | full
fv_all | desktopWeb | add | full
galaxie_greek_hebrew_mnemonic | linux | add | full
galaxie_greek_hebrew_mnemonic | macos | add | full
galaxie_greek_hebrew_positional | desktopWeb | add | full
galaxie_greek_hebrew_positional | mobileWeb | add | full
galaxie_hebrew_positional | mobileWeb  | remove | full
gandhari | mobileWeb  | remove | full
[...]

PR #3054, opened 2024-08-27
[itrans] Address mobile layouts #2793

LornaSIL:
These are the 6 keyboards in `itrans` for Issue #2793.

I hope it's not too many to check at once, but I found all the issues that needed addressing were similar since they were by the same author.


DavidLRowe (approved):

DavidLRowe:
I see that the examples have an empty Note="" but assume that's not a problem. (I guess that's where "name of language" or whatever would go.)

LornaSIL:
I found the examples in the documentation, but there wasn't anything that indicated the meaning. I could guess, but I was afraid to.

PR #3055, opened 2024-08-27
[gautami_thamizh_tamil] New keyboard that deprecates old keyboard

LornaSIL:


DavidLRowe:
gautami_thamizh_tamil.php and welcome.htm are UTF-16 LE with BOM. Probably better if they could be UTF-8. Otherwise LGTM.

DavidLRowe (approved):

PR #3064, opened 2024-08-30
[paleohispanic] add flick input layout

mareuo:
image
paleohispanic


keyman-server:

LornaSIL:
There's a merge conflict. You'll need to remove the change you made to `korean_phonetic.kpj`.

Also, I don't really see how your flicks will work unless you add rules for the `T_xxx` IDs in your .kmn file. Have you actually tested the keyboard?

mareuo:
phn
rules
I tested the layout in browser, the flick input is not working yet, it can move, but the letters are not printing.

LornaSIL:
There are several issues that complicate this keyboard. 
1. First of all we must remember the script is not in Unicode so probably no other fonts will work with this keyboard.
2. You need to have the font installed in order for the characters to show up, and even then we cannot guarantee in what environments the font will work.
3. The font IS working in the Keyman Android App, but will not work outside of the app on Android because you cannot install a font for general use on an Android. I think iOS is different, but I don't know for sure.
4. I have found that in the test environment (browser testing) the font doesn't always show on the keycaps so you either just have to look at what the output is, or you can test within Keyman Developer.

Since this is an experimental keyboard and an unencoded script I'm not sure you should be spending too much time enhancing this keyboard with flicks.

Remember there is still a merge conflict as well that needs resolving.

PR #3011, opened 2024-08-01
[fv_hulquminum] - mobile spacing, upgrade to v17 dev

xyzhelen:

PR #3067, opened 2024-09-03
[fv_all] Update package version to 13.0

darcywong00:
This is a major bump of fv_all.kmp to 13.0 which incorporates all of the FirstVoices keyboard updates up to July 31, 2024.

Note: does not include #3011

### DIFF
--- a/release/packages/fv_all/HISTORY.md
+++ b/release/packages/fv_all/HISTORY.md
@@ -1,5 +1,8 @@
 # fv_all Keyboard Package
 
+## 13.0 (3 Sep 2024)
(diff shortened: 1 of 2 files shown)


LornaSIL:
There were a bunch of small changes July 31. Do those not matter for HISTORY.md? I'm not quite sure how the `fv_all` package works.

darcywong00:
fv_all is what the mobile apps include, but the user determines which of the keyboards gets installed.

With so many minor keyboard updates this past year, I've generally just highlighted if a new keyboard is added to the package.

PR #3072, opened 2024-09-03
[l-m targets] Add tablet for most but not for multi_pak_phonetic

LornaSIL:
Addresses some of the keyboards between l-m #2793
Adds tablet:
lao_2008_basic
maltese
manichaean
mro_phonetic
myancode_san

Removes .js file from multi_pak_phonetic keyboard as it would be too complicated to quickly implement.

Skipped mozhi_malayalam because it looks too complicated to quickly implement.


DavidLRowe:
For myancode_san keyboard:
- HISTORY.md has version 1.3
- .kmn file has version 1.2

Seems as though HISTORY.md could change to 1.2. (I made that change and also reordered the items in HISTORY.md.)

DavidLRowe (approved):

LornaSIL:
Thanks for catching that!

PR #3080, opened 2024-09-09
[colchis_latin] Update language tag

LornaSIL:
I think/hope I have the langtag correct.


LornaSIL:
@Nnyny has to update the lexical-model to use the same language tag, so it might be best to approve this and the lexical-model at the same time.

DavidLRowe:
Changes approved. Waiting to merge until @Nnyny updates the lexical model.

Nnyny:
Lexical model PR: https://github.com/keymanapp/lexical-models/pull/275

PR #3110, opened 2024-09-19
[galaxie_mnemonic] Remove kvks as keyboards are mnemonic

LornaSIL:
Relates to #2793 

This was definitely complicated. If you look at the history the targets have changed multiple times. I hope I got it right.


mcdurdin:
I'm just wondering if it makes sense to remove the .kvks entirely, beacuse it is still useful for US English base keyboard and largely for UK English, even if it gets increasingly unhelpful for other base keyboards (this is also something in the back of my mind that we want to support in the future -- translating the .kvk alongside the .kmx)

DavidLRowe (approved):

LornaSIL:
@mcdurdin I can revisit it if you wish, but what's there was so inconsistent I just finally had to decide something. Currently, it's not available to use online: https://keyman.com/keyboards/galaxie_greek_mnemonic

but the .kvk was in the keyboard .kmp
I've redone the 2 keyboards and 1 package 3 times on my computer as I've tried to figure out what to do.

DavidLRowe (approved):

mcdurdin:
OK, let's leave it for now. Mnemonic keyboard support should improve in upcoming releases -- cross-platform support and also hopefully we can address kvk at the same time, and get it working better without lots of engineering.

PR #3117, opened 2024-09-22
[Todhri] new keyboard

mauruo:


keyman-server:

LornaSIL:
It will be nice to have a Todhri keyboard on our site, but we are going to want a Todhri font before we can approve this. Otherwise it would be very frustrating to people to just see square boxes. Can you point me to where I can download it and add it to the repo? At this point I'm not finding a built font on Noto. I've found a repo, but not a ttf.

If you decide to add the font to our shared repo, that is fine, but make sure to use the script code and not the whole name of the script. So, create a `todr` folder within the `noto` folder. Thus the font would be in `release\shared\fonts\noto\todr\NotoSansTodhri-Regular.ttf`

Also, make sure to add the font to the touch layout as well.

Then, in the touch layout one thing is that you need to remember to add **Next Layer:** settings. So, for the default layer, shift key, set the Next Layer to **Shift** and on the shift layer, set the shift key to default. On the rightalt layer, set the shift key to go to shift. All the rightalts seem to be set correctly.

In the .kps set the language tags to `sq-Todr` and `als-Todr`. I'll add those to our BCP 47 tag list so in the future we'll recognize those.

Thanks.

LornaSIL:
I did finally find where the `NotoSerifTodhri-Regular.ttf` font is if you want me to slot that in instead.

mauruo:
yes

PR #3114, opened 2024-09-20
[FV hailzaqvla] New font in package, new code points

HopsAndHops:
One last straggler :)

* Updated to new code points for lambda, lambda with stroke, and capital labmda
* Packaged with Noto Heiltsuk Sans Font
* Added warning about characters not working correctly prior to new characters being officially released

PR #3125, opened 2024-09-25
[fv_all]: Add NotoHeiltsuk font and code points for fv_hailzaqvla

darcywong00:
Version bump in fv_all.kmp to include the update from #3114 

Could also hold off till #3011 gets merged too.


darcywong00:
I'll proceed without waiting on #3011 so we can get an update out to the FirstVoices apps

darcywong00:
Rebased to include #3011 so this is ready to review

DavidLRowe (commented):
RSLGTM

DavidLRowe (approved):

PR #3136, opened 2024-10-04
[gautami_thamizh_tamil] update PDF for online help

DavidLRowe:
Copied the PDF file from the "doc" folder and placed it in the "source/help" folder (replacing the existing PDF file), so that it will be accessible for the PHP file that provides the online help.


LornaSIL (approved):
LGTM

PR #3158, opened 2024-10-25
[manga] add four keyboards

DavidLRowe:
Four keyboards for Manga language: Latin script and Ajami (Arabic script) for both QWERTY and AZERTY layouts.

("manga_latin_qwerty" was originally "manga" and "manga_ajami_qwerty" was originally "manga_ajami". I used a lot of copy/paste and think I caught all the places needing Latin changed to Ajami, or QWERTY changed to AZERTY or both, but I kept finding more.)

No documentation at the moment (targeting "experimental").


LornaSIL (approved):
I didn't think you could reference the shared fonts in release. But, it built, so I learned something new.

PR #3160, opened 2024-10-28
New minor version

gsghyd:
Minor changes in key assignments and documentation.


keyman-server:

gsghyd:
Ready for review.

LornaSIL:
Thanks for this PR.
- The mobile layout is significantly different and I'm not sure it's been completely redesigned. Can you double-check? It only has Latin characters on the keycaps and some of them output Latin and some output Tamil so it doesn't seem correct to me.
- The .kps file includes a file `gautami_thamizh_tamil\source\..\DOC\gautami_thamizh_tamil.pdf` that I believe you removed. If you still want to include the .pdf you could reference it from the `help` folder. The way it is now, the keyboard will not compile.
- I'm not very excited about using Word to output html, as it is so messy. In the future it might be best to produce a clean html.
- In the .php file you'll need to remove lines 7-22:
```
<meta http-equiv=Content-Type content="text/html; charset=unicode">
<meta name=Generator content="Microsoft Word 15 (filtered)">
<title>Gautami Thamizh/Tamil</title>
<style>
</style>
<meta charset=utf-8>
<meta name=viewport content="width=device-width, initial-scale=1">
</head>

<body lang=EN-IN link="#0563C1" vlink="#954F72" style='word-wrap:break-word'>

<div class=WordSection1>

<div>

<h1 style='margin-right:-72.0pt'><span lang=EN>Gautami Thamizh/Tamil</span></h1>
```
(We don't need the `<h1>` because the title is autogenerated from the `pagename` and it is just duplicated.)

- Further down you have an `<h1>` for the **Vowels**. You could change that to an `<h2>` because the help file shouldn't include any `<h1>`s.
- Further down, **Consonants** is a `<p>`. You might consider making that an `<h2>` to be consistent with the **Vowels** header.
- In **Package Details** you might want to add the `welcome.htm` file back in there so that it will display when the package is installed.

image

gsghyd:
Hi Lorna,
(1) I have not worked on the mobile layout at all! Any changes are quite
inadvertent.
(2) I will fix the PDF reference in the .kps file.
(3) I understand your concern and I too would prefer not to use MSWord to
create html files, but unfortunately I know no better.
(4) I would really prefer not to mess around in html and php files. Would
it be possible to have the "help" link call up a PDF file with all the
pertinent information as well as keyboard layouts? I don't like the way the
autogenerated help file looks. I would like to have more control over what
is displayed in response to a click on the "help" link.
I just uploaded Gautami Bangla/Bengali. I suppose it will have all the same
problems.
Thanks and regards,
Gautam

gsghyd:
Hi Lorna,
All done with Gautami Thamizh/Tamil. Hopefully I have been able to make the
changes you asked for. Let me know.
Regards,
Gautam

gsghyd:
I have tried to fix the .php and .kps files in Gautami Bangla/Bengali to
the extent I can.
I hope they are okay now.
Gautami Devanagari coming up soon!

LornaSIL:
@gsghyd sorry for not checking this yet. I'm hoping I can look at it tomorrow.

PR #3161, opened 2024-10-28
[taiwan_austronesian] update to keyboard

Cathaylaboratory:
ready for reviewing


LornaSIL:
There are just two issues.
- There's a welcome.htm in the help folder. Please delete that file. It's there from a previous PR, but it shouldn't be there.
- In the php file (in help) there are two `<h1>` headers. They don't look right in the help files because the autogenerated header is an `<h1>`. Both of those should be changed to `<h2>`

Thanks.

Cathaylaboratory:
I have tried to fix the issue. Could you have a look?

LornaSIL (approved):
Thanks for making those changes. LGTM.

PR #3163, opened 2024-10-29
[abyssinica_font] Update font and keyboards that use typetuned fonts

LornaSIL:
This updates to the newest version of Abyssinica SIL. However, the gff keyboards needed to change because TypeTuner uses a different naming scheme than previously. So, we've updated
-  `AbyssinicaSIL-Connected-Regular.ttf ` to `AbyssinicaSILConnected-Regular.ttf `
- `AbyssinicaSILAlternate1-Regular.ttf ` to `AbyssinicaSILAlternateOne-Regular.ttf `
- `AbyssinicaSILAlternate2-Regular.ttf ` to `AbyssinicaSILAlternateTwo-Regular.ttf `

- Removed copyright dates everywhere except in LICENSE.md
- Removed version from README.md to make future updates easier
- Added `welcome.htm` to the **Details** tab in the .kps
- Update all keyboards to use Keyman Developer 17.0 format, cannot update keyboard package for `gff_gurage_and_amharic` because that still requires Keyman Developer 16.0 format.
- No changes were made to keyboards, kvks, mobile layout, etc.

@dyacob just for FYI.


mcdurdin:
> * Update all keyboards to use Keyman Developer 17.0 format, cannot update keyboard package for `gff_gurage_and_amharic` because that still requires Keyman Developer 16.0 format.

I think you mean cannot update keyboard project .kpj here? It is not a problem to leave these projects in 16.0 format :grin:

LornaSIL:
That one is in the packages folder. I never know what to call it.

LornaSIL:
> I think you mean cannot update keyboard project .kpj here? It is not a problem to leave these projects in 16.0 format 😁

This was mostly explanatory for the reviewer because when there are multiple .kmn files in a kpj it cannot be updated. Since it took me awhile to remember this, I felt the explanation would help, but perhaps there was not enough information to make it helpful.

mcdurdin:
It was just that you wrote 'keyboard package' rather than 'keyboard project'.

LornaSIL:
But...I don't know what to call it when it's in the `packages` folder. It's
a different beast with multiple keyboards.

mcdurdin:
OK, so the 'keyboard project' is the .kpj file. Even when it is in the 'packages' folder. The package file is always the .kps file (and we can call them multi-keyboard packages if they have more than one .kmx in them, perhaps). We could call them multi-keyboard projects but it's a little ambiguous! 

These beasts just exist to support the few keyboard packages where we have more than one keyboard in the package. They are kinda wedged in a bit awkwardly alongside our normal single-keyboard-per-package approach but I don't think we can go stop supporting them at this point.

The folder name 'packages' is probably a misnomer. 'bundles' would have been good, except that we used to use that for Keyman+keyboard bundled into a single .exe installer...

PR #3166, opened 2024-10-30
[galaxie_hebrew] Update Galaxie Hebrew keyboards to use DISPLAYMAP

LornaSIL:
- Updated the two Galaxie Hebrew keyboards to use the `&DISPLAYMAP`.
  - Added KbdHebr as the OSKFont
  - Removed 25cc from OSK and touch layout 
  - Added Ezra SIL for DisplayFont
- Updated the two Galaxie keyboard packages that include both the Greek and Hebrew keyboards.


LornaSIL:
Thanks for catching that! It would not be a happy experience for the users.

DavidLRowe (approved):

PR #3170, opened 2024-10-31
[dives_akuru_inscript] New keyboard for Diak

LornaSIL:
Had to add a font for this keyboard in the noto folder.


DavidLRowe:
I don't see the .js file in the .kps file list. Is that an oversight? Otherwise looks fine.

LornaSIL:
Thanks for catching that. I had started this as an LDML keyboard (which doesn't support mobile), but when I realized I couldn't associate a font with it, and the OSK was square boxes, I switched to a regular .kmn structure instead of LDML.

DavidLRowe (approved):

PR #3171, opened 2024-10-31
[Linb_Cyprt] Remove grc from Linb and Cprt keyboards

LornaSIL:
Based on CLDR and langtag changes.


mcdurdin:
See also:
* https://github.com/keymanapp/keyman/issues/9099#issuecomment-2436756320

mcdurdin:
Should we be removing `grc`, or changing it to `grc-Cprt`?

LornaSIL:
> Should we be removing grc, or changing it to grc-Cprt?

I don't actually know what to do since I'm seeing comments on langtags and CLDR. I thought the CLDR one said certainly grc didn't use Cprt and then on langtags they discovered that maybe it did.

DavidLRowe (approved):

mcdurdin:
I don't know the answer either! According to random Google search, yes: https://crewsproject.wordpress.com/2017/09/08/ancient-literacy-and-cypriot-mercenaries/#:~:text=Instead%20of%20using%20the%20Greek,scholars%20but%20not%20well%20understood).

Issue #3191, opened 2024-11-08
bug(itrans_odia): examples need spaces between keys


mcdurdin:
https://github.com/keymanapp/keyboards/blob/56b7537010bd846c8a6c5ce47fb4e2380f96136a/release/itrans/itrans_odia/source/itrans_odia.kps#L137C1-L159C18

Currently, it renders like this:
![image](https://github.com/user-attachments/assets/02f0cf32-f2c0-416e-bd8e-d13bb94b8193)

But it should have each letter as a separate key.

See the specification at https://help.keyman.com/developer/18.0/context/edit-language-example#toc-key-sequence

LornaSIL:
Fixed by #3193

PR #3193, opened 2024-11-08
[examples] Fix up .kps files for correct structure

LornaSIL:
Addresses #3191 as well as other keyboards with issues.

Since this is only displayed on the webpage and not in a keyboard package I don't think a version bump is necessary.


mcdurdin:
> Since this is only displayed on the webpage and not in a keyboard package I don't think a version bump is necessary.

At this point, that is correct. At some point in the future, we may show language examples in the Keyman UI.

DavidLRowe (approved):

PR #3200, opened 2024-11-14
[triqui_itunyoso] Initial submission of Itunyoso Triqui keyboard

kaylashames:


keyman-server:

kaylashames:
Ready for review

kaylashames:
Hi Lorna,

I couldn't quite figure out how to add the font properly. I used Roboto-Regular.ttf from https://fonts.google.com/specimen/Roboto . If you could help me with that, that would be great.

I could also use Noto Sans - would I have to go change that in Keyman developer and reupload?

LornaSIL:
Our preference is to not have just a mobile layout. If you could develop the rules needed for a desktop layout and On-Screen keyboard that would be best.

- Long keyboard names are sometimes problematical and I generally prefer we not use the word "Keyboard" in a keyboard name. It's somewhat redundant so we know it's a keyboard and on mobile devices the user may only see the word Keyboard and not the important words of what the keyboard is. So, I'd like the name changed to `Itunyoso Triqui` and remove the word "Teclado" in the keyboard name. Of course feel free to use it elsewhere. It's nice to have Spanish documentation for the users.
- The .kmn file is referencing `Keyboard.keyman-touch-layout` which I asked you to delete so the keyboard won't compile. Oops. I assumed that the file `triqui_itunyoso.keyman-touch-layout` was the correct one. That is in fact the name that should be used. So, I'm not sure if the code in `Keyboard.keyman-touch-layout` is correct or if the code in  `triqui_itunyoso.keyman-touch-layout` is correct. Keep the name `triqui_itunyoso.keyman-touch-layout` but make sure the source is correct for what you want, then change the filename in the .kmn file to `triqui_itunyoso.keyman-touch-layout`

- When I changed the mobile layout to `triqui_itunyoso.keyman-touch-layout` I get these warnings:
```
triqui_itunyoso.kmn:481 - warn KM020AD: Other rules which reference this key include CAPS or NCAPS modifiers, so this rule must include NCAPS modifier to avoid inconsistent matches
triqui_itunyoso.kmn:481 - warn KM020AD: Other rules which reference this key include CAPS or NCAPS modifiers, so this rule must include NCAPS modifier to avoid inconsistent matches
triqui_itunyoso.kmn:66 - warn KM020AB: nul must be the first statement in the context
triqui_itunyoso.kmn:481 - warn KM020AD: Other rules which reference this key include CAPS or NCAPS modifiers, so this rule must include NCAPS modifier to avoid inconsistent matches
triqui_itunyoso.kmn:481 - warn KM020AD: Other rules which reference this key include CAPS or NCAPS modifiers, so this rule must include NCAPS modifier to avoid inconsistent matches
triqui_itunyoso.kmn:481 - warn KM020AD: Other rules which reference this key include CAPS or NCAPS modifiers, so this rule must include NCAPS modifier to avoid inconsistent matches
triqui_itunyoso.kmn:445 - warn KM020AD: Other rules which reference this key include CAPS or NCAPS modifiers, so this rule must include NCAPS modifier to avoid inconsistent matches
triqui_itunyoso.kmn:481 - warn KM020AD: Other rules which reference this key include CAPS or NCAPS modifiers, so this rule must include NCAPS modifier to avoid inconsistent matches
triqui_itunyoso.kmn:481 - warn KM020AD: Other rules which reference this key include CAPS or NCAPS modifiers, so this rule must include NCAPS modifier to avoid inconsistent matches
```

Although they are just warnings, they will affect the user experience.

- The `LICENSE.md` file has this: `© 2024 Kayla Shames`. It needs to have the word "Copyright" at the beginning of the line. `Copyright © 2024 Kayla Shames`
- The .kps file needs to have `LICENSE.md` added to the list of files. Then on the **Details** tab add it in to the **License file:** field:
- 
image

Also, as in the image above, add welcome.htm in and add a reasonable description. That description is what will be used on the Keyman keyboard home page for this keyboard to help people know if they want to use it.

Usually a welcome file will just explain how to use the keyboard. You have a very long file which describes how the keyboard is used with a lexical model. I assume you are submitting that also at some point? I'm not sure if all of that really belongs in the welcome file.

For any keyboard in the `release` section we require an online help file. It might end up looking something like this help file: https://help.keyman.com/keyboard/basic_kbdhe/1.1/basic_kbdhe

We generally take the content of the welcome.htm file. That would be anything between the first `<h1>` (not including the `<h1>` and the end, but not including `</body>` and `</html>` and putting a php header on it:
```
<?php 
  $pagename = 'Itunyoso Triqui Keyboard Help';
  $pagetitle = $pagename;
  require_once('header.php');
?>
```

This file would be called `triqui_itunyoso.php` and you would put it in the `help` folder: `triqui_itunyoso/source/help/`triqui_itunyoso.php`

As discussed earlier, it would be preferable to use Noto Sans instead of Roboto. I can help with that if the above things can be resolved.

kaylashames:
Hi Lorna,

Thank you for the comments. I will work on implementing these. I will submit the lexical model as well.

There are a few reasons I only included a mobile keyboard. First, the majority of speakers of this language only use a mobile phone, and computers are not very common in the area. Second, because the language is just Latin script with accents, it didn't seem necessary to create a separate desktop keyboard. It is fairly convenient to type Itunyoso Triqui using a standard Latin keyboard. On mobile, it is harder, due to the quantity of accents needed and their high functional load, so a purpose-built Triqui keyboard has more of a use case. Let me know your thoughts.

LornaSIL:
Another argument for a desktop layout is if someone were to use an external
keyboard, like via Bluetooth, it wouldn't work unless you include a desktop
keyboard.

kaylashames:
How would I set the font to Noto Sans in Keyman Developer? Should I download the file and mimic the /shared/fonts/ structure on my own device, or is there a better way to do it?

I think I will just include the mobile layout. If there is a need for a computer/on-screen/bluetooth keyboard layout in the speech community, I will add those at a future date.

DavidLRowe:
Yes, mimicing the shared/fonts/ structure is awkward, but works. See the "Including fonts in your keyboard package" heading on the https://help.keyman.com/developer/keyboards/github/step-2 help page.

(https://help.keyman.com/knowledge-base/?id=106 has information on downloading a folder from GitHub. That might potentially be useful if you need to download multiple files for the font.

kaylashames:
Ready for review

kaylashames:
Done.

LornaSIL (approved):
I think it's all good now!

PR #3203, opened 2024-11-17
New keyboard: Gautami IndiTran

gsghyd:
For IAST/ISO-15919 transliteration and IPA transcription of Indian languages and scripts. 
Ready for review.


keyman-server:

gsghyd:
READY FOR REVIEW.

LornaSIL:
- There are two welcome.htm files. Can you delete the one you aren't wanting to use? 
- Then, you should verify the one referenced in the .kps file is the one you *do* want to use.
- Under **Keyboards** tab you could select `welcome.htm` so it shows up when the user installs the keyboard.
- Under **Details** language tag, there is no need to include `.001` as a region. Just use `la`.

LornaSIL (approved):
I'm going to approve it. For IPA keyboards we have used `und-Latn` but it will be interesting to see if the Latin language `la` works for you. We might want to switch to `und-Latin` later. I'm not sure. It's transcription of Indic languages so not exactly the same thing.

mcdurdin:
In the future, we may be able to move to a more precise BCP 47 tag for transcription, using the `t-` subtag prefix; see https://datatracker.ietf.org/doc/html/rfc6497 for the elaborate details. (We can't do that today because Keyman restricts to language-script-region for compatibility reasons, but we want to improve that, see keymanapp/keyman#12012)

PR #3206, opened 2024-11-22
Add files via upload

okandale:


okandale:
Ready for review.

LornaSIL:
Thanks for the update @okandale! There are some file conflicts that must be resolved before we can review the request. When we did a big update of Keyman in July we did update a lot of the keyboards. I'm wondering if you pulled the latest version of the files before you did your update? Those need resolving before we can approve.

Another issue is that you have committed files which should not be committed. All the compiled files should not be committed. So, please delete the following files from the PR:
- build/laz.js
- build/laz.kmx
- build/laz.kvk
- build/test.js
- build/test.kmx
- laz.kpj.user (this is just a temporary file on your system)
- test.kmn (not needed)

The `laz.kpj` is including `test.kmn` in the list of files. You need to remove `test.kmn` from the list of files. A better solution is to upgrade to Keyman Developer 17.0 so that it only references the .kmn and .kps files that are in the source folder.

In addition, your `laz.kmn` should be updated. This statement: `store(&VERSION) '15.0'` can stay that way, you had bumped it to 17.0 but that statement refers to the minimum version of Keyman it will work with and it will certainly still work with 15.0

This statement is the version number of the laz keyboard: `store(&KEYBOARDVERSION) '1.0.2'` That's the one where the version needs to be bumped to `1.0.4` is what I think your HISTORY.md file is indicating.

PR #3208, opened 2024-11-22
Update readme.htm

gsghyd:
Typo removed from file. Wrong version of PDF was being displayed.

### DIFF
--- a/release/gautami/gautami_inditran/source/readme.htm
+++ b/release/gautami/gautami_inditran/source/readme.htm
@@ -16,7 +16,7 @@
 <h1>Gautami IndiTran</h1>
 <p>Copyright © 2024 Gautam Sengupta</p>
 <p>This keyboard is for IPA transcription and IAST/ISO-15919 transliteration of Indian languages and scripts. Please download and read the document
-  <a href="https://help.keyman.com/keyboard/gautami_inditran/1.0/gautami_inditran.pdf">gautami_inditran.pdf</a> 
+  <a href="https://help.keyman.com/keyboard/gautami_inditran/1.0.1/gautami_inditran.pdf">gautami_inditran.pdf</a> 
   before you start using this keyboard. Any symbol required for transcribing or transliterating any Indian language or script but unavailable on this keyboard may kindly be reported to the <a href="mailto:gsghyd@gmail.com">author</span></a>. Thank you for your help.
 </p>
 </body>


keyman-server:

LornaSIL:
Unfortunately, any time you update a file that is part of the keyboard package you must also update the version number in the .kmn and add a HISTORY.md entry. I'm sorry!

gsghyd:
READY FOR REVIEW.

gsghyd:
> Unfortunately, any time you update a file that is part of the keyboard package you must also update the version number in the .kmn and add a HISTORY.md entry. I'm sorry!

DONE. READY FOR REVIEW. Bumped version number in .kps as well. Is that necessary?

LornaSIL:
@gsghyd I think you may have made the changes, but forgot to commit them?

gsghyd:
> @gsghyd I think you may have made the changes, but forgot to commit them?

I see no pending commits. I see the message "this branch is 5 commits ahead ...". Since the earlier pull request was pending, I didn't create a new one.

mcdurdin:
> I see the message "this branch is 5 commits ahead ..."

This means that you have 5 commits on your local branch that have not yet been pushed to this branch on GitHub. You should run `git push` to push them to GitHub.

DavidLRowe:
I notice that you include the .pdf file in the help folder, but don't reference it from the .php file. 

Some background information (which you may already know):
- The readme.htm file provides information about the keyboard at the beginning of the installation
- The welcome.htm file provides "help" information packaged with the keyboard that is displayed at the end of installation and whenever the user requests help.
- The .php file (and the "help" folder) provides the online "help" information for this keyboard

I wonder if it would be better to:
- include the .pdf file in the welcome folder (in addition to the copy in the help folder)
- include that .pdf file in the list of files in the package (.kps file)
- reference the .pdf file from welcome.htm file (this would be a relative reference to a local file, something like: `<a href="gautami_inditran.pdf">gautami_inditran.pdf</a>` since the .pdf file is included in the package)
- either change the reference in the readme.htm to be the same as in the previous point, or remove the reference entirely
- include a similar reference to the .pdf file in the .php file (this will also be a relative reference as in the previous point, since the .pdf file is included in the help folder) and then the .pdf file will be accessible from the online help

If any of this is unclear, please ask followup questions.

gsghyd:
Hi Marc and David,
- I didn't initiate a new "pull" request because I saw that the old one was
still pending. Should I start a new one now, maybe after closing the old
one?

   - either change the reference in the readme.htm to be the same as in the
   previous point, or remove the reference entirely.

- I want the PDF to be accessible from readme.htm. I suspect that a
"relative reference" to the PDF from readme.htm is not going to work since
at the beginning of the installation it - the PDF -  is not in place yet.
Please let me know if this is correct. If so, I will not make any further
changes to the readme file.
- It's not clear to me why it would be necessary or expeditious to include
two copies of the PDF in the package, since I could reference the copy of
the PDF in the HELP folder from welcome.htm if I wished to. I am still in
the process of finding my way around. So, kindly bear with me.
Thanks and regards,
Gautam

mcdurdin:
Gautam,

You can 'push' additional commits to an open pull request -- with `git
push` (or equivalent in a git GUI tool). So you don't need to open a new
pull request in that situation.

The compiled package has all files in the same folder. When you open the
readme, all the files in the package are available, so a PDF link would
work. You should not have two copies of the same PDF in the package,
although you may need two copies of the PDF in the repository -- one for
the package, and one for the help site.

Marc

gsghyd:
Hi Marc,
Thanks for letting me know about "git push". Regarding the rest of the
issues, if I understand you correctly, my best option would be to leave the
PDF in the HELP folder if it is referenced by the PHP. Otherwise, it
wouldn't matter where it is located since "the compiled package has all
files in the same folder" and I could access it from readme.htm or
welcome.htm if required. Am I missing something here?
Regards,
Gautam

gsghyd:
Will initiate new pull request after makiing more changes.

mcdurdin:
Yes, I think you have that all correct.

PR #3213, opened 2024-11-25
Itunyoso Triqui Font change

kaylashames:
Change font of Itunyoso Triqui keyboard from Tahoma to Noto Sans


keyman-server:

kaylashames:
Ready for review

LornaSIL:
Thanks @kaylashames. When there are any changes made which affect the keyboard package - like if you change the font or change any rules in the keyboard or the content of the readme.htm or welcome.htm, the version number of the keyboard must be "bumped". At a minimum, you would need to update the `HISTORY.md` and the `triqui_itunyoso.kmn` file. You'll need to record what changed in this version in the `HISTORY.md` file and just update `store(&KEYBOARDVERSION)` in the .kmn file.

The reason for this is that our build system will offer the new version of the keyboard to the users if the version gets bumped. Otherwise, the current users would never be offered this improved version of the keyboard.

mcdurdin:
> The reason for this is that our build system will offer the new version of the keyboard to the users if the version gets bumped. Otherwise, the current users would never be offered this improved version of the keyboard.

Furthermore, we never publish over existing published versions of keyboards, so if the version number is not bumped, the changes will never be made available to anyone!

kaylashames:
Should be good now!

LornaSIL (approved):
LGTM

LornaSIL:
Also, close your current PR where you are changing the README.md. Please just delete the version from the README.md so it doesn't need updating each time. The only places we want to be changing the version is in HISTORY.md and the .kmn.

The only place we really want a year in the copyright is in LICENSE.md. Otherwise it's more work to maintain the metadata.

PR #3217, opened 2024-11-28
[gautami_inditran] Update

gsghyd:
Adjusted image size in .php file.


keyman-server:

gsghyd:
READY FOR REVIEW.

LornaSIL:
It looks like you've updated the php file for the inditran keyboard and that would be fine, but then the tamil keyboard was updated in this same branch and thus the pull request.
The problem is you deleted two folders called `help` and `welcome` and uploaded the contents of two folders called `HELP` and `Welcome` There is a difference in casing which is a significant difference in programming and in Linux operating systems. So, those need to be reverted back to the correct lowercase forms. 

If you can separate the 2 keyboard PRs it will be easier to review. 

Today and tomorrow are an American holiday (I'm American) and I will likely not be able to review them as quickly as you might like.

gsghyd:
Hi Lorna,
Sorry, I didn't mean to rush you and I wasn't aware of the holidays in the
US. I assumed that something had gone wrong with my pull request preventing
it from being processed.
Best,
Gautam

LornaSIL:
Not a problem! We can't know all the holidays around the world!

gsghyd:
Errors in pull request.

Issue #3221, opened 2024-12-02
bug: keyboard search includes sil_euro_latin when searching for "han script"


darcywong00:
I was testing localizing keyboard searches and didn't understand why **sil_euro_latin** appeared in the list when searching for `s:han` (han script)

Repros on the live site:
https://keyman.com/keyboards?q=s%3Ahan

![Image](https://github.com/user-attachments/assets/2df65e08-4428-4515-bf51-13993d063158)

I'm not finding "han script" in the api.keyman.com blob
https://api.keyman.com/keyboard/sil_euro_latin

mcdurdin:
Note: I haven't verified the other tags. It'd be good for a package build to verify the script tags are the same for all referenced BCP 47 languages (and hint on mismatches)

mcdurdin:
After implementing a test in kmc-keyboard-info against keymanapp/keyman#12752, I got the following report:

```
$ ./build.sh | grep KM09011
aramaic_hebrew.kpj - hint KM09011: The script 'Syrj' associated with language tag 'amw-Syrj' does not match the script 'Syrc' for the first language in the package.
arbore.kpj - hint KM09011: The script 'Latn' associated with language tag 'amf' does not match the script 'Zyyy' for the first language in the package.
batak.kpj - hint KM09011: The script '<param>' associated with language tag 'btk-Batk' does not match the script 'Batk' for the first language in the package.
baybayin.kpj - hint KM09011: The script 'Latn' associated with language tag 'tbw-Tglg' does not match the script 'Tglg' for the first language in the package.
baybayin.kpj - hint KM09011: The script 'Latn' associated with language tag 'hnn-Tglg' does not match the script 'Tglg' for the first language in the package.
baybayin.kpj - hint KM09011: The script 'Latn' associated with language tag 'bku-Tglg' does not match the script 'Tglg' for the first language in the package.
bu_phonetic.kpj - hint KM09011: The script '<param>' associated with language tag 'und-Latn' does not match the script 'Latn' for the first language in the package.
basic_kbdinuk2.kpj - hint KM09011: The script 'Cans' associated with language tag 'iu' does not match the script 'Latn' for the first language in the package.
basic_kbdiulat.kpj - hint KM09011: The script 'Cans' associated with language tag 'iu' does not match the script 'Latn' for the first language in the package.
[...]

PR #3223, opened 2024-12-02
[sil_el_ethiopian_latin] Update language tags

LornaSIL:
Addresses 1 kbd in https://github.com/keymanapp/keyboards/issues/3221
Note that the messages about script `'Zyyy'` did not make sense to me. I just reviewed the language tags to make sure they were correct and fixed the ones that were not.

### DIFF
--- a/release/sil/sil_el_ethiopian_latin/HISTORY.md
+++ b/release/sil/sil_el_ethiopian_latin/HISTORY.md
@@ -1,6 +1,10 @@
 SIL EL - Ethiopian Latin Change History
 =======================================
 
+1.4.3 (2 Dec 2024)
+-------------------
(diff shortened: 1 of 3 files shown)


DavidLRowe (approved):

mcdurdin:
> Note that the messages about script `'Zyyy'` did not make sense to me. I just reviewed the language tags to make sure they were correct and fixed the ones that were not.

@srl295 and I talked about this one last night (I didn't get a chance to document yet). It arises because `mul` has no real script associated, so it gets the `Zyyy` "Code for undetermined script" tag by default. We should change `mul` to `mul-Latn` (sorry, I know that means another change for this keyboard)

PR #3234, opened 2024-12-03
[langtag_others] Update keyboards for most remaining langtag issues

LornaSIL:
Addresses some kbds in https://github.com/keymanapp/keyboards/issues/3221


release/j/jawa/source/jawa.kps
```diff
@@ -1,148 +1,151 @@
...
+      <Languages>
+        <Language ID="id-Java">Indonesia (Aksara Jawa)</Language>
+        <Language ID="jv-Java">Jawa (Aksara Jawa)</Language>
+        <Language ID="kaw-Java">Kawi (Aksara Jawa)</Language>
+        <Language ID="mad-Java">Madura (Aksara Jawa)</Language>
+        <Language ID="sas-Java">Sasak (Aksara Jawa)</Language>
+        <Language ID="su-Java">Sunda (Aksara Jawa)</Language>
+        <Language ID="osi">Osing (Aksara Jawa)</Language>
+        <Language ID="tes">Tengger (Aksara Jawa)</Language>
+      </Languages>
```
mcdurdin, line 134: This is curious to me -- all the languages in the list here had `Java` script subtag so I don't know why we were getting the hint:

```
jawa.kpj - hint KM09011: The script 'Java' associated with language tag 'jv-Java' does not match the script 'Latn' for the first language in the package.
jawa.kpj - hint KM09011: The script 'Java' associated with language tag 'kaw-Java' does not match the script 'Latn' for the first language in the package.
jawa.kpj - hint KM09011: The script 'Java' associated with language tag 'su-Java' does not match the script 'Latn' for the first language in the package.
jawa.kpj - hint KM09011: The script 'Java' associated with language tag 'osi-Java' does not match the script 'Latn' for the first language in the package.
jawa.kpj - hint KM09011: The script 'Java' associated with language tag 'tes-Java' does not match the script 'Latn' for the first language in the package.
```

I would like to investigate this before we commit this change.

LornaSIL:
Many of these didn't have a tag in langtags. I had to keep langtags open
and make additions there based on some of the langtags in keyboards.

mcdurdin:
> Many of these didn't have a tag in langtags. I had to keep langtags open and make additions there based on some of the langtags in keyboards.

Even so, with the explicit subtag, the script should have always been `java`. I need to check to make sure there isn't a logic bug in the compiler!

mcdurdin:
> Even so, with the explicit subtag, the script should have always been `java`. I need to check to make sure there isn't a logic bug in the compiler!

And ... there it is. If the tag as given in the package isn't found in langtags.json, e.g. in this case, `id-Jawa`, then the compiler gets confused and uses instead the default script subtag. Will resolve this, and re-report just to make sure we are not adding extra confusion to the mix.

LornaSIL:
I reviewed the new list and think these changes are correct. So, if someone can review it that would be helpful.

DavidLRowe (approved):

PR #3235, opened 2024-12-03
[sil_euro_latin] Update to keyboard

### LINKED ISSUE #2957: [sil_euro_latin] Update help for unmodified apostrophe `'`
The EuroLatin  Keyboard help (3.04)  has a lot of diacritics composed by typing a punctuation or symbol before a vowel. It seems these rules are written that if you press the punctuation or symbol key twice, the diacritic  does not appear. But this information is not mentioned in the keyboard help file. 

This is a useful  feature, for example in the French word **l'enfant**. You want apostrophe, then e, not e acute. 

A user who asked this question: https://community.software.sil.org/t/control-when-character-is-inserted/8971

### LINKED ISSUE #3013: [sil_euro_latin] don't auto-switch to default layer when on numeric layer
In German the comma is used as decimal separator. When on the numeric layer and typing digits followed by a comma, the keyboard switches back to the default layer. If you type a period instead, it stays on the numeric layer. 

Looking at the other keys on the numerical layer, they are all related to numbers and math, and so I think it would make more sense to disable automatic layer switching for the numeric layer. Typing any key on the numerical layer stays on the numerical layer. To go back to the default layer you press the layer button.

LornaSIL:
* Update language tags
* Update documentation for punctuation

Finished fix for #3221 
Fixes #3013


release/sil/sil_euro_latin/source/sil_euro_latin.keyman-touch-layout
```diff
@@ -2377,6 +2377,7 @@
                 "id": "K_1",
                 "text": "1",
                 "sp": 0,
+                "nextlayer": "numeric",
```
mcdurdin, line 2380: I would prefer that this be implemented in the .kmn rather than in the touch layout -- in the `begin postkeystroke` code path. The logic is there for digits already:

```
    c Okay, let's stay on the numeric layer if we are there already
    platform('touch') if(&newLayer = "") if(&layer = 'numeric') any(digit) > context
```

So we just need to create a new store `digits-and-punct` that includes the other outputs that should not switch layers, and use that there, I think.

Also, I don't think all the punctuation keys should be keeping us on the numeric layer, e.g. '"' should definitely go back to default. We need to consider each punctuation character carefully; the following are my starting thoughts:

`0`-`9` -- stay numeric
`.` and `,` -- these should go back to the default layer unless preceded by a digit.
`@` -- IMO go back to default layer, used most commonly probably in email addresses
`#` -- Debatable, the online elite use hashtags 😆 
`<`, `>` -- Debatable again... can be used for maths but also many other things. perhaps if preceded by digits plus optional whitespace?
`«`, `»` -- these are quotation marks, so go back to default
`+`, `-`, `*`, `/`, `^`, `%`, `=`, `≠` -- all of these have multiple uses (particularly `-` and `/`!). I don't think treating them as maths symbols only makes sense. If preceded by digits and optional whitespace?
`(` and friends -- these are much more commonly used in text than for numbers, I think. Should not stay in numeric layer
`"`, `!`, `` ` ``, `~`, `¡`, `?`, `¿`, `;`, `:`, `…` and `&` -- these are not numeric related. Go back to default layer.

So, in essence, I don't really agree with @ermshiperete's original suggestion, and I missed the issue when it was opened. We can add more nuance to the layer switch by sticking in the numeric layer when in the middle of a digit sequence and mathematical operators are used, but apart from that I think we need to drop back to default layer.

The rules would be something like (needs careful crafting):
```
    store(math-operator) ',.=+-*/' c ... and others
    store(spacing-math-operator) '=+-*/' c ... and others, but not ./,
    platform('touch') if(&newLayer = "") if(&layer = 'numeric') any(digit) any(math-operator) > context
    platform('touch') if(&newLayer = "") if(&layer = 'numeric') any(digit) ' ' any(spacing-math-operator) > context
```

release/sil/sil_euro_latin/source/sil_euro_latin.keyman-touch-layout
```diff
@@ -2767,11 +2800,13 @@
               {
                 "id": "U_0027",
                 "text": "'",
+                "nextlayer": "numeric",
                 "hint": "`",
                 "sk": [
                   {
                     "text": "",
```
mcdurdin, line 2807: this seems wrong

mcdurdin (changes requested):
Sorry to be a downer on this but I think we need some very careful analysis on nextlayer changes. I agree more nextlayer nuance would be good to have but currently this change is a step backwards (less nuanced!)

release/sil/sil_euro_latin/source/sil_euro_latin.keyman-touch-layout
ermshiperete: I'm fine with fine-tuning the rules when we do a layer switch. However, I think `:` should stay in numeric layer if preceded by a digit since that's used for time.

LornaSIL:
I'm reverting the changes to the mobile layout so this PR will not address #2957 as I do not feel confident in how to do it. It only address langtags and documentation.

LornaSIL:
@mcdurdin I won't be making the changes to the numeric layer in this PR so if you can check it just for the tagging and also documentation of `''` (I just updated the prefix image which only displayed a single quote) I'd appreciate it.

mcdurdin:
I removed `fixes #2957` from the original comment, hope that is okay. The documentation update looks fine.

release/sil/sil_euro_latin/source/sil_euro_latin.keyman-touch-layout
mcdurdin: OK, I will pull my comment into #3013, and let's try and spec it in detail.

mcdurdin (approved):

DavidLRowe (approved):

mcdurdin:
@LornaSIL wrote:
> I'm reverting the changes to the mobile layout so this PR will not address #2957 as I do not feel confident in how to do it.

Then I wrote:
> I removed `fixes #2957` from the original comment, hope that is okay. The documentation update looks fine.

Ugh. We both meant #3013. I'll fix this up. #2957 is fixed by this, #3013 isn't.

PR #3260, opened 2024-12-11
[indonesia] Fix touch layout for zxq

### LINKED ISSUE #3247: Bug on z layout
hi, im using indonesian layout keyboard, theres a problem with z caps, which if i double press z caps, the print isn't zz, but z and change to x change to q.

please repaired this 🙏

thanks before

LornaSIL:
Fixes #3247


DavidLRowe:
I think HISTORY.md needs a change as well. Otherwise looks good!

DavidLRowe (approved):

PR #3266, opened 2024-12-27
SLP1 - Roman keyboard

Harshavardhan-Pandurangan:
The Sanskrit Library Phonetic ASCII package includes two keyboards: one to enter Devanagari Unicode, the other to enter Roman Unicode with diacritics.  These include all accents required for The R̥gveda.  Both keymaps follow the Sanskrit Library ASCII encoding described in brief at https://sanskritlibrary.org/help-text.html, and in full in Linguistic Issues in Encoding Sanskrit, Appendix B available in a PDF linked to at the bottom of that webpage.  The Romanization followed in the slp1_roman keyboard is ISO 15919 except e and o are represented as in IAST without the macron.  Capital characters are obtained by wrapping the character sequence in curly brackets.


keyman-server:

LornaSIL:
This pull request is in better shape than your previous one. I haven't done a thorough check, but these are the issues I can see immediately.
- if you could remove the version number from the `README.md` file then that file won't need to constantly be updated. It is sufficient to include the version in the `HISTORY.md` file and in the .kmn file
- The .kvks file is empty. When you open the .kmn file and go to On-screen keyboard tab you can import the on-screen keyboard. That will make it more useful for your users. Otherwise the OSK will be blank.
- We do not allow distribution of proprietary fonts. That means that the `times.ttf` font may not be distributed here. It is a Monotype font with a restricted license. You'll need to remove the font and change the font in the .kps.
- The Sanskrit2020.ttf font is allowable because it appears to be under the Open Font License. However, we do not wish to have a lot of different font files with the same name in our repo, so when a font seems like it will be useful for several different keyboards, we prefer to add them to our `release\shared\fonts` section of the repo and then it can be used for various keyboards. I'll go ahead and add it there. When that's done I'll suggest what to do next.
   - I was looking in this repo: https://sourceforge.net/projects/advaita-sharada-font/files/Devanagari/ and it appears that `Sanskrit2020U.otf` is a more up-to-date version of the font. Is that version acceptable to you?

Harshavardhan-Pandurangan:
How can i ensure that `times.ttf` is used for the keyboard? Is there any other way to reference licensed fonts that are available for free. And yes both the current font for `Sanskrit2020` and the newer `Sanskrit2020U` version are fine with us.

LornaSIL:
The .kvks file is blank. You should "fill from layout" using the directions I listed in the other PR. Then you can select "Times New Roman" in the Font menu. If someone has Times New Roman installed it will use that font. If it's not installed (which seems unlikely) then it will fall back to something else.

PR #3274, opened 2025-01-09
[dakota_smsc] Initial submission of the dakota_smsc keyboard

butanial:
Thank you Lorna for your help and patience!


keyman-server:

DavidLRowe:
Thanks for your submission. It's in pretty good shape. Here are some comments on the file contents:

LICENSE.md
- The Copyright line needs a name (of an individual or organization). From the dakota_smsc.kmn file, it seems that should be: Shakopee Mdewakanton Sioux Community
- The date should be 2025 (or include a range that ends with 2025, since it seems you've been working on this keyboard for some time. 
- Perhaps you want something like: Copyright © 2024-2025 Shakopee Mdewakanton Sioux Community

README.md
- The "Keyboard Homepage:" needs to include the keyboard project folder name:
https://keyman.com/keyboards/dakota_smsc
- Once the keyboard is approved and merged, information about it will appear at this address.

dakota_smsc.kmn
- In the &TARGETS store, please use just 'any' (since that implies all the other platforms): store(&TARGETS) 'any'
- In the &COPYRIGHT store, we encourage you to omit the date, so that it doesn't have to be changed when updating the file:
store(&COPYRIGHT) 'Copyright © Shakopee Mdewakanton Sioux Community'
- In the &NAME store, I'd suggest 'SMSC Dakota'. Since this is the name under which the keyboard will be published, it would be better to include the language name.
- I think you can delete the line:
store(&KMW_EMBEDCSS) 'dakota_smsc.css'
since it refers to a file that is not included in the keyboard files submitted. (Most keyboards don't use embedded CSS and don't need this line or the .css file.)

DavidLRowe:
I have some comments on the keyboard contents. It generally looks very good. That's quite an impressive array of emojis!

For the touch layout, you have both a tablet and a phone layout, but they seem to be very similar. In our experience, it is a lot of extra work to keep identical tablet and phone layouts synchronized. Since Keyman will use a tablet layout if a phone layout isn't available, and vice versa, it makes sense to maintain just one and remove the other. 

There are a few differences in the two that you have. For example, for the tablet layout, default layer, there are longpress keys for ! and ? on the period key, but the phone layout doesn't have these. (Also note that there's a third empty longpress key on the period that shows a blank and doesn't produce any output, which might be confusing for a user.)

I'd suggest that you make sure everything you want is on the phone layout, then delete the tablet layout. Keyman will then use the phone layout for both phones and tablets. 

You have two different ways to enter a "t" with a dot above. Typing a "t" then a dot above produces "ṫ" and typing the longpress produces "ṫ". Although they look the same, the first is U+‎0074 (LATIN SMALL LETTER T) followed by U+‎0307 (COMBINING DOT ABOVE) and the second is U+‎1E6B (LATIN SMALL LETTER T WITH DOT ABOVE). You may want to include additional statements in the .kmn file that will always generate the latter form (both for desktop layout and touch layout).

DavidLRowe:
PS I hope the feedback is not too daunting! We hope to help you make the keyboard the best it can be.

butanial:
Thank you David, I appreciate the feedback! I want this to be a good keyboard also. It's my first keyboard and my coding skills aren't what they used to be. 😉 I've made the changes requested.

DavidLRowe:
All the changes look good, except for the change in the longpress "t" with dot above. You changed the ID from U_1E6B to U_1E6B[_0307]. (Following the syntax for multiple Unicode values, that should really be U_1E6B_0307, but surprisingly it works with the square brackets.) This ends up producing a "t" with dot above (U+1E6B) and then puts a dot above (U+0307) that, which is not what you want ("t" with two dots stacked above).

The question of different sequences applies not just to "t", but to all the vowel + acute accent and consonant + dot above combinations.

I see several possibilites:
1. Put the longpress "t" with dot above back to having ID U_1E6B and let the keyboard create different sequences for all those characters. 
2. Change all the longpress sequences to create multiple values, that is, U_0074_0307 for "t" with dot above, U_007A_0307 for "z" with dot above, U_0061_0301 for "a" with acute accent, etc.
3. Add code in the .kmn file that will detect sequcenes such as U+0074 U+0307 and replace them with the appropriate single value (U+1E6B). (For an example, see the bottom of the [manga_latin_qwerty.kmn](https://github.com/keymanapp/keyboards/blob/master/experimental/m/manga_latin_qwerty/source/manga_latin_qwerty.kmn) file, where basevowel plus combining character sequences are converted to individual characters.) 

And another possibility is to do the first option now, then work on number 3 in a later version.

butanial:
Hi David, I opted for changing it to U_0074_0307. Thank you!

DavidLRowe:
Hmm. Although I wrote about the "t" key, that was just a representative. The comment applies to the five vowels, plus all seven consonants: cghpstz, as well as the upper case of those twelve.

To implement the code to normalize the keyboard output, select the "Layout" tab on the left, then the "Code" tab on the bottom. That will show you the underlying Keyman rules represented by the keyboard image in the Design view. If you then paste in the following code at the bottom of the existing file:

```
match > use(NFC)

group(NFC)

store(basevowel) 'aeiouAEIOU'
store(acute) U+00E1 U+00E9 U+00ED U+00F3 U+00FA U+00C1 U+00C9 U+00CD U+00D3 U+00DA
any(basevowel) U+0301 > index(acute, 1)

store(basecons) 'cghpstzCGHPSTZ'
store(dotted) U+010B U+0121 U+1E23 U+1E57 U+1e61 U+1E6B U+017C U+010A U+0120 U+1E22 U+1E56 U+1e60 U+1E6A U+017B
any(basevowel) U+0307 > index(dotted, 1)
```

whenever one of the keys matches, the "NFC" group will be applied to combine any base and diacritic pair in the list with a single code point. This will make the output of your keyboard consistent (for both desktop and touch layout).

Note that if you do paste in that code, you'll no longer be able to use the Design view. (You'll get a message: "The Design tab is disabled because the keyboard file is too complex to represent and modify visually. You must make any further changes to the keyboard directly in the Code tab.")

I'll approve this keyboard as it stands (with some modifications to the PHP help file) and you can see about adding the normalization code when you make the next version (1.6 perhaps).

DavidLRowe (approved):

DavidLRowe:
The .kps file had `<GraphicFile>..\smsc_logo.png</GraphicFile>` but the smsc_logo.png file was not present, so the package could not be built. I removed the reference to the missing file. (But then it still failed.)

DavidLRowe:
Also had to remove the file reference from later in the .kps file. Trying again!

DavidLRowe (approved):

DavidLRowe:
Thanks for your submission and your patience as we worked to get it through the system!

We did modify the .php file and the .kps file. You'll want to download copies of those to make sure the versions you have are up-to-date. You can use these links and click the download icon for each:
- https://github.com/keymanapp/keyboards/blob/master/release/d/dakota_smsc/source/dakota_smsc.kps
- https://github.com/keymanapp/keyboards/blob/master/release/d/dakota_smsc/source/help/dakota_smsc.php

Oops. I see now that the .php file references the images, so we'll need a copy of those images in the "help" folder. I'll submit a PR to fix that.

butanial:
Thank you for all your help and approving the keyboard! Our community is very excited!

PR #3275, opened 2025-01-10
[winchus_php] Changes to php to match welcome file and display online

LornaSIL:
Help file isn't displaying properly:
https://help.keyman.com/keyboard/winchus/0.9.7/winchus

Also, the contents seem to be from a previous version of documentation so I pulled in contents from welcome.htm.

html structure is bad, but I didn't attempt to clean that up.


DavidLRowe (approved):
RSLGTM

alex-larkin:
Thanks so much! 

I had noticed that the margins were super wide on mobile, and know that it has a lot of issues, but hadn't had a chance to work on it. I really appreciate your improvement! And I know that I still have a lot of work to do on it. Especially, the CSS and getting all the tables inserted.

PR #3277, opened 2025-01-13
update mjh_hudum only

gregeck:
Initially updated all 5 keyboards.
Then realized that there would be no way for keyman to pull one keyboard at a time.
Reverted all changes to 5 keyboards so as to be consistent with 
  https://github.com/keymanapp/keyboards/tree/master/experimental
Moved m/mjh_hudum to mjh/mjh_hudum
Uploaded changes only specific to mjh_hudum
Will do 4 individual pull requests for aligali, todo, sibe, and manchu after this hudum pull request executed successfully


keyman-server:

LornaSIL (approved):
The keyboard does build. Since we did a move from "m" to "mjh" I'm finding it a bit confusing. 
- Also, there were some "white space" only changes to your archived .kmn files in other projects. Try to only make changes in one keyboard per pull request
- There was a test file you added that needed deleting. I think we had to do that before. You should probably delete it. It was in the "keyboards" folder.
- I changed the version for `1.01` to `1.0.1` as that is the way we prefer versioning to happen. So, for future keyboards use that scheme.
- I moved the recent change in HISTORY.md from the bottom to the top as we prefer to see the latest change to a keyboard at the top (not the bottom)

gregeck:
Noted with thanks  on all four mjh_hudum items.
Should I go on to the other four keyboards?

LornaSIL:
Try another one.

mcdurdin:
> * I changed the version for `1.01` to `1.0.1` as that is the way we prefer versioning to happen. So, for future keyboards use that scheme.

Please note -- this is not just a preference. `1.01` is treated the same as `1.1` -- prefix zeroes are ignored. See https://semver.org for more detail.

PR #3287, opened 2025-01-15
added releases/a/arabic_olpc

Caliph3ubaidUrRe7man:


keyman-server:

LornaSIL:
I've just had a quick look at your keyboard. In general your pull request is in really good shape. I have a few comments that would be helpful to think about.

- I looked at the OLPC link. Is this the only OLPC keyboard you plan to submit? If so, the naming is fine. If you think you will submit more, I would recommend changing the foldername/filenames to `olpc_arabic` (reverse the word order) and put it under a new folder in `release` called `olpc`. That way all the OLPC keyboards would be grouped together. That's just a suggestion for you to consider if you plan to submit more OLPC keyboards.

These next issues need to be done, they aren't just suggestions.
- Please delete the following files from the Pull Request:
   - arabic_olpc.kpj.user (this is just a temporary file)
   - `build\arabic_olpc.js` (these are compiled files and our build system will rebuild this file)
   - `build\arabic_olpc.kmx` (these are compiled files and our build system will rebuild this file)
   - `build\arabic_olpc.kvk` (these are compiled files and our build system will rebuild this file)
- In the .kmn file, could you go to the On-screen keyboard tab and import the layout? The .kvks file is empty and that is not helpful to the users of the keyboard. Then you'll need to recommit the .kvks file
- In the .kps file, you should add the language tag in the **Keyboards** tab. I suspect it would just be `ar` unless you know this keyboard supports other languages. The mobile keyboard won't even install if you don't have a language tag.
  - It would also be helpful to add more description in the **Details** tab about why someone would want to use the keyboard. That information will show up at the top of the keyboard home page like this one: https://keyman.com/keyboards/sil_euro_latin

LornaSIL:
We need to get through one keyboard before you start adding more keyboards. Please address removing the other keyboards as well as the build files.

PR #3288, opened 2025-01-15
[dakota_smsc] Fix help file

LornaSIL:
Help file doesn't display. `<style>` doesn't have and end.

Also removed some of the keyboard installation instructions since those aren't recommended to include. The online help for Keyman will always be the most up-to-date.


DavidLRowe (approved):
Thanks for cleaning up after me!

DavidLRowe:
@butanial You'll need to pick up the changes in the .php file.

PR #3293, opened 2025-01-17
chore: upgrade to kmc 17.0.333

mcdurdin:
One relevant change in 17.0.333:

* fix(developer): filter incorrect fonts out of .keyboard_info (keymanapp/keyman#12913)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,54 +9,54 @@
       "version": "1.0.0",
       "license": "MIT",
       "dependencies": {
-        "@keymanapp/kmc": "17.0.332"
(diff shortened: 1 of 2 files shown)


DavidLRowe (approved):
RSLGTM

darcywong00:
Do we need to bump the multi-keyboard packages to fix keymanapp/keyman#12852?

mcdurdin:
> Do we need to bump the multi-keyboard packages to fix [keymanapp/keyman#12852](https://github.com/keymanapp/keyman/issues/12852)?

No, because .keyboard_info files are refreshed without version bumps.

PR #3299, opened 2025-01-23
Easy Arabic Keyboard

Caliph3ubaidUrRe7man:

PR #3300, opened 2025-01-23
[easy_arabic] Revises keyboard PR

LornaSIL:
Replaces #3299 

- Changed the name to remove the word `_keyboard` as that is redundant.
- Added help file
- Put fonts in `release\shared` folder so other keyboards will benefit from the font.

@Caliph3ubaidUrRe7man


DavidLRowe (approved):
RSLGTM

PR #3303, opened 2025-01-24
chore(taigi_viet_telex): fix missing index parameter

mcdurdin:


LornaSIL (approved):
Oops. Thanks! LGTM

mcdurdin:
This was picked up by the v18 compiler -- earlier versions of the compiler didn't find it...

Issue #3296, opened 2025-01-22
Fix version numbers


LornaSIL:
@mcdurdin we have these version numbers for some of the keyboards. Do I need to "fix" those?
1.01
0.51
1.21
1.05
2.0.25
1.26

Keyboards:

- Udi(SIL)
- ጉራጌ (Gurage) Legacy
- Easy Chakma
- Vietnamese Telex
- Burushãski Girminas
- Burushaski Girminas (بروشسکی)

mcdurdin:
Only `1.01` and `1.05` are potentially ambiguous. They should either be `1.1` or `1.0.1`, and `1.5` or `1.0.5` respectively.

Each component of the version number is an integer, so `1.01` is major version `1` and minor version `1`.

The components are called major, minor, patch. https://semver.org/ describes how we try and interpret the version numbers (although for keyboard purposes, the semantic meaning of the components is not very important -- just that version numbers go up).

dyacob:
I thought we fixed this for the Gurage keyboard a year ago or more -??  It must have been for another one.  Unless someone is poised to make the Gurage Legacy version fix, I can do so this weekend.

LornaSIL:
Please do! I didn't have time last week. If memory serves, we did the other
one, not legacy.

LornaSIL:
https://keyman.com/keyboards/rac_shina also needs version adjustment.

LornaSIL:
Are you going to fix the gurage one @dyacob or do you want me to?

dyacob:
@LornaSIL, thanks for asking, I took an unexpected side journey with COVID, which has passed, but now I'm behind on everything and catching up.  Please do make the fix if you have any spare moments. Thank you.

LornaSIL:
I'm sorry about your side journey! I wish you good healing.

PR #3305, opened 2025-01-27
[versions] Fix versioning for several keyboards

LornaSIL:
Fixes most of #3296 except for Gurage one which @dyacob will fix


DavidLRowe:
Thanks! Looks great except that burushaski_girminas_perso_arabic.kps has version 3.0.0, whereas HISTORY.md and .kmn file have 2.0.0. I was tempted to change the .kps value to 2.0.0, but wanted to check with you first, so I'll let you make that change.

DavidLRowe (approved):

PR #3319, opened 2025-02-05
[irula_tamil_winscript] New keyboard

LornaSIL:
Still a work in progress as I need more feedback.


DavidLRowe (approved):

DavidLRowe:
I removed `</head>` from the .php file and change the version number in HISTORY.md to match that in the .kmn file.

PR #3323, opened 2025-02-07
Tengwar

Caliph3ubaidUrRe7man:


LornaSIL:
In general this looks very good.
- I think we will want the tengwar font in experimental as well, so `experimental/shared/fonts/tengwar/TengwarFormalCSUR.ttf`
- That will mean updating your .kps file to reference the correct path.
- The other issue I discovered in one of your other keyboards is that the graphics aren't displaying properly in the online help file because they are too large. If you could resize the graphics in the `help` folder to be between 800-900 pixels they will display better.
- I would really like the description everywhere changed from `Tengwar Keyboard--As seen on https://freetengwar.sourceforge.net` to `Tengwar Keyboard--As seen on https://freetengwar.sourceforge.net. This keyboard is using Private Use Area codepoints since Tengwar is not in Unicode.`

I haven't tested the keyboard yet.

LornaSIL:
You can try that, but I don't think the php code respects resizing that way. I meant to actually resize the images.

mcdurdin:
> You can try that, but I don't think the php code respects resizing that way. I meant to actually resize the images.

That would work, because PHP is just a wrapper for HTML

PR #3334, opened 2025-02-11
[tengwar_quenya] Rename keyboard and delete prior tengwar files in experimental

LornaSIL:
Add online help for `legacy/t/tengwar` keyboard since the one in experimental got added to the `legacy/t/tengwar` keyboard (that is copied to another repo, so somehow that other repo needs updating and I'm hoping this does it).

The `legacy/t/tengwar` keyboard currently has the osk as well as all the metadata added to the home page (https://keyman.com/keyboards/tengwar). I'm hoping that will revert to the metadata it should have once we rename the keyboard in `experimental`.

Replaces #3332


LornaSIL:
For the record, the included fonts are under OFL and GNU.

DavidLRowe (approved):

PR #3335, opened 2025-02-11
[programmer_dvorak] New keyboard

Caliph3ubaidUrRe7man:


LornaSIL:
I'll be taking a few days to focus on other work. I'll try to check this in a few days.

LornaSIL:
In general this looks to be in good shape. There is one significant issue I think you will want to fix. The word "programmar" is misspelled and should be "programmer" (at least I don't think this was an intentional spelling). That means in all the filenames and foldername it needs correcting as well as all the keyboard names in the files.

One other thing. When I keyboard does not contain a mobile layout, the `.js` file should not be included in the files in the `.kps` file.

LornaSIL:
Thanks. Can you remove the folder and files for `programmar_dvorak`?

Caliph3ubaidUrRe7man (commented):
whats the point of the keyboard if its only on the web?

LornaSIL:
> whats the point of the keyboard if its only on the web?

The `desktop` target encompasses "Windows Linux and macOS". So, list Desktop, Windows, Linux, macOS is redundant.

LornaSIL (approved):
LGTM

Issue #3333, opened 2025-02-10
bug(tengwar): two keyboards with the same name


mcdurdin:
```
$ find . -name 'tengwar'
./experimental/shared/fonts/tengwar
./experimental/t/tengwar
./legacy/t/tengwar
```

This has caused the upload to fail due to the duplication.

The keyboard under experimental/t/tengar needs to be renamed unless it is a keystroke-compatible replacement for legacy/t/tengwar (which is longstanding).

Secondary - we should be able to detect these overlaps at build time (looks like detection is only in place for experimental/release, and does not check legacy folder).

DavidLRowe:
@Caliph3ubaidUrRe7man Unfortunately, the "tengwar" keyboard we approved recently will need to be renamed (unless, as noted above, it is keystroke for keystroke the same as the legacy one, in which case I believe we could deprecate the legacy one in favour of the new one).

Caliph3ubaidUrRe7man:
what would you like to rename it?

mcdurdin:
> Secondary - we should be able to detect these overlaps at build time (looks like detection is only in place for experimental/release, and does not check legacy folder).

I have moved this to #3338.

mcdurdin:
This was supposed to be fixed in: 

* #3334 

But upload is still failing but with a different error now:

```
07:04:30   Uploading /c/BuildAgent/work/837a1c4afc30068e/keyboards/legacy/t/tengwar/
07:04:30   parse error: Expected separator between values at line 18, column 19
07:04:30   parse error: Expected separator between values at line 18, column 19
07:04:30   parse error: Expected separator between values at line 18, column 19
07:04:30   parse error: Expected separator between values at line 18, column 19
07:04:30   Failed to locate /c/BuildAgent/work/837a1c4afc30068e/keyboards/legacy/t/tengwar/build/
07:04:30   Aborting with error 1
```

Investigating.

PR #3339, opened 2025-02-12
fix: add missing comma in tengwar.keyboard_info

mcdurdin:
The patch in #3336 missed a comma, which broke the release build. (The files in legacy/ are more fragile and do not have many build checks, as they generally do not get modified.)

Fixes: #3333
Emerging-From: #3336


LornaSIL (approved):
So sorry! It's been awhile since I edited a keyboard_info file. LGTM.

PR #3356, opened 2025-02-21
[sil_kcho] Initial creation

darcywong00:
Keyboard submission on behalf of users who requested the following:

Keyboard for Kcho languages:  (Mün Chin `mwq` and Daai Chin `dao` languages in Myanmar)
Outputs NFC when possible.

The language group has this extra vowel (accessed with Desktop keys)
vowels

and any vowel can have the following 4 tones:
tones

The Right-Alt layer has the keys that got replaced.

Note: Hopefully the online help osk works when this gets released...


LornaSIL (approved):
LGTM

PR #3367, opened 2025-03-08
[jerriais] add keyboard for Jèrriais

simlrh:
Hello, first time contributor wanting to add a lexical model and a keyboard layout for the Jèrriais language to Keyman. Apologies for anything that I haven't done correctly.

The build system gives a warning that the BCP 47 tag `nrf-je` is not canonical and should be `nrf`. I'm reluctant to use `nrf` because this tag is for "Norman", of which there are at least 3 variants with different orthographies. This keyboard wouldn't be useful to a Guernésiais speaker for example.

Brief context, Jèrriais is a romance language spoken in the British Isles, so we need an easy way to enter diacritics but we still want to use a UK keyboard layout. Hence the minimal desktop keyboard.


keyman-server:

DavidLRowe:
https://www.ethnologue.com/language/nrf/ claims "Guernésiais is an endangered indigenous language of Guernsey and Jersey. It belongs to the Indo-European language family. Direct evidence is lacking, but the language is thought to be used as a first language by the elderly only. It is not known to be taught in schools." 

It sounds as though you would say that the orthography of Jèrriais is distinct enough from Guernésiais that the two could not use the same keyboard (and presumably also would need different lexical models).

DavidLRowe:
As for the pull request itself, it is in very good shape. The only request is that you complete the On-Screen Keyboard (OSK). In Keyman Developer you should be able to click on the "On-Screen" tab on the left, then tick the "Auto-fill from underlying layout" at the bottom, then use the "Fill from layout" button at the top. This will avoid a user seeing a blank keyboard when asking for the on-screen keyboard.

simlrh:
> It sounds as though you would say that the orthography of Jèrriais is distinct enough from Guernésiais that the two could not use the same keyboard (and presumably also would need different lexical models).

Actually I'm looking at some examples of other Norman dialects and it is *possible* that this keyboard might be useful to them - I don't know enough about them to say for sure. But certainly they would need a different lexical model, because the words are different.

I will fix the OSK as soon as I can. Thanks!

simlrh:
Ready for review

simlrh:
I've also looked further into the `nrf` issue, and it seems there's quite a lot of confusion about what it refers to. Wikipedia uses `nrf` for all Norman languages. [SIL]( https://iso639-3.sil.org/code/nrf) says just Jèrriais and Guernésiais. Ethnologue confusingly says that Guernésiais is the indigenous language of Jersey (probably because it expects one code to be associated with one language name, which isn't the case if `nrf` refers to Jèrriais & Guernésiais). This [IANA list](https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry) includes Serquiais.

As for this keyboard, it could potentially be used for Guernésiais but definitely not for Serquiais, so I think the simplest thing to do is keep it as `nrf-je`.

DavidLRowe:
One minor change: With the Packaging tab (jerriais.kps), Details tab on the left, in the Welcome file field, select "welcome.htm". (Sadly, the New Project feature currently fails to fill in that field.) Everything else looks great!

Your nrf-JE proposal sounds reasonable.

https://www.ethnologue.com/contributor-program/ contains information on contributing updates to Ethnologue. It sounds as though you could provide insights into the language situation on Jersey. (It also discusses receiving free access to Ethnologue data, if that's something that interests you.)

DavidLRowe:
I've been informed that the Welcome file field currently doesn't need to be filled in. If a "welcome.htm" file is present, it is automatically picked up. You can fill in that field in a later revision.

DavidLRowe (approved):

DavidLRowe:
Keyboard is available at https://keyman.com/keyboards/jerriais

PR #3378, opened 2025-03-17
fix(khmer_ldml): encoding of `<` should be `&lt;`

mcdurdin:
Should not need a new version of the keyboard, just to pass Keyman Developer 18.0 checks.


LornaSIL (approved):
LGTM

PR #3380, opened 2025-03-17
fix(winchus): fix broken U_2B1B touch layout key

mcdurdin:


alex-larkin:
Thanks so much Marc! I'm just so curious (😂), with all of the keyboards and everything going on, how did you happen to find that spurious space?

LornaSIL:
He's doing a lot of testing for Keyman 18.0

mcdurdin:
The new 18.0 compiler found it for me. It's doing more validation of keyboards and picking up a number of things that slipped past us all.

alex-larkin:
Neat! I figured it was something like that.

PR #3385, opened 2025-03-18
[slp1_roman] Updated keyboard rules and added touch

### LINKED ISSUE #3314: bug(slp1_roman): Invalid virtual key sequences in stores
This is not valid .kmn syntax. `[...]` is treated as a virtual key, so this will not be doing what you expect. In Keyman Developer 18, these are detected as invalid, because they are not valid key codes:

```
slp1_roman.kpj - info KM05002: Building release/s/slp1_roman/slp1_roman.kpj
slp1_roman.kmn - info KM05002: Building release/s/slp1_roman/source/slp1_roman.kmn
slp1_roman.kmn:76 - error KM0200A: Invalid token found
slp1_roman.kmn - info KM05007: release/s/slp1_roman/source/slp1_roman.kmn failed to build.
slp1_roman.kpj - info KM0500C: Project release/s/slp1_roman/slp1_roman.kpj failed to build.
```

_Originally posted by @mcdurdin in https://github.com/keymanapp/keyboards/pull/3298#discussion_r1938785904_

LornaSIL:
Fixes #3314 

In addition to the `store` issue, the keyboard included a `.js` file. Rather than deleting that file from the .kps I went ahead and added a tablet layout based on the OSK.


mcdurdin (changes requested):
This seems like a change in intended behaviour. The intent seems to be to treat `U+0061 U+0304` as a single element. Might be better to replicate the grouped characters in rules, even though it's a bit wordy, e.g. for the first grouped characters in vowels store (U+0061 U+0304):

```
any(vowels) + U+007E > context U+0303
U+0061 U+0304 + U+007E > context U+0303
U+0061 U+0304 U+0301 + U+007E > context(1) context(2) U+0303
```

I have not done a full analysis at this point so some of the cases may be covered by other rules. I guess I would expand out all the examples into rules first -- there are only 5 vowels rules and 1 consonant rule so it's not going to be too dramatic to expand. Then review the expansions to determine which are covered by other rules (perhaps most of them?)

If this seems like too much effort at this point, perhaps @Nnyny could lend a hand too? (This is also a good candidate for LDML in the future given the attempted use of grouped chars which work well in LDML transforms.)

LornaSIL:
I started to do something like your suggestion, but I ended up realizing
prior rules were outputting the combining marks and it didn't appear the
store rules were being utilized.

LornaSIL:
If @Nnyny wants to close this and tackle it differently, that's fine with me. I did spend a good bit of time yesterday reviewing the keyboard and the documentation and testing both the online one and my changes and I don't feel like there is a regression in actual behavior, but that's up to you.

mcdurdin:
> I did spend a good bit of time yesterday reviewing the keyboard and the documentation and testing both the online one and my changes and I don't feel like there is a regression in actual behavior, but that's up to you.

Okay, sorry, I didn't realize you'd already done this investigation to that level of detail. As you are confident that there is no regression, let's just go ahead with your changes and not make @Nnyny redo that work!

mcdurdin (approved):
LGTM

Issue #3390, opened 2025-03-21
bug(taigi_viet_telex): invalid 'index()' statement


mcdurdin:
https://github.com/keymanapp/keyboards/blob/master/release/t/taigi_viet_telex/source/taigi_viet_telex.kmn#L517

Last `index()` statement is missing an offset. v18.0 compiler detects this, v17.0 misses it (many thanks to @markcsinclair for QA improvements to the compiler!)

PR #3391, opened 2025-03-21
chore: fix broken `index()` in taigi_viet_telex.kmn

mcdurdin:
Fixes: #3390


LornaSIL:
Does the keyboard version need bumping or is there no change in behavior?

mcdurdin:
> Does the keyboard version need bumping or is there no change in behavior?

Yes, that rule would have not worked correctly, so I will bump the version. Sorry I didn't do that earlier.

DavidLRowe (approved):

PR #3393, opened 2025-03-21
[ho_latin] New keyboard

Singitur:
This is a keyboard for Ho, which is also written using the Latin alphabet in addition to Warang Citi.


keyman-server:

Singitur:
Ready for review

keyman-server:

LornaSIL:
Thanks for your submission.
Some comments:
- The LICENSE.md file must have the word Copyright" in it. So, line 3 needs to be `Copyright © 2024-2025 Mangu Purty`. And you should change or extend the year to 2025. I know that sounds fiddly, but our build system checks for the word `Copyright`.
- The .php file must be formatted with the correct header. Replace lines 1-28 with this:
```
<?php
  $pagename = 'Ho Latin Keyboard Help';
  $pagetitle = $pagename;
  $style = <<<EXTRA
  .touch {width: 40%; }
  .symbol{width: 25px; }
  .keyacute{height: 12px; width: 12px;}
  .key{height: 12px; width: 18px;}
EXTRA;
  require_once('header.php');
?>
``` 

- You have two welcome.htm files. You should delete the one that is not correct, or if they are duplicates, delete the one in the `source` folder and all associated graphics. Just keep the graphics in the `welcome` folder.
- I don't understand the need for this file `ho_latin-help.htm`. It seems to be the same as the `welcome.htm`.
- `ho_latin-codes.txt` could be deleted. It is empty and serves no purpose.
- In the .kps file:
   - (under **Files**) you must include the graphics for the welcome.htm file or it will not display properly.
   - You must also include `LICENSE.md` in the list of **Files**.
   - In the **Details** tab you must include a description of the keyboard. This will show up on the home page of the keyboard.
- In the .kmn have you tested the keyboard? I am not getting the behavior as documented in the OSK. Sometimes people think if they modify the OSK it will change the rules of the keyboard, but it does not. You must include the rules in the .kmn in order to produce the characters that you have documented in the OSK. For example, if I try to type the p with acute, I get an "f"
This is the same for other keys such as the backtick to get the t with dot below. 
image

You'll need to add rules for each of those, like this:
```
+ "p" > U+1E55
+ "P" > U+1E54
```

One last thing, you have rules for semicolon so typing `a;` turns into "á". There's nothing wrong with a rule like that and it works fine. I would just recommend providing a way for the user to type a semicolon when they want it. You could put it on an unneeded key. Since you have characters on ALT SHIFT you could perhaps add it to that layer. Then make sure to document it. I think you will need to reimport your OSK after you adjust all the rules. Then select **Auto-fill underlying layout** and then **Fill from layout**.

I know this is a lot of feedback, but it's great you are producing a keyboard the Ho people can use.

Singitur:
Thank you for the detailed review. I have made the necessary changes.
- [x] "Copyright" included in LICENSE.md file.
- [x] Formatted .php file.
- [x] No duplicate welcome.htm files.
- [x] ho_latin-help.htm deleted.
- [x] ho_latin-codes.txt deleted.
- [x] Graphics for the welcome.htm file, LICENSE.md, HISTORY.md and README.md added in the .kps file. Keyboard description added. 
- [x] OSK modified, output is in line with the documentation.
- [x] `ALT + ;` produces `;`, the same goes for other symbols (backtick and tilde).

LornaSIL (approved):
Looks good. Thanks for updating those issues so quickly.

Issue #3331, opened 2025-02-10
Khmr: Convert Khmer keyboards to use Busra instead of Khmer Mondulkiri


LornaSIL:
#3330 This issue should be approved before we begin the process.

These are the Khmr keyboards, but not all of them will use Mondulkiri:
https://keyman.com/keyboards/khmer_angkor
https://keyman.com/keyboards/sil_jarai
https://keyman.com/keyboards/khmer_advanced
https://keyman.com/keyboards/krung
https://keyman.com/keyboards/sil_brao
https://keyman.com/keyboards/sil_bunong
https://keyman.com/keyboards/sil_khmer
https://keyman.com/keyboards/basic_kbdkhmr
https://keyman.com/keyboards/basic_kbdkni

Also, do not modify the Displaymap settings (and the KbdKhmr font) for keyboards that use that:
```
store(&DISPLAYMAP) '../../../shared/fonts/kbd/kbdkhmr/KbdKhmr.json'
```

PR #3397, opened 2025-03-24
[busra_kbds] Change Khmr keyboards to use Busra instead of Mondulkiri

LornaSIL:
Making the Khmr keyboards consistently use "Busra" instead of "Khmer Mondulkiri" or "Khmer Busra".
One help file `easy_chakma` also referenced "Khmer Mondulkiri" unnecessarily.

Fixes: #3331


LornaSIL:
Fixes #3331

DavidLRowe (approved):

DavidLRowe:
I note that in khmer_advanced.kps the version listed is 1.0.1 (rather than the updated 1.1.0). Since the "FollowKeyboardVersion" option is present, I think that the package will pick up the updated version from the .kmn file.

Similarly khmer_angkor.kps has version 2.1 (rather than the updated 2.1.1).

@mcdurdin Having these old version numbers in the .kps file is confusing. Do you think it would be feasible to delete any existing version number when the "Package version follows keyboard version" box is ticked? (And not only at the moment that the box is ticked, but also when writing out the file, delete any existing version number if the "FollowKeyboardVersion" option is set.)

mcdurdin:
> @mcdurdin Having these old version numbers in the .kps file is confusing. Do you think it would be feasible to delete any existing version number when the "Package version follows keyboard version" box is ticked? (And not only at the moment that the box is ticked, but also when writing out the file, delete any existing version number if the "FollowKeyboardVersion" option is set.)

I have opened keymanapp/keyman#13576

mcdurdin:
> Fixes #3331

This can go in the PR body to automatically link and close the parent issue (if it is in a comment, then we lose the automation)

PR #3400, opened 2025-03-25
[masaba] This is a Bambara keyboard that uses the Masaba script.

oyousuf:
This is a Bambara keyboard that uses the Masaba script, a script that is used in western Mali. Masaba has not yet been encoded into Unicode.


keyman-server:

oyousuf:
Ready for review.

LornaSIL:
Thanks for submitting these keyboards! Again, I have a lot of feedback.
- It would be useful for you to update to Keyman Developer 17.0. Some of the issues I'm finding could be resolved with the update.
- However, because I think you created these keyboards last year or previously, our build structure has changed. So, here are changes because of our build structure:
   - Please delete `masaba.keyboard_info`. The information in this file should all be placed in the .kps
   - In the `LICENSE.md` make sure the word `Copyright` is added and since you made the keyboard in 2024, extend the copyright date to include 2025: `Copyright © 2024-2025 Oreen Yousuf`
- The graphics in the welcome.htm file are simply too large to display the whole keyboard layout. Either you should resize the graphic image, or change the size with html. You could change the image reference to include a width like this: 
`Keyboard Layout`
-  For all the descriptions in: README.md, readme.htm, welcome.htm, masaba.kps, masaba.php, please add a statement similar to what we did for lakeside: `The Masaba script is not yet included in Unicode.`
- When you open `masaba.kpj` with Keyman Developer 17.0 it will ask
image
If you go ahead and say "Upgrade project" you will see it has retained some old history that needs fixing. (The new project file reduces the complexity of the .kpj file and you end up with less errors.) However, now we have to fix some things. You can see it still has the `nti.kmn` file that you were probably working on associated with the keyboard:
image
Click on "Options" (to the right of `nti.kmn` and say "Remove from Project"
image
- Then click on "Upgrade Project" and it should upgrade just fine.
- There are a lot of messages in the messages window when I try to compile. This is because you are using some IPA characters on the left side of the `+`. Everything on the left side of the `+` should be a keyboard key you can type. The hardware keyboard doesn't know about eng, left hook n, open o, epsilon. So, I think you have duplicate rules:
```
U+E009 + "e" > U+E008 c pɛ
U+E009 + "ɛ" > U+E008 c pɛ
```
Where you can delete the one with the epsilon because that one will never fire. There are quite a few like that. Those need to be resolved.

- Have you tested the mobile layout? I'm not seeing a way to access the layers that are called `nasal*`.
- In the .kps file
   - Please delete `masaba.ico` (that file is compiled into the .kmx)
   - On the **Keyboards** tab, the BCP-47 code should be `bm-Zzzz` (`bm` would mean `Latn` script). Once the script is added to Unicode it will get a proper ISO 15924 code)
   - We are going to need the pdf renamed so it doesn't have any spaces in the name. Then you'll need to remove the existing pdf from the list of files and readd it to the file list at this current location. (Right now it's looking somewhere else on your system). Maybe "MasabaTypingManual.pdf` would be a good short name.
   - You'll need to remove the font from the list of files and then re-add the font from this current location. (Right now it's looking somewhere else on your system). Then remember to add it back on the **Keyboards** tab.
- `masaba.php` needs a few changes too:
   - Remove this line `<div id='osk-tablet' data states='default shift numeric'>` since you don't have a tablet layout. We may revisit this after you've updated the mobile layout.

I haven't tested the keyboard yet. I'll let you fix these issues first. It will be great to get this keyboard in the system and hopefully will help the user community.

oyousuf:
Thanks for the very detailed feedback, it really helps.

I think I implemented all these changes.

Thanks for the bit about ɛ and ɔ also! I didn't know non-ASCII wasn't allowed. I assumed I had to make sure the right thing was outputted if someone with a Bambara keyboard typed ɛ,ɔ,ɲ,ŋ. I've removed all of these rules

The mobile keyboard works online and on my local phone also (the latter I simply downloaded the .kmp). All of the "nasal" layers are activated by clicking a vowel. The vowels are placed where the English vowels are normally on a QWERTY keyboard

LornaSIL:
Everything looks good except there is no `masaba.ico` file in the `source` folder. You could either add that file, or comment out that line in the .kmn file.

I'd appreciate it if you removed the version number in `README.md` as well. That way that file doesn't always have to be updated. It's sufficient to have the version in the .kmn and in the `HISTORY.md` file.

PR #3401, opened 2025-03-25
[tsakonian] Update keyboard and Catrinity font

### LINKED ISSUE #3268: Improvements on Tsakonian keyboard
Following Kostakis' orthography for the Tsakonian language, the characters used for Tsakonian sounds th, ph and kh use an accent mark instead of a dot. These should be modified to τ̇, π̇ and κ̇ and their capital counterparts. Additionally, the mark τ͡σ is absent from the keyboard. 

Could these be added? Happy to provide further details on the matter.

LornaSIL:
Fixes #3268 

- Updated keyboard to add requested keystrokes.
- Changed font to use shared Catrinity and updated version of font to latest version.
- Improved documentation to make it easier to maintain and improved html.


DavidLRowe (approved):
I updated the version in the .kps file (even though I think it gets overridden by the version from the .kmn file).

PR #3402, opened 2025-03-25
[lac_viet] new keyboard submission

Cathaylaboratory:
ready for reviewing


keyman-server:

LornaSIL:
Mostly it looks good, but we do need the warning messages fixed. It should not be hard. The issue is with these rules:
```
any(low_uw) any(aa) + 'a' > 'u'index(a, 1)
any(low_uw) any(aa) + 'A' > 'u'index(a, 1) 
any(upper_uw) any(aa) + 'a' > 'U'index(a, 1)
any(upper_uw) any(aa) + 'A' > 'U'index(a, 1) 
```

These stores have 6 characters in them:
```
store(low_uw) 'ưứừửữự'
store(upper_uw) 'ƯỨỪỬỮỰ'
```
and this store:
`store(a) 'âấầẩẫậÂẤẦẨẪẬ'`
has 12 characters. 

From what I can tell, you could change the above rules to use a different index store. This is what I think you want altho I cannot tell exactly what result you are wanting:

```
any(low_uw) any(aa) + 'a' > 'u' index(low_aa, 1)
any(low_uw) any(aa) + 'A' > 'u' index(upper_aa, 1) 
any(upper_uw) any(aa) + 'a' > 'U' index(low_aa, 1)
any(upper_uw) any(aa) + 'A' > 'U' index(upper_aa, 1) 
```
The rules you have currently cannot possibly be giving the user the results desired. This is probably the kind of change you'll need to make for the other keyboards that have this problem.

LornaSIL:
There are still warnings that must be addressed.

Cathaylaboratory:
What warnings?

LornaSIL:
As I mentioned above, there are warnings about the any and index stores being different lengths. If you compile your keyboard within Keyman Developer and look in the messages folder it will tell you all the warnings.

Cathaylaboratory:
I know they are in different length but actually it works in that way.

LornaSIL:
You'll need to fix them. I suggested a solution above. The stores must be the same length when you use `index`.

Cathaylaboratory:
I mean this sentence work in that way, it works like ưa + a > uâ ứa + a > uấ, when it is lower case, it don’t need to be the same length.

Cathaylaboratory:
Changing it is meaningless and only cause inconvenience

LornaSIL:
It's not meaningless. The stores must have the same number of characters in them to work properly. If you fix it in your keyboards your future keyboards will be improved.

Cathaylaboratory:
The letters outside this range simply will never been matched. If I need to make the number same. I need to set a new store(), which add the complexity of the codes.

LornaSIL:
If they will never be matched, then remove them.

Cathaylaboratory:
I have fixed the issue.

Sent with [Proton Mail](https://proton.me/mail/home) secure email.

Cathaylaboratory:
I have changed the code Please merge the request.

LornaSIL (approved):
Thanks for making those changes. Similar changes will need to be made in updates to current keyboards and in future keyboards.
Please note there is no online help file, but we do not require that in the experimental keyboards.

PR #3410, opened 2025-03-31
[kbd_3221] Adjust language tags for Issue 3221

LornaSIL:
Relates to #3221 

This still does not address basic_kbdinuk2, basic_kbdiulat nor kreative_supersymbol as I do not know how to address langtags in multi-script keyboards.


DavidLRowe (approved):

mcdurdin:
This is a good update, thank you. Should we be informing the original keyboard authors of the fixes?

Runeboard:
>During our testing for the upcoming Keyman Developer 18 release we came across inconsistent language tags where either the language code didn't exist, or the script tag wasn't consistent with other script tags in the keyboard. These included changes to the language tags in your keyboard(s). Feel free to take a look, and please especially remember to pull the changes before making further changes to your keyboard(s). If you don't know how to download a single keyboard, this page is helpful: https://help.keyman.com/knowledge-base/?id=106

Was this just for the Tsakonian keyboard? What even is the correct language tag for that?

LornaSIL:
@Runeboard when you created the keyboard you tagged the languages for the keyboard with `grk` and `tsd`. Unfortunately, `grk` isn't a valid language tag, so we changed it to `grc` and `el`. We just recently started letting our build system validate the language tags in the build. It also checks for consistency that two different scripts aren't selected. You are welcome to remove those if you wish, but the language tag needed to be a valid official tag.

LornaSIL:
> Where can I find a list of all official tags please?

Keyman uses a langtags.json file and when you use a bad tag, you should be getting a warning when you compile the .kps and a suggestion of a better language tag. But, I'm not sure what happens if you make a language tag up.

The langtags file uses the 3 letter ISO 639-3 codes, but if there is a 2 letter code, that is the preferred code. The the script code is used from ISO 15924. 

There's a default script for many of them, so for `eng-Latn-US` the minimal tag would be `en`.

This text file might be helpful: https://ldml.api.sil.org/langtags.txt

The problems we get are when there is not a defined language or script and then it's kind of an art to figure it out. We used `und` (Undetermined) if there is no language code and we use `Zzzz` (unencoded) for the unencoded scripts.

In this case I believe you used an ISO 639-5 code (`grk`) which the language tags don't use.

LornaSIL:
@jagdouglastavani This is an old Pull Request by me that is already merged. I don't see that you have submitted a pull request.

Runeboard:
> The langtags file uses the 3 letter ISO 639-3 codes, but if there is a 2 letter code, that is the preferred code. The the script code is used from ISO 15924.
> 
> There's a default script for many of them, so for `eng-Latn-US` the minimal tag would be `en`.
> 

Thank you. As Tsakonian is written in the Greek script, would the correct code be `tsa-CG`?

PR #3411, opened 2025-03-31
[copyright_php] Tidy up keyboards php files including removing copyright dates

LornaSIL:
- Removed SIL copyright entries
- Removed years for other copyright entries
- Removed extra `<body>` and `<html>` tags
- Removed unnecessary `<h1>` tags
- Changed summary of *Basic* keyboards where it said `1.1 generated from template.`
- General cleanup where I saw things.

Since this is only help files, there is no version bump.


DavidLRowe:
sil_cheyenne.php still has a `<head>` section. Is that intentional? It seems it was removed from other .php files.

taigi_viet_telex.php seems to have been deleted. It that intended?

DavidLRowe (approved):
Thanks for all this work!

PR #3415, opened 2025-04-02
[tirhuta]: Convert help to php

darcywong00:
Addressing broken help deployment
https://github.com/keymanapp/help.keyman.com/actions/runs/14203996698/job/39797302200

Converts the tirhuta help file to php


darcywong00:
Ah, already done

PR #3417, opened 2025-04-02
[php_more_fixes] More fixes to online php files

LornaSIL:
- Primary changes were to remove or add extraneous `<div>` and `</div>` codes where they made the online help display poorly.
- Also removed some table of contents that were duplicates of the sidebar.
- Renamed one help file that wouldn't work because the name was misspelled.
- Removed quite a few `<h1>` elements because they were the same as the title of the page and look bad.

No changes to keyboards.


DavidLRowe:
The only question I had was on eo_plus.php where you added a `<head>` element (with a link to a styleshee?). I don't know enough to say whether this is right or wrong! I'll approve it and you can merge the PR if the eo_plus.php file is right as is or else update the PR and I'll look at it again.

DavidLRowe (approved):
approved (but see comment)

LornaSIL:
A change I made, maybe yesterday, negatively affected the display for that page.

release/e/eo_plus/source/help/eo_plus.php
```diff
@@ -3,6 +3,10 @@
 $pagetitle = 'Esperanto Plus Keyboard Help';
 require_once('header.php');
 ?>
+<head>
+    <link rel="stylesheet" href="eo_plus.css" />
+</head>
```
mcdurdin, line 8: We shouldn't have two `<head>` sections in the .php. Instead we should add, directly under `$pagetitle = ...`:

```
$pagestyle = "@import url('eo_plus.css');";
```

A similar example (Google fonts stylesheet but same principle): https://github.com/keymanapp/help.keyman.com/blob/5494edbfceae6a357fcac5087072200b8bc5ccc1/keyboard/sil_vai/0.6/sil_vai.php

PR #3421, opened 2025-04-03
[php_head] Move head css into php header

LornaSIL:
I'm just doing this for two help files to see if I have the syntax correct.
If correct, there are 41 other keyboards with the same issue.


mcdurdin (approved):
Yes, this looks correct. The `@import` rule must be the first rule (or rules) in the `$pagestyle` content, which is what you have done here.

Thanks!

PR #3433, opened 2025-04-09
Initial moro language keyboard 

rc-swag:
Still some many items TODO.

Question: Should usage of U+0110 / U+0111 be retained, or should the keyboard layout adopt U+A7C7 / U+A7C8

This version has both using the brackets and backslash key caps, as well as using sequences, need to decide prefered method.

Touch layout has a character on the second row that is on the top row on the desktop version.

Keyboard documentation.


keyman-server:

LornaSIL:
There are likely are fonts that support U+0110 / U+0111 than U+A7C7 / U+A7C8, but I assume you are checking with the community about that decision.

I'd like to see the keyboard NOT include a local copy of the Catrinity font. There is a Catrinity in `\keyboards\release\shared\fonts\catrinity` that you can reference in the .kps. That one is much newer than the one you committed (v2.150) which is an example of why we don't want multiple copies of fonts (with different versions) in the repo if at all possible.

rc-swag:
Thanks Lorna I will move the PR from Draft to Ready for review and comment here when it is ready but I am not there yet.

I am meeting with the representative from the community and the driver behind the keyboard and will have some more updates based on that information.

andjc:
It may be better to determine what reaction [L2/25-067](https://www.unicode.org/L2/L2025/25067-bowl-struck-variants.pdf) gets. I suspect that Moro falls into the same category where U+0110 / U+0111 is the correct casing pair rather than U+A7C7 / U+A7C8. 

It is also worth noting that to match older Moro primers both sets U+0110 / U+0111 and U+A7C7 / U+A7C8 would need an alternative glyph, i.e. Moro usage differs from reference glyphs of both sets.

rc-swag:
Issues still attempting to resolve U+0110 / U+0111 vs U+A7C7 / U+A7C8. It does appear from the references @andjc has provieded that U+0110 / U+0111  is the way to go. The mention in L2/19-179 also seems to be incorrect I think. 

> “The Moro language of Sudan distinguishes dental // from alveolar /d/. Several orthographies exist for Moro; one writes the voiced dental as ꟑ, one writes it as đ, and one as ḏ. Texts exist which use the TAU GALLICUM.”

I am not sure this is correct as the orthography given to me has all three different characters d (U+0064) ḏ (U+1E0F) and ꟈ(U+A7C8) 
Whereas the Unicode note suggest that it is either ḏ (U+1E0F) or ꟈ (U+A7C8) not both. 

The person I worked with in terms of glyph shape aggreed with a document Andrew had. I have pasted an image here as I haven not been able to find a font that matches.
image

However, I am trying to get him to introduce me to someone he knows that worked on the Bible translation and I believe some other lanuage surveys. 

In summary I believe we will go with U+0110 / U+0111, however I have not been able to find a font that has the preffered Glyph variation. 

The other issue I have is Keyboard layout for the Mobile 
For the desktop I had 
image
But for the mobile I had 
image

Not the ñ is on the second row as the first row was crowded. I have not designed actual userful keyboard layouts before so I am looking to @LornaSIL or @DavidLRowe  for guidence if this could be done better. The Layout is in this PR.

The Moro language speaker I have been working with liked the Mobile layout however I don't think he has used the desktop much. I will look to meet with him next week.

The Main Todo items.

- [ ] Get in contact with ex language consultant for input on U+0110 / U+0111 and preffered glyps
- [ ] Find a font that renders the prefered glyphs
- [ ] Advice from @LornaSIL  and @DavidLRowe on the place ment of ñ on the mobile and desktop layouts
- [ ] Documentation

LornaSIL:
A7C7/A7C8 are fairly new to Unicode. I can see why they might have used it, but I'm surprised anyone picked up on using those characters since they are new.

We can provide a font for distribution for Moro if they want to use 0110/0111 with the bar through the bar through the d. I don't see a need for a variant for the hooked r; at least based on the screenshot they are using the capital form of the hooked R. If they want the lowercase form we can produce a font that includes that variant as well. Our fonts do support the variants, it just isn't straightforward to people to know how to use them.

andjc:
Regarding fonts:

I suspect the primers for Moro used an 8-bit fonts generated from SIL Encore. When I worked on Moro on Keyman we adapted the Charis SIL font, which had the lowercase variant for hooked R, and swapped it for the default glyph.

And modified 0110/0111 to suit. This could be added as a variant glyph, modified versions of A7C7/A7C8 should work.

Youripaste:
> Lorna has sent me a font with the changes to the default Charis SIL font.
> 
> The reason for my latest pause is actually quite intersting. I was int contact by someone from SILA (Australian affiliate of SIL Global) who has been working with members of the Moro community living in different parts of the world who are also interested in a Keyman keyboard, amoung other literacy projects. We are now trying to get the two groups together shortly.

YOU are lazy. Thats the true Mr. Swag! Moro community is disapiring every day! keyman is killing the language and taking keyboard hostage!

LornaSIL:
Are you still working on this keyboard?

LornaSIL:
Thanks! The main thing I see now is that you have different copyright holders. Make sure to choose the same one everywhere. 
- LICENSE.md and readme.htm have: `Copyright © 2025 Ross Cruickshank & Benjamin Angalo`
- *.kmn and *.kps have: `Copyright © SIL Global`
- 
Feel free to change it to what is appropriate. Personally I'd like the copyright info removed from readme.htm so it doesn't need to be updated in case the copyright holder changes, but that's not required, it just needs to be consistent.

Almost there!

LornaSIL (approved):
Yay! Congrats!

PR #3437, opened 2025-04-10
[mende_kikakui] WebOSK help needs to be shorter

LornaSIL:
Online WEBOSK help has to be shorter in order not to not cover up the main keyboard page.


DavidLRowe (approved):
RSLGTM

PR #3442, opened 2025-04-11
May I do two updating at once?

Cathaylaboratory:
Because their code are basically same. If I find one bug in one keyboard, that means other keyboard also need to fix these same bug. In this way can save our time. The phonology map and keyboard document of Hong Kong keyboard need a change. I want to change the khuych spelling into khuich because I found uych spelling is not necessary. It is a little bit urgent becuase the phonology map can cause chaos for the user if I don't update it soon.
Another bug it the quă can not be generated sperately and smoothly. It will affect both Vietnam and Hong Kong keyboard, make them less reliable and inperfect.
In Hoisan and Lac Viet keyboard the same issue exist. But their spellings are simple and don't have such combination, those bugs will not be matched in Hoisan and Lac Viet keyboard if they are used only in their script. So they are not very urgent. I choose to update them until I found 5 6 bugs.
Ready for reviewing. 
Thanks


keyman-server:

LornaSIL:
Only include one keyboard per pull request.

PR #3452, opened 2025-04-22
vietnam keyboard updating

Cathaylaboratory:
ready for reviewing
There is a bug during the generation and this bug always let the keyboard system get collapse in my phone.
I found it out and fixed it.
It seems like that if a rule is partial contradict with other rule, changing the rule orders will affect the utility of the rule
it can works however it will cause bugs sometimes, let the system collapse in my phone

### DIFF
--- a/release/v/vietnam/HISTORY.md
+++ b/release/v/vietnam/HISTORY.md
@@ -1,6 +1,11 @@
 vietnam Change History
 ====================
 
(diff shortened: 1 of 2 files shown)


keyman-server:

LornaSIL (approved):
I made some changes, so make sure to pull those changes before you revise it again.

PR #3453, opened 2025-04-23
[php_head] Attempt to fix some of the php head and css elements

LornaSIL:
A few of the keyboards that need fixing. If this works, I'll do all the Senegal ones next.


DavidLRowe (dismissed):
My only comment is on `$pagename = 'sil_senegal_bqj_azerty';`. Others have `$pagename = 'Blah Blah Blah Keyboard Help';`. Is it worth changing the name?

Otherwise looks fine (though I have limited knowledge of PHP).

I'll approve and you can decide whether or not to change the name for sil_senegal_bqj_azerty.

DavidLRowe (approved):

PR #3458, opened 2025-04-25
[sil_tepehuan] Fix whitespace in help file

darcywong00:
Fixes the failed keyboard help deployment on https://github.com/keymanapp/help.keyman.com/actions/runs/14653975660/job/41125613577

> PHP Parse error:  Invalid indentation - tabs and spaces cannot be mixed in ./keyboard/sil_tepehuan/1.1/sil_tepehuan.php on line 28

Screenshot of the key sequences

image

Fixes: keymanapp/keyman#13756

PR #3461, opened 2025-04-25
[sil_tai_dam_typewriter] Fix whitespace in help file

darcywong00:
Similar to #3458 which fixed deployment of php help file to help.keyman.com

Fixed invalid mix of tab/whitespace in the heredoc CSS syntax.

I had to consolidate the font line here. 
@mcdurdin - is that valid syntax?

### DIFF
--- a/release/sil/sil_tai_dam_typewriter/source/help/sil_tai_dam_typewriter.php
+++ b/release/sil/sil_tai_dam_typewriter/source/help/sil_tai_dam_typewriter.php
@@ -57,7 +57,7 @@ $pagestyle = <<<END
           top: 3px;
         }
         
-        .key .keycap {          
+        .key .keycap {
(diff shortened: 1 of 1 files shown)


LornaSIL (approved):
LGTM

release/sil/sil_tai_dam_typewriter/source/help/sil_tai_dam_typewriter.php
```diff
@@ -66,11 +66,7 @@
         }
         
         .keytext {
-          
-					font:
-						9pt
-							 
-						"Tai Heritage Pro"; 
+          font: 9pt Tai Heritage Pro;
```
mcdurdin, line 69: Should be: `font: "9pt Tai Heritage Pro";`

PR #3462, opened 2025-04-25
[3460_php] Remove tabs in mixed space plus tabs

### LINKED ISSUE #3460: audit PHP help files using heredoc syntax and mixing tabs/spaces
See #3459 

We've had a few recent deployment failures of keyboard help files flagging

> PHP Parse error: Invalid indentation - tabs and spaces cannot be mixed in ./keyboard/sil_tepehuan/1.1/sil_tepehuan.php on line 28

Might need to audit other keyboard help files that have this too:

From a brief regex search (there may be more):

bj_cree_east_james_bay
bj_cree_woods
bj_mista_washa_cree
bj_ojii_cree

galaxie_greek_mnemonic (also has a nested keyboardchart.php file)

Syntax seems wrong?
gautami_inditran
autami_thamizh_tamil

nepali_romanized

LornaSIL:
Fixes #3460 

Also, https://help.keyman.com/keyboard/gautami_devanagari/1.0/gautami_devanagari is not displaying properly. The only thing that seemed wrong is it had a BOM so I saved it as UTF-8 without a BOM.


darcywong00:
thanks @LornaSIL!

PR #3464, opened 2025-04-26
korea keyboard

Cathaylaboratory:
ready for reviewing


keyman-server:

LornaSIL:
LICENSE.md - missing copyright information
HISTORY.md - incomplete sentence on who created it.

Double-check that all the other information is correct please.

Cathaylaboratory:
I have changed

Sent with [Proton Mail](https://proton.me/mail/home) secure email.

LornaSIL:
Mostly it looks good, but the HISTORY.md file is for the wrong keyboard. The LICENSE.md needs to have the copyright date extended to include 2025. So, `Copyright © 2024-2025 cathaylab` or since it's a new keyboard, perhaps it should just be `Copyright © 2025 cathaylab`

Cathaylaboratory:
OK I have changed

Sent with [Proton Mail](https://proton.me/mail/home) secure email.

LornaSIL:
@Cathaylaboratory we are unfortunately not going to be able to accept this keyboard in the `keymanapp/keyboards` repository. The .kmn files has 263,000+ lines of code and the .kmn syntax is just not designed to handle keyboards of that size. 
Your previous `hakka` keyboard had the largest number of rules at 72,000+ rules which was already a bit too large.

You have several options. You may keep your keyboard as is, and distribute in on your website or Github repo. Or you could write a smarter keyboard that uses stores and chaining rules.

Keep in mind that we also won't be able to accept updates to keyboards such as `hakka` that increase the number of rules even more, so it would be wise for you to figure out more about how to write smarter keyboards to reduce the number of rules.

I'm sorry, but we'll be closing this PR for the above reasons.

Cathaylaboratory:
Why you can not upload my keyboard?
If actually can compile into package if you patient enough.

mcdurdin:
Unfortunately, it's just too big:

* korea.js is 17 megabytes, which is just HUGE for a Javascript file. This would be a problem for web keyboard users on slow or expensive connections.
* it will have problems with memory on many devices, causing out-of-memory crashes.
* at best, it will be very slow to load. On slower Android devices, it would almost certainly cause ANR (application not responding) errors as the system tries to load the keyboard.

The slow build is also a minor problem for our build systems, because it takes nearly 5% of the overall build time, which is dramatically higher than other keyboards, and this negatively affects all of the hundreds of contributors to this repository. We probably wouldn't reject the keyboard on this basis, but it does have an impact.

Cathaylaboratory:
So can I just delete the web keyboard function? So the issue will be solved

Cathaylaboratory:
I mean I am using the keyboard on my iphone 13mini and it don’t have any sort of collapsing.

Cathaylaboratory:
260000 rules is large. But actually for todays phone or computer in 2025, it is not large enough for let our device collapse. My labtop and google pixel phone can also use this keyboard without any issues.

mcdurdin:
> it don’t have any sort of collapsing

previously, you [wrote](https://community.software.sil.org/t/keyman-phone-keyboard-disappear-bugs/10089):

> Hello, I found my vietnam keyboard sometimes disapper in the phone. Sometimes it suddenly shutdown and recover after 2 seconds, sometime it disappear, I need to reload my keyboard.

It may not crash immediately, but it will eventually as the operating system reclaims resources.

Cathaylaboratory:
They are different. The vietnamese keyboard collapse is sometimes because of some rules are actual contradict. Such as uoi + i and oi + i rule. The system will be confused by those two cases sometimes and cause pause, I believe it work in this way. But the korean keyboard they don’t contradict like vietnamese one. They are not contradict because they are characters.

Cathaylaboratory:
The korean keyboard it working.

mcdurdin:
> The system will be confused by those two cases sometimes and cause pause, I believe it work in this way

You are completely wrong. "Contradictory" rules do not cause the app to crash. It just means one of the rules will never fire. On the other hand, a keyboard that takes too much memory _will_ cause the program to crash.

I'm going to leave this discussion here and won't be responding any further. We've given you feedback on why the keyboard cannot be accepted into the repository. You are free to distribute the keyboard by other means, but it won't be going into this repository.

PR #3465, opened 2025-04-28
feat: Keyman Developer 18.0.235

mcdurdin:
Note: please do not merge before 30 April 2025, as Keyman team are in team meetings, so will not be available to support.


package.json
```diff
@@ -24,6 +24,6 @@
   },
   "homepage": "https://github.com/keymanapp/keyboards#readme",
   "dependencies": {
-    "@keymanapp/kmc": "17.0.333"
+    "@keymanapp/kmc": "18.0.235"
```
darcywong00, line 27: Should we also npm update the dependencies to 18.0?

mcdurdin: What do you mean?

darcywong00: ah nvm

darcywong00 (approved):
lgtm

PR #3469, opened 2025-04-28
[style_php] Some of the remaining php files with style elements

LornaSIL:
This isn't all of them, I may try to finish this later today if I have time.


DavidLRowe (approved):
LGTM (in my limited understanding!)

PR #3470, opened 2025-04-29
[gondi_dev] Fix heredoc END in help file

darcywong00:
Fixes help deployment fail on https://github.com/keymanapp/help.keyman.com/actions/runs/14715278035/job/41297103574

coming out of #3469 

> PHP Parse error:  syntax error, unexpected end of file, expecting variable or heredoc end or "${" or "{$" in ./keyboard/gondi_dev/1.1.2/gondi_dev.php on line 269

PR #3472, opened 2025-04-29
attempt to fix remaining heredoc syntax

darcywong00:
Follows #3469 and #3470

Attempting to fix remaining heredoc syntax for other PHP files.
Affecting help deployment https://github.com/keymanapp/help.keyman.com/actions/runs/14722448206/job/41318603973

> PHP Parse error:  syntax error, unexpected end of file, expecting variable or heredoc end or "${" or "{$" in ./keyboard/gondi_gunjala/1.1.3/gondi_gunjala.php on line 131
xargs: php: exited with status 255; aborting


mcdurdin (approved):
LGTM

(Note: prior to PHP 7.3.0, `END` had to be in the first column. I think we are on PHP 7.4.0 currently, so should be okay)

PR #3475, opened 2025-04-29
fix: use extended regex instead of Perl regex to support Apple grep

mcdurdin:
Fixes: #3308


jahorton (approved):
Thanks; build.sh is working great now on my Silicon Mac.

PR #3483, opened 2025-05-03
[Ugewekeyboard]second submission 

RyanCecil1:
an all inclusive ewe keyboard developed using the standard qwerty layout for typing in the ewe language spoken in various regions of Ghana and the diaspora


DavidLRowe:
Thanks for your interest in contributing. 

There seems to be a space in the folder name "ewe ". This is not permitted. You'll need to correct that to "ewe" (with no trailing space).

There also seem to be some unrelated files ga*.* included. See "[Files changed](https://github.com/keymanapp/keyboards/pull/3483/files)" heading above to see exactly what files are included in this pull request.

If you are targeting both "desktop" (computers including laptops) and "touch" (phones and tablets), then you are essentially covering all platforms, so it is better to just use: `store(&TARGETS) 'any'` (and remove all the other choices). 

When choosing the language code, pick the simplest form. In this case "ee" is the best choice.

You can make changes to this pull request, or close it and start with a fresh one. I see that you've also submitted keyboards for Akan and Ikposso, but they have some similar problems. I'd encourage you to get one keyboard through the entire process before submitting others. You'll save yourself time in the long run.

PR #3493, opened 2025-05-07
[winchus] added dash — & improved touch layout

alex-larkin:
* Added long dash (U+2014, —) to all platforms
* Fixed an issue with some flicking keys not working
* Reordered many long-press keys for consistency and convenience.
* Updated documentation

Note: I experimented with adding a Spanish keyboard in the hopes of adding spell check, but it didn't go anywhere. I don't desire that to be a part of the keyboard at this time, but wasn't sure how to remove it. Could you please help me ensure an `espanol.kmn` didn't make it into the PR? Thanks!

PR #3501, opened 2025-05-08
[winchus] Copy missing activar-boton-teclado-360.png help file

darcywong00:
Follows #3493 

Copies missing png file which is causing keyboard help deployment to fail on keymanapp/help.keyman.com#2042 and  keymanapp/help.keyman.com#2043


DavidLRowe (approved):

DavidLRowe:
Thanks for catching this, @darcywong00 . @LornaSIL pointed out another correction that was needed, so I included that in this PR.

Issue #3468, opened 2025-04-28
chore: remove "Keyboard" from name of "GFF Amharic Keyboard"


mcdurdin:
Also GFF Amharic Classic -- there may be other keyboards with the same issue, using "Keyboards" in the name of the keyboard not recommended.

(Observed during Keyman team meetings)

(I can do an audit later)

LornaSIL:
Also `GFF Gurage Keyboard`

mcdurdin:
I did a search across the generated .keyboard_info files and found the following which had the name 'Keyboard' in their name -- coming out of the .kps and/or the .kmn.

I think that the appropriate solution is to remove 'Keyboard' but there are some such as 'BU Keyboard' for which that change is slightly difficult, because that would leave it as just 'BU'. Similarly 'Pan Africa (for US Keyboards)' -- perhaps we should leave that even though it ends up as an unwieldy 'Pan Africa (for US Keyboards) keyboard' in the search results.

We don't need to do a version bump, as the change is not very significant for end users, but will help the keyboard search metadata.

I also searched for 'package' and 'keyman', but there were no additional keyboards listed with those terms in their name.

```
56 results - 56 files

africaus.keyboard_info:
  2    "id": "africaus",
  3:   "name": "Pan Africa (for US Keyboards)",
  4    "description": "This is the US layout version of the SIL Pan Africa keyboard. The keyboard follows the US QWERTY key arrangement. For a mnemonic keyboard designed to run on any hardware layout, use 'Pan Africa (Deadkeys)'. \r\n\r\nThis keyboard was developed to type the Latin-script languages of Africa. It was created using data received from SIL entities in Benin, Burkina Faso, Cameroon, Chad, C\u00F4te d'Ivoire, Nigeria, Senegal and Togo. Where possible, the keystroke sequences that the majority of the entities seemed to prefer was used.\r\n\r\nThe Keyman source file is provided as a part of the package for convenience in modifying the keyboard.\r\n\r\nFor the official release of this SIL keyboard package, visit:\r\n  http://scripts.sil.org/AfricanKeyboard1",

bu.keyboard_info:
  2    "id": "bu",
  3:   "name": "BU keyboard",
[...]

PR #3518, opened 2025-05-12
[3468_keyboard] Update some names with Keyboard in name

LornaSIL:
Still in progress while we figure out what to do about legacy.


mcdurdin:
Let's leave legacy and deprecated keyboards alone :grin: less work for you I hope

LornaSIL:
Ready for review @mcdurdin 
In particular, take a look at `fv_all` and `galaxie_greek_hebrew_mnemonic`. Both of those are packages and I changed "Keyboards" to "Package" for the first one and just removed "Keyboard" in the 2nd one.

LornaSIL:
Fixes #3468

legacy/l/laounicodebasic_on_thaikbd/laounicodebasic_on_thaikbd.keyboard_info
```diff
@@ -1,6 +1,6 @@
 {
   "id": "laounicodebasic_on_thaikbd",
-  "name": "Lao Unicode on Thai Keyboard layout",
+  "name": "Lao using Thai layout",
```
mcdurdin, line 3: I'm not sure if we should change this one, because 'Lao Unicode' has been a name in popular use to differentiate from other Lao (Unicode) keyboards. Oof.

```suggestion
  "name": "Lao Unicode on Thai Keyboard layout",
```

mcdurdin (dismissed):
Just one to revert (it's a little bit weird), but apart from that LGTM

DavidLRowe (approved):

PR #3525, opened 2025-05-15
[urdu_dvorak] pressing any key in shifted-state of mobile keyboard,  unshifts the state to default

Caliph3ubaidUrRe7man:


LornaSIL:
@Caliph3ubaidUrRe7man can you make sure you've tested this keyboard to see if you've implemented it the way you want before I check your PR? I'm thinking you are waiting for the keyboard to get online before you do a thorough testing.

LornaSIL (approved):
Actually, this keyboard won't compile because you manually edited the mobile layout. That's a pretty fragile thing to do because you either are missing commas or have extra commas. I'm not able to figure it out for you. It's better to use Keyman developer to create your mobile layout.

PR #3532, opened 2025-05-19
[soninke_n_ti] N'ti script keyboard for the Soninke language

oyousuf:
This is a Soninke-language keyboard using the N'ti script. The script was created in 1966 and is used in Mali, Congo (Brazzaville), and Ivory Coast. The script is not yet encoded in Unicode.


keyman-server:

oyousuf:
Ready for review

LornaSIL:
As a reminder, the folder and filename structure must be the same. If the keyboard filenames are `soninke_n_ti` Then the folder name must be under `s` and named `soninke_n_ti`. 

The file in the help folder must also follow the filename system so it should be `soninke_n_ti.php`

I see a temporary file that should be removed `source/~$iTypingManual.docx`

Regarding fonts. Versioning of fonts will be important for future iterations. The font here is `001.000` which I think it equivalent to version `1.000` (I don't think the leading zeros are considered in the version). In the future, if the font is updated, you would want to increment to `1.100` or something like that. Keyman keeps track of version numbers and if a version number in a future version of the font is less than the current version it won't install the new version of the font, so just keep that in mind for future updates.

LornaSIL:
Unfortunately, not all of the files got moved from the original `experimental/nti` folder. Everything in the root of `nti` needs to move to the root of `soninke_n_ti`. The whole `nti` folder should be deleted and removed from the Pull Request. I think you moved everything else over.
- In the .kmn file there are a lot of duplicated rules. It would be good to remove the duplicates so you don't get the warnings. For example, this message `soninke_n_ti.kmn:209 - hint KM020AE: This rule will never be matched as the rule on line 17 takes precedence` actually refers to line 208 in the keyboard and line 17. One of them is using the glyph and ones uses the USV. Both rules shouldn't be there.
- In the .kps file
   -  the name of the package is `Sooninkanxanne`. Is that what you want? Should that be the name that is used in the .kmn? The naming in the various files seems a bit inconsistent. Most of the files seem to use `Soninke N'ti`.
   - the language tag should be `snk-Zzzz` as the tag you have used would indicate Latin script.
- In the .php file the desktop osk syntax should be:
```
<h2>Desktop Keyboard Layout</h2?
</div>
``` 
I can keep checking Pull Requests, but I'm not able to merge any PRs right now as we work out some issues with our downloads server.

oyousuf:
Thanks for catching the old nti folder and duplicate .kmn code.

LornaSIL (approved):
LGTM. Thanks for fixing the issues. The font won't be displayed until I get the font uploaded to another site.

PR #3535, opened 2025-05-22
chore: update rsync settings for build agents

mcdurdin:


darcywong00 (approved):
Lgtm.
So the new vars are configured for the action?

mcdurdin:
> So the new vars are configured for the action?

USERPROFILE is a Windows env var. RSYNC_HOME is configured on the agents.

PR #3538, opened 2025-05-22
chore: try fixing quotes in rsync command

darcywong00:
Follows #3537

We're trying to fix the rsync call to downloads.keyman

I wonder if the  the whole `--rsh` line should be in a set of quotes.


darcywong00:
Replaced by #3540

PR #3542, opened 2025-05-23
[bidayuh] update rules

Nnyny:


LornaSIL (approved):
LGTM. I did add more text to the HISTORY.md so we understand what the actual change was.

PR #3552, opened 2025-05-29
[kazakh_arabic_full] new keyboard for kk_Arab

MuratKaribay:
This is a keyboard that does not use shift, since Arabic is not a bicameral script, and the user needs the full layout, for this purpose it was created.


keyman-server:

MuratKaribay:
@LornaSIL in fact, users want a layout without shift, and there are a couple keyboards that already do that.

If I add this touch layout to kazakh_arabic as a tablet layout, will it be accepted as an additional layout for keyboards? (for example: Gboard)

Here are some examples where the shifter is not used:

IMG_20250530_020742
IMG_20250530_020703
IMG_20250530_020635

MuratKaribay:
I added the layout in kazakh_arabic as tablet to test it, but couldn't find how to switch from phone to tablet in my phone, maybe if the user can't switch layouts, then probably not necessary, I guess.

DavidLRowe:
Keyman detects the type of device (phone vs tablet) and uses the best available. So, if both phone and tablet layouts are included in the keyboard, only the phone layout would be available on a phone. 

I have not looked at your layouts, so perhaps this question is not sensible. Is it possible that your changes for eliminating the shift layer could work on the phone? Then you could go back to having just one layout (phone) that works on both platforms.

MuratKaribay:
If possible, I would like to have two different layouts with and without the shifty layout. But if only one is needed, then let it stay that way.

PR #3553, opened 2025-05-29
Added key sequence to type ordinal numbers from user feedback

oyousuf:


LornaSIL:
Thanks!
- You'll need to always retain the previous history in HISTORY.md and add (at the top) the changes you are making. So add an entry for the new version that says what kind of update you are making. "Add keystrokes for ordinal numbers" or something like that.
- Has the version of the font and typing manual changed? If not, you shouldn't recommit them. 
- The .kmn needs the version number changed to 1.0.1
- Please remove your changes to the .php file. We had to commit a fix last time because of the apostrophe in the name. You should pull that change so you don't recommit the wrong title.
- Please remove the doc file that starts with a `~`. That's a temporary Word file that we don't want in the repo.

I think that's it.

oyousuf:
Thanks for the feedback. I fixed all those changes I believe. The font and typing manual did change so I had to recommit them.

PR #3548, opened 2025-05-26
63 - Naijatype experimental

rowbory:
Added 3 codepoints (for Kanuri and Kamuku), improved small caps behaviour, updated docs (languages and some details). Incorporated the previous docs CSS changes. Thanks.

PR #3563, opened 2025-06-09
[andika] Update keyboards using Andika to new version

LornaSIL:
Add new version of Andika

Update keyboards using Andika to include new typefaces

This does not include any of the keyboards using Andika subsets.

This PR will replace PR #3548 as I pulled in the changes @rowbory was trying to make.


DavidLRowe (approved):
LGTM

PR #3573, opened 2025-06-14
[Dagaare] initial upload

RyanCecil1:
first upload of the UG Dagaare Keyboard


keyman-server:

RyanCecil1:
thanks for your time and input as well.

LornaSIL:
You would still need to upload each keyboard separately in a separate PR for each. I'd rather you did one at a time so that you will learn from each keyboard submission and fix the issues we find so by the 2nd or 3rd keyboard we won't find issues. (This is not a requirement of our build system, but I have found it is easier to fix issues one at a time and learn from the process so that we don't have to start all over. Also, it's easier to make mistakes when you change or upload a bunch of keyboards at once.)

We do have the concept of keyboard bundling, but we don't really recommend doing that anymore. I would especially encourage you to consider how necessary it is. Would someone using Dagaare really feel the need to type in another language? If it's a majority language maybe, but it seems unlikely they would want to type in all 10 other languages. It's a bit complicated and by the time you submit all 10 keyboards I'll probably have to assist you in creating a bundle. Keyboard bundles go in: [release/packages](https://github.com/keymanapp/keyboards/tree/master/release/packages)

An alternate idea is that in a keyboard (such as Dagaare) you can use the "related" field in the .kps file. You can see how it shows up on this keyboard: https://keyman.com/keyboards/gff_tigrinya_eritrea You can either choose a keyboard you want to "deprecate" with that keyboard (the older versions of the same keyboard that had a different name) or you can choose to "relate" a keyboard without deprecating it. In the example above there are 4 deprecated keyboards. Then, in this keyboard it just shows a "related" keyboard (for IPA): https://keyman.com/keyboards/nlci_ipa

PR #3574, opened 2025-06-16
[lavanya] Updated backspace handling. Removed support for touch layout

prashanth-lavinna:
1. Updated backspace handling. When using backspace (for example, after typing k+i, if you press backspace on i, it shows క instead of క్). This has been fixed. Now inherent vowel in not maintained even when the backspace is used. 

2. There are better touch layout for phones. Gboard stands out for Telugu. Rather than pushing for an inferior tool, I've decided to remove support touch layout.

### DIFF
--- a/release/l/lavanya/HISTORY.md
+++ b/release/l/lavanya/HISTORY.md
@@ -1,10 +1,10 @@
 Lavanya Change History
 ====================
 
-1.0 (2025-05-06)
 
 2.0 (2025-06-16)
 - Added backspace handling for vowel matras
 
+1.0 (2025-05-06)
 ----------------
 * Created by Prashanth
--- a/release/l/lavanya/HISTORY.md
+++ b/release/l/lavanya/HISTORY.md
@@ -1,7 +1,6 @@
 Lavanya Change History
 ====================
 
-
 2.0 (2025-06-16)
 - Added backspace handling for vowel matras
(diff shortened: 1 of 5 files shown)


LornaSIL:
I'm checking on this change. In general we don't like to remove behavior from a keyboard and removing the touch layout does remove behavior.

LornaSIL:
One option would be to keep the touch layout, but in the description and readme file we could add "There are other touch layouts which provide a better experience. For example..." and we could link to one you think is good.

prashanth-lavinna:
Hello Lorana,

I made the changes to reflect your feedback.
> I opted for touch screen support.
> I updated readme and description to point to Gboard.

Prashanth

LornaSIL (approved):
Thanks for doing this. We had decided you couldn't remove the touch layout so existing users wouldn't be affected. I think what you've added as a pointer is helpful.

LornaSIL:
It looks like we forgot to bump the keyboard version in the .kmn file. Because of that, the new version is not actually online. I can make that change for you and hopefully it will be okay. In the future, the version in the .kmn must be increased or the keyboard will not be rebuilt.

prashanth-lavinna:
Thank you Lorana. I will keep in mind to rev up the .kmn file. Have a great
day.

Prashanth

PR #3576, opened 2025-06-17
GFF Amharic 3.2.2

dyacob:
This is the GFF Amharic Keyboard v3.2.2:

- fixes to glow keys for alternative/secondary letters.
- alternative/secondary letters are now presented on their respective layers, the primary letter then appears as the hint.
- active key shade reset to sadis-key color.


DavidLRowe:
Does the documentation (phone layer images, for example) need to change to reflect the changes you've made?

DavidLRowe (approved):
Thanks for the updates. I wasn't sure if document changes were needed or not.

PR #3578, opened 2025-06-20
GFF Harege Fidelat v1.1.2

dyacob:
This is the GFF Harege Fidelat v1.1.2:

* CSS fixes for K_SHIFT keys.


DavidLRowe:
Some gff_mesobe_fidelat files slipped into the PR. They should be removed from this PR.

The end of the year range in LICENSE.md file needs to be updated to 2025.

dyacob:
That was quite the branch-leak!  Thanks for catching that.  I believe it has been set right now and the LICENSE.md year updated.

PR #3583, opened 2025-06-22
[Modi KaGaPa Phonetic] Initial Submission

abhi-deshpande:
Adding files for Modi KaGaPa Phonetic.


LornaSIL:
In addition to the kinds of changes needed from your other keyboard feedback, I have a bit more.
The language tags need to include the script since without a script code it indicates the Devanagari script which would not be true for this keyboard. So, they should be:
`mr-Modi`, `hi-Modi`, and `sa-Modi`.

Also, for a script that doesn't have much digital support, I usually recommend including a font with the keyboard. In our repo, we have a folder called `shared` and we have included the Noto Modi font here: \release\shared\fonts\noto\Modi
I recommend including that font in your list of files in the .kps. Then, also include the font on the **Details** tab.
In addition, I recommend using that font in the .kvks and touch layout files. That will give the user a better experience. (I didn't recommend doing this for your other keyboard since the Devanagari script has such good support).

LornaSIL (approved):
Thanks for making these changes! LGTM

PR #3587, opened 2025-06-22
Karabolpoular2, a update of Arabic keyboard karampoular 

sven-oly:


keyman-server:

LornaSIL:
Is this keyboard just an update to the previous version? If so, it's best
just to update the existing one. We don't use versions in the filenames. If
this is a completely different keyboard, then please rename it without
including the version in the filename. Thanks!

LornaSIL:
You don't want a version number in the filename. The filename will be part
of the url forever even when you update the keyboard.

LornaSIL:
At a minimum the folder name the keyboard is in needs to be renamed from `karambolpoular_version2` to `karambolpoular2`. But it feels to me like you are using the `2` as a version. It might be best to change the name of the folder and filenames to `karambolpoular_harmonized` or something like that which will distinguish it from the other keyboard. 
However, if you really want the keyboard name to be `KARAMBOLPOULAR2` then I guess the folder and filenames can stay as `karambolpoular2`

I haven't checked the keyboard thoroughly yet. I'd like the folder name to be resolved first. As always, we don't require documentation when a keyboard is in `experimental` but it certainly would help the users to have graphics displayed in the welcome.htm file so they know how to type.

Since this is an initial submission, it would make sense if the keyboard version could be version 1.0 instead of version 2.0.

In the .kmn the targets header should be set to `any` rather than listing out all the targets.

sven-oly:
Thank you for your help, Lorna. I will close this PR and then create a new one carambolfulbhe with the new keyboard.

PR #3603, opened 2025-06-30
maint: add `.editorconfig` file

ermshiperete:


.editorconfig
```diff
@@ -0,0 +1,15 @@
...
+indent_style = space
+indent_size = 2
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.md]
+max_line_length = 160
+
+[{Makefile,rules}]
+indent_style = tab
```
mcdurdin, line 15: We shouldn't need these rules in this repo, but they don't hurt

mcdurdin (approved):
This may be a little confusing to the average keyboards repo user because Keyman Developer doesn't use this file. But LGTM otherwise

PR #3605, opened 2025-07-03
[fulfulde_borgu_ajami_azerty] New keyboard

LornaSIL:
A similar QWERTY version will be submitted later.


DavidLRowe:
The .kmn file has:
```
store(&LAYOUTFILE) 'fulfulde_borgu_ajami_azerty.keyboard-layout'
```
But I would have expected `.keyman-touch-layout` (rather than `.keyboard-layout`), both in the &LAYOUTFILE statement and the filename for the touch layout.

Is it intended for the touch layout to have Latin displayed on the keys, but have the keys produce Arabic characters? Wouldn't we want the key labels to match the output even in "experimental"?

I also noted the different formatting of this file:
```
"layer": [
  {
    "id": "default",
    "row": [
      {
        "id": 1,
        "key": [
          {"id": "K_Q","text":"q"},
          {"id": "K_W","text":"w"},
```
instead of the usual:
```
"layer": [
  {
    "id": "default",
    "row": [
      {
        "id": 1,
        "key": [
          {
            "id": "K_Q",
            "text": "q"
          },
          {
            "id": "K_W",
            "text": "w"
          },
```
Is that an intentional change? I think it would make reviewing files easier, particularly if the order of the fields was stable (either alphabetical or a specific order, such as id, text, pad, width, sp, for example).

PR #3606, opened 2025-07-04
GFF Mesobe Fidelat - KPS Description Update for Web Preview.

dyacob:
This update is made to enhance the keyboard's web description at keyman.com as per the issue and discussion here:
https://github.com/keymanapp/keyman.com/issues/589

Markdown text with an image reference has been added to the KPS description field, it is _not_ essential for the KMP file. Accordingly, the keyboard version number remains the same.

The Markdown text has been tested in a separate markdown file and was verified to work. It is assumed to work just as well under the keyman.com website. An adjustment may follow if the web conversion at keyman.com does not appear as expected.  If all goes well, the companion GFF Harege Fidelat keyboard will likewise be updated.

PR #3612, opened 2025-07-07
Mesobe Fidelat Splash Image Resize

dyacob:
This update adds a **new** "splash" image to accompany the keyboard description text.  The image is reduced in size to 940 pixels width to better fit into the 960px container at the keyman.com website.  The original image used would automatically reduce in size when viewed from a mobile phone (see screenshot 1):

Screenshot_20250707_172228_Chrome

However, from a desktop/laptop web browser the image would not shrink and would overflow the container (see screenshot 2):

Screenshot 2025-07-07 at 5 50 08 PM

I'm hoping that a 940px width will work well in both environments  However, bear with me, this will probably require 2 PRs to introduce a new image and be able to link to it.  I think this is the case because the image markdown link in the Description field is to a commit in _my_ github account's fork and branch of the keymanapp/keyboards repository and _not_ to a commit in the keymanapp/keyboards repo.   When I later delete _my_ branch or fork, I assume the link reference will break.  Presumably commit data does not persist after these deletions.

If this is correct, then I would need to create a follow-up PR that has the Markdown link into the  keymanapp/keyboards repository's commit history after the new image file has been merged into the repo. Can you confirm this @mcdurdin ?


keyman-server:

DavidLRowe (approved):
I think that link should be stable.

mcdurdin:
> If this is correct, then I would need to create a follow-up PR that has the Markdown link into the keymanapp/keyboards repository's commit history after the new image file has been merged into the repo. Can you confirm this @mcdurdin ?

It should be stable, but in any case, as I noted in https://github.com/keymanapp/keyboards/pull/3606#issuecomment-3186668419, I no longer feel like we need to use the commit sha for image links. Sorry for the change in direction but hopefully simpler long-term

Issue #3601, opened 2025-06-27
[jorai] On the Jorai keyboard (for jra-Latn), "stroke" missing


imnasnainaec:
A word collection for Jorai was done with The Combine, and there's an oddity with one of the "with stroke" characters (https://github.com/sillsdev/TheCombine/issues/3858), so we're digging into keyboard options for Jorai.

According to https://scriptsource.org/cms/scripts/page.php?item_id=wrSys_detail_sym&key=jra-Latn and the data in The Combine, Jorai has four "with stroke" characters: ƀ (U0180), đ (U0111), Ƀ (U0243), and Đ (U0110). I don't see any of them, or a "with stroke" modifier, on https://keyman.com/keyboards/jorai.

Also, the Keyman keyboard is not listed in https://scriptsource.org/cms/scripts/page.php?item_id=wrSys_detail_font&key=jra-Latn.

LornaSIL:
Those strokes appear to be on the RALT keys. See the help documentation for the keyboard:
https://help.keyman.com/keyboard/jorai/1.0.4/jorai

It is apparently in that resource pack, but that download doesn't seem to produce anything useful :(

LornaSIL:
I do think the search mechanism on keyman.com is probably your best way to find keyboards for a language. ScriptSource hasn't updated the keyboard links in a few years.

imnasnainaec:
Thank you, Lorna! Yes, the keys are on the RALT layer (https://help.keyman.com/keyboard/jorai/1.0.4/jorai#toc-ralt).

The RALT layer isn't working for me on either https://keyman.com/keyboards/jorai or https://keymanweb.com/#jra-latn,Keyboard_jorai. (Currently on macOS 14.5 with Chrome 138.0.7204.92, but also previously on Windows 11 with Chrome).

LornaSIL:
We updated the keyboard. It should appear online within an hour or two. Let us know if it doesn't fix the issue for you.

DavidLRowe:
Not sure why merging #3615 didn't close this issue, but closing it now manually.

LornaSIL:
I tested the keyboard and the RALT keys are working for me now.

PR #3615, opened 2025-07-08
[jorai] Update jorai so RALT keys will work on any platform

LornaSIL:
The rules were written before we had good support for macOS or Linux.


LornaSIL:
Fixes #3601

PR #3619, opened 2025-07-11
sil_tchad_qwerty

roger-nadoumngar:
we add property: vertical-align: baseline in the Td.css


keyman-server:

LornaSIL:
Hi @roger-nadoumngar It looks like you added that file to the root of this folder `https://github.com/keymanapp/keyboards`. It needs moving from there. It actually needs to be committed into the keyboard folder in the `welcome` folder where the existing one is. And you probably need to add it to the `help` folder as well since it's in both locations. If you add it to the `welcome` folder, that will be added to the keyboard package.

Any time you make a change to a file that is part of a keyboard package you need to bump the keyboard version as well. Since this is a tiny change, you can bump the version by one increment. I think we are currently at 5.3.3, so you could make it 5.3.4. You'll need to change the version in the HISTORY.md and in the .kmn file.

You should discuss with your team when would be an appropriate time to switch to using the new Charis font (which doesn't have SIL in the name anymore). If you wanted to do that, the link in your .css would be:
`https://fonts.languagetechnology.org/fonts/sil/charis/web/Charis-Regular.woff2` and
`https://fonts.languagetechnology.org/fonts/sil/charis/web/Charis-Regular.woff`

Hope this is all clear.

LornaSIL:
I'm sorry, but the version in the .kmn needs to be bumped, not just HISTORY.md.

LornaSIL (approved):
I was able to update the version in the .kmn so I can approve it. Make sure that version number gets bumped when you make changes to files that are included in the keyboard package.

PR #3625, opened 2025-07-16
[korean_9_keys] korean 9 keys updating

Cathaylaboratory:
ready for reviewing.
I deleted those rules for old Hangul. Because I found the 9 keys structure is very hard to handle so much letters


keyman-server:

LornaSIL:
Thanks for reducing the number of rules. When I compile the keyboard there are many, many messages
```
korean_9_keys.kmn:522 - hint KM020B0: The store referenced in index() is longer than the store referenced in any()
```
Basically, starting at line 521, you use a store with 9 characters:
```
store(upper_mid) 'YUIHJLBNM'
```
And you have a store with 14 characters:
```
store(g) '교겨갸고거기규구그가개게걔계'
```
The message is referencing this rule and the following rules:
```
'ㄱ' + any(upper_mid) > index(g, 2)
```
You should not index a rule with a different number of characters in the store. Yes, I suspect the first 9 characters will index properly. You should make the number of character match so you don't get these hundreds of messages. These messages are there in order to help you make your keyboard work better for the users.

LornaSIL (approved):
Thank you for fixing those! Looks good.

LornaSIL:
I just realized your online help file isn't displaying. Can you submit a new PR that renames `korea_9_keys.php` to `korean_9_keys.php` ?
You don't need to do any version bump because that file is not part of the keyboard package. Thanks.

Cathaylaboratory:
Sorry about this mistake.
I have replaced the file.

BTW I updated the the korean dubeolsik keyboard at the same time.
Is it OK?

PR #3629, opened 2025-07-17
[greek_tonizo_1_0_5] Submission of keyboard files

tonizo2025:


tonizo2025:
Hi Lorna,
Thanks a lot for your help and suggestions.
Here is what I have done and am currently testing:
1. I changed Alt to LAlt.2. I changed the  CTRL ALT rules to RALT  3. I changed the .KVKS code referencing the Lalt keys to <layer shift="LA">4. I changed  the .KVKS code referencing the Ralt keys to <layer shift="RA">Is this correct?
The keyboard works OK in test mode.With the keyboard locally installed the OSK works fine for Lalt, but not for Ralt.  What am I doing wrong?
Note: The CAPSLOCK does not work in OSK. Is this a problem?
Thank you in advance for your help.  
Amaryllis


    On Friday, 18 July 2025 at 17:41:48 EEST, Lorna Evans ***@***.***> wrote:  
 
 LornaSIL left a comment (keymanapp/keyboards#3629)
The new rules regarding CTRL + ALT don't match what is on the OSK. Also, using ALT doesn't match what is in the OSK.

We strongly recommend using RALT rather than LALT so that application shortcuts don't get confused with the keyboard keys.

Anyway, these rules starting at line 143:
[CTRL ALT K_A]
do not match what you have on the .kvks file. In the .kvks file you have LALT so the OSK never displays for LALT, or CTRL+ALT
My recommendation is to change the CTRL ALT rules to RALT and then change the OSK to reference the layers for RALT and not LALT.

If you want to keep [CTRL ALT K_A] then you should change the OSK to not distinguish between left and right alt and in the code the layer should be <layer shift="CA"> and not <layer shift="LA">

There are also rules that reference just ALT: [Alt K_A]

You probably need to resolve your rules to LCTRL + RALT and ALT to RALT. Then, I think you need to do fill from layout again. I know that may mean fixing up the keys like q where you have multiple characters on the keycaps, but right now the OSK is definitely not accurate.

tonizo2025:
I want to upload the new keyboard which works fine on my computer, apart from the onscreen kbd for  RALT which does not work. I would like to sort this out and publish the new version of my keyboard.
Should I send you the .kmn file? 

    On Wednesday, 24 September 2025 at 18:06:49 EEST, Lorna Evans ***@***.***> wrote:  
 
 LornaSIL left a comment (keymanapp/keyboards#3629)
It makes sense to close this since the keyboard has been updated since this PR was created. However, I don't think the changes you were trying to make have been incorporated. Do you still want to add those changes in a new PR based on the current source? Or is there something you would like me to do to update the keyboard? The HISTORY.md file for this PR said:
⦁   Addition of left and right pointing double angle quotation marks used as quotation marks in Greek texts

⦁   Modification of the apostrophe key to produce U+02BC (Modifier letter apostrophe)

⦁   Addition of lower-case vowels without automated breathings on the same positions, but with the Ctrl and Alt keys pressed (extremely useful for corrections)

So, those changes have never been incorporated. I cannot remember what the problem was though.

LornaSIL:
Oh, okay. Maybe you can upload the zip file to google drive or dropbox or something similar and then send the link here. Or you can try attaching a zip here, but I'm not sure you are allowed.

tonizo2025:
https://www.tonizo.gr/various/greek_tonizo.zip
 

    On Wednesday, 24 September 2025 at 22:37:27 EEST, Lorna Evans ***@***.***> wrote:  
 
 LornaSIL left a comment (keymanapp/keyboards#3629)
Oh, okay. Maybe you can upload the zip file to google drive or dropbox or something similar and then send the link here. Or you can try attaching a zip here, but I'm not sure you are allowed.

LornaSIL:
The RALT keys seem to work for me on the desktop computer. 
On the touch layout the alt layer doesn't work. If you go to that layer and rename it to `rightalt` then it works. (Select the "alt" layer, click on **Edit...** and then rename the layer `rightalt`.

Is that what you are talking about, or is there a problem with RALT on the desktop?

LornaSIL:
Is the CTRL key also selected?

LornaSIL:
I'm afraid I don't know. You may need to raise an issue rather than
continuing to write on this closed pull request. That way we can tag the
developers.

tonizo2025:
Thanks a lot. 

    On Thursday, 25 September 2025 at 02:25:17 EEST, Lorna Evans ***@***.***> wrote:  
 
 LornaSIL left a comment (keymanapp/keyboards#3629)You can submit it as a new Pull Request. I haven't double checked to make
sure you took the latest files from the repo. Otherwise it seems to build
fine. I won't be available to check it until Monday, but it's
possible @DavidLRowe has time.

PR #3632, opened 2025-07-18
gff_amh_7 v1.6 - Minimal update to support replacement of older fonts

dyacob:
This version is created to support the objective of removing the "Washra" font collection which has been superseded by the "Waldba" collection, that are already in the Keyman repository.


mcdurdin:
.woff/.woff2 fonts can be used for mobile devices (which use a web renderer for their OSK). However, there is little benefit in including the file itself in the package as well as .ttf. 

.woff is supported in the `WebOSKFonts` and `WebDisplayFonts` sections for use in KeymanWeb.

For further discussion, see also:
* https://github.com/keymanapp/keyman/issues/11686

PR #3637, opened 2025-07-21
[phonetic_farsi] Update copyright and contact details

alliancecrusader:


LornaSIL:
Normally I would question changing the license holder, but since you are the one who made the original commit, this is fine.

However, any time a change is made to a file that is distributed within a keyboard package, the version must be bumped.

In this case you should bump the version in the .kmn to something like `2.0.1` and add a line to `HISTORY.md` with the version and date and explain what changed. I don't know why HISTORY.md doesn't have an entry for `2.0`. If you remember what changed, it would be good to add an entry for that version as well.

alliancecrusader:
Hi, I've bumped the version in both the .kmn and .kp
image
Also, the entry for version 2.0 in HISTORY.md is present from what I can see, I have no issue viewing it on my end.

LornaSIL (approved):
Looks good. Thanks.

PR #3638, opened 2025-07-21
[Sisaali] New keyboard

RyanCecil1:
changed CTRL to RALT and corrected the shift layer as well as the alt layer changes.


RyanCecil1:
thanks

PR #3659, opened 2025-07-30
GFF Ge'ez Keyboard v2.0.3

dyacob:
This is the version 2.0.3 GFF Ge'ez Keyboard.  Updates include:

* Removed the `NotSerifEthiopic-Regular.woff2` font file from the KPS.
* Fixed array length warning in the `.kmn` file, fixed earlier in GFF Amharic.
* Expansion of the Description field.

This is a fairly small update.  While installing the keyboard for a friend today, I noticed the packaged `.woff2` file.  This update removes it and takes the opportunity to add other small changes (such as for the Description field).

I noticed that when searching for "Ge'ez", or "Geez" at the Keyman website that this keyboard does not appear in the search results.  I have added these two terms to the Description in the hopes that the keyboard becomes discoverable.  If not, I would want to investigate how searching works.


keyman-server:

LornaSIL:
Can you double-check that the OSKFont and DisplayFont in the .kps are both set to the same font that is used in the .kvks (which you just changed to Abyssinica SIL)?

LornaSIL (approved):
Yes, thanks. I think this is correct. When you do the next keyboard you can "relate" it to this one if you want.

PR #3661, opened 2025-08-01
[php_bold] clean up php files that make help headers use bold

LornaSIL:
All the bold headers were bothering me. Some of the styles were unnecessarily overwriting what our standard h1 and h2 use.


DavidLRowe (approved):
LGTM -- Glad you understand this PHP stuff!

PR #3677, opened 2025-08-15
netgkb

nileshbandhiya:
Minor updation in netgkb.kps file for updation at keymanweb.com where example shown how to write.

### DIFF
--- a/release/n/netgkb/HISTORY.md
+++ b/release/n/netgkb/HISTORY.md
@@ -1,6 +1,10 @@
 netGujarati Keyboard Layout Change History
 ====================
+1.0 (2025-08-15)
+----------------
+* Updated netgkb.kps file for the example line for Keymanweb.com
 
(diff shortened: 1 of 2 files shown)


nileshbandhiya:
I am beginner in GitHub and using web environment for edits. Kindly guide me if I am doing wrong.

LornaSIL (approved):
Your changes are fine. 
Explanation for any "wachers": Usually when a change is made to anything in the keyboard package we require a version bump, but in this case, the example text only is used on Keyman Web so that is why we aren't requiring a version bump.

DavidLRowe:
@nileshbandhiya You asked: "Why request is automatically blocked?" We are very grateful for contributions like yours from the wider Keyman community. Unfortunately, for security reasons, we have to review pull requests before accepting them. Thanks for your patience.

PR #3680, opened 2025-08-19
Update Orma keyboard

sewhite:
The Orma community has requested some changes:

1) the t-ejective should be represented by t cedilla, not t with hook. (Lower and upper case)
2) the saltillo should be acccessible in the desktop keyboard
3) on mobile, the saltillo should be a longpress on the q key.

### DIFF
--- a/release/o/orma/source/orma.kps
+++ b/release/o/orma/source/orma.kps
@@ -1,55 +1,57 @@
-<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0"?>
 <Package>
   <System>
     <KeymanDeveloperVersion>18.0.238.0</KeymanDeveloperVersion>
     <FileVersion>7.0</FileVersion>
   </System>
   <Options>
-    <ExecuteProgram></ExecuteProgram>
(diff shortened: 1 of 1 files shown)


sewhite:
I don't see where in the kps file it says either "updated July 2025 by Kevin Warfel" or "Keyboard for the Orma language of Kenya." I have updated the  kps file to include license.md. Did that change register? (I dragged the modified file onto this page, and it said  "Copy" and I let it go, but I don't see any difference. 

Where do I find the orma.php file?

sewhite:
OK,  I think I fixed the KPS and the PhP now.

LornaSIL:
Thank you! Unfortunately I don't see that you've committed the files yet.

LornaSIL (approved):
I had to update the .kps to get the LICENSE.md file in the correct location. I also updated the description field.

sewhite:
Thanks Lorna. How soon will the updated keyboard be downloadable?
Steve White
International Language Technology Consultant

LornaSIL:
Looks like it's there now.

PR #3682, opened 2025-08-21
[sil_kcho] Add key to output combining tilde, and handle NFD combining diacritics

darcywong00:
A user requested the following updates to the sil_kcho keyboard

* Change `~` key from outputting combining acute to combining tilde (a new tone mark)
* Use `_` key to output combining acute
* Add rules for handling NFD combining diacritic input (Mainly FieldWorks interaction with the `ä`,  `Ä`, `ü` and `Ü` vowels)

I also added the new tone marks to the touch layout keyboard.


LornaSIL (approved):
LGTM

PR #3705, opened 2025-09-06
 pre-vedic sanskrit keyboard

vvasuki:
Tested at https://sanskrit-coders.github.io/input/keyman_pre_vedic/test/


keyman-server:

vvasuki:
Ready for review

LornaSIL:
In general your files look good. We do require a copyright holder, so "none" is not acceptable.
The bigger issue is that when we have a non-alphabetic folder name, like `itrans` we assume those keyboards are all by one organization. That means that even though you based your keyboard on the `itrans` one, we need you to remove `itrans` from the name and filename and put it in an alphabetic folder. So, you might wish to call it `devanagari_sanskrit_pre_vedic` and put it in the `d` folder. 

Once you do that I'll check the keyboard more carefully. I'm sorry, I know it's a good bit of work when you have to rename files and all the internal references in the .kmn and kps.

vvasuki:
Renamed the keyboard and the files (using a more proximal standard, OPTITRANS).

> I initially had that same question. However, all the other keyboards in that folder were done by one group, and we think we should keep those files together.

Something is wrong with the semantics here - Dropping them into a separate (sub?)folder named sanskrit-documents might've been better than seemingly tying up a standard with a group.

mcdurdin:
> Something is wrong with the semantics here - Dropping them into a separate (sub?)folder named sanskrit-documents might've been better than seemingly tying up a standard with a group.

The grouping of keyboards is somewhat adhoc. In general, when we have a group of keyboards maintained by a single author/org, then we try and place them together by use of a prefix in the keyboard id. These keyboards had the prefix itrans, and that made sense given they were all related to itrans.

I am not altogether opposed to having keyboards from other authors in the same group if it makes more sense to do so. We just haven't done so before, so we need to adjust our internal guidelines.

vvasuki:
> It looks like all the `RALT` and `RCTRL` rules were removed from the .kmn file (when comparing it to the itrans version of the keyboard. Is that what you wanted?

Yes - wouldn't be used at all (the keyboard is designed only/ mainly for non-touch interfaces).

> If so, then do you want to remove the `rightalt` and `rightctrl` layers on the touch layout? Right now those layers don't output anything because there are no rules in the .kmn to output those keys.

I thought that touch-layout was independent of kmn.  Removed those. 

> The `LICENSE.md` should just have the year `2025` at this point, not `2025-2030`. We'll bump that if there are changes to the keyboard in a subsequent year.

Done.

PR #3717, opened 2025-09-10
chore: remove redundant servervars reference

mcdurdin:
This is just a maintenance cleanup; I noticed that some older .php files had a reference to servervars.php, which is not needed.


LornaSIL (approved):
LGTM

PR #3727, opened 2025-09-17
[php_tables] Fixup some of the help files - tables and graphics

LornaSIL:


DavidLRowe (approved):
I was going to ask:

> easy_arabic: should the four `*.png` files be removed from source/help/ ?
urdu_dvorak: should the eight `mobile*.png` and `typing*.png` files be removed from source/help/ ?

but I see that you had removed those then added them back.

I don't understand, but I'll approve this as is!

LornaSIL:
So, the kps file pulls those graphics from the help folder, which I didn't realize. I did away with the need for the graphics in the help file, but I didn't want to update the keyboard itself so I added it back in.

PR #3728, opened 2025-09-19
Add initial version of linear_b_full_syllabary

WseriesY:
Add the initial version of the Linear B Full Syllabary Signs keyboard.


WseriesY:
Ready for review

WseriesY:
Sincerely thank you for your reply!

There I have found that our existing Linear B keyboard doesn't have some special syllables, such as "dwo", "pte", "twe", et cetra. They are part of the Linear B syllabary system.
So I made a new Linear B keyboard to make sure that these special signs can be typed successfully.

> Already available: https://keyman.com/keyboards/linear_b, whats needed is linear A
> 
> image
>

WseriesY:
Oh, I see.
I have realized how the situation is by your comments. Thanks for your reply.

PR #3729, opened 2025-09-22
[optitrans_devanagari_sanskrit_pre_vedic] nukta vyanjanas

vvasuki:
Fixing a bug.


LornaSIL:
Each time you do an update, besides the bug fix, you must bump the version number in the .kmn and update the HISTORY.md file to show what was changed. This gives us a version history, but more importantly it will then offer the updated keyboard to existing users of this keyboard.

vvasuki:
> Each time you do an update, besides the bug fix, you must bump the version number in the .kmn and update the HISTORY.md file to show what was changed. This gives us a version history, but more importantly it will then offer the updated keyboard to existing users of this keyboard.

Done

LornaSIL (approved):
LGTM. Thank you!

PR #3730, opened 2025-09-22
[php_pagestyle] Have been using wrong command in php files

LornaSIL:
It looks like sometimes I recommended people use `$style` instead of `$pagestyle` which affects whether the correct font (and table settings) are used.


DavidLRowe (approved):
Thanks!

PR #3734, opened 2025-09-24
[php_iconsize] Attempt to fix icon header on help pages

LornaSIL:


release/l/lao_2008_basic/source/help/lao_2008_basic.php
```diff
@@ -2,7 +2,7 @@
   $pagename = 'Lao 2008 Basic Keyboard Help';
   $pagetitle = $pagename;
   $pagestyle = <<<END
-  img {width:100% ;max-width:750px;}
+  img {width:100%; max-width:750px;}
```
mcdurdin: This overrides all images on the page, which is why it's going wrong. We should limit it to specific images

release/sil/sil_kmhmu/source/help/sil_kmhmu.php
```diff
@@ -1,14 +1,12 @@
...
-  table.display tr td { font: 10pt Tahoma; border: solid 1px #ccccff; padding: 4px; text-align: center}
-  table.display tr th { font: bold 10pt Tahoma; border: solid 1px #ccccff; padding: 4px; text-align: center}
-  table.display { border-collapse: collapse; width:640px;}
+  .kmhmu {font-family: "Leelawadee UI","Kmhmu MX","Kmhmu OT"; font-size: 12pt; }
+  .display { border-collapse: collapse; width:640px;}
   img {width:100% ;max-width:750px;}
-  th.narrow {width:40px;}
-  th.medium {width:100px;}
+  .narrow {width:40px;}
+  .medium {width:100px;}
```
mcdurdin, line 8: These changes make the CSS rules less specific, so more likely to go wrong. Again, the img rule is the problematic one

release/t/tibetan_direct_input/source/help/tibetan_direct_input.php
mcdurdin, line 1: I'm not sure what is happening in this file?

mcdurdin (dismissed):
I think the main issue here is with the overly broad `img` CSS rules, which apply to all images on the page -- including the Keyman icon itself.

We can help you get a local copy of help.keyman.com running so you can preview changes, if that would help. Or you could give this kind of work to @Meng-Heng or @Nnyny, who already have local help sites running, making testing and verification much less painful.

release/t/tibetan_direct_input/source/help/tibetan_direct_input.php
LornaSIL: This one had too many sections and you can see at the bottom of the page (at Mobile/Tablet Keyboard Layout) it's all messed up:
https://help.keyman.com/keyboard/tibetan_direct_input/1.4/tibetan_direct_input

LornaSIL:
Thanks @mcdurdin I've made changes, in this case removing the images and using the osk command :)
In the future I will either get @Nnyny or @Meng-Heng to help or I will try to get my local setup working again. 

FYI: I'm gradually reviewing every help page for problems which is why I find it easier to just fix things when I see them rather than waiting on someone else. But, that is a problem when I can't test it.

release/sil/sil_kmhmu/source/help/sil_kmhmu.php
LornaSIL: I removed them and used the osk syntax.

release/l/lao_2008_basic/source/help/lao_2008_basic.php
LornaSIL: I removed and used osk syntax

LornaSIL:
@mcdurdin Even though I changed this PR, it says it can't be merged because you requested changes.

PR #3735, opened 2025-09-26
feat: Add Mara keyboard layout

Laitei40:
This pull request introduces a new keyboard for the **Mara language (ISO 639-3: `mrh`)**, created by Laitei.

The keyboard is designed to be intuitive for users by building on the standard English (QWERTY) layout. It provides a straightforward way to type the special characters required for Mara.

**Key Layout Changes:**

* The **Q** key has been remapped to produce **â** (unshifted) and **Â** (shifted).
* The **X** key has been remapped to produce **ô** (unshifted) and **Ô** (shifted).
* The **J** key has been remapped to produce the digraph **ch** (unshifted) and **CH** (shifted).

The package includes a comprehensive `README.md` with bilingual descriptions (English and Mara), visual layouts for both desktop and touch platforms, and full documentation. All files have been compiled and tested successfully.

#### **Contributor Checklist**

* [x] The keyboard is located in the correct folder: `release/m/mara/`.
* [x] The source files (`.kmn`, `.kps`, etc.) are in the `source/` subfolder.
* [x] The compiled keyboard package (`.kmp`) is in the `build/` subfolder.
* [x] A detailed `README.md` and `welcome.htm` file are included.
* [x] The keyboard has been tested and is fully functional.


DavidLRowe:
Thank you for this pull request. It's well put together. We do, however, need some changes because of how our build system is configured.

(1) Folder/filename conventions:
We require the folder name and the base of many of the filenames to be the same. You have project folder name "mara" (which is good), but "mara_keyboard.kpj" which doesn't match. (We discourage the use of "keyboard" in the name for the keyboard, since it is redundant and may end up showing "Mara Keyboard keyboard" in some contexts.) 

Please change the file names from mara_keyboard*.* to mara*.* so that they match the project folder name. You will also need to change the file references within some of the files (.kmn, .kps, .kvks) to match the renamed files.

Also the &NAME field in the .kmn file should be 'Mara' (rather than 'Mara Keyboard').

(2) There are several informational files:
- The README.md file is for keyboard authors and developers. 
- The readme.htm file is packaged with the keyboard and provides a brief description of the keyboard as the user begins the installation process.
- The welcome.htm file is packaged with the keyboard and is intended to provide usage information.

The usage information that you have put in the README.md file should be moved to the welcome.htm file. You may want to make a "welcome" folder (under the "source" folder) where you can put the welcome.htm file and the images you currently have in the "screenshots" folder. (Otherwise put the welcome.htm file and the images it references in the "source" folder.)

(3) Online help file:
Under the "source" folder you will need to add a "help" folder and a "mara.php" file. This provides the online help. The "mara.php" file can be created from the welcome.htm file (once you have finalized that) following the instructions at https://help.keyman.com/developer/keyboards/phphelpfile . Note particularly the "Creating a keyboard help document from welcome.htm" and "Dynamically Constructed Keyboard Images" sections. 

The welcome.htm will have references to the .png images. If you keep those references in the mara.php file, you'll need to make a copy of those .png files in the "help" folder (in addition to the ones used by welcome.htm). If instead you use the dynamically constructed keyboard images option, you won't need to copy the .png images to the "help" folder. 

(4) On-Screen Keyboard (OSK)
For the desktop layout, the .kvks file (that produces the OSK display) has nothing for the "3" key on the shift layer, so the OSK has a blank key. The keyboard, however, produces "#" for that key. Either the .kvks file (and the PC_Shifted.png image file) should change to include the "#" on the "3" key of the Shift layer, or else the .kmn file should have an additional rule `+ [SHIFT K_3] > nul` to suppress output for that key combination. 

Hopefully, you'll be able to make these changes fairly easily, but if you have any questions, please ask! Thanks again for your contribution.

DavidLRowe:
Your changes to the filenames generally looks good, but there are a few changes needed so that everything works in our automated build system.

The "welcome.php" in the "help" folder needs to be renamed to "mara.php".

In that "mara.php" file, the syntax for displaying keyboard images needs to be corrected:

Desktop layout:
```
</div>
```

Phone layout:
```
</div>
```

If you use that syntax for dynamically created keyboard images, then you can delete the unused *.png and *.jpg images from the "help" folder.

Typically we don't reference the LICENSE.md file in the PHP help file. You can include the copyright statement `Copyright © Laitei` but we strongly recommend not including the date, since otherwise, when the keyboard is updated, this file would need to be updated just to change the year.


You indicated that you were putting the welcome.htm file and the *.png and *.jpg images that it references into a "welcome" folder (under the "source" folder). That is not a requirement. Some authors find it easier to keep the files together in a folder. Others find it easier to put everything in the "source" folder. So it is fine to leave the welcome.htm and the four image files in the "source" folder.

Although the .kvks file was updated to include "#", I don't see that the change to update the PC_Shifted.png image was pushed to your branch to be included in this pull request.

Thanks for your patience!

DavidLRowe:
You still need to update PC_Shifted.png (after the addition of '#' on the '3' key) because that image is used in welcome.htm (which does not use dynamically generated images).

In the mara.php file, the `<h1>Mara Keyboard</h1>` line needs to be removed, since the header already supplies that heading, so the current file would result in a duplicated heading.

Additionally, I think that the `<style>` section of mara.php can be removed (that is all the text from `<style>` through `</style>`, including those markers). It is not used by the PHP file.

A minor suggestion for the welcome.htm file: You might want to move the "Keyboard Layout" section (and the keyboard images) up in the document so that they come before the "Links" heading. This will let users see them without going all the way to the end of the document. But that's a minor point.

Laitei40:
> You still need to update PC_Shifted.png (after the addition of '#' on the '3' key) because that image is used in welcome.htm (which does not use dynamically generated images).
> 
> In the mara.php file, the `<h1>Mara Keyboard</h1>` line needs to be removed, since the header already supplies that heading, so the current file would result in a duplicated heading.
> 
> Additionally, I think that the `<style>` section of mara.php can be removed (that is all the text from `<style>` through `</style>`, including those markers). It is not used by the PHP file.
> 
> A minor suggestion for the welcome.htm file: You might want to move the "Keyboard Layout" section (and the keyboard images) up in the document so that they come before the "Links" heading. This will let users see them without going all the way to the end of the document. But that's a minor point.

A châ, ei pahthâ haw. (Okay, I understand now.)

Here is a reply written in English for you to post on GitHub.

---

Thank you for the follow-up and the clear feedback! I have now pushed all the requested changes.

To confirm:
* The `PC_Shifted.png` image has been updated to show the '#' key.
* In `mara.php`, the duplicate `<h1>` heading and the unused `<style>` section have both been removed.
* In `welcome.htm`, I have also moved the "Keyboard Layout" section up for better visibility, as you suggested.

Everything should be correct now. Please let me know if there is anything else needed.

DavidLRowe (approved):

PR #3737, opened 2025-09-28
[wyandot] Add experimental Wyandot keyboard

MCGallaspy:
Adds an experimental keyboard for the Wyandot language.

### DIFF
--- a/experimental/w/wyandot/source/wyandot.kvks
+++ b/experimental/w/wyandot/source/wyandot.kvks
@@ -5,6 +5,120 @@
     <kbdname>wyandot</kbdname>
     <flags>
       <displayunderlying/>
+      <usealtgr/>
     </flags>
   </header>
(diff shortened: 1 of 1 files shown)


LornaSIL:
In general this looks very good for a keyboard going in experimental.

For the .xml file that you've submitted, can you tell me how you want to use that? You would either want an ldml keyboard to just have an xml file, or you would want to use the .kmn file which it looks like you are using. If the xml file isn't needed, perhaps you could delete it.

We also highly recommend keyboards to use RALT instead of just ALT. That way users won't have as many conflicts with applications that use the Left ALT key for shortcuts.

MCGallaspy:
Thanks for the review! I'll remove the xml file, no reason to include it. I generated it using Keyman Developer, not realizing that it was extraneous. I'll also change to use RALT.

PR #3744, opened 2025-10-04
Follow-up fixes for Mara keyboard

Laitei40:
This PR adds the final updates for the Mara keyboard that missed the last merge. It includes:
- Updated phone layout screenshots.
- Fixes to the package source file to include all images.


keyman-server:

DavidLRowe:
You'll need to update the version number (otherwise the system won't build the new version or offer it to users). That needs to be done in the .kmn file, changing `1.0` to `1.0.1`, for example:
`store(&KEYBOARDVERSION) '1.0.1'`

But for the benefit of humans, a new entry needs to be made at the top of the HISTORY.md file describing the changes for the new version. So, add (above the `1.0` entry): 
```
1.0.1 (2025-10-03)
--------------------
* {description of the change}
```

Laitei40:
Ready for review.

LornaSIL:
The language tag should just be `mrh`. The second set of 4 letters refers to the script code and there is no script called "Mara". `mrh` resolves to `mrh-Latn-IN` and we want the minimal tag which is `mrh`.

Laitei40:
> The language tag should just be `mrh`. The second set of 4 letters refers to the script code and there is no script called "Mara". `mrh` resolves to `mrh-Latn-IN` and we want the minimal tag which is `mrh`.

----

Thanks for the explanation, that makes perfect sense. I've made the correction to use the minimal tag `mrh`, since, as you pointed out, the script and region are correctly inferred. Appreciate the help!

DavidLRowe:
Thanks for catching that, @LornaSIL . 

@Laitei40 the commit you pushed has `mrh-Latn-IN` but it needs to be `mrh`.

PR #3750, opened 2025-10-10
[sil_yi] change mobile layout

DavidLRowe:
The touch layout is quite different. The 47 layers allow any of the 1100+ Yi characters to be typed with just two keystrokes.

The "extras" folder has the spreadsheet and Python script used to generate the .keyman-touch-layout file.

Marked as Draft pending final approval from Dennis.


DavidLRowe:
This PR removes the "phone" layout (which was a standard QWERTY layout and used the .kmn rules to construct syllables by typing between 2 and 5 Latin characters followed by space) and adds a "tablet" layout with multiple layers allowing any Yi character to be composed with just two touches (though longpress or multitap may be required for the radicals and some punctuation).

PR #3764, opened 2025-10-28
[geezword_tigrinya] Fix help CSS indentation level

darcywong00:
Fixes CSS indentation of the geezword_tigrinya help file which blocked help deployment on keymanapp/help.keyman.com#2312 and keymanapp/help.keyman.com#2313

> PHP Parse error:  Invalid body indentation level (expecting an indentation level of at least 2) in ./keyboard/geezword_tigrinya/18.0/geezword_tigrinya.php on line 6


LornaSIL:
There's also an issue of mixing spaces and tabs for indentation. I don't think this one had that problem, but I need to remember to check that also!

PR #3768, opened 2025-11-03
[runeboard] Version 4

Runeboard:
Mostly key repositionings. If there are any problems, please just say.


LornaSIL (approved):
I think it was done correctly. Thank you so much.
It would also be wise to update these other keyboards using the font. Do you want to update those keyboards, or should I?
carian
elymaic
tsakonian
old_english

Runeboard:
Would you mind updating the other keyboards please? That way, I can check how you've done things later if I run into any future problems with confidence that I'm relying upon the work of someone much more experienced in this field.

PR #3770, opened 2025-11-04
[wyandot] Add touch keyboard for Wyandot

MCGallaspy:
This builds but does not install on Keyman Android. I cannot for the life of me figure out why. Can anyone please help?

The specific error message I get when attempting to install the .kmp file is "Keyboard package has no touch-optimized keyboards to install."

### DIFF
--- a/experimental/w/wyandot/source/wyandot.keyman-touch-layout
+++ b/experimental/w/wyandot/source/wyandot.keyman-touch-layout
@@ -1,0 +1,729 @@
+{
+  "phone": {
+    "font": "Tahoma",
+    "layer": [
+      {
+        "id": "default",
+        "row": [
+          {
+            "id": 1,
+            "key": [
+              {
+                "id": "K_Q",
+                "text": "q"
+              },
+              {
+                "id": "K_W",
+                "text": "w"
+              },
+              {
+                "id": "K_E",
+                "text": "e",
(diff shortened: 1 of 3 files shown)


DavidLRowe:
I'll note that your touch layout won't produce output for the `ą` longpress key on `a` for example. The key on the touch layout has ID of `T_new_244` assigned by the Keyman Developer program. Unless you put a rule in the .kmn file that is triggered by `[T_new_244]`, no output will be produced. But a better solution is to change the ID to `U_0105` which will cause the touch layout to output U+0105. (And so on for the other longpress keys.)

I'll try to look at your question about no touch layout tomorrow.

DavidLRowe:
I wonder if the problem is the fact that the version number has not been incremented. It may be that Keyman sees that it already has version 1.0 of this keyboard installed. You could try changing the version in the .kmn file to 1.1 (for example) and rebuilding. (You should also change the HISTORY.md file with the new version information.) You'll need to update the version number information eventually for this keyboard revision to be accepted. 

An alternative for testing would be to uninstall the keyboard from your Android device, then try installing the updated version. I think in that case it shouldn't matter that the updated keyboard has the same version as the existing one (because the existing one will have been removed).

LornaSIL:
See new PR

PR #3776, opened 2025-11-07
[runeboard] Version 4.2

Runeboard:
The prior mobile layout didn't function as intended so I recreated it using Developer.
Other changes are minor.


Runeboard:
> The ID in the touch layout is invalid `U_16EB U_16EB U_16EB`. It should be `U_16EB_16EB_16EB`

Fixed. Thanks for spotting that.

PR #3782, opened 2025-11-11
Soninke N'ti - removed unused punctuation

oyousuf:
Removed 2 punctuation marks that the community told me they aren't using. Added 1 other punctuation mark (inverted exclamation mark) that they are using. Fixed one glyph error in the font. Updated typing manual to reflect these changes.


LornaSIL:
- Update HISTORY.md so it is like the following. The goal is to have the version and date with the changes below it.

```
1.0.2 (2025-11-11)
----------------
* Removed 2 punctuation marks not used by people. Added exclamation mark used by people.

1.0.1 (2025-05-29)
----------------
* Added keystrokes to output ordinal numbers.

1.0 (2024-07-23)
----------------
* Created by Oreen Yousuf

```

- Can you undo the change to the php file? The online help files don't look great with `<h1>` elements because the title of the page is an h1 and it's unnecessary to include the title "Keyboard Layout". If you feel it's necessary, you also need to fix the `</h2?` to `</h2>`
- You'll need to remove the temporary file that got added: `~$iTypingManual.docx` (that is a transient file in Windows when you have the Word document opened)
- If you want to commit your Word document which is used for the pdf, you may, but we would prefer you put it in an `extras` folder so we know it's not part of the package.

oyousuf:
Thanks for catching those errors. I think I fixed them all now.

PR #3784, opened 2025-11-14
[la rarler] Initial submission of la rarler keyboard

dualmeower:


keyman-server:

LornaSIL:
Thanks for your submission!
I haven't tested the keyboard yet, but you'll need some changes made first. Our build system requires a very strict file structure. Right now you have a folder called `la rarler` which has a space in the folder name. Spaces are not allowed and the folder name should be the same as the filenames, so you should rename the folder `la_rarler`.

- We expect the `la_rarler` folder to contain these files which you have not created:
   - HISTORY.md
   - LICENSE.md
   - README.md
   - `la_rarler.kpj` 
- Then create a `source` folder and move the rest of your files into that folder.
- In the end you should have:

```
release\t\la_rarler\HISTORY.md
release\t\la_rarler\README.md
release\t\la_rarler\LICENSE.md
release\t\la_rarler\la_rarler.kpj
release\t\la_rarler\source\la_rarler.ico
release\t\la_rarler\source\la_rarler.keyman-touch-layout
release\t\la_rarler\source\la_rarler.kmn
release\t\la_rarler\source\la_rarler.kps
release\t\la_rarler\source\la_rarler.kvks
release\t\la_rarler\source\welcome.htm
release\t\la_rarler\source\readme.htm
```
- The copyright holder doesn't make sense to me (`la .ynis.`) It looks like something from a hacked font. I think you'll need to give us an actual name.
- I see that the language `jbo` is a constructed language. Unfortunately `Zbal` is not a valid script code and we do not allow private use script codes. You are welcome to use `jbo-Zzzz`. Zzzz represents an uncoded script. https://unicode.org/iso15924/iso15924-codes.html
- The font you have called for is in your own computer so it needs to be added somewhere such as in a `la_rarler\font` folder. Then, you'll need to reference it in the .kps file from that folder rather than your downloads folder.

- After reviewing all of this, because of the private use codes for this script, you will actually be required to put this keyboard in our `experimental` section rather than in `release`. You can change all my comments above that say `release` to `experimental`.
- Also, please delete the file `la_rarler.kmn` that is in the root of the `keymanapp/keyboards` folder.

There are other things that need modifying, but I guess this is enough for you to start with. You may find it easier to close this Pull Request and fix up these issues before submitting your keyboard to the `experimental` section.

For reference, this keyboard might be helpful for you to review since it is also for a constructed language with an uncoded script. https://github.com/keymanapp/keyboards/tree/master/experimental/kreative/kreative_sitelenpona_ucsur

LornaSIL:
@dualmeower I'm closing this because we cannot accept the keyboard as it is. Feel free to update your keyboard per the comments above and resubmit it. There are quite a few complications as the comments suggest so I'm not sure this keyboard is suitable to be in the repo anyway.

PR #3783, opened 2025-11-12
[sil_taang] Convert from TaangU.kmn and add additional punctuation keys

darcywong00:
Fixes #3779 

This converts TaangU.kmn keyboard which @mhosken created and makes the following updateS:
* Rename to sil_taang keyboard package for Rumai Palaung language `rbb` for desktop targets
* Restore punctuation keys for <kbd>(</kbd>, <kbd>)</kbd>, <kbd>-</kbd>,<kbd>=</kbd>, and <kbd>+</kbd>. Those currently output Burmese characters not needed for the language community.
* Add <kbd>[</kbd> and <kbd>]</kbd> to the shift layer. Those currently output Burmese characters can also be removed

PR #3785, opened 2025-11-18
[sil_taang] Add touch layout

darcywong00:
Follows #3783 for the sil_taang keyboard
User also requests a touch layout. I made 3 layers: default, shift, numeric

* Press the top left key (dotted circle) to combine consonants.
* longpress on the top row of keys on the default layer to get numbers.
* Several punctuation characters are on the longpress of the "." key.

User is happy with the layout and can access all their characters

Test-bot: skip


LornaSIL:
Looks good. The .php file references a "phone" layout, but it's a "tablet" layout. Can you fix the php file?

darcywong00:
> The .php file references a "phone" layout, but it's a "tablet" layout. Can you fix the php file?

ok, I ended up changing the touch layout from "tablet" to "phone"

PR #3787, opened 2025-11-20
[amazigh_latin] Link to deprecated keyboards

darcywong00:
Does #3781 (release/a/amazigh_latin) deprecate legacy/a/amazigh_lat and legacy/a/amazigh_lat.kmn?

If so, this updates some DEPRECATED.md files so the keyboard search will show the linking

I also wasn't sure if we need to update the legacy/...keyboard_info files


darcywong00:
I notice there's a build warning
```bash
amazigh_latin.kpj - warn KM09012: Neither the listed BCP 47 tag 'grr-Latn' nor the language subtag 'grr' were found in langtags.json
```

from the .kps
```xml
        <Language ID="grr-Latn">Taznatit (Latin)</Language>
```

PR #3794, opened 2025-11-22
[shanghai_jamo] updating

Cathaylaboratory:
ready for reviewing


LornaSIL:
Maybe I'm missing it, but I don't see that you updated the *.kmn file. Even if there are no changes to the keyboard itself, you'll need to bump the keyboard version to what you did in HISTORY.md.
- It looks like you didn't pull my changes before you made these changes. Two months ago I had to rename your php file because it had a space in the name. I may have made other changes as well. Please pull the changes and then make changes to the php file based on the one already in the repo. The file cannot have a space in the name, it must match the folder/filenames.

PR #3803, opened 2025-11-27
[shan_thailayout] Initial submission of shan_thailayout keyboard

ninenik:


DavidLRowe:
Thanks for adding the shan_thailayout.php help file. This file should either use the .png images from welcome.htm (with `` but not both.
- If you decide to use the images (as you are doing in welcome.htm), then please remove the `<div id='osk...>` lines from the .php file.
- If you decide to use the special keyboard images syntax, then please
  - remove the `` is not available for use in welcome.htm, so these lines should be removed from that file. That will leave the display of the .png images, which will make help available for those using the keyboard.

DavidLRowe:
Someone will look at it on Monday. Thanks for your patience!

release/s/shan_thailayout/source/shan_thailayout.kps
```diff
@@ -0,0 +1,91 @@
...
+      <Description>File tablet_layout.png</Description>
+      <CopyLocation>0</CopyLocation>
+      <FileType>.png</FileType>
+    </File>
+    <File>
+      <Name>help\shan_thailayout.php</Name>
+      <Description>File shan_thailayout.php</Description>
+      <CopyLocation>0</CopyLocation>
+      <FileType>.php</FileType>
+    </File>
```
darcywong00: This file `shan_thailayout.php` will be used for the online keyboard help at help.keyman.com, so it can be removed from the .kps file with the Keyman Developer package editor.

ninenik: Done. I have removed the .php file from the package list.

release/s/shan_thailayout/source/shan_thailayout.keyman-touch-layout
```diff
@@ -0,0 +1,798 @@
...
+            ]
+          },
+          {
+            "id": 4,
+            "key": [
+              {
+                "id": "K_SHIFT",
+                "text": "*Shift*",
+                "width": 160,
+                "sp": 1
```
darcywong00: The shift key on each of the layers don't seem to function. I think you need to assign a "next layer" for each of the shift keys.

default layer +  <kbd>shift</kbd> = shift layer
shift layer + <kbd>shift</kbd> = default layer?
alt layer + <kbd>shift</kbd> = ? layer

darcywong00: My other comment about the touch layout is that there's quite a few unassigned keys on each layer: ~6 on default layer, 3 on shift layer, 2 on alt layer).

It would be good to "touch-optimize" the layouts by removing the blank keys, and re-distributing the keys on the rows. This way on smaller devices, the keys aren't too small to type.

Ideally, the number of keys on the rows wouldn't be changing going from one layer to another (so it won't be jarring to the users)

ninenik: Thanks for the review. I have updated the touch layout based on your feedback:

Fixed Shift Keys: I have assigned the "next layer" for the shift keys, so they now switch layers correctly.
Optimized Layout: I have removed the unassigned/blank keys to improve the layout on smaller screens.
Updated Image: I have also updated the tablet_layout.png to reflect these changes.

Ready for review.

DavidLRowe:
Thanks for your work in addressing the issues that @darcywong00 pointed out. I have just a few more comments:

On the Packaging, Details tab, for the "Welcome file" field, please select your "welcome.htm" file.

The desktop layout has some empty keys on the on-screen keyboard (OSK) image. Currently those are unmentioned in your .kmn file and so they produce what is on the US English (QWERTY) keyboard layout  (`7tz/~Y for the default and shift layers).

- If you want that behavior, then you should update the OSK accordingly by adding the character to the OSK.
- If you don't want that behavior you should add a rule to make the key produce nothing. For example:
```
store(unused) "`7tz/~Y"
+ any(unused) > nul
```

The following are optional:
You've addressed the problem of the shift key not returning from the shift layer (and likewise for the alt layer). There are a few minor changes that might make the keyboard easier to use. 

- Currently on the alt layer, the shift key does nothing. You could make it activate the shift layer. Likewise on the shift layer, the alt key does nothing. You could make it activate the alt layer.
- Currently on the shift layer, the shift key has a "Key type" value of "special". You could make it "special (active)" and then the shift key on the shift layer will show as activated. Likewise for the alt layer.

DavidLRowe:
BTW: You will always see a "build failed" message when you add a new commit to this pull request. Someone on the team needs to manually trigger the build and we generally wait until the submission is reviewed before doing that.

release/s/shan_thailayout/source/shan_thailayout.keyman-touch-layout
```diff
@@ -0,0 +1,747 @@
...
+                "id": "K_5",
+                "text": "ၻ"
+              },
+              {
+                "id": "K_6",
+                "text": "ု"
+              },
+              {
+                "id": "K_7"
+              },
```
darcywong00, line 38: > The desktop layout has some empty keys on the on-screen keyboard (OSK) image. Currently those are unmentioned in your .kmn file and so they produce what is on the US English (QWERTY) keyboard layout (`7tz/~Y for the default and shift layers).

The touch layout also currently has empty keys for 
* <kbd>7</kbd> and <kbd>t</kbd> on the default layer
* <kbd>Y</kbd> on the shift layer

darcywong00: shan-mobile-default

The last key on each row will have a width to fill all the remaining space.
If you don't want these extra-wide keys, you have some options:
* re-distribute more keys (for example, some keyboards move the backspace key to above the <kbd>enter</kbd> key.
* create "spacer" keys in Keyman Developer that have blank text, and a width. See
 https://github.com/keymanapp/keyboards/blob/b5598bb8f340a0932f9e3429773cc1be7b943570/release/sil/sil_euro_latin/source/sil_euro_latin.keyman-touch-layout#L855-L860

ninenik: I have addressed all the feedback from @DavidLRowe and @darcywong00:
Ready for review.

DavidLRowe (approved):

DavidLRowe:
Thanks for your patience. The keyboard should be available before too long at https://keyman.com/keyboards/shan_thailayout

PR #3822, opened 2025-12-16
[KbdArabHarmattan] Add font for OSK use

LornaSIL:
This font can be used for any of the keyboards that use Harmattan. The design is preferable than Scheherazade for some of the African languages.


DavidLRowe (approved):
RSLGTM

PR #3837, opened 2025-12-23
[spanish_telex] updating

Cathaylaboratory:
Ready for reviewing


LornaSIL:
I can approve this, but I have two comments that I'd like you to think about first.
- Just using the `ALT` key is problematical for use in many applications and for cross-platform use. It would be better to use `RALT` instead.
- There is no documentation adjustment so no will know about these ALT/RALT keys and how to access the quotes. You could consider updating the welcome.htm and online help file.
- 
Since this is in experimental, I won't require you to make the changes, but it would be helpful to your users.

PR #3840, opened 2025-12-31
[linear_b_full_syllabary] Second Amendment of Linear b full syllabary

WseriesY:
Recently I found that the Right Alt layer did not work. I have amended some codes to recover the on-screen RALT layer.
Thanks for checking my second repull!


WseriesY:
Ready for review

LornaSIL:
Mostly the pull request looks good. 
- You should increase the version number in the .kmn and document the new version number and what has changed in the HISTORY.md file. Otherwise, without a version bump these changes will not be offered to the users.

PR #3842, opened 2026-01-08
Fix Spacebar (U+0020) issue in v1.0.3

Laitei40:
### Summary
This pull request fixes an issue introduced in v1.0.3 where the Spacebar stopped producing a space character.

### Details
- Explicitly handles the Space character (U+0020) in the keyboard source
- Restores normal typing behavior across platforms
- Aligns with updated Keyman engine behavior

### Notes
- No layout or character changes
- No impact on existing mappings other than Space handling

### DIFF
--- a/release/m/mara/source/mara.kmn
+++ b/release/m/mara/source/mara.kmn
@@ -18,7 +18,7 @@ group(main) using keys
 c =========================
 c LETTERS (NCAPS = normal)
 c =========================
-+ [K_SPACE] > U+0020
+
 + [NCAPS K_A] > 'a'
 + [NCAPS K_B] > 'b'
 + [NCAPS K_C] > 'c'
@@ -196,5 +196,3 @@ c =========================
 c =========================
 c FALLBACK (CRITICAL)
 c =========================
-
-nomatch > use(main)


release/m/mara/source/mara.kmn
mcdurdin, line 1: This keyboard has a recursive loop for unmatched keystrokes, which is why the space character is not working. After 50 recursions, in order to prevent a hang, Keyman Core aborts rule processing, and ignores the key event altogether.

So, you should delete line 200 (`nomatch > use(main)`), and then line 21 (`+ [K_SPACE] > U+0020`) will not be necessary, as it will be handled by default processing.

laiteichozah: Thank you very much for the clear explanation.

That makes sense — the nomatch > use(main) rule was indeed causing a recursive loop for unmatched keystrokes, which explains why the space key stopped working and why Keyman Core aborted processing after reaching the recursion limit.

I will remove line 200 (nomatch > use(main)) as suggested. With that rule gone, line 21 (+ [K_SPACE] > U+0020) is no longer necessary and can safely be removed, allowing the space key to be handled by default processing.

I appreciate you taking the time to analyze the issue and explain the underlying behavior so clearly. Thank you for the help.

mcdurdin: You're welcome!

laiteichozah: I’ve now removed line 200 (nomatch > use(main)), and also removed line 21 (+ [K_SPACE] > U+0020) as discussed.

With these changes, unmatched keystrokes are handled by default processing and the space key works correctly.

Thank you again for your guidance.

LornaSIL (approved):
Hope it works this time for you!

PR #3849, opened 2026-01-13
[sil_colombia_venezuela] New keyboard: sil_colombia_venezuela

joserosas-silamericas:
### Summary
This PR adds a new unified keyboard for Indigenous languages of Colombia and Venezuela.

### Details
*   **ID:** sil_colombia_venezuela
*   **Name:** Teclado Unificado Indígena (Colombia y Venezuela)
*   **Languages:** Supports 60+ languages (Wayuunaiki, Barí, Yukpa, Sikuani, etc.) using BCP 47 tags.
*   **Copyright:** SIL Global.
*   **Features:** Includes Andika font, touch layout, and rotation/deadkey input methods.


joserosas-silamericas:
Ready for review

joserosas-silamericas:
Thanks for the instructions. 
**I just didn't make one of the changes** you mention, about the Latn tags for each language. **Actually, I would like to ask you to allow me to leave it as it is. Let me explain why:** 

The thing is, I have finished my .kmn code and have been testing it since December. I had different types of problems and, after investigating the reason, I understand that it was because, at the Windows level, the OS gets confused about which base language to associate with the keyboard, despite all the other settings. What happened to me in the tests on three laptops was that after a period of use (minutes or hours), the Keyman keyboard would deactivate and the layout would switch to US English (which is very confusing to use in South America), the name of the keyboard in the Windows language bar would change to something strange, and when I clicked on it, the menu would open on the left side of the screen. This could not be solved by restarting; the only way was to either delete and reinstall (after doing this 3 or 4 times, it no longer worked either), or open Keyman Configuration and try to link it to Spanish as the base. 
Looking for a solution on the Web, I found this space to specifically indicate that each language must be linked to a Latn layout to prevent Windows from “getting confused.” Since I applied this, I haven't had the problem again. 

(On a separate note, on other keyboards that Spanish-speaking colleagues made locally, this issue of Windows ‘confusing’ the keyboard layout is so common that it has discouraged us from even reaching this point of seeking to publish it. I believe that adding Latn in our case could fix the problem for others).

Pepe.

PR #3851, opened 2026-01-13
[rac_2] Tidy up files, add KbdArab.json, add KbdArabScheherazade

LornaSIL:
- Tidying up first half of rachitrali keyboards
- Add KbdArab.json file for OSK display
- Add KbdArabScheherazade font for OSK display
- Changed some of the keyboards to use Scheherazade New rather than Scheherazade


DavidLRowe:
Another huge job! I've approved it and you can merge when you like.

PR #3854, opened 2026-01-14
Update karambolpoular

sven-oly:
The author has requested an update to this layout in experimental. I attempted to create a new layout under the name CarambolPoular, but reviewers suggested that this should be done as an update to this layout.


LornaSIL:
This is VERY confusing. The author said he wanted `karambolpoular`
deprecated in favor of `carambolfulbhe` which we added. Until we have
better clarity I'm not inclined to update  `karambolpoular`. See:
https://keyman.com/keyboards/carambolfulbhe

PR #3863, opened 2026-01-27
[linear_b_full_syllabary] Update for RALT

WseriesY:
Formerly I was suggested that I should start my new version (1.2) with the files already in the repo rather than the v1.0 files, and I should correct the faults in HISTORY.md. 
Here I have tried to fix the problems above, and tested the keyboard on all of my devices. Thank you for your check!


LornaSIL (approved):
So, it looks fine. I think the issue was that the keyboard version wasn't bumped so your changes in v1.1 never made it to the build. Sorry for missing that when we checked it!

PR #3866, opened 2026-02-02
[athinkra_mwangwego] Keyboard v1.1

dyacob:
This is an initial submission of the _**experimental**_ GFF Mwangwego desktop keyboard.  It uses the provisional encoding of the Mwangwego script that is slated for Unicode v19.0. A font meeting the encoding requirements is included with the package.


LornaSIL:
Here's some initial feedback in no particular order.
- In the root, remove the `gff_mwangwego.keyboard_info` file as we no longer use that. 
- In the .kmn file select "desktop" and "web" as targets. No need to select the OSes separately. I'm assuming you want web since you included the .kvk in the .kps
- In the .kmn you should remove the target for the mobile layout. I guess I'm okay with you committing the `keyman-touch-layout` since you've put some work into it, but your documentation says this is a desktop layout so comment out this line:
`store(&LAYOUTFILE) 'gff_mwangwego.keyman-touch-layout'`
- I'm not sure what this file is for `gff_mwangwego.css`. It's empty so you could blow it away and not reference it in the .kmn.
- .kps
   - BCP47 tags. What you've used would default to Latin I would think. You should use `Zzzz`, so the tags would be `ngl-Zzzz`, etc.
   - My preference is not to include the copyright years in the .kmn and .kps files. Just include the years in LICENSE.md
   - You'll need a description on the **Details** tab. It can be the same as what you have in `README.md`.
   - The website link would be better to say `https://keyman.com/keyboards/gff_mwangwego`
   - On the **Keyboards** tab while I was testing, I think I typed in the information for the font, make sure you do too.
- You've committed `Mwangwego-Typing-English.log` which was a log file created when you were making the pdf. That should be removed.
- If this image is not used, you could remove it: MwangwegoBanner.png
- I'm not sure this file needs to be in the repo. If so, maybe move it to extras? build-command.txt
- The README.md file indicates you are using provisional assignments. In the Word document and pdf it says "Unicode Supplementary Private Use Area-A.". You should change that to talk about the provisional assignments.
- One last thing, for keyboards for scripts not in Unicode yet I usually ask people to make it clear that codepoints can still change. So, I'd like the documentation in the README.md, welcome.htm, pdf, and the description in the .kps to say something like:
   - This is an experimental keyboard for the Mwangwego script.  It requires the included “Mwangwego Book” font which in turn encodes the script under its provisional encoding for Unicode 19. See the typing manual for full details of the typing rules. **Since this keyboard uses a provisional encoding, those codepoints can change. Any documents created using this keyboard could become obsolete if the codepoint assignments change.**  

I think that's it for now.

dyacob:
@LornaSIL , I believe each items has been addressed now.

Some work advanced further on the mobile keyboard, and it should be complete by the next update.  The CSS file has some content now, too.  I've kept them in the repository for safekeeping, though they are not yet build targets.

LornaSIL:
There's an `release/athinkra` folder. If you want to create an `experimental/athinkra` folder I would be okay with that. So far this is only the second `athinkra` keyboard though, so it's not necessary to distinguish. I think we thought more would be coming.

PR #3877, opened 2026-02-13
[manchu_cyrillic] keyboard

Cathaylaboratory:
Manchu cylliric keyboard submission.
This is the keyboard for Manchu cylliric script


keyman-server:

LornaSIL:
Thanks for you PR. Because the word Cyrillic is misspelled in the folder and filenames, I think it would be best to rename the folder and the filenames from:
`manchu_cylliric` to `manchu_cyrillic` 

The content of all the files should also be reviewed to be spelled "Cyrillic"

I see the php file has a header from a previous keyboard, so check that kind of thing too.

Remember we have a *very* strong preference for using `RALT` and not just `ALT` since that works better in applications and especially cross-platform.

I didn't really check the keyboard yet, but I believe these issues should be fixed first.

PR #3890, opened 2026-02-24
[tay_latn] New keyboard

LornaSIL:
New keyboard based on old kmn one in ScriptSource.


DavidLRowe:
tay_latin.php and welcome.htm
- first list item seems to duplicate the third item, but maybe I'm missing something.
- second list item has "...and o, O, u, or followed..." which seems to be lacking "U" between "or" and "followed". 
Otherwise looks fine.

PR #3937, opened 2026-03-28
[cantonese_telex] updating

Cathaylaboratory:
ready for reviewing


keyman-server:

LornaSIL:
Is this PR ready for review? When you first submitted it you said it was ready and then you updated a bunch of times.

It also concerns me that this keyboard is in `release` but you keep changing the keyboards significantly (including which characters are output!). I would expect keyboards in `release` to be more stable than this keyboard seems to be.

Cathaylaboratory:
Sorry, this time I find I still need to apply small changes on the keyboard after I upload it. I think this time it should be OK now.
I changed the ṳ to ŭ because I found the ṳ doesn’t appear correctly on many platforms. The ŭ works better. And the ŭ letter is not using new tone marks or new strokes.
I believe as since I update the keyboard, much problems have already been sorted out. So the frequency of updating of this keyboard will significantly go down in the future.

LornaSIL (approved):
LGTM

Cathaylaboratory:
Thank you.

PR #3964, opened 2026-04-20
[khmer_traditional] keyboard submission

Meng-Heng:
This keyboard has a new font that requires a final confirmation to be able to use, and there's additional typing testing before it's ready for a full review.


keyman-server:

LornaSIL:
The keyboard itself looks pretty good. One thing is that the font isn't selected in the .kps in the **Keyboards** tab.

The big thing though is that the font does not have a clear Open license, so we cannot accept those two fonts. They need to be licensed under OFL, GPL, MIT, or Apache would be okay. It cannot say freeware, or be empty. These fonts have an empty license field.

Meng-Heng:
Hello, @LornaSIL. We've decided to take the font out and use KbdKhmer & Busra. Please review.

LornaSIL:
@Meng-Heng This mostly looks good. There are two remaining issues:
- The OSK should be modified to support `RA` and `SRA`, rather than `A` and `SA`. You can make that change in the code view rather than reimporting the OSK if you want. For example, change `<layer shift="A">` to `<layer shift="RA">` and `<layer shift="SA">` to `<layer shift="SRA">`
- The other is an issue I hadn't realized we have. We have the old version of Busra in the repo and the new version. You should use the newer fonts from here: `keyboards\release\shared\fonts\sil\busra` rather than the fonts from `\release\shared\fonts\khmer\khmerbusra` I will do a separate PR to remove these old fonts to avoid future confusion. (I just did a check and there aren't any other projects using them.)

release/k/khmer_traditional/source/help/khmer_traditional.php
```diff
@@ -0,0 +1,35 @@
...
+
+<p>
+   The Khmer Traditional keyboard is designed to facilitate a seamless transition from the standard Microsoft
+   Khmer keyboard layout to the Keyman platform. It preserves the canonical character inventory and key mappings
+   familiar to users across both desktop and mobile environments, thereby minimizing cognitive load and supporting continuity
+   in typing Khmer alphabets.
+</p>
+
+<p>
+    A distinctive feature of this keyboard is the inclusion of the character ឲ, which is mapped to the <kbd>Right</kbd> + <kbd>Alt</kbd> + <kbd>y</kbd>
```
darcywong00: I would combine Right+Alt to one key

```suggestion
    A distinctive feature of this keyboard is the inclusion of the character ឲ, which is mapped to the <kbd>Right Alt</kbd> + <kbd>y</kbd>
```

Same block on welcome.htm too.

release/k/khmer_traditional/source/welcome/welcome.htm
```diff
@@ -0,0 +1,62 @@
...
+<h3>Default</h3>
+<img src="DesktopU_.png" alt="Desktop Default layer of Khmer Traditional Keyman keyboard">
+
+<h3>Shift</h3>
+<img src="DesktopU_S.png" alt="Desktop Shift layer of Khmer Traditional Keyman keyboard">
+
+<h3>Alt</h3>
+<img src="DesktopU_A.png" alt="Desktop Right Alt layer of Khmer Traditional Keyman keyboard">
+
+<h3>Shift Alt</h3>
```
darcywong00: > The OSK should be modified to support RA and SRA, rather than A and SA.

Maybe these headers should update to `Right ALT (unshifted)` and `Right ALT Shift`

release/k/khmer_traditional/source/welcome/welcome.htm
```diff
@@ -0,0 +1,62 @@
...
+<p>
+   The Khmer Traditional keyboard is designed to facilitate a seamless transition from the standard Microsoft
+   Khmer keyboard layout to the Keyman platform. It preserves the canonical character inventory and key mappings
+   familiar to users across both desktop and mobile environments, thereby minimizing cognitive load and supporting continuity
+   in typing Khmer alphabets.
+</p>
+
+<p>
+    A distinctive feature of this keyboard is the inclusion of the character ឲ, which is mapped to the <kbd>Right</kbd> + <kbd>Alt</kbd> + <kbd>y</kbd>
+    key combination. This character is not currently represented in the standard Microsoft Khmer layout,
```
Meng-Heng:
```suggestion
    A distinctive feature of this keyboard is the inclusion of the character ឲ, which is mapped to the <kbd>Right Alt</kbd> + <kbd>y</kbd>
    key combination. This character is not currently represented in the standard Microsoft Khmer layout,
```

LornaSIL:
The keyboard looks really good. There's one other issue, and I'm sorry I didn't notice it earlier. Since you are using `KbdKhmr.ttf` in the .kps, you need to add a header statement in the .kmn:

`store(&DISPLAYMAP) '../../../shared/fonts/kbd/kbdkhmr/KbdKhmr.json'`

That file gets compiled in the .kmx so there is no need to add that file to the .kps.

FYI: The `KbdKhmr.ttf` is a PUA encoded file which gives better onscreen display for combining marks. But, the `KbdKhmr.json` file tells Keyman how to display Unicode codepoints to the PUA codepoints.

Meng-Heng:
Thank you. This should be ready for your review.

LornaSIL (approved):
LGTM

PR #3967, opened 2026-04-24
[remove_old_busra] Remove old Busra and Mondulkiri

LornaSIL:
No keyboards are currently using the old version of Busra (new version is in a different location).
No keyboards are using Mondulkiri (replacement font is Busra)
Removing the fonts to avoid confusion so they will not be selected.


DavidLRowe (approved):
RSLGTM

PR #3968, opened 2026-04-27
[moro_language] add moro translation for help - welcome

rc-swag:
Adds the Moro translation for the Moro Keyboard help.

Also fixed the table for keyboard rules in moro.php which was not displaying correctly.

Replaced an incorrect unicode character in the table.


LornaSIL:
A change to `KbdHebr.sh` crept in here. That should be removed. And, do remember even if it's just a change to documentation, the keyboard version will need bumping in the HISTORY.md and .kmn.

LornaSIL:
Thanks for your submission. When a file that is part of the keyboard package (like welcome.htm) gets changed, then the .kmn version must be incremented and a description of the change in HISTORY.md should also be made toward the top of the file (not the bottom).

LornaSIL (approved):
Great, glad to see the use of the features. LGTM.

PR #3986, opened 2026-05-15
Add Mising (Latin) keyboard layout

Vivian5G56:
This keyboard layout is designed to support the Mising Agom Kébang (MAK) orthography. It includes full support for special characters and long-vowel markers (such as é and í) to assist with linguistic documentation and community use.


keyman-server:

Vivian5G56:
Ready for review

LornaSIL:
I'll check it next week. However, since this is a keyboard repo, we find including the word `keyboard` in the keyboard name and/or ID to be redundant. Please make that change to remove that word from the folder, filenames, and keyboard name. Thanks!

Vivian5G56:
Thanks for the feedback. I will update the filenames and the internal keyboard name to remove the word "keyboard" as requested, and upload the updated version soon.

Vivian5G56:
I have successfully updated the project folders and IDs from 'mising_keyboard' to just 'mising' to remove the redundancy.                                                                                                                                                   Thanks !

LornaSIL:
Thanks for fixing up the naming. Here are some other things to adjust:
- The .kvks file needs you to import the layout (right now it's blank). Even if it's the same as the standard keyboard, if you leave it blank the keys will also be blank which is confusing to the user.
- In the .kmn, make sure the targets is just `any`. Don't include the other targets since `any` covers them all.
- The .php file needs all the style information within the php header. You can look at an example of that in other php files in the repo. Here's one example: https://github.com/keymanapp/keyboards/blob/master/release/sil/sil_cheyenne/source/help/sil_cheyenne.php

Vivian5G56:
Thank you for the feedback and guidance! I have successfully made all three adjustments:

Updated the .kmn file targets to any.

Populated the .kvks file layout so the visual keys are no longer blank.

Moved the CSS styling inside the $pagestyle variable in the mising.php file header to match the template.

LornaSIL:
It looks very good. One last thing, could you please remove the version number from the welcome.htm and the mising.php file. Those are places the version isn't necessary and it's just one extra place we would have to update for any future changes. We prefer the version is only in the HISTORY.md and .kmn file.

Vivian5G56:
I have removed the version numbers from both welcome.htm and mising.php as requested. 

Thank you!

LornaSIL (approved):
LGTM

Vivian5G56:
Thank you so much for the guidance and for merging the keyboard I really appreciate your help throughout this process.

LornaSIL:
@Vivian5G56 You are very welcome. You made it easy.

PR #3998, opened 2026-05-28
[hanifi_rohingya_alternate] new keyboard

JustaRed:
This alternate Hanifi layout changes positions of keys for an easier and more intuitive typing experience in Hanifi.

Included are:

- .kmn and .kps files
- Two fonts: Noto Sans Hanifi Rohingya and Noories One
- welcome.htm and keyboard guide
- two .pngs for the keyboard guide
- .ico file for the keyboard


keyman-server:

LornaSIL:
Thanks for fixing things up. There are still some issues that should be resolved, but hopefully now that everything is in the right location it will be easier.

- In the .kmn file you need to reference the on-screen keyboard and the touch layout. You should add these header statements:
```
store(&VISUALKEYBOARD) 'hanifi_rohingya_alternate.kvks'
store(&LAYOUTFILE) 'hanifi_rohingya_alternate.keyman-touch-layout'
```
Also, in the .kmn there are some `nul` rules, but you have earlier rules that define those keys:
```
+ [K_E] > nul
+ [K_U] > nul
+ [K_F] > nul
+ [K_G] > nul
+ [K_J] > nul
+ [K_K] > nul
+ [K_L] > nul
+ [K_M] > nul
```
You probably meant to make the SHIFT keys nul instead. You could change the above rules to this:
```
+ [SHIFT K_E] > nul
+ [SHIFT K_U] > nul
+ [SHIFT K_F] > nul
+ [SHIFT K_G] > nul
+ [SHIFT K_J] > nul
+ [SHIFT K_K] > nul
+ [SHIFT K_L] > nul
+ [SHIFT K_M] > nul
```
- The copyright in the .kmn is not the same as elsewhere. Change that from `Just_Red` to `Rohingyan`
- You should change the target to "any": `store(&TARGETS) 'any'`

- Next we will look at the touch layout.
   - When the phone and tablet layout are the same, we only want one. You can delete the phone layout since it is the same as the tablet. Then, you only need to manage one touch layout and it will work on both phone and tablet.
   - Next, the current layout will not let you change layers. You should select the **SHIFT** key and change **Next Layer** to "Shift"
image
  - Then, go to the Shift layer and select the **SHIFT** key and now change **Next Layer** to "Default". This will allow the user to change from default to shift and back.
image

- Now, we have several issues with fonts. We like to only include one version of a font in the repo, for that reason we have a folder for shared fonts. The Hanifi ones are here: `keyboards\release\shared\fonts\noto\Rohg` You should remove the fonts and the font folder from your pull request and use the shared fonts. That way we aren't distributing multiple versions of the same font. Also, we cannot accept the font "RohingyaNooriesOne.ttf" because that font doesn't have a license and we require fonts in our keyboard packages to be under an open license, such as the Open Font License, MIT, or GPL.

Now, we'll look at the .kps
- You need to add `LICENSE.md` to the list of files
- You'll need to remove the two fonts from the list of files and then add the fonts from the above Rohg folder.
- Go to the **Keyboards** tab and add in the font to **Keyboard font** and **Display font**
- The BCP-47 tag is incorrect. Please change that to `rhg-Rohg` (`Hani` represents "Han (Hanzi, Kanji, Hanja)" which I'm sure you don't want!). You could use the minimal tag which is just `rhg`.
- Now, go to the **Details** tab. The copyright information again should be changed to `Rohingyan`
- Fill in the welcome, readme, and license information as below:

image

Those files are all used in displaying the information in the keyboard package as well as on the website.

- We also require any keyboard that goes in our `release` folder to include a help file. This generally has the same information as what is in the welcome file. That file is used as an online help.
   - Create a `help` folder within `source`
   - Create a file called `hanifi_rohingya_alternate.php`
   - Open that file in a text editor and paste the following text into that file.

```
<?php 
  $pagename = 'Hanifi Rohingya Alternate Keyboard Help';
  $pagetitle = $pagename;
  require_once('header.php');
?>

<p>
  This is a Hanifi keyboard which lays out all the keys in a more understandable, suitable way, which also allows for space for some letters not included in the normal Hanifi keyboard.
</p>

</div>
```

JustaRed:
As I am new to keyboard creating and publishing, how do I use the shared Hanifi fonts? Also I have made fixes in my repo.

LornaSIL:
I updated the .kps to use the shared fonts. I'll finish checking it soon.

LornaSIL (approved):
LGTM. Thanks! Should be available shortly.

PR #3999, opened 2026-05-29
Made some difficult keys easy, and removed some unused characters.

ghost:


keyman-server:

ghost:
ready for review

LornaSIL:
Thanks for your submission. 
- You have an ID in the new touch layout `T_new_384` that doesn't produce anything because there is not rule for firing it. You might want to change it to `U_0640`. I'm not sure why an error isn't being produced!
- Since you significantly changed the touch layout you might want to regenerate your graphic images that are in the `help` folder. Don't change the file names so that the welcome.htm and help file don't need to be updated.

Otherwise, the changes are fine.

ghost:
ready for review

LornaSIL:
For better display of combining marks we have two special fonts designed for the OSKfont for Arabic script. We should not use Harmattan for the `<OSKFont>`. We need to keep `\kbd\kbdarab\KbdArabScheherazade.ttf` as the font used for display. Or, if you like the Harmattan font for the display font we could use `\kbd\kbdarab\KbdArabHarmattan.ttf` but we should not change to just "Harmattan". When people are using the keyboard it will still use Amiri.

ghost:
sounds good.

LornaSIL (approved):
Thanks for fixing those issues! LGTM.

PR #4008, opened 2026-06-04
[mising_php] Revise help file again for better display

LornaSIL:
I did actually test this locally this time. All the extraneous code was making it display strangley.

### DIFF
--- a/release/m/mising/source/help/mising.php
+++ b/release/m/mising/source/help/mising.php
@@ -4,7 +4,6 @@
 
   $pagestyle = <<<END
   /* ── Base ── */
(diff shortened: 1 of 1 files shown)


release/m/mising/source/help/mising.php
```diff
@@ -6,63 +6,7 @@
   /* ── Base ── */
   *, *::before, *::after { box-sizing: border-box; margin: 0; padding: 0; }
```
mcdurdin:
```suggestion
```

Generally, top-level changes like this should not be included in CSS for docs

mcdurdin (dismissed):

mcdurdin (approved):

PR #4016, opened 2026-06-10
[tai_lanna_thai] Update the name, but not the id

### LINKED ISSUE #3689: [tai_lanna_thai]: rename keyboard discussion
Hello, @LornaSIL and @Jayasiri-Jayasaro.

Could we rename the keyboard?

> From Ben:
> I might suggest calling it 'Tham Lanna' instead of 'Tai Lanna', since 'Thai Lanna' is an alternative name for Fakkham (https://en.wikipedia.org/wiki/Fakkham_script)

Original post: https://typo.social/@keyman/115065717599732504.

LornaSIL:
Fixes #3689 
We chose not to change the ID.


LornaSIL:
@jahorton I understand you are willing to check PRs. Marc and I decided we could rename the keyboard, but not the folder/files.

mcdurdin (approved):

PR #4027, opened 2026-06-15
[mara] Update Keyboard to 1.0.7 (Spacebar Fix)

Laitei40:
This pull request updates the Mara keyboard to version 1.0.7 to resolve a critical bug affecting basic typing functionality.

### **Changes in this release:**

* **Spacebar Fix:** Fixed an issue where the spacebar was failing to output spaces. The `nomatch > use(main)` fallback rule was unintentionally swallowing the keystroke because the space key was not explicitly defined in the ruleset. 
* Added `+ [K_SPACE] > U+0020` to the `.kmn` source file to resolve the issue.
* Updated the `HISTORY.md` and package version numbers to `1.0.7`.

Build tested successfully locally.


keyman-server:

release/m/mara/source/mara.kmn
mcdurdin: The `nomatch > use(main)` line is incorrect: it leads to a recursive call to the `main()` group, which gets caught eventually by Keyman's safety valve on recursive calls, so it doesn't crash, but it does mean you'll get no output on the keystroke. This is why you were seeing spacebar not working -- it wasn't matched in the group, and then the `nomatch` rule caused it to recurse and eventually safely abort, albeit without any output.

To fix this, delete both of these rules.

```suggestion
```

mcdurdin (dismissed):

release/m/mara/source/mara.kmn
Laitei40: Hi @mcdurdin,

Thank you so much for the detailed explanation! That makes perfect sense as to why the spacebar was failing to output anything without actually crashing the keyboard. 

I have applied your suggested change and deleted both the `nomatch > use(main)` rule and the explicit spacebar mapping. The unmapped keys are now passing through normally. 

I've updated the PR with this fix for version 1.0.7. Thanks again for catching this and for your help!

LornaSIL (approved):
LGTM

LornaSIL (approved):
LGTM

PR #4043, opened 2026-06-29
[arabic_flick] Flick Layout for Arabic

ghost:


ghost:
should the k_left and k_right in touch layout be a default key, special, or something else?

keyman-server:

LornaSIL:
Thanks for your submission.

I think your users would have a better experience if you use `RALT` instead of just `ALT` in your rules. That is because many applications use `ALT` for shortcut keys. `RALT` also works well cross-platform.

The font you have chosen is one that isn't well documented, and I don't think it's what you actually want.
I'll try to explain it.

- Because Arabic (and other scripts) has combining marks that don't always display well by themselves on the visual keyboard, we developed a way for better display.
- We created a special font that should ONLY be used for display. It is PUA encoded. We have the one you used called `KbdArabScheherazade.ttf`.
- The USER will not even want to know about that font. You would still select the font the user wants `\release\shared\fonts\sil\scheherazade_new\ScheherazadeNew-Regular.ttf
- In order to use this special font, you need to do the following:
   - Add this header into your .kmn: `store(&DISPLAYMAP) '../../../shared/fonts/kbd/kbdarab/KbdArab.json'` (This .json file is what maps Unicode characters to PUA characters for that special font).
   - The .kvks should still use "Scheherazade New"
   - The .kps should include all the Scheherazade New files as well as `KbdArabScheherazade.ttf` in the list of files.

image

   - Then, on the **Keyboards** tab, you would have something like this:

image

You can read more about this method here:
- https://help.keyman.com/knowledge-base/?id=115

If you don't want this complication, you can just not use `KbdArabScheherazade.ttf` at all and switch to `ScheherazadeNew`. The user will have a better experience if you do the steps above, but it is a bit fiddly to implement correctly, and it's not well documented.

LornaSIL (approved):
LGTM

mcdurdin:

This keyboard has now been published and will shortly be available at: https://keyman.com/keyboards/arabic_flick.

PR #4046, opened 2026-06-30
[halang] Add telex style keyboard

poetcin:
halang keyboard first pull request


keyman-server:

poetcin:
Ready for review

LornaSIL:
Thanks for your submission! I haven't tested the keyboard yet, but here are a few comments:

- The LICENSE.md needs the copyright year `Copyright © 2026 HEAR the WORD`. (That is the only place we want the year, please don't add the year elsewhere.)
- Any keyboard in `release` needs to have some documentation on the keyboard layout. So, in the `welcome.htm` you could add a table with the keystrokes that are different from normal Qwerty, or you could add graphics with the layout.
- I noticed your .kvks only displays the default layer. You should reimport the .kvks layout.
- We also require any keyboard in `release` to have an online help file.
   - You would add a `help` folder within source. Within the `help` folder create a file called `halang.php`. 
   - Add this into `halang.php`:

```
<?php 
  $pagename = 'Halang Keyboard Help';
  $pagetitle = $pagename;
  require_once('header.php');
?>

<p>Halang Telex Style keyboard for the Halang language of Vietnam. </p>

<p>Type "a", plus a "w" to produce the "a" with an inverted breve above it.</p>
<p>Type a vowel twice to produce the vowel with a circumflex above it.</p>
<p>Type "o" or "u" followed by a "w" to produce the "ơ'" or "ư".</p>
<p>Type "uo" followed by a "w" to produce "ươ".
```

You should get the idea. The same kind of help should go into the `welcome.htm`.

- It would be helpful to use this description in `readme.htm`, `welcome.htm` and the .kps `Halang Telex Style keyboard for the Halang language of Vietnam.` instead of `Halang Telex Style Keyboard Ver 1.1`. We don't really want the version in there and a bit more description is always helpful to people who are trying to decide if a keyboard is appropriate for what they need.

Once you make these changes I'll test the keyboard, but hopefully this won't be too hard. You've got the files in the right location!

poetcin:
All requested updates have been made. 
The .kvks file appeared normal on my end, but I recopied and pushed it again.
Ready for review.

LornaSIL:
Thanks! There are just a few more changes. The keyboard compiles fine, but I have some suggestions for that too.

When the keyboard is compiled into the .kmp it is "flattened" in that no folders are in there. So, your welcome.htm file shouldn't not have a path to the `help` folder for the graphics because there will not be a help folder in the .kmp.

- Move the graphics out of the `help` folder into `source`
- Remove the path to the `help` folder in `welcome.htm`
- In readme.htm make the same adjustment as you did elsewhere for the description of the keyboard. (That is the first thing people will see when installing the package.)
- The OSK isn't technically accurate. There aren't any rules in your .kmn to type uppercase "F", "Q", "W" to get combining marks, so either you need to add those rules, or you should modify the OSK because it just outputs "F", "Q", "W". I think adding to the rules so that the user doesn't get frustrated would be best. You can add extra rules to handle the casing, or I'll suggest an alternate way of handling it below.
- In the .kps file make sure to add the graphics into the list of files so that they will be included in the package.
- In the halang.php file, add these lines at the bottom. They will automatically produce the layout of the keyboard.

```
</div>

<h2>Mobile/Phone Keyboard Layout</h2>
<p>Due to the size and number of keys, some characters are hidden in the long press. 
	Press and hold on the key with a little dot on the top right to reveal and use them.</p>

</div>
```

For the .kmn rules, in order to let the "W" or "w" work the same way, you could add a store that includes both keystrokes. See the first line below, it lets the user type either uppercase or lowercase.

```
    store(w) [K_W] [SHIFT K_W]
```

Then, change the last 4 rules to:
```
    c Special ươ vowel Diacritic rules - Uppercase and Lowercase
    'uo' + any(w) > 'ươ'
    'Uo' + any(w) > 'Ươ'
    'uO' + any(w) > 'ưƠ'
    'UO' + any(w) > 'ƯƠ'
```
That means whether the user types uppercase or lowercase "w" the rule will work. The other way you have to have at least 4 extra rules. 

You can do something similar for "f" and "q".

I can accept the rules as you have them, but I think making it less frustrating to your users would help. (I had to read the code in order to figure out why what I was typing didn't work and you don't want your users to have to do that.)

poetcin:
All requested updates have been made.
Ready for review.

LornaSIL:
Thanks! I'm sorry I wasn't clear, but the `halang.php` file should be moved back into a folder called `help`. The `help` folder needs to be in the `source` folder: `halang/source/help/halang.php`.

Everything else is fine, so just make that one change, don't move your graphics anywhere.

poetcin:
Moved halang.php into the help/ directory.
Ready for review.

LornaSIL (approved):
LGTM

mcdurdin:

This keyboard has now been published and will shortly be available at: https://keyman.com/keyboards/halang.

PR #4080, opened 2026-07-29
[sil_lisu] Update keyboards to remove 02D7 and use Lisu Bosa font

LornaSIL:
This change was triggered by #4051 but in the end I didn't make the change that was being requested because my research on the script indicates that the regular hyphen shouldn't be used.


darcywong00 (approved):
lgtm

mcdurdin:

The following keyboards have been published:
* https://keyman.com/keyboards/sil_lisu_basic
* https://keyman.com/keyboards/sil_lisu_standard

PR #4092, opened 2026-08-08
Add isv_latin: Interslavic (Latin), standard orthography

radoslove:
First Interslavic keyboard in the catalogue.

Standard orthography only — č š ž ě plus the punctuation Interslavic texts use („ ” – —). The extended alphabet is deliberately absent: it is not written in new text, and this matches the author's Windows (.klc), macOS (.keylayout) and Android layouts, which carry exactly the same four letters.

Touch: no AltGr exists on a touchscreen, so the letters sit on longpress of their base key — hold `c` for `č` — the same mnemonic as AltGr+C on desktop. The longpress keys are touch-only T_* keys with their own rules rather than a re-use of the RALT rules, because a touch subkey cannot reach the rightalt layer. Both sets are generated from one table so they cannot drift apart.

Language: isv-Latn (ISO 639-3 `isv`), resolves natively. Tested on iPadOS 26 via Keyman: install, typing, longpress, capitals. Builds warning-clean with kmc 18.0.249.

Sources and the generator: https://github.com/radoslove/keyboard-interslavic


keyman-server:

LornaSIL:
Thanks for your keyboard submission. It looks to be in very good shape. There are a few things you should update:

- go to the "On-Screen" keyboard tab (in the .kmn) and select "Auto-fill underlying layout" and then select "Fill from layout". That way users who choose to use the On-Screen keyboard will see what they should type. This is also what we would use in our online help file. More on that below.
- In the .kmn you can select the target of "any" and unselect all the others.
- In the touch layout, when the tablet and phone have the exact layout, just remove one of them. In this case, remove the tablet one. That makes it much easier to maintain your keyboard when you make changes. The phone layout will work on tablet, etc.
- Any keyboard that goes into the `release` section is required to have an online help file. You would create a `help` folder within `source`. Add a file called `isv_latin.php` in that folder. You would likely want to include most of the content you have in `welcome.htm` but you need to use the php header and styles. (You may not need the the "data-states" since you've documented the keyboard with a table.) You can read about it here: https://help.keyman.com/developer/keyboards/phphelpfile If you need help with this, I will look at it after you submit what you think will work.

radoslove:
Thanks for the review — all four are done, plus one thing that turned up while I was in there.

**On-Screen keyboard.** The `.kvks` was an empty stub, so the OSK rendered blank and there was no way to discover that `č` is AltGr+C. It now carries the underlying US layout on the base and shift layers together with this keyboard's own RALT / SHIFT+RALT cells, and sets `displayunderlying` and `usealtgr`. I generate the sources rather than use the IDE (I'm on macOS), so I took the shift-state codes — `""`, `S`, `RA`, `SRA` — from existing catalogue keyboards rather than guessing; the touch-layout vocabulary doesn't carry over.

**Targets.** Now `any`.

**Tablet form.** Removed — it was byte-identical to `phone`. One caveat in the interest of being straight with you: my on-device iPad pass was run *with* the tablet form present, so phone-on-tablet is the documented fallback rather than something I have re-confirmed on hardware. I'll re-test on the iPad; say the word if you'd rather I confirm that before this goes in.

**Online help.** `source/help/isv_latin.php`, following the spec you linked. Most of the content is `welcome.htm`'s, including the iOS 16+ Full Access notice, which is the single thing users most need to be told. I did keep the `data-states` blocks — with the `.kvks` now populated they render something useful, and the phone block shows the swipe layout that the table can't. Happy to take a round on this file; I'd rather you look than have me guess.

One extra you didn't ask for: `HISTORY.md` listed 1.0 only, while the keyboard had already reached 1.5. I've reconstructed the intermediate entries from the generator's own history and called this round 1.6, since the OSK fix is user-visible.

Everything except the `.php` and `HISTORY.md` is generated from one table by [`build_keyman.py`](https://github.com/radoslove/keyboard-interslavic/blob/main/build_keyman.py), so the desktop and touch sets can't drift apart. Builds warning-clean with kmc 18.0.249 — there is one hint, that flicks want Keyman 17+, and that one is deliberate: `minKeymanVersion` stays at 10.0 so older installs keep longpress rather than being locked out for a purely additive feature.

radoslove:
Ready for review

LornaSIL:
I think it looks good. There is no need to bump the version each time you make a change, unless it's been distributed. But, it's fine to do it this way.
I wonder if you would want to change the description in the .kps to what you have in the readme.htm:
```
Tipkovnica za medžuslovjansky jezyk — standardna ortografija: č š ž ě.
Interslavic keyboard — standard orthography: č š ž ě.
```
If you do, I could make that change for you. The reason I suggest it, is that is the text which will show up on the keyboard home page, so it would be nice to be in the language.

Otherwise, it's fine.

LornaSIL:
Did you want to update the description?

radoslove:
Yes, thank you — that is the better call, and the reason you gave is the convincing one: the catalogue page is exactly where the language should speak for itself.

Pushed in [`f2407e6f`](https://github.com/keymanapp/keyboards/pull/4092/commits/f2407e6faefe60e056d90457aac4fb73f6e31bed). The description now reads:

```
Tipkovnica za medžuslovjansky jezyk — standardna ortografija: č š ž ě.
Interslavic keyboard — standard orthography: č š ž ě.
```

That is the `readme.htm` wording verbatim — Interslavic first, English as a gloss. And thank you for the note on versions; I have left this one alone, since nothing has been distributed since the last bump.

Ready for review.

LornaSIL (approved):
LGTM

keyman-server:

This keyboard has now been published and will shortly be available at: https://keyman.com/keyboards/isv_latin.

PR #4104, opened 2026-08-19
chore: disable auto-commenter until fix for PR search algorithm lands

mcdurdin:
Note, also fixes a typo in build.sh _ vs -.

This is not currently working correctly; we need to retrieve the previous commit on master rather than the information from TeamCity which shows commits on the branch.

Test-bot: skip


SabineSIL (approved):
RSLGTM

PR #4106, opened 2026-08-19
fix: use previous successful build to determine last commit to check in auto-commenter

### LINKED ISSUE #4105: bug: auto-commenter selects wrong range of commits
/resources/build/version/teamcity-get-build-changes.ts is not currently working correctly. We need to retrieve the previous successful build and use its last commit instead, I think? Currently the range it gives relates to the branch which is merging, which can end up dragging in unrelated branches.

mcdurdin:
Fixes: #4105


darcywong00 (approved):
lgtm

PR #4126, opened 2026-09-12
fix(mbya_guarani): remove critical whitespace from .php

mcdurdin:
The `<<<END` heredoc syntax in PHP is very sensitive to trailing whitespace, and the resulting error message is quite unclear.

Fixes: #4124

PR #4125, opened 2026-09-12
maint: replace missing PHP variable name

mcdurdin:
Post PHP upgrade, help site complained about missing variables in these pages.

PR #4129, opened 2026-09-14
[mbya_guarani_php] Remove more extra space

LornaSIL:


darcywong00:
Ah, this is also handled on #4125 in commit 9442b5ec907bb7cd264437d3d648dff3fb199d10

mcdurdin:
Yes, I resolved this in #4126, which @LornaSIL merged into #4125, which was not at that time approved. Should be able to close this.

